PR #7, opened 2020-06-19
Zc/upload directory

corps:
Combines my upload directory and @coleshaw 's etna-js extraction work.

Todo:

- [x] Carefully test and merge https://github.com/mountetna/metis/pull/82/files into master metis
- [x] Pull master metis into monoetna
- [x] Change monoetna's ci so that metis js + webpack tests run against monoetna's version of etna packages.
- [x] Merge my work here https://github.com/mountetna/metis/pull/81 into this branch
- [x] Get all green and cool.
- [x] Change monoetna to automatically sync to downstream child repos upon merge.


coleshaw:
It looks like `jest` doesn't find any tests:

```
 In /app
  82 files checked.
  testMatch: **/__tests__/**/*.js?(x),**/?(*.)(spec|test).js?(x) - 0 matches
  testPathIgnorePatterns: /node_modules/ - 82 matches
Pattern:  - 0 matches
Hash: bc1843769a08e03bceef
Version: webpack 3.12.0
```

Is it not searching deep enough in the folder structure, or looking in the wrong subtree?

coleshaw:
Oh nevermind, I see that the monorepo does tests for each subproject, so I just happened to find the entry for a subproject with no UI tests.

corps:
> Oh nevermind, I see that the monorepo does tests for each subproject, so I just happened to find the entry for a subproject with no UI tests.

Good point though, I could make each test be a separate sub header.  It was simply convenient to do it all under one header.

corps:
Next, I'll add feedback from https://github.com/mountetna/metis/pull/81/files and close that PR

graft:
Can you describe how the last item works?

corps:
Sure.
On commit to master, the github action would be triggered.  It would then issue, for each subtree repo, something like `git subtree push --prefix=the-thing`.  What git subtree essentially does is linearly scan the entire repo's history, finding all commits that had file chnages inside of `the-thing`.  It creates a new temporary history of _only_ the changes in monoetna that actually applied to `the-thing`.  It uses commit messages from the corresponding `git subtree pull` to align that commit history of monoetna to the history of the upstream `the-thing`.  Based on that history alignment, it attempts a normal diff merge, and commits a new message upstream based on that.  

Here's an example of a `git pull` that contains the commit hash that aligns future git subtree pushes. https://github.com/mountetna/monoetna/commit/a0a1a149ce1ac87cf245b5aca8f823c654eecdc0 

As a result, it will look and function, for all purposes, like `git merge` from this repo to each of the child repos.

If there is any failure that an automatic merge can't solve (in the case that someone makes a merge in the child repo but it hasn't yet made it back to the mono repo), the job will just fail and will require a manual pull.

Once we're all doing most of our development from the monorepo, that won't be much of a concern however.

graft:
Neat, seems pretty simple and straightforward actually.

PR #9, opened 2020-06-24
Cs metis upload

coleshaw:
This PR changes several things:

* Fixes Metis HMAC params to allow for author fields, flagged in a comment on mountetna/metis#78. The corresponding Magma change will be in a separate PR, given the weird branch-iness of the current workflow.
* Extracts the dispatch logic for the upload action and worker in `etna-js`, into a separate method (to be used in Timur).
* Allows for CORS upload requests in `etna-js`.
* Configure Metis to allow for CORS requests.


metis/lib/server/controllers/upload_controller.rb
```diff
@@ -58,6 +52,21 @@ def upload
 
   private
 
+  def create_upload(params)
```
corps, line 55: So, I noticed that if a user starts a download but doesn't complete it, then tries to upload _a different file_ to the same path, the uploading fails because it re-uses the existing `Upload` model, which may have a `next_blob_hash` that doesn't match the new file that is replacing it.

Shoudl we change this behavior to more of a `create_or_reset_upload`?  If the upload does exist, update its file_by_position and next_blob* params to defaults.

This would allow starting an incomplete download over with a new file.
Should we consider changing this to a "create or reset" like behavior?  If the upload exists, it should be reset to resume from the 0 byte position.  Client'

corps:
Looks like a legitimate test failure in metis btw.

coleshaw:
> Looks like a legitimate test failure in metis btw.

Hm, that's interesting. Passes for me on macOS, I wonder if it's an OS-related artifact of the order in which files are fetched. I'll look into making the test more robust.

metis/lib/server/controllers/upload_controller.rb
coleshaw: Hm...that's an interesting scenario. That seems like maybe a browser component problem, not a controller problem? I'll play around with it tomorrow, but wouldn't `find_or_create` get the existing `Upload` (partially completed), and then reset `next_blob_hash`?

```
    Metis::Upload.find_or_create(
      file_name: params[:file_path],
      bucket: params[:bucket],
      metis_uid: metis_uid,
      project_name: params[:project_name]
    ) do |f|
      f.author = Metis::File.author(params[:user])
      f.file_size = 0
      f.current_byte_position = 0
      f.next_blob_size = -1
      f.next_blob_hash = ''
    end
```

I wonder if the browser skips the `authorize_upload` route and instead tries uploading directly back to `upload_blob`, which I could see causing an issue with `next_blob_hash`.

Does it make sense if I make another issue around this use case, since I think it's not directly related to this specific upload-from-Timur feature, but seems like a bug somewhere?

etna/packages/etna-js/utils/fetch.jsx
```diff
@@ -34,7 +34,7 @@ export const headers = (...types) => {
 export const json_fetch = (method) => (path, params) =>
   fetch(path, {
     method,
-    credentials: 'same-origin',
+    credentials: 'include',
```
graft, line 37: Is there a good reason for this? The same-origin policy is much safer.

coleshaw: Without CORS credentials, Timur can't make a request to Metis to upload things...

coleshaw: Note that the `magma_api` request also uses CORS credentials. The alternative is to not share the upload components  or inject the fetch utility or something....seems messier.

metis/lib/server/controllers/upload_controller.rb
graft: Since upload from Timur is to a magma-issued unique temp path resuming may be broken anyway, as the same file would goto a new path on upload retry.

As to not resuming correctly, the upload_start controller only has the file size as a hint as to what the user will send. It could reject if this mismatches but otherwise it should try to resume.

etna/packages/etna-js/utils/fetch.jsx
graft: The same-origin policy ought to work for metis as it does for magma, as long as the domain (e.g. ucsf.edu) is the same.

coleshaw: So if I return that to `same-origin`, Metis tries to call Janus to authenticate the user, and rejects the request. Note that [when Timur calls Magma](https://github.com/mountetna/monoetna/blob/master/timur/lib/client/jsx/api/magma_api.js#L12), it also has to use `include` for `credentials`, not  `same-origin` -- do you have an example of where Timur calls Magma using `same-origin` policy and it works?

```
Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://janus-dev.etna-development.org/login?refer=http%3A%2…339c187cd380a16c6273%26X-Etna-Id%3Dmagma%26X-Etna-Headers%3D. (Reason: CORS request did not succeed).
```

Screen Shot 2020-06-25 at 10 15 49 AM

metis/lib/server/controllers/upload_controller.rb
coleshaw: Good point, the Timur filenames are unique for every `::temp` request. So perhaps that issue is a Metis-specific upload scenario rather than via Timur.

etna/packages/etna-js/utils/fetch.jsx
corps: > The same-origin policy ought to work for metis as it does for magma, as long as the domain (e.g. ucsf.edu) is the same.

Yeah, as far as I know this isn't true.  SOP is in consideration of the entire FQN, including subdomain. https://en.wikipedia.org/wiki/Same-origin_policy#Origin_determination_rules is a good workflow to check

graft: You're right, this is not the case; we relax it on the server end using Etna::CrossOrigin, but the client doesn't relax it right now.

We can instead apparently [relax the same-origin policy](https://developer.mozilla.org/en-US/docs/Web/Security/Same-origin_policy) in the client by setting e.g. `document.domain = 'ucsf.edu'`. I'm not sure this is actually a significant improvement but in these matters I'd prefer to be as conservative as possible.

corps: Interesting, I didn't know you could actually alter the domain directly.

graft: Relevant issue I just wrote up which might obviate some of this discussion: https://github.com/mountetna/etna/issues/60

coleshaw: Modifying `document.domain` seems like a pretty neat idea, but doesn't seem to work in practice (Like Zach, that's new to me). I wonder if because the APIs are still in subdomains and aren't on the top-level domain.

I set Timur to use `document.domain = document.domain.split('.').slice(1).join('.');`, which for local dev for me becomes `etna-development.org`. And then I changed the `magma_api` calls to use `credentials: same-origin` instead of the current `includes`, and now I get a `401` Unauthorized. So I think there are still issues authenticating across services, where credentials have to be included.

corps: The alternative could be to read values from cookies and set them on requests explicitly as headers?
Then we don't have to worry SOP as it applies to cookies.

graft: Yeah, there's no reason we can't do this already in advance of mountetna/etna#60 - the auth layer already respects headers, the client is able to read the token in JS. Doesn't have to be immediately, though.

metis/lib/server/controllers/upload_controller.rb
```diff
@@ -58,6 +52,21 @@ def upload
...
+      project_name: params[:project_name]
+    ) do |f|
+      f.author = Metis::File.author(params[:user])
+      f.file_size = 0
+      f.current_byte_position = 0
+      f.next_blob_size = -1
+      f.next_blob_hash = ''
+    end
+  end
+
```
graft, line 69: If this has to move out of the main controller method I'd prefer to keep methods like this on the relevant model, like `Metis::Upload.fetch` or something, rather than adrift in the controller class (where it becomes confusing who they are "helping").

metis/lib/server/controllers/upload_controller.rb
```diff
@@ -71,7 +80,19 @@ def upload_start
       metis_uid: metis_uid,
     ).first
 
-    raise Etna::BadRequest, 'No matching upload' unless upload
+    if !upload
+      require_params(:project_name, :bucket_name, :file_path)
+      hmac = @request.env['etna.hmac']
+      upload = create_upload({
+        file_path: @params[:file_path],
```
graft, line 87: Hmm, I am confused by this - above in authorize we create the upload; here we find an existing upload, then if it does not exist we call create_upload - if we can do that safely, doesn't this mean we no longer need to do it in authorize at all?

Also - here you call create_upload, which uses `find_or_create`, but we know there is no upload to be found, so create() suffices - but can we not just dispense with the Upload.where above and use create_upload just once in this controller?

metis/lib/server/controllers/upload_controller.rb
```diff
@@ -71,7 +80,19 @@ def upload_start
       metis_uid: metis_uid,
     ).first
 
-    raise Etna::BadRequest, 'No matching upload' unless upload
+    if !upload
+      require_params(:project_name, :bucket_name, :file_path)
```
graft, line 84: Note that you don't have to require_params for things that are already in the route, as they are guaranteed to be present. This is maybe a more future-proof way to write the controller, though.

etna/packages/etna-js/utils/fetch.jsx
coleshaw: :+1:

metis/lib/server/controllers/upload_controller.rb
coleshaw: Oh yeah, I see what you're saying. We could move Upload creation to `upload_start` in all cases, since it's not really required in `authorize` -- there we could just do the checks and return the URL.

coleshaw: @graft , like so?

coleshaw:
Yeah, I figured that would catch it, but adding this would be a little more defensive in case the route definition changes.

graft:
Could also defend with a spec, which you should do anyway, all routes should have specs covering auth states.

metis/lib/server/controllers/upload_controller.rb
coleshaw: Recorded in mountetna/etna#61. I think this is a client-side issue.

PR #10, opened 2020-06-27
First pass at ability to download folders

corps:
G2G as per demo.

I cleaned up some babel stuff by upgrading the babel version and simplifying some configs.
I also did some scaffolding for having ui tests in metis as well.  I'm pondering if we want to move some of the devDependencies of etna-js into etna, especially those that concern testing?  Otherwise we're forced to duplicate npm installs of every test dependency in etna-js for every project that wants to borrow testing code from etna-js.


.github/workflows/master.yml
```diff
@@ -47,7 +47,7 @@ jobs:
       # Clone the repo again configured with the GITHUB_TOKEN that won't trigger another workflow :'(
       - uses: actions/checkout@v2
         with:
-          path: ../derp
+          path: derp
```
corps, line 50: Unrelated fix to the automatic subtree syncing.  Ask me about it later :P

development-apache/Makefile
```diff
@@ -0,0 +1,30 @@
+# SHELL := /bin/bash
```
corps, line 1: I created a make file here so that you can restart / down / up / logs the development apache server more easily

```
make -C development-apache restart
```

docker/Makefile
```diff
@@ -40,4 +40,4 @@ ps: docker-compose.yml ## Shows ps of all projects' containers
 
 .PHONY: logs
 logs: docker-compose.yml ## Shows logs of all running projects' containers
-				@ docker-compose logs
+				@ docker-compose logs -f
```
corps, line 43: This will make the logs command tail by default, which is honestly more useful.

etna/packages/etna-js/jest.config.js
```diff
@@ -0,0 +1,20 @@
+module.exports = {
```
corps, line 1: I extracted this out of the package.json#jest config object once it started to get bigger.  This would allow us to add comments and piece together / share config in the future easier than it being part of the package.json payload.

etna/packages/etna-js/spec/setup.js
```diff
@@ -1,7 +1,10 @@
+import 'regenerator-runtime/runtime';
```
corps, line 1: Just centralizing this so all tests benefit.

etna/packages/etna-js/utils/asserts.jsx
```diff
@@ -2,12 +2,18 @@ export function assertIsSome(thing) {
...
 
   if (keys.length === 1) {
-    const key = keys[0];
-    const value = thing[key];
+    const value = thing[keys[0]];
     if (!value) {
       throw new Error(`Expected ${key} to be present, value was ${value}`);
     }
     return value;
+  } else if (keys.length > 1) {
```
corps, line 10: Convenience method I'm writing to help hunting down stray null bugs faster.

assertIsSome({ my, inputs, here }) => { my, inputs, here } and verify each key is true-y.

metis/babel.config.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
```
corps, line 1: Extracting this out means that webpack and jest can share it more easily.

metis/lib/client/jsx/actions/file_actions.jsx
```diff
@@ -1,19 +1,62 @@
...
+  );
+}
+
+export const downloadFilesZip = ({ folder_name, files }) => (dispatch) => {
+  // a stable identifier to share to cache some downzip work when re-clicked.
+  const downloadId = Array.from(files.map(({ file_hash }) => file_hash).join(".")).reduce((s, c) => Math.imul(31, s) + c.charCodeAt(0) | 0, 0)
+  const folderParts = folder_name.split("/");
+  const zipName = folderParts.join("--") + ".zip";
+
+  return downZip.downzip(downloadId, zipName, files.map(({ size, download_url, file_path }) => ({
```
corps, line 38: This section will unfortunately remain harder to effectively e2e since node has no effective service worker test utilities.

metis/lib/client/jsx/api/files_api.jsx
```diff
@@ -1,16 +1,19 @@
 import { json_get, json_delete, json_post } from '../utils/fetch';
 
+// TODO: These really ought to be refactored into a service class that knows about metis' domain
+// in various environments, rather than assuming the location of the currently viewed page contains
+// these endpoints.
 export const postRetrieveFiles = (project_name, bucket_name, folder_name) =>
-  json_get(`/${project_name}/list/${bucket_name}/${folder_name}`);
+  json_get(`${window.location.origin}/${project_name}/list/${bucket_name}/${folder_name}`);
```
corps, line 7: Required as our node-fetch stub in jest requires all fetch calls to include protocol and server (origin)

metis/webpack.config.js
```diff
@@ -88,5 +87,14 @@ module.exports = {
       filename: 'public/css/metis.bundle.css',
       allChunks: true,
     }),
+    new CopyPlugin([
```
corps, line 90: ServiceWorkers must be sideloaded as seperate files, so we are forced to generate the asset outside the bundle.

etna/packages/etna-js/upload/api/__tests__/upload_api.spec.jsx
```diff
@@ -3,10 +3,6 @@ import * as api from '../upload_api';
 import { mockStore, stubUrl, cleanStubs } from '../../../spec/helpers';
 
 describe('upload api', () => {
-  afterEach(() => {
-    cleanStubs();
-  });
-
   it('postAuthorizeUpload makes a request to the authorize upload route', () => {
     global.fetch = require('node-fetch');
```
coleshaw, line 7: nit: Should be able to get rid of this `fetch` assignment, since you now assign it in `setup.js`?

coleshaw:
Duplicating installs seems a waste. If we move the dependencies, would we have to do `npm install` twice in `etna` to test `etna-js`? Once at the root level and once in `packages/etna-js`?

etna/packages/etna-js/upload/api/__tests__/upload_api.spec.jsx
corps: Yup

corps:
> Duplicating installs seems a waste. If we move the dependencies, would we have to do `npm install` twice in `etna` to test `etna-js`? Once at the root level and once in `packages/etna-js`?

Oh, it's not like that.  It's just that we have to explicitly add the same test dependencies both in `etna-js` and `metis` / `timur`, unless we move them out of `devDependencies` into `dependencies` instead.

development-apache/docker-httpd.conf
```diff
@@ -555,23 +555,25 @@ ServerTokens Prod
...
-    ProxyPass / http://janus_app:3000/
+    ProxyPass / http://janus_app:3000/ disablereuse=on
     ProxyPassReverse / http://janus_app:3000/
 
     ServerName janus.development.local
 
     SSLEngine on
     SSLCertificateFile /root/certs/janus.development.local.pem
     SSLCertificateKeyFile /root/certs/janus.development.local-key.pem
+    Header add Service-Worker-Allowed "/"
```
corps, line 566: These have been reflected already in cbc-chef repo as well.  Necessary for service workers to be allowed on any path of the metis program.  If we do not include this header, we would be forced to serve the downzip-sw.js file in the `/` directory instead of the `/js` directory because service workers by default are not allowed to control pages that are not children of their own paths.  Check documentation on Service-Worker-Allowed and service worker scopes for more info.

etna/docker/app/docker-entrypoint.sh
```diff
@@ -4,6 +4,7 @@ set -e
 : "${RANDOM_MAX:=500}"
 [ -n "$DEBUG" ] && echo "Running: $@"
 export PATH="/app/node_modules/.bin:/app/vendor/bundle/$RUBY_VERSION/bin:$PATH"
+cp npmrc ~/.npmrc || true
```
corps, line 7: I am working soon on a way to 'share' these docker pieces between projects.  Right now I have to manually duplicate new logic to all projects.

metis/spec/setup.js
```diff
@@ -0,0 +1,15 @@
...
+const raf = (global.requestAnimationFrame = (cb) => setTimeout(cb, 0));
+
+const Enzyme = require('enzyme');
+const EnzymeAdapter = require('enzyme-adapter-react-16');
+// Setup enzyme's react adapter
+Enzyme.configure({ adapter: new EnzymeAdapter() });
+
+global.fetch = require('node-fetch');
+nock.disableNetConnect();
+nock.emitter.on('no match', req => {
```
corps, line 13: Helpful for debug purposes; when a request is made in a test that does not match a stub, information on that request is logged so that you can quickly figure out which urls & payloads need stubbing.

etna/packages/etna-js/upload/api/__tests__/upload_api.spec.jsx
corps: Taken care of.

coleshaw:
Looks good to me based on the demo. Should we create a separate "tech debt" issue to move the `etna-js` dependencies?

corps:
> Looks good to me based on the demo. Should we create a separate "tech debt" issue to move the `etna-js` dependencies?

https://github.com/mountetna/monoetna/issues/13

PR #12, opened 2020-06-30
Add documentation for Timur about new UI buttons

coleshaw:
Documentation for the 3 COMET V1 buttons.

Not sure if the images will link correctly, but this format is the [GItHub image markdown for relative images](https://gist.github.com/TT--/14260aef6c0e31fca5b37e7cb3c53020).


coleshaw:
Images now render correctly using kramdown syntax. Also updated with info about the Timur modal for inputting the Metis path.

graft:
On my screen the pictures are centered and don't wrap text, which makes the progression hard to follow. A general layout scheme for this kind of image progression might be nice...

I also think you should exclude the biological context here - the files could really be of any sort, a PDF, a PPTX, etc., and aren't necessarily biological data; we are just "managing files". It also makes it harder to see in your example images which attributes are relevant, since there are other, non-file attributes in the image which are only distractions. I would be more inclined to use constructed examples in a mock interface rather than show the actual IPI record.

coleshaw:
I've got the images + text in a two-column layout now, like below. Does this look better?

Screenshot from 2020-07-08 19-54-39

PR #14, opened 2020-07-02
Uploader now attempts reset when blob returns a 422

corps:
As per discussion in channel, this makes 3 changes to allow for both logical file upload resume and also reset in the case that a differing file begins an upload in the same location.

1.  When the client receives a 422 from a blob action, it will (up to timeout) restart the upload by passing reset: true to upload_start.
2.  When the server receives upload_start with reset true, it will reset the upload
3.  When the server receives a file_size differing from an existing upload to upload_start, it will also restart the upload.

Includes some test refactoring / cleanup on the client side as well.


etna/packages/etna-js/spec/helpers.js
```diff
@@ -11,15 +11,34 @@ export const stubUrl = ({
...
-  nock(host)
-    [verb](path, request)
-    .reply(status, response, {
-      'Access-Control-Allow-Origin': '*',
-      'Content-type': 'application/json',
-      ...headers
-    });
+  let nocked;
+
+  return new Promise((resolve, reject) => {
```
corps, line 18: `nock` by default doesn't have a sane way to order request mocks, which is important if you want to be certain that requests come in a particular order.

This minor change makes the return value of this function more useful; it's a promise that is only fulfilled when the given mocked response is actually served, allowing one to use promises to chain the order together in which you expect requests to come.  Doesn't break any existing usages.

etna/packages/etna-js/spec/helpers.js
```diff
@@ -11,15 +11,34 @@ export const stubUrl = ({
...
+        'Access-Control-Allow-Origin': '*',
+        'Content-type': 'application/json',
+        ...headers
+      });
+  });
 };
 
 export const cleanStubs = () => nock.cleanAll();
+
+global.expect.extend({
```
corps, line 35: Useful extension to expect that allows own to ensure that a given promise does resolved.  Used in combination with the above to assert that a given mocked response is actually fulfilled.

etna/packages/etna-js/spec/setup.js
```diff
@@ -1,10 +1,31 @@
...
+      options = args[0];
+    }
+  }
+
+  const reqString = stringifyRequest(options, bodyString);
+  console.error('got request with no match', reqString);
+})
+
+// nodejs equivalent
+global.FormData = URLSearchParams;
```
corps, line 31: This fixes another issue I noticed where we were unable to match form data requests correctly.  It seems that in nodejs the correct replacement is URLSearcParams

etna/packages/etna-js/spec/setup.js
```diff
@@ -1,10 +1,31 @@
...
 const EnzymeAdapter = require('enzyme-adapter-react-16');
 // Setup enzyme's react adapter
 Enzyme.configure({ adapter: new EnzymeAdapter() });
 
-global.fetch = require('node-fetch');
\ No newline at end of file
+global.fetch = require('node-fetch');
+
+nock.disableNetConnect();
+nock.emitter.on('no match', (...args) => {
```
corps, line 15: Useful error message when a request is made that doesn't match a mock.

etna/packages/etna-js/upload/api/upload_api.jsx
```diff
@@ -8,4 +8,4 @@ export const postAuthorizeUpload = (base_url, project_name, bucket_name, file_pa
 
 export const postUploadStart = (upload_url, request) => json_post(upload_url, { action: 'start', ...request });
 export const postUploadCancel = (upload_url, request) => json_post(upload_url, { action: 'cancel', ...request });
-export const postUploadBlob = (upload_url, request) => form_post(upload_url, { action: 'blob', ...request });
+export const postUploadBlob = (upload_url, request, checkStatus= true) => form_post(upload_url, { action: 'blob', ...request }, checkStatus);
```
corps, line 11: Useful when custom logic is needed around certain status codes; allows optionally that the promise succeed with the entire response object.

etna/packages/etna-js/upload/workers/uploader.jsx
```diff
@@ -155,7 +157,7 @@ export default (uploader) => {
     // private methods
 
     reset: () => {
-      uploader.timeouts = 0; // The number of times an upload has timed out.
+      uploader.timeouts = {}; // The number of times an upload has timed out.
```
corps, line 160: Localize timeouts to a per url basis in the case of particular bugs.

etna/packages/etna-js/upload/workers/uploader.jsx
```diff
@@ -174,14 +176,18 @@ export default (uploader) => {
...
-        uploader.reset();
+    timeout: (upload) => {
+      const { url } = upload;
+      const { timeouts, maxTimeouts } = uploader;
+
+      timeouts[url] = timeouts[url] || 0;
+
+      if (++timeouts[url] >= maxTimeouts) {
         uploader.dispatch({ type: 'UPLOAD_TIMEOUT', upload });
-        return false;
```
corps, line 182: This boolean was incorrect; timeout should return true when there is a timeout, and false when there is not.  Before, it would stall immediately after any one failure.

etna/packages/etna-js/upload/workers/uploader.jsx
```diff
@@ -174,14 +176,18 @@ export default (uploader) => {
 
     remove: (upload) => uploader.dispatch({ type: 'REMOVE_UPLOAD', upload }),
 
-    timeout: () => {
-      ++uploader.timeouts;
-      if (uploader.timeouts == uploader.maxTimeouts) {
-        uploader.reset();
+    timeout: (upload) => {
+      const { url } = upload;
```
graft, line 180: Does timeout here actually mean any kind of failure? I.e. a 422 lands here, but is not a "timeout"...

etna/packages/etna-js/upload/workers/uploader.jsx
```diff
@@ -174,14 +176,18 @@ export default (uploader) => {
...
+      const { timeouts, maxTimeouts } = uploader;
+
+      timeouts[url] = timeouts[url] || 0;
+
+      if (++timeouts[url] >= maxTimeouts) {
         uploader.dispatch({ type: 'UPLOAD_TIMEOUT', upload });
```
coleshaw, line 186: Does anything catch this action and display a message or log it? I don't see anything, but it might be nice to display something.

etna/packages/etna-js/upload/workers/uploader.jsx
```diff
@@ -174,14 +176,18 @@ export default (uploader) => {
 
     remove: (upload) => uploader.dispatch({ type: 'REMOVE_UPLOAD', upload }),
 
-    timeout: () => {
-      ++uploader.timeouts;
-      if (uploader.timeouts == uploader.maxTimeouts) {
-        uploader.reset();
```
coleshaw, line 180: Removing the call to `reset()` seems like it might change the behaviour of the uploader -- have you tested re-uploading the same file after it times out?

etna/packages/etna-js/upload/workers/uploader.jsx
corps: I also was slightly confused by this terminology.  I think the original author's intent was for timeout to be like "gating multiple failed requests" to prevent server being inundated with requests that cannot succeed.  So yes, any error is the approach I went with since the idea I believe is to generally gracefully pause for any reason that the server continuously rejects requests.

corps: @graft What would you prefer the terminology and behavior be by the way?  I can see how the terminology is confusing, but I definitely think pausing downloads when _any_ general error repeats itself 5 times is worth having to prevent a bug from forcing retries constantly and hurting the server.

corps:
Some more minor fixes coming

etna/packages/etna-js/upload/workers/uploader.jsx
corps: Good call, I locally am resetting at this point.  Haven't pushed.

etna/packages/etna-js/upload/workers/uploader.jsx
corps: Yeah, so... it doesn't look like it does anything right now lol.  But I am keen on it pausing the download and notifying the user, which I think I can add.  My plan is to do some deeper refactoring that will make this easier.

graft:
I can't find this comment to reply to any more, alas. I think "retry" seems
more accurate than "timeout", since a time limit is not really involved but
a retry limit is.

corps:
Temporarily closing this so it doesn't get merged yet, will reopen when ready.

PR #15, opened 2020-07-02
Alert user when service workers cannot initialize

corps:
Small change that alerts the user when their browser does not support service workers and they attempt a download by zip.


coleshaw:
CI getting some sort of permissions error for creating the `etna-js` links?

PR #18, opened 2020-07-10
Add timur

coleshaw:
Adds the Timur application to `monoetna`. Unfortunately without a running Magma instance, it's not quite useful for development yet...

Also updates the Metis `rack` Gemfile.lock version since I was running into a conflict in the `metis_app` container.

This PR also pulls in the latest versions of Metis, Magma, and Timur from the separate repos (I ran `./bin/pull-subtrees`).

WIP because still need to resolve conflicts...


corps:
Nice dude!  I can help to resolve conflicts / setup magma to try to get this in place.

coleshaw:
Thanks, that would be great. I’m a little stumped as to why the NPM tests fail, because the webpack container has modules installed in order to serve the UI...Haven’t had a chance to noodle on it further.

timur/docker-compose.yml
```diff
@@ -0,0 +1,62 @@
...
+    depends_on:
+      - timur_db
+
+  timur_webpack:
+    <<: *timur_base
+    environment:
+      <<: *app_environment
+      RUN_NPM_INSTALL: "true"
+    command: >
+      sh -c "./bin/timur create_routes && npm run webpack"
```
coleshaw, line 48: For Timur, we have to run the `create_routes` command to get the latest `routes.js` file on the UI side, before calling webpack. Note that this requires us to install the Ruby gems in the Webpack container, which is different than the other ones.

timur/config.ru
```diff
@@ -9,6 +9,8 @@ require_relative 'lib/server'
 
 Timur.instance.configure(YAML.load(File.read('config.yml')))
 
+# Used primarily by development servers.  These are served via apache in production.
+use Rack::Static, urls: ['/css', '/fonts', '/images', '/js'], root: 'public'
```
coleshaw, line 13: Lets Apache serve the static files, so they aren't processed by Etna Auth middleware.

timur/docker-compose.yml
```diff
@@ -0,0 +1,62 @@
+version: "3.4"
+volumes:
+  app-gems:
+    driver: local
+  timur-db-data:
```
coleshaw, line 5: Namespace the volumes by app name, so we don't run into conflicts with the other containers.

timur/docker-compose.yml
```diff
@@ -0,0 +1,62 @@
...
+  environment: &app_environment
+    COVERAGE: "true"
+    DATABASE_HOST: "timur_db"
+    TIMUR_ENV: "development"
+
+services:
+  timur_app:
+    <<: *timur_base
+    ports:
+      - 3002:3000
```
coleshaw, line 37: Change the host port we map to (3002), so it's different than the other containers.

timur/config.ru
corps: :+1:

timur/config.yml.test
```diff
@@ -1,3 +1,15 @@
+<<<<<<< HEAD
```
corps, line 1: Looks like a simple merge conflict here.

timur/docker-compose.yml
corps: :+1:

timur/docker-compose.yml
corps: Yup, sounds good!

timur/npmrc
```diff
@@ -0,0 +1 @@
+prefix = /app/.npm-packages
```
corps, line 1: :+1: This ends up being necessary for the `npm link` inside the container.  `npm link` has lots of "gotchas" unfortunately so it hasn't been as robust as I had hoped, but that reminds me... there was on change I made that helps with that.

timur/docker/app/docker-entrypoint.sh
```diff
@@ -0,0 +1,22 @@
...
+
+  rm -f tmp/pids/*.pid
+  if [ -z "$SKIP_DB_WAIT" ]; then
+    dockerize -wait tcp://timur_db:5432 -timeout 60s
+    ./bin/timur migrate
+  fi
+fi
+
+if [ -n "$RUN_NPM_INSTALL" ]; then
+  npm install --unsafe-perm
```
corps, line 19: Originally, to setup development links between the monorepo etna-js and metis, I had a make target `setup-links`.  However, I found out that `npm install` obliterates any `npm link` that you setup, so any container that would come up would incidentally destroy those links.

The fix i introduced into metis is to have, in this section here, a new section

```
  if [ -z "$MASTER_BUILD" ]; then
    echo 'Not master build, performing link to local etna-js...'
    npm link ../etna/packages/etna-js
  fi
```

The basic idea is that by default, containers actually do setup the `npm link` to the etna-js of the local checkout.  Only in master builds (in which we are create the special  etna-js artifact branch) do we not setup the npm link, so that it runs a final test against the package.json "as is" (since that is how it would run in production).  

If you decide to go this direction, too, to finish this setup, add a volume entry into the timur docker-compose for `- ../etna:/etna:cached` in addition to the above block here.

timur/webpack.config.js
```diff
@@ -1,74 +1,79 @@
...
     new webpack.DefinePlugin({
-      'process.env': {
-        NODE_ENV: JSON.stringify(env.NODE_ENV),
-      }
-    })
+      "process.env": {
+        NODE_ENV: JSON.stringify(env ? env.NODE_ENV : "development"),
+      },
+    }),
+    new EventHooksPlugin({
```
corps, line 72: We don't need this part for timur, this is metis only for the zip download part.  Clearing this will fix the tests.

corps: Oh wait, you need it for copying the routes piece?
Ah, so the file needs a `fs` import at top and the filename part needs to be changed.

coleshaw: No, you're right, I'll take this out. I added this because I thought I needed it to copy the `routes` file, but I don't think that's actually necessary.

PR #21, opened 2020-07-14
WIP Dockerized magma

corps:
First pass at dockerizing magma.  Currently just clones magma-ipi locally into `projects` for  the simplest approach.  Didn't want to get anything more fancy until we have a better idea of migrating off the ipi models and how we might eventually want to share model data with developers (without checking the fixtures in publicly, that is)

I'm still running into an issue 

```
~/monoetna/magma]$ docker-compose -p monoetna up

...
Migrating ipi
magma_app_1  | /app/vendor/bundle/2.5.7/ruby/2.5.0/gems/sequel-5.28.0/lib/sequel/extensions/migration.rb:179:in `block in reverse': irreversible migration method used, you may need to write your own down method (Sequel::Error)
```

Any ideas?


magma/docker/app/docker-entrypoint.sh
```diff
@@ -0,0 +1,26 @@
...
+  rm -f tmp/pids/*.pid
+  if [ -z "$SKIP_DB_WAIT" ]; then
+    dockerize -wait tcp://magma_db:5432 -timeout 60s
+
+    for project in $(cat config.yml | grep ':project_path:' | sed -e 's/.*:project_path:\s*//g'); do
+      echo $project
+      bin/magma create_db "$(basename $project)"
+      MAGMA_ENV=test bin/magma create_db "$(basename $project)"
+      bin/magma migrate "$(basename $project)"
+      MAGMA_ENV=test bin/magma migrate "$(basename $project)"
```
coleshaw, line 18: For Magma, you also have to call the `bin/magma global_migrate` command to catch all migrations.

coleshaw: I also think you don't need to call `bin/magma migrate` with a project name. I feel like create_db also doesn't require a project name.

magma/docker/app/puma.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+export METIS_ENV=development
```
coleshaw, line 3: Should be `MAGMA_ENV`?

magma/docker/app/docker-entrypoint.sh
```diff
@@ -0,0 +1,26 @@
...
+  bundle check || bundle install -j "$(nproc)"
+
+  rm -f tmp/pids/*.pid
+  if [ -z "$SKIP_DB_WAIT" ]; then
+    dockerize -wait tcp://magma_db:5432 -timeout 60s
+
+    for project in $(cat config.yml | grep ':project_path:' | sed -e 's/.*:project_path:\s*//g'); do
+      echo $project
+      bin/magma create_db "$(basename $project)"
+      MAGMA_ENV=test bin/magma create_db "$(basename $project)"
```
coleshaw, line 16: I'm a little confused here. You set the `MAGMA_ENV` to `test` for these calls .... but shouldn't that just reflect the container `MAGMA_ENV` value? So when running the container for development, you may not have the `ipi` schema to run the migrations?

magma/docker-compose.yml
```diff
@@ -0,0 +1,51 @@
...
+  volumes:
+    - .:/app:cached
+    - magma-app-gems:/app/vendor/bundle:rw
+    - magma-log-dir:/app/log:rw
+    - magma-tmp-dir:/app/tmp:rw
+    - ./projects:/app/projects
+    - ../etna:/etna:cached
+  environment: &app_environment
+    COVERAGE: "true"
+    DATABASE_HOST: "db"
```
coleshaw, line 26: Does this need to be a `magma_d ` host to not conflict with the Metis DB host?

magma/docker-compose.yml
```diff
@@ -0,0 +1,51 @@
...
+    magma_ENV: "development"
```
coleshaw, line 27: Does this need to be `MAGMA_ENV` instead of `magma_ENV`?

magma/Makefile
```diff
@@ -0,0 +1,21 @@
+include ../make-base/stubs.mk
+
+export MAGMA_PROJECTS=$(shell cat config.yml | grep ':project_path:' | sed -e 's/.*:project_path:\s*//g')
```
coleshaw, line 3: This is picking up both the "development" and the "test" projects from the `config.yml` file, which was unexpected. Does that mean that we should only have "development" configs in `config.yml` and keep test configs elsewhere? Otherwise it seems like we'll get the test `labors` project in development and the `ipi` project in test?

magma/docker/app/docker-entrypoint.sh
coleshaw: I took out these four lines and the tests pass....but running the dev container results in DB errors. I think it's related to the shell grep / sed above finding dev and test projects.

```
Finished in 2.13 seconds (files took 0.96826 seconds to load)
258 examples, 0 failures

Randomized with seed 20558
```

magma/Makefile
coleshaw: Or use something like [yq](https://github.com/mikefarah/yq) to pick just the ones that match the environment? Not sure what the best approach would be here to get tests and dev to work independently.

magma/docker/app/docker-entrypoint.sh
corps: Yeah, I was lazy and added the invocations twice so that I could test both the test and development environments at once was all, so we can leave just one in.

magma/docker-compose.yml
corps: Yup

magma/docker-compose.yml
corps: Search and replace artifact, yup.

magma/Makefile
corps: I actually figured it was ok to have both in both environments anyways, which is why I didn't bother doing anything more complicated.
 
Ok, I'll go ahead and try parsing only the currnent environment's projects.

magma/docker/app/docker-entrypoint.sh
```diff
@@ -3,23 +3,31 @@ set -e
 
 export PATH="/app/node_modules/.bin:/app/vendor/bundle/$RUBY_VERSION/bin:$PATH"
 
+function findEnvConfig() {
```
coleshaw, line 6: We'll have to standardize that `test` config comes second, right?

coleshaw:
So I don't think this is an issue explicitly with this PR, but I'm now getting an error on the Apache container (I can file a separate issue if needed). This happens after running `docker image prune -a` to clean out old images, and then `make up`.

```
Building apache
Step 1/4 : FROM library/httpd:2.4.43
2.4.43: Pulling from library/httpd
Digest: sha256:387f896f9b6867c7fa543f7d1a686b0ebe777ed13f6f11efc8b94bec743a1e51
Status: Downloaded newer image for httpd:2.4.43
 ---> ccbcea8a6757
Step 2/4 : WORKDIR /app
 ---> Using cache
 ---> d48c83ddef7a
Step 3/4 : RUN apt-get update
 ---> Using cache
 ---> d38067d54fc4
Step 4/4 : RUN apt-get install -y libapache2-mod-xsendfile
 ---> Running in 49966eb4db03
Reading package lists...
Building dependency tree...
Reading state information...
E: Unable to locate package libapache2-mod-xsendfile
ERROR: Service 'apache' failed to build: The command '/bin/sh -c apt-get install -y libapache2-mod-xsendfile' returned a non-zero code: 100
make[3]: *** [../make-base/docker-compose.mk:17: build] Error 1
```

Do you think there's a version conflict somewhere or is the image misconfigured, since it says it's a newer version of `httpd`? What version are you running?

magma/docker/app/docker-entrypoint.sh
corps: Yeaaaa... you're not wrong.  I could use `yq` by adding it into the container, I just avoided that because it's an additional download into the container, but maybe that's a better approach in the end.

corps: > So I don't think this is an issue explicitly with this PR, but I'm now getting an error on the Apache container (I can file a separate issue if needed). This happens after running `docker image prune -a` to clean out old images, and then `make up`.
> 
> ```
> Building apache
> Step 1/4 : FROM library/httpd:2.4.43
> 2.4.43: Pulling from library/httpd
> Digest: sha256:387f896f9b6867c7fa543f7d1a686b0ebe777ed13f6f11efc8b94bec743a1e51
> Status: Downloaded newer image for httpd:2.4.43
>  ---> ccbcea8a6757
> Step 2/4 : WORKDIR /app
>  ---> Using cache
>  ---> d48c83ddef7a
> Step 3/4 : RUN apt-get update
>  ---> Using cache
>  ---> d38067d54fc4
> Step 4/4 : RUN apt-get install -y libapache2-mod-xsendfile
>  ---> Running in 49966eb4db03
> Reading package lists...
> Building dependency tree...
> Reading state information...
> E: Unable to locate package libapache2-mod-xsendfile
> ERROR: Service 'apache' failed to build: The command '/bin/sh -c apt-get install -y libapache2-mod-xsendfile' returned a non-zero code: 100
> make[3]: *** [../make-base/docker-compose.mk:17: build] Error 1
> ```
> 
> Do you think there's a version conflict somewhere or is the image misconfigured, since it says it's a newer version of `httpd`? What version are you running?

That's, super odd.  I'm using an explicit version, so httpd's version did not change.  Unfortunately, I am installing xsendfile via apt with these two commands

```
RUN apt-get update
RUN apt-get install -y libapache2-mod-xsendfile
```

Which I suppose is non deterministic if the apt index is not working.  Try running in the root monoetna dir `docker-compose build --no-cache` just to see if there's some issue with that.

corps: Oh I see, yeah it says from your output that a newer image was downloaded.  That would mean they publish a new image for the existing version, which is odd...

PR #23, opened 2020-07-15
Metis client retry

coleshaw:
Could probably make this more robust, but I ran into a problem uploading, where VPN kicked me off and then I was getting an error about "Wrong byte position" for a file when doing a `put` with the Metis Client. I suspect this was the file I was in the middle of uploading when kicked off the VPN, so it was a partial upload and somehow the `current_byte_position` was corrupted. 

This PR instructs the client to send the `reset` flag one time to try and re-start the given upload. If the upload still fails, the client will not retry again and the user will have to investigate why the upload is failing for some other reasons.

WIP: Need to do more testing, and want to make the max number of retries configurable.


metis/bin/metis_client
```diff
@@ -134,11 +134,31 @@ class MetisClient
...
 
+  def retry_upload(upload_path, upload)
+    number_attempts = MetisConfig[:max_retries] || 1
+    for retry_number in 1..number_attempts do
+      response = json_post(upload_path,
+        file_size: upload.file_size,
+        action: 'start',
+        next_blob_size: upload.next_blob_size,
+        next_blob_hash: upload.next_blob_hash,
+        reset: true
```
corps, line 152: NIT:  Can we only call reset explicitly on a 422 response?  If we get a temporary 503, say, from an apache restart or something, it'd suck to have to restart a 200gb file that could have just been retried.

corps: The 422 will come from blob_upload, so this would require more state handling.  If it's too much trouble don't worry.

coleshaw: That's a good idea. I think I'll have to add the reset-call to blob_upload anyways, since the error doesn't just happen on upload_start.

coleshaw:
Changing this out of WIP status, I think it works. When testing locally, if I kill the server or ctrl-c the upload process via the client, when I re-start up the client the upload will 1) fail, 2) reset, and 3) retry. On the retry it will pick back up again.

One thing I added was a `delete_partial!` method on the upload. I realized that if the database was out of sync with the temporary file on disk, you would never be able to reset because of the mismatch in `current_byte_position`. That would lead to forever a `Wrong byte position` exception in `upload_controller`. So it seems that on file reset, we also need to remove the partially uploaded blob. Not sure if that will lead to excessive re-uploading of data, but it does allow you to re-continue a corrupted upload.

PR #24, opened 2020-07-17
Zc/timur attribute filter ui

corps:
UI Work for the attribute filtering.

1.  Adds column on left side to fit the attribute list
2.  Introduces a TreeView component that can handle both the simple case (a list of options) and potential future use cases (such as grouped attributes, or other tree view pages).

Does not hook it up to state, will sync with @coleshaw on how to approach that.


corps:
One goal of this PR, too, was to introduce style-loader and css that sits next to components.  Especially useful for etna-js components that may be shared and want to come with their own (optional) css parts.

PR #25, opened 2020-07-17
Integrate attribute filter

coleshaw:
This integrates the work from #24 with the attribute state work.

New behaviours:
* In Timur Search, after you select a Table (model) from the dropdown menu, a list of attributes should appear on the left of the screen. By default all of the attributes should be checked.
* User can de-select / select from the set of attributes.
* Clicking the `Search` button fetches only the indicated attributes.
* When results are displayed on the page, selecting / unselecting attributes does not change the data displayed in the table.
* Clicking Search or navigating the pages in the table (via `<` or `>` buttons) will fetch new data for cached pages, if a new set of attributes has been selected. If the set of attributes does not change, then page caching works as before (navigating to a previously fetched page does not re-fetch the data, except after navigating back to page 1).
* Selecting a new model should update the list of attributes, with all attributes checked.

I should add more tests around this...but I think this is good enough to get feedback on.


coleshaw:
@corps , this includes an update to the `etna-js` test that was failing ... it looked like the `lastSelectionsByCallback` method is supposed to pick all leaf nodes, but the test was missing one from the expected list. If I mis-interpreted, please let me know.

etna/packages/etna-js/components/__tests__/TreeView.spec.jsx
```diff
@@ -0,0 +1,193 @@
...
+            ['Top Level 1', 'Inner 2'],
+            ['Top Level 1', 'Inner 2', 'Inner Inner 1'],
+            ['Top Level 1', 'Inner 2', 'Inner Inner 2'],
+            ['Top Level 1', 'Inner 2', 'Inner Inner 3'],
+            ['Top Level 1', 'Inner 3'],
+            ['Top Level 2']
+          ]);
+
+          expect(test.lastSelectionsByCallback).toEqual([
+            'Inner 1',
```
coleshaw, line 182: This test was failing, so I added this node because it seems like a leaf and should be included in the results from `lastSelectionsByCallback`?

magma/lib/magma/attributes/collection.rb
```diff
@@ -7,7 +7,7 @@ def query_to_payload(link)
     end
 
     def query_to_tsv(value)
-      query_to_payload(value).join(", ")
+      value.join(", ")
```
coleshaw, line 10: This fixes a bug in Magma and should already be in `master`. Pulled in because of `bin/pull-subtrees` I think.

timur/jest.config.js
```diff
@@ -0,0 +1,19 @@
...
+    CONFIG: {
+      project_name: 'labors'
+    }
+  },
+  testURL: 'http://www.fake.com',
+  transformIgnorePatterns: ['node_modules/(?!(etna-js)/)'],
+  moduleNameMapper: {
+    '^service-worker-loader!': '<rootDir>/__mocks__/service-worker-loader.js',
+    '^.*[.](css|CSS)$': 'identity-obj-proxy',
+    '^react$': '<rootDir>/node_modules/react'
```
coleshaw, line 13: Pulled the `jest` config into its own file per current recommended practices. Included an entry in `moduleNameMapper` for `react`, to avoid the "Hooks" error.

timur/lib/client/jsx/components/model_viewer.jsx
```diff
@@ -1,49 +1,63 @@
...
-        Header: att_name,
-        accessor: att_name,
-        Cell: ({value, original}) => <AttributeViewer
-          template={template}
-          document={original}
-          value={value}
-          attribute={ template.attributes[att_name] }
-        />
-      }
+  const display_attribute_names = restricted_attribute_names
```
coleshaw, line 27: Add a parameter to `model_viewer` to only display certain attributes from the model template.

timur/lib/client/jsx/components/search/search.jsx
```diff
@@ -2,138 +2,330 @@ import Pager from '../pager';
...
+  ['Top Level 2']
+];
+
+const spinnerCss = css`
+  display: block;
+  margin: 2rem auto;
+`;
 
-class Search extends Component {
+export class Search extends Component {
```
coleshaw, line 41: Export the raw component (unwrapped with `redux` `connect`), to make some tests a little bit easier to do.

PR #26, opened 2020-07-20
Issue193/magma retrieve question error

coleshaw:
This PR addresses mountetna/magma#193. It catches a `Magma::QuestionError` in the `/retrieve` endpoint, which can be thrown when no results are returned for a paginated query (hence the requested page cannot be found). Also adds a test for the condition.


magma/lib/magma/attributes/collection.rb
```diff
@@ -7,7 +7,7 @@ def query_to_payload(link)
     end
 
     def query_to_tsv(value)
-      query_to_payload(value).join(", ")
+      value.join(", ")
```
graft, line 10: This addresses an unrelated issue that we discovered the other day?

coleshaw: Yeah, this PR should really only be 1 commit and not include this code change, but I'm not sure how to get it looking clean. This change is already merged into [monoetna master](https://github.com/mountetna/monoetna/blob/master/magma/lib/magma/attributes/collection.rb). I think it's some issue with my workflow, since this change was in the standalone repository and had to get pulled in via `pull-subtrees`. I'll play around with it and see if I can get this to reflect just the one chnage.

coleshaw:
@graft , fixed it, I think because the other subtree changes were merged in via another PR, so merging in master made this cleaner.

coleshaw:
Closing this cruftiness.

PR #27, opened 2020-07-20
Filter actions

coleshaw:
@corps , here is the current state of the filter actions I've implemented. I'm assuming (hoping?) that the UI components you generate have a way of sending state of three items: the attribute, the operator, and the value. i.e. for a filter like `ice_time>0.5`, the attribute would be `ice_time` (column name), the operator is `>`, and the value is `0.5`. I am then massaging those into a single string to send to Magma. If it is easier to just send out the string representation of the entire filter parameter, let me know and I'll tweak the actions.

Thanks!


corps:
Perfect, I'll start building off this!  @coleshaw

PR #28, opened 2020-07-22
WIP Timur filters ui

corps:
WIP!  Still working on this, but getting gloser.

Refactored the Search Page out into a functional component, and broke up the search parts into separate components to reduce complexity.  Added some generalized utilities and improved UI experience.

Still needs the actual filter generator UI piece, but getting close.


coleshaw:
Looks nice! I aspire to write code this elegant ... 

Small nits:

* When you switch tables, it still shows the text `Found 1 records in` -- should that be removed from the UI when a new table is selected? 
* The pagination controls on the results table have disappeared. I think this is just a flag that gets passed to `ModelViewer`
* If you select a table, and then click the toggle arrow on the left of the Search label (to hide the "add / remove attributes" control), the table SelectInput gets reset to `------`, but the `selectedModel` still seems present in state because I can search. I would expect that SelectInput selected option to not change.
* The TSV button doesn't seem to get the same `attribute_names` as the Search button. If I do a Document search and only select attributes "description" and "project", the table looks correct in the browser. But the TSV button also includes a "name" column, which it shouldn't.

corps:
More improvements on this.  The query builder works!  There are still some missing parts:

1.  Column ordering is still not consistent.  I get the feeling that we may need to create a selector that can transform the attribute_names == 'all' => a list, and then a selector that does the ordering.
2.  Pagination did disappear!  I need to take a closer look.
3.  Yes, so the SelectInput component did not accept a `value` input initially, so it was only showing mounted state.  I need to hook up selectedModel to the value prop of SelectInput.  That's why it seems to visually reset when it collapses.
4. TSV button issue.
5.  Overall, style could be improved, but I got the layout and basic feel right.  I think text / minor adjustments should come.
6.  Currently no way to switch back to advanced search.  State exists, just don't have a component to swap -- defaults to the query builder only.

coleshaw:
Looks pretty good, I think we should merge, deploy to staging, and get feedback on it.

PR #30, opened 2020-07-24
Magma metis communication

coleshaw:
In order for Magma to be able to talk to Metis (for the file / image attribute linking feature in Timur), the Docker network has to be reconfigured. This PR maps the `metis.development.local` hostname inside the Magma container to the Apache container's IP address, so that Magma can send requests to Metis.

Also adds the `rootCA.pem` to the Magma container so that Metis's SSL cert is recognized as valid.


corps:
I intend to implement a more comprehensive solution not requiring this manual configuration soon.  Thank you for this fix!

PR #31, opened 2020-07-24
Further Timur Search Improvements

corps:
1.  Added summary by column selection
2.  Columns now sorted (parent, identifier, other)
3.  Fix(?) for caching bugs.
4.  Prevent unselecting required columns (parent, identifier, etc).


coleshaw:
The fixed columns look nice, and the # of selected attributes is very helpful! I still see the problem with caching and changing columns / attributes, so I'll file that as an issue to address later.

Also, one nit that could also be addressed later with the other feedback: the CSS difference between the unselectable columns and the selectable ones in the modal was pretty subtle to my eyes, it would be nice to have a stronger differentiation.

corps:
@coleshaw Good call on the caching issue.  I'd love to spend more time to write an e2e test for this, but I don't think I can get that in.  I'll quickly try changing it again to see if I can fix it by hand.

coleshaw:
@corps , I don't think we should let the caching issue block this PR..

corps:
> Looks good to me. I think the small nit from standup today was to show `identifier` as the first column in search results and `parent` as second column.
> 
> I think the other things I commented are lower priority for this week.

Got it, just added in last commit.

corps:
> @corps , I don't think we should let the caching issue block this PR..

I managed to fix it anyways lol.

> @corps , I don't think we should let the caching issue block this PR..

You are probably totally right, I incidentally ended up with a minor fix based on your original approach 
https://github.com/mountetna/monoetna/pull/31/commits/e66e232bc832cd6febd5e81718833fdf2bb43837#diff-5224f4e3a28e21161364afc236b1a014

PR #32, opened 2020-07-27
Order timur view

coleshaw:
In the default Timur `Browse` view, this PR attempts to sort attribute records by natural sort if they are Arrays of strings. This results in a view like:

Screenshot from 2020-07-27 10-36-03

Any string or non-array attribute values should be unchanged.


timur/lib/client/jsx/components/browser/browser_pane.jsx
```diff
@@ -3,51 +3,86 @@ import * as React from 'react';
...
-            record={record}
-            value={ record[attribute_name] }
-            revised_value={ (attribute_name in revision) ?  revision[attribute_name] : record[attribute_name] }
-            attribute={pane.attributes[attribute_name]}
-            {...other_props}
-          />
-        }).filter(_=>_)
-      }
+const BrowserPane = ({record, revision, pane, ...other_props}) => {
+  var collator = new Intl.Collator(undefined, {
```
coleshaw, line 46: Adding a collator to use down below.

timur/lib/client/jsx/components/browser/browser_pane.jsx
```diff
@@ -3,51 +3,86 @@ import * as React from 'react';
...
+      )}
+      <div className='attributes'>
+        {Object.keys(pane.attributes)
+          .map((attribute_name) => {
+            return (
+              <PaneAttribute
+                key={attribute_name}
+                record={record}
+                value={
+                  Array.isArray(record[attribute_name])
```
coleshaw, line 68: If you ignore the formatting changes, this is the only change to the component logic.

graft:
Could this logic be moved into CollectionAttribute? Are there other circumstances in which this sort would apply?

coleshaw:
> Could this logic be moved into CollectionAttribute? Are there other circumstances in which this sort would apply?

I think `CollectionAttribute` is a good place to start where we can catch most of the other instances where this would be helpful.

timur/lib/client/jsx/components/attributes/collection_attribute.jsx
```diff
@@ -1,40 +1,51 @@
...
+}) => {
+  if (mode != 'edit') {
+    var collator = new Intl.Collator(undefined, {
+      numeric: true,
+      sensitivity: 'base'
+    });
+    return (
+      <div className='attribute'>
+        <div className='collection'>
+          {(value || []).sort(collator.compare).map((link) => (
```
coleshaw, line 26: Changed to sort before mapping.

timur/lib/client/jsx/components/attributes/collection_attribute.jsx
```diff
@@ -1,40 +1,51 @@
...
+  mode,
+  value,
+  revised_value,
+  document,
+  template,
+  attribute,
+  reviseDocument
+}) => {
+  if (mode != 'edit') {
+    var collator = new Intl.Collator(undefined, {
```
coleshaw, line 19: new Collator to use below in the sort method.

timur/lib/client/jsx/components/attributes/collection_attribute.jsx
corps: Nit:  This sort is a mutation on the original attribute, meaning that visiting this page could change the behavior of the state implicitly.  Would it be possible to
1.  sort in the reducer somehow
2.  make a copy and then sort?

```
const sortedCollection = useMemo(() => [...value].sort(collator.compare), [value]);
```

coleshaw:
I think this would be a little messier to try and sort in the reducer because all the different attribute types would be mixed in together there, and since it's just a presentational change I kept it in the `collection_attribute` component.

PR #33, opened 2020-07-27
Fixing Put command to actually use dir in last argument

corps:
Reported by lenny -- ensures that the folder path used as the last argument of put is actually the relative directory used for the upload.

### DIFF
--- a/metis/bin/metis_client
+++ b/metis/bin/metis_client
@@ -1,5 +1,5 @@
 #!/usr/bin/env ruby
-#
+
 require 'optparse'
 require 'resolv'
 require 'json'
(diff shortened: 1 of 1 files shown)


corps:
Also adds ability to ignore existing files.

coleshaw:
This would be a great feature -- I just got bit by this and had to recursively delete a bunch of folders. Kind of painful.

However, I also remember there was a discussion between @graft  and Arjun  about the target folder format?  Something about it should be a Metis path instead of a folder  ... ?

graft:
[Here](https://github.com/mountetna/metis/pull/68#issuecomment-631638340) we agreed on implementing it as `put <input files> <remote dir>`. [Here](https://github.com/mountetna/metis/pull/68/commits/4fea24aeacc5af7af7c39431b251fe18d9e0a481) is where I claim to support this behavior, but neglect to add a test proving this (I ran into incompatibility between the Etna::Auth and Etna::TestAuth layers and was too lazy to resolve them).

I also agree it would be nice if put would eventually respect a metis path in the place of (or rather in addition to) <remote dir>.

I am curious what use case requires the use of `put -x`, to preserve instead of overwrite a non-matching file size, any examples? Lenny was inquiring about how to prevent his puts from repeating uploads, but I think the existing logic ought to skip identical files on re-put; but it's not clear to me what kind of snafus can actually result in the wild.

corps:
> I am curious what use case requires the use of `put -x`, to preserve instead of overwrite a non-matching file size, any examples? Lenny was inquiring about how to prevent his puts from repeating uploads, but I think the existing logic ought to skip identical files on re-put; but it's not clear to me what kind of snafus can actually result in the wild.

I'm actually not sure, to be honest.  I can remove it.  I didn't get many other details, possibly Lenny isn't aware of the current behavior?

PR #35, opened 2020-07-28
Issue#94 add pauseUpload action and tie to the worker

coleshaw:
This PR hooks the `UploadControl` component's Pause button into the upload dispatcher, and addresses mountetna/metis#94. Before this change, the pause button was not dispatching actions, and it was not possible to pause uploads.

To test, you can verify that the pause button now pauses large uploads when clicked, and the play button that replaces it continues the upload.

### DIFF
--- a/etna/packages/etna-js/upload/actions/upload_actions.jsx
+++ b/etna/packages/etna-js/upload/actions/upload_actions.jsx
@@ -1,31 +1,42 @@
-import { postAuthorizeUpload } from '../api/upload_api';
+import {postAuthorizeUpload} from '../api/upload_api';
 import {errorMessage, message as showDialog} from './message_actions';
-import {AddUploadCommand, CancelUploadCommand, UnpauseUploadCommand, Upload} from "../workers/uploader";
+import {
+  AddUploadCommand,
+  CancelUploadCommand,
+  UnpauseUploadCommand,
+  Upload,
+  PauseUploadCommand
+} from '../workers/uploader';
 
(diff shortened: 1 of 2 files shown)


etna/packages/etna-js/upload/actions/upload_actions.jsx
```diff
@@ -34,45 +45,72 @@ export const fileSelected = ({ file, folder_name, bucket_name }) => (dispatch, g
...
-}
+};
 
-export const unpauseUpload = ({ upload }) => (dispatch) => {
+export const unpauseUpload = ({upload}) => (dispatch) => {
   dispatchUploadWork(dispatch, UnpauseUploadCommand(upload));
-}
+};
+
+export const pauseUpload = ({upload}) => (dispatch) => {
```
coleshaw, line 94: This is the only method added to the actions for this PR.

metis/webpack.config.js
```diff
@@ -11,14 +11,20 @@
...
   context: path.resolve(__dirname),
   resolve: {
     extensions: ['.js', '.jsx'],
     alias: {
-      'font-awesome': path.join(__dirname, 'node_modules/@fortawesome/fontawesome-free')
+      'font-awesome': path.join(
+        __dirname,
+        'node_modules/@fortawesome/fontawesome-free'
+      ),
+      react: path.join(__dirname, 'node_modules/react'),
```
coleshaw, line 25: Added these three aliases to prevent multiple React instances from being found.

PR #37, opened 2020-07-28
Re-enable bin/console

graft:
The removal of Magma::Client in favor of Etna::Client resulted in the loss of `bin/console`, which was still trying to setup the magma client and would crash. Removed here.


coleshaw:
Looks good, sorry I missed that when I took out Magma Client.

PR #38, opened 2020-07-28
Fixing inability to download root folder

corps:
Pretty simple fix here.


graft:
mysterious... how does it work?

coleshaw:
So amazingly, this worked for me. I'm curious as to how that simple change made the bucket download work, too. I'm guessing it just lists all files and folders inside the bucket?

One question, is it supposed to assign the ZIP file the name of the bucket? I got ".zip", but it would be nice if it were `<bucket-name>.zip`.

corps:
The fix is that `folder_name` is actually undefined for the root folder, so I gave it a default value.

I could include the bucket name instead in this case, I'll see if I can thread the state through then.

PR #39, opened 2020-07-29
Support whitespaces escaped in magma retrieve

corps:
1. Support whitespace in retrieve endpoints, escaped with '_'.  Supports escaping '_' as well to preserve ability to use those.
2. Advanced search shows column filtering

### DIFF
--- a/magma/lib/magma/retrieval.rb
+++ b/magma/lib/magma/retrieval.rb
@@ -183,6 +183,8 @@ class Magma
         match, att_name, operator, value = term.match(FILTER_TERM).to_a
         raise ArgumentError, "Filter term '#{term}' does not parse" if match.nil?
 
+        value = unescape(value)
+
         att = attributes.find{|a| a.name == att_name.to_sym}
         raise ArgumentError, "#{att_name} is not an attribute" unless att.is_a?(Magma::Attribute)
         case att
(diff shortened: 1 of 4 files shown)


graft:
A consideration: Most of this "escaping" problem results from jamming the filter into a single string, which is really not necessary. we could instead submit the filter as a JSON array of terms, [ { attribute_name, operator, value } ]. This is the first thing Magma already does when it receives the string; the filter could easily be made to recognize it is receiving an array instead of a string and behave accordingly.

If we pass an array of terms there is no problem with spaces - a string containing spaces can be clearly delineated, and we can avoid defining our own escape syntax.

corps:
> A consideration: Most of this "escaping" problem results from jamming the filter into a single string, which is really not necessary. we could instead submit the filter as a JSON array of terms, [ { attribute_name, operator, value } ]. This is the first thing Magma already does when it receives the string; the filter could easily be made to recognize it is receiving an array instead of a string and behave accordingly.
> 
> If we pass an array of terms there is no problem with spaces - a string containing spaces can be clearly delineated, and we can avoid defining our own escape syntax.

True.  In which case, how does the advanced search work, does it only allow you to edit the first of the predicate terms?

graft:
If by "advanced search" you mean the old-school filter input string (and not /query on magma), you could have the endpoint continue to support the space-delimited string syntax and just pass the filter input string as given to this. It wouldn't allow you to put in spaces, but you can use `.` anyway which is ok.

PR #42, opened 2020-07-30
Set the filename for the tsv response

corps:


graft:
Could we compute something more informative for the default filename? e.g. `#{project_name}-#{model_name}-#{date}.tsv`

PR #45, opened 2020-07-31
Update etna gem for testing

coleshaw:
While trying to add more MetisClient tests, I ran into a couple issues with Etna that made it difficult to execute the Metis tests. This PR addresses those. Two changes:

1) `Etna::TestAuth` will validate both "test" signatures as well as "real" HMAC signatures. Previously only a "test" signature (the string `'valid'`) was accepted. This now allows testing of metis_client PUT situations where a "real" HMAC signature is in the workflow.
2) `Etna::ParseBody` will gracefully update params for a `multipart/form-data` request with an empty body. This should only every happen in a test scenario. Previously it would throw an exception and break the test, but this at least lets the test continue. It should not affect non-test behavior.

**NOTE:** I'm putting in this PR separate from the Metis changes, because the updated Metis tests require an updated, published Etna gem. So unfortunately it's not possible to do them in a single PR.

### DIFF
--- a/etna/etna.gemspec
+++ b/etna/etna.gemspec
@@ -1,6 +1,6 @@
 Gem::Specification.new do |spec|
   spec.name              = 'etna'
-  spec.version           = '0.1.13'
+  spec.version           = '0.1.14'
   spec.summary           = 'Base classes for Mount Etna applications'
   spec.description       = 'See summary'
   spec.email             = 'Saurabh.Asthana@ucsf.edu'
--- a/etna/etna.gemspec
+++ b/etna/etna.gemspec
@@ -1,6 +1,6 @@
 Gem::Specification.new do |spec|
   spec.name              = 'etna'
-  spec.version           = '0.1.14'
+  spec.version           = '0.1.13'
   spec.summary           = 'Base classes for Mount Etna applications'
   spec.description       = 'See summary'
   spec.email             = 'Saurabh.Asthana@ucsf.edu'
--- a/etna/lib/etna/parse_body.rb
+++ b/etna/lib/etna/parse_body.rb
@@ -23,7 +23,7 @@ module Etna
         )
       when %r{multipart/form-data}i
         params.update(
-          Rack::Multipart.parse_multipart(env)
+          Rack::Multipart.parse_multipart(env) || {}
         )
       end
       # Always parse the params that are url-encoded.
--- a/etna/lib/etna/test_auth.rb
+++ b/etna/lib/etna/test_auth.rb
@@ -74,7 +74,7 @@ module Etna
   end
   class TestHmac < Hmac
     def valid?
-      @test_signature == 'valid'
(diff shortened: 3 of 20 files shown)


etna/lib/etna/parse_body.rb
```diff
@@ -23,7 +23,7 @@ def call(env)
         )
       when %r{multipart/form-data}i
         params.update(
-          Rack::Multipart.parse_multipart(env)
+          Rack::Multipart.parse_multipart(env) || {}
```
coleshaw, line 26: When there is no multipart data (but the `multipart/form-data` header is included), this part would throw an exception trying to update the parameters because `Rack::Multipart.parse_multipart()` would return `nil`. Really that should only happen in a test environment, but this addition lets the process continue cleanly.

etna/lib/etna/test_auth.rb
```diff
@@ -74,7 +74,7 @@ def approve_hmac(request)
   end
   class TestHmac < Hmac
     def valid?
-      @test_signature == 'valid'
+      @test_signature == 'valid' || super
```
coleshaw, line 77: This lets the `Etna::TestAuth` middleware also validate that "real" HMAC signatures are valid. The scenario happens when a Metis-signed HMAC URL is used in a test -- i.e. when POSTing to the `upload_url` returned by `authorize_upload`. Without this change, `valid?` will always return `false` in such a scenario. This change lets `TestAuth` handle test HMAC signatures as well as "real" ones.

etna/lib/etna/parse_body.rb
graft: Hmm why is this only possible in a test environment?

coleshaw: I guess I would only expect it to happen in a test environment (or if someone intentionally constructs such a message). I wouldn't expect something like that to happen within one of our applications though.

etna/etna.gemspec
```diff
@@ -1,6 +1,6 @@
 Gem::Specification.new do |spec|
   spec.name              = 'etna'
-  spec.version           = '0.1.13'
+  spec.version           = '0.1.14'
```
coleshaw, line 3: Bump the version to get automatic gem publishing from the GH action.

etna/spec/controller_spec.rb
```diff
@@ -54,7 +54,7 @@ class Server < Etna::Server; end
       @app = setup_app(
         Arachne::Server,
         [ Etna::TestAuth ],
-        test: { log_file: @log_file }
+        test: { log_file: @log_file, hmac_keys: {etna: 'key'} }
```
coleshaw, line 57: Because of the call to `super` in `Etna::TestAuth#valid?`, we now have to provide an `hmac_keys` config for tests so that `Etna::Auth#valid?` does not throw an exception.

PR #46, opened 2020-08-03
add some tests around Metis Client...

coleshaw:
This PR adds some testing around the Metis Client. Ideally the tests would be more ignorant of the underlying implementation, but there are some challenges to testing Metis Client:

1) Without Apache X-Sendfile module in the loop, any GET request for a file does not actually send the file back. So the file is created by the client, but with 0 bytes.
2) [WebMock does not support multipart/form-data bodies](https://github.com/vcr/vcr/issues/295#issuecomment-20181472), either in stubbing or in mocking Net HTTP Post requests. So we can't Post blobs to the Metis API in CI testing right now, so the new test just makes sure that the APIs are called correctly.

So this PR is meant as a lets-have-something testing step, that should help us catch some Metis Client bugs. But to make it better we would probably have to run smoke-tests on a server somewhere instead of via CI.

### DIFF
--- a/metis/Gemfile.lock
+++ b/metis/Gemfile.lock
@@ -22,7 +22,7 @@ GEM
     debase-ruby_core_source (0.10.9)
     diff-lcs (1.3)
     docile (1.3.2)
-    etna (0.1.12)
+    etna (0.1.14)
       jwt
       multipart-post
       net-http-persistent
--- a/metis/bin/metis_client
+++ b/metis/bin/metis_client
@@ -710,13 +710,16 @@ EOT
           break
         rescue MetisClient::UploadError => e
           puts "#{e.message}"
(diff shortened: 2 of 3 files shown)


corps:
> So this PR is meant as a lets-have-something testing step, that should help us catch some Metis Client bugs. But to make it better we would probably have to run smoke-tests on a server somewhere instead of via CI.

So, I am quite close to having an e2e all services running (including apache) integration testing setup working for CI.  We can soon actually have smoke tests running on server-like-CI in the near future.

coleshaw:
> So, I am quite close to having an e2e all services running (including apache) integration testing setup working for CI. We can soon actually have smoke tests running on server-like-CI in the near future.

That will be awesome!

PR #47, opened 2020-08-04
Metis Thumbnails

coleshaw:
This PR enables thumbnail support in Metis for PNG, JPEG, and TIFF files via a `thumbnail=<anything>` URL param when used with the `download_file` route. Thumbnails are generated on the fly and served to the client. I've hardcoded the thumbnail width to be 240px, and the library keeps the aspect ratio.

Not sure if there are other image types we should support?

It also updates the `image_attribute` component in Timur to take advantage of this feature and shows it in the UI, like the previous behavior.

The Metis feature relies on the [vips](https://github.com/libvips/libvips) library, and so there is a related update to mountetna/cbc-chef#18.

### DIFF
--- a/metis/docker/app/Dockerfile.development
+++ b/metis/docker/app/Dockerfile.development
@@ -18,6 +18,8 @@ RUN apt-get update \
       libopenblas-base \
       libcurl4-openssl-dev \
       libssl-dev \
+      libvips \
+      libvips-dev \
       parallel \
       postgresql-client-10 \
       --no-install-recommends &&\
--- a/metis/docker/app/Dockerfile.development
+++ b/metis/docker/app/Dockerfile.development
@@ -19,7 +19,6 @@ RUN apt-get update \
       libcurl4-openssl-dev \
       libssl-dev \
       libvips \
-      libvips-dev \
       parallel \
       postgresql-client-10 \
       --no-install-recommends &&\
--- a/metis/lib/server/controllers/download_controller.rb
+++ b/metis/lib/server/controllers/download_controller.rb
@@ -14,17 +14,24 @@ class DownloadController < Metis::Controller
 
     raise Etna::Error.new('File not found', 404) unless file && file.has_data?
 
-    puts @params
     if @params.fetch(:thumbnail, nil)
-      puts file.data_block.location
       thumbnail = Vips::Image.thumbnail(file.data_block.location, 240)
-      puts thumbnail
-      puts thumbnail.filename
-      puts MimeMagic.by_path(file.file_name)
+      mimetype = MimeMagic.by_path(file.file_name)
+
+      case mimetype.to_s
+      when /jpe?g$/
+        buffer = thumbnail.jpegsave_buffer
+      when 'image/tiff'
+        buffer = thumbnail.tiffsave_buffer
+      when 'image/png'
+        buffer = thumbnail.pngsave_buffer
+      else
+        raise Etna::Error.new("Thumbnails not supported for mimetype #{mimetype}", 422)
+      end
       return [
         200,
-        { 'Content-Type' => MimeMagic.by_path(file.file_name) },
-        [ thumbnail.tiffsave_buffer ]
+        { 'Content-Type' => mimetype },
+        [ buffer ]
       ]
     end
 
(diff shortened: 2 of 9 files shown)


metis/lib/server/controllers/download_controller.rb
```diff
@@ -11,10 +14,34 @@ def download
 
     raise Etna::Error.new('File not found', 404) unless file && file.has_data?
 
+    if @params.fetch(:thumbnail, nil)
+      thumbnail = Vips::Image.thumbnail(file.data_block.location, 240)
+      mimetype = MimeMagic.by_path(file.file_name)
+
+      case mimetype.to_s
+      when /jpe?g$/
+        buffer = thumbnail.jpegsave_buffer
```
corps, line 23: Where do these buffers get saved btw?  Is the buffer consistent per path or unique on each call?

corps: Or, is the thumbnail entirely in memory?

graft: It looks like Rack streams the object (see line 34 below), so it's all in memory. This leads me to wonder how much memory this uses and how efficient the process is, as some of these TIFF and JPEG files can get quite large (if they are showing an H&E stain of a tumor at high magnification e.g.)

coleshaw: It's entirely in memory. I'm not sure how to measure (tips?), but I manually tested with a 1GB TIFF file, and it streamed back really quickly, and I would say it's not obvious that it's generating the thumbnail on the fly. vips is more performant than ImageMagick (you can check [their benchmarks](https://github.com/libvips/libvips/wiki/Speed-and-memory-use)).

You can test it on timur-test-1.ucsf.edu, just upload a giant image to metis-test-1.ucsf.edu...

coleshaw: Of course, testing it locally, it's highly dependent on my laptop's specs. Not sure how those compare to the VMs.

corps: The transformation may itself be streamable, which means it wouldn't require much memory.  It would just read chunks from disk, do the transform for that chunk, and pass it over the wire before continuing.

That said, it _could_ be CPU intensive if this is happening for many thumbnails without any caching.  Is it possible for us to atleast set an ETag and observe If-None-Match to ensure clients atleast can attempt to cache the resource?
We could use the basename of the chunk (it's computed md5) as the unique identifier.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag

corps: Oh you know what, it isn't streaming _from the source to the network_.  The buffer contains the full transform, after which it streams the buffered response.  That means that it probably _can_ be CPU intensive, especially since it does the work eagerly.

graft: Of note: the search page on timur will render thumbnails, so this might trigger 200 x number-of-image-columns simultaneous thumbnail requests.

corps:
One other q: do we need to add installing vips to the Dockerfile as well?

metis/lib/server/controllers/download_controller.rb
coleshaw: Oh, I'll test that...200 would be a lot to process on the fly.

coleshaw:
That change should be included in this PR already.

metis/docker/app/Dockerfile.development
```diff
@@ -18,6 +18,7 @@ RUN apt-get update \
       libopenblas-base \
       libcurl4-openssl-dev \
       libssl-dev \
+      libvips \
```
corps, line 21: Got it, thanks!

coleshaw:
You two were both right. Definitely CPU intensive when dealing with lots of images, and with a 200 page size, it's not very manageable. So I'll rethink this, maybe the "generate thumbnail" route with Magma or Timur making the request is the way to go, or having Metis automatically generating thumbnails for any uploaded image files, on upload_complete.

graft:
I think this method is still viable if 1) the route caches to disk and 2) the cache can be pre-populated by a lazy worker (similar to md5 computation etc.).

coleshaw:
Yeah, those are good ideas. Basically we provide our own image caching layer. I'll re-open this when I have a chance to try that.

PR #48, opened 2020-08-06
add command to delete orphaned data blocks

coleshaw:
I think a "delete orphaned data blocks" command might be helpful as we transition to the new Magma-Metis data deletion scheme, and maybe as a regular garbage-collection task to run. There are already orphaned data blocks that we should remove and not back up (see Kanban card). 

This PR adds a Metis command to detect and remove orphaned data_blocks, where orphaned means the data_block is not referred to by any Metis::File. It does not delete the zero-byte data_block, which is a generic pointer (like `/dev/null`) and could have no file references at any given moment though we don't want to remove it.

The new behavior is that:

1) The data_block is removed from disk.
2) The Database entry data_block is kept (with corresponding md5_hash), but has a `removed` boolean flag set to `true` in the database. This can then be used for auditing the data deletion and preventing re-uploads.

Also adds some tests around the commands (copied from Magma).

I think the new Magma -> Metis annotation scheme will overtake this in time and may make this command irrelevant once all the data blocks are clean and consistent.

### DIFF
--- a/metis/bin/delete-orphan-data-blocks
+++ b/metis/bin/delete-orphan-data-blocks
@@ -1,0 +1,8 @@
+#!/bin/bash
+
+# Should put this in CRONTAB?
+source ~/.bashrc
+
+cd $(dirname `dirname "$0"`)
+echo $(date)
+bin/metis delete_orphan_data_blocks
--- a/metis/bin/remove-orphan-data-blocks
+++ b/metis/bin/remove-orphan-data-blocks
@@ -5,4 +5,4 @@ source ~/.bashrc
 
 cd $(dirname `dirname "$0"`)
 echo $(date)
-bin/metis delete_orphan_data_blocks
+bin/metis remove_orphan_data_blocks
--- a/metis/config.yml.template
+++ b/metis/config.yml.template
@@ -29,7 +29,6 @@
 
 :development:
   :log_file: log/error.log
-  :md5_log_file: log/md5_deletions.log
   :auth_redirect: https://janus.development.local
   :token_name: JANUS_DEV_TOKEN
   :token_algo: RS256
--- a/metis/config.yml.template
+++ b/metis/config.yml.template
@@ -1,7 +1,7 @@
 ---
 :test:
   :log_file: /dev/null
-  :secret_key: 
+  :secret_key:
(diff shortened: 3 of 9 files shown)


corps:
Makes sense.
Would it make sense to also collect an md5 manifest log of deleted orphaned blocks, just for audit purposes?  Such a log could be backed up safely and is in no way PI or sensitive.

coleshaw:
That's a great idea, thanks! I'll add that in.

graft:
Rather than sticking this output in the logger, which is somewhat impermanent, may we instead simply delete the data, update description, and set an empty hash value, leaving the empty block in the table?

graft:
This also doesn't seem to destroy the actual data on disk.

coleshaw:
Hm, let me fix the disk issue, I didn't realize `delete` didn't hook into a filesystem action, sorry.

> Rather than sticking this output in the logger, which is somewhat impermanent, may we instead simply delete the data, update description, and set an empty hash value, leaving the empty block in the table?

Would just setting the md5_hash value to empty string remove our ability to do an audit of what was deleted (which was the goal of logging the deleted blocks)? Would it make sense to have a boolean flag in `DataBlock` instead, to record that it was deleted from disk, but keep the hash present?

graft:
Yeah... An added benefit of the "removed" flag could be gatekeeping re-insertion of the same data.

coleshaw:
Oh, that's a great idea!

Should we set this up to run on a cron schedule, less frequent than the archiving?

If this PR looks okay up to the removal command, I'll like to put the gatekeeping in a separate PR.

corps:
@coleshaw 
So I'm working on the s3 changes, and I realized a general concept that would be useful for both your problem (marking a datablock as "removed") and also for my problem (marking the archiving method for a datablock based on historical associated projects) would be to associate "metadata" with data blocks. 

I'm working on a PR that adds this concept in general.  What if we shared that data model to solve both our problems?

coleshaw:
Sure! I'll hold off on merging this then, and we can look at the block metadata PR first.

coleshaw:
Going to merge this in per comments in #50 -- we'll keep `removed` as an explicit attribute on `DataBlock`.

PR #49, opened 2020-08-06
Zc/docker build improvements

corps:
A big push on moving to docker deployments, while also significantly improving the docker development experience.

Goals:
1.  Use base images and builds to reduce the massive disk space overhead of default docker-compose behavior (which has unique images for every project, even though the common parts do not need duplication).
2.  Reduce the degree to which excessive builds are executed, leverage better docker caching.
3.  Greatly increase docker test speeds in CI by pushing and pulling cached images
4.  Fully build self contained images for each application that in theory could be deployed.  Hook up CI to build and push these (even if they aren't being deployed yet).
5.  apache should server app assets like in production, rather than relying on the rack middleware.
6.  Common parts of docker configuration to all apps should be centralized, not duplicated.
7.  etna gem now auto updates from monorepo and changes Gemfiles to point to github refs.  This means that both etna-js and etna can be developed and pushed in the same commit as downstream consumers of those changes (!!). 

I recognize that I probably spent too much time into this, so I plan to move my priorities back to the metis and follow up work.  I really wanted to push on this, because the design decisions here will help me follow up on the documentation I am working re proposal for docker deployment.  My current tech spec is more about limitations of our current docker and less on the specifics of moving towards docker deployment, because I hadn't overcome this hump.

Seperately, I have a different branch that proves using chef + kitchen to deploy docker to CentOS docker images themselves; the result would be the ability to fully reproduce a deployment locally and even have CI test chef changes.  Once that is mature,  we can be ready to attempt a docker deployment when we have spare cycles.


archimedes/Gemfile
```diff
@@ -1,7 +1,7 @@
 source 'https://rubygems.org'
 
 # etna application/server gem
-gem 'etna'
+gem 'etna', git: 'https://github.com/mountetna/monoetna.git', ref: ''
```
corps, line 4: Going to remove this, was an aborted attempt to modify archimedes, but I decided to scope changes to just `timur` `metis` `janus` and `magma` for now.

corps:
Notably, all though these changeset looks big, it includes 2/3 of its size __in deletes__.

There is actually a huge simplification happening here.  I'll start annotating with comments whats going on.  I also intend to update and improve README.

.github/workflows/branches.yml
```diff
@@ -12,56 +12,36 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - id: cache-packages-volumes
```
corps, line 15: This was an old attempt at caching in github actions to improve circle ci performance.  It was simply WAY too complicated for the value it was adding.  The new approach is simpler: explicitly build images that cache from existing ones in dockerhub that we push at the end of our build process.  There will be some downloading overhead, but this downloading will be bundled, composed, and smaller than the individual processes that were running before.

.github/workflows/branches.yml
```diff
@@ -12,56 +12,36 @@ jobs:
...
-      - run: cd docker && docker-compose -p monoetna push || true
-        if: steps.cache-docker.outputs.cache-hit != 'true'
-        name: Push images to cache location
-      - name: Load volumes from cache
-        run: |
-          docker run --rm -v /tmp:/backup/ `docker volume list -q | egrep -v '^.{64}$' | egrep 'node-modules|gems' | awk '{print "-v " $1 ":/mnt/" $1}'` alpine tar -xf /backup/packages-volumes.tar -C /mnt
-        if: steps.cache-packages-volumes.outputs.cache-hit == 'true'
       - name: Run metis test suite
         id: metis_first_test_run
+        env: &release-env
```
corps, line 17: This will instruct the `build_image` and related build scripts to build / cache from a `master` image in the dockerhub `etnaagent` repo which I have established (free, etna-agent@ucsf.edu email).

.github/workflows/branches.yml
```diff
@@ -12,56 +12,36 @@ jobs:
...
+          PULL_IMAGES: 1
+          IMAGES_PREFIX: etnaagent/
+          IMAGES_POSTFIX: :master
         run: |
-          make -C metis test
+          make -C metis release
```
corps, line 22: I now distinguish between `test` and `release` targets.  `test` is meant for quick iteration with docker-compose.  It does not build finished images outright, but relies on the docker-compose to pull in code via volumes.  That is much faster for development, but less sincere for "testing" the release worthiness of the images.  Thus, I've added this new target `release` which directly involves built images.

.github/workflows/master.yml
```diff
@@ -35,10 +35,13 @@ jobs:
       - uses: actions/checkout@v2
         with:
           fetch-depth: 1
-      - name: Run smoke tests
-        env:
-          MASTER_BUILD: 1 # Will cause tests to use the true package artifacts, not local source.
-        run: make test
+      - name: Test + Push Release
```
corps, line 38: Once we merge to master, we actually set PUSH_IMAGES to allow pushed our new image to dockerhub.  This will allow future builds to use it as a reference cache.  I don't PUSH on PRs since they tend to involve more rapid changes.  We only want to push "canonical" images that are actually a result of merged code.

This won't currently trigger a deploy anywhere, this is mostly here now for the purpose of CI build caching.

.gitignore
```diff
@@ -59,3 +59,4 @@ build-iPhoneSimulator/
 
 .docker-build-mark
 downzip-sw.js
+.npmrc
```
corps, line 62: In general, due to the unfortunate way in which `npm link` works currently, we are required to drop `npmrc` files that configure the directory for npm linking correctly that end up in the individual repositories.  We don't actually want these files to be committed, however, as they would confuse / break non docker usage of those apps.  I plan to write away this problem in the future (switch to yarn over npm), but this is my stopgap.

Makefile
```diff
@@ -12,60 +12,63 @@ help: ## Display help text
...
-restart: ## Restarts running services for a project
-				@ echo Run this within a specific app context, ie: make -C metis restart
+	@ echo Run this within a specific app context, ie: make -C janus migrate
 
 .PHONY: test
 test: ## Runs all projects' tests
-				@ set -e && for project in $(projects); do make -C $$project test; done
-
-.PHONY: setup-links
-setup-links: ## Sets up local development links for npm (and one day ruby) packages to use local copies of files.
```
corps, line 62: No longer necessary.  links are automatically handled behind the scenes without any interaction necessary from developers.

Makefile
```diff
@@ -12,60 +12,63 @@ help: ## Display help text
...
-				@ for project in $(projects); do make -C $$project setup-links; done
+	@ set -e && for project in $(projects); do make -C $$project test; done
 
-.PHONE: tag
-tag:  ## For ci, alters the project docker-compose.yml such that images contain specified repository tagging
```
corps, line 66: No longer necessary with new `build_image` approach.

Makefile
```diff
@@ -12,60 +12,63 @@ help: ## Display help text
...
-				@ make -C docker tag
+.PHONY: release
+release: ## Builds static docker images staged for release, runs tests against them, and pushes them to dockerhub (requires PUSH_IMAGES=1)
+	@ set -e && for project in $(projects); do make -C $$project release; done
 
 .PHONE: clean
-clean:  ## Drops all data volumes, wiping all development databases and caches, and rebuilds all docker images cleanly.
-				@ make -C docker clean
+clean:  ## Cleans many dangling docker references, recovering much disk space.
```
corps, line 67: Highly recommend running this after this branch hits master, as it will clear up a ton of wasted disk space from the way `docker-compose` creates unique images for each project (rather than sharing base images).

bin/child-npm-projects
```diff
@@ -1,35 +0,0 @@
-#! /usr/bin/env bash
```
corps, line 1: This whole file was never that useful anyways.  We can safely assume for now that all "shareable" npm projects exist in the `etna/packages` directory and be done with inferring it with logic like this.

bin/mk-dev-cert
```diff
@@ -0,0 +1,27 @@
+#! /usr/bin/env bash
```
corps, line 1: Little script to generate new development SSL certs using `mkcert` behind the scenes.  Notably ensures that we're always using the same rootCA for sanity purposes.

bin/update-package-references
```diff
@@ -19,11 +19,16 @@ DIR="$( cd -P "$( dirname "$SOURCE" )" && pwd )"
```
corps, line 19: This file is run in `master.yml` right now on merges to master.  Its job is to update package.json and Gemfile references to etna to specific commits to the monorepo instead.  This removes the need to upload to npm or ruby gems to issue a deploy.

bin/update-package-references
```diff
@@ -19,11 +19,16 @@ DIR="$( cd -P "$( dirname "$SOURCE" )" && pwd )"
 
 cd $DIR/..
 
-lastCommitted=$(git log --pretty=oneline --abbrev-commit | grep "Updated package.json references to artifacts" | head -n 1 | cut -d' ' -f1)
+git log --pretty=oneline --abbrev-commit | tail -n 10
```
corps, line 22: This logic is supposed to shortcircuit the work if no changes to `etna/` were found in the commits since the last `etna reference updates` commit.  It works, sometimes, but I've noticed it being over eager.  I've added these additional lines as a means to help debug that issue and plan to remove them when I've got this nailed down.

bin/update-package-references
```diff
@@ -44,43 +49,80 @@ function mkArtifactSha() {
     cp -r $DIR/../.git ./
     git checkout --orphan "$artifactName-$SHA"
     cp -r $orig/$targetDir/* ./
+    rm -rf packages || true
```
corps, line 52: When creating an artifact, the size of the resulting commit can be quite large if it also always includes the inner packages directory.  Removing this here is a simple reduction in artifact creation.

bin/update-package-references
corps: This will also, ALL GO AWAY once deployments switch to using docker itself (since the images will have already bundled etna alongside the code).  This exists primarily as a gap to allow our current deployment process to continue to run naively while we get the docker foundations in place.

bin/update-package-references
```diff
@@ -44,43 +49,80 @@ function mkArtifactSha() {
...
     git add .
     git commit -m "Artifact $artifactName for sha $SHA"
+#    git push origin HEAD --force
```
corps, line 55: Removing

development-certs/certs/timur.development.local-key.pem
```diff
@@ -0,0 +1,28 @@
+-----BEGIN PRIVATE KEY-----
```
corps, line 1: All development certs have been pulled out into their own top level directory, and are now baked into the common `etna-base` images.  This means we don't have to duplicate these in a few different places, as they are immediately shareable.

docker/.dockerignore
```diff
@@ -0,0 +1,17 @@
+src/node_modules
```
corps, line 1: This file is super important and will be worth documenting in my updated `docker/README.md`.

In order to keep docker images slim and prevent accidental development artifacts from slipping it (potentially with secrets), this file instructs docker to filter which files are copied into build images similar to gitignore and npmignore.

docker/development-psql/Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM postgres:10.12
```
corps, line 1: Variants exist now for postgres 10 and 9.  Timur and magma use postgres 10 right now just so that they have access to the db_seeds generated from the vm (which used postgres 10). 

We'll, in the future, need to carefully upgrade  postgres across all apps to 10.

docker/docker-compose.etna-app.shared.yml
```diff
@@ -0,0 +1,90 @@
+
```
corps, line 1: One of the nice benefits of the changes I've made here include being able to extract the common docker-compose.yml logic that had previously been duplicated many places into a common shared file.

Technically, this still gets expanded explicitly into each project's yml file (see default_compose script) through a merge process, we are now able to add or change defaults here and have them propogate equally across other yml files easily.  yml by default does not include `include` or `require` type mechanics, so this is the closest we can get to that.

docker/etna-apache/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM library/httpd:2.4.43
```
corps, line 1: Base apache image used for the two types of apache services.  I need to provide some additional documentation on my propose development / production networking approach.

docker/etna-base/entrypoints/build.sh
```diff
@@ -0,0 +1,68 @@
+#!/usr/bin/env bash
```
corps, line 1: The previous 'ad-hoc' entrypoints have been scratched in favor of shareable, consistent entrypoints that handle the two main configurations of docker images: "building" via installation and preparation of assets, and "development" configuration such as linking etna and waiting for database readiness before starting.

While most etna applications can easily share logic for these entrypoints, there are some notable exceptions where special logic is required.  To aid in this, both build.sh and development.sh support "hooks" that projects provide.  `magma` and `janus` use these currently.

README.md covering `etna-base` and `entrypoints` coming next.

PR #50, opened 2020-08-07
Adding metadata model

corps:
A model for attaching arbitrary metadata rows to data_blocks.  Some use cases include indicating when a block has been deleted by reaper, or in my case, indicating when a data_block would be need to be archived in s3 vs tape or other means (based on wether it was associated with a certain project at some point, etc).

Mirrors similar concepts in s3.

Why a hash like table and not just rows on data_block?
Mostly because the idea for metadata rows is that they aren't necessary for most data_block rows in the first place; they are intended as lightweight state for special data cases, and aren't considered canonical or necessary for the function of data blocks in a general sense. 
They allow for a ton of maintenance use cases in a simple manner, and could potentially allow for apis and other integrations to also attach arbitrary specialized information about the data sources (like tagging PII or source info) without needed to directly integrate that functionality canonical into metis itself.


graft:
Hmm, while this is generic, it might not be appropriate for things we want to be canonical, like "removed". Other tags seem useful - for example, you could keep a history of links and unlinks to this data block here - but would you expect other behaviors (for example, showing the file version history or forbidden previously removed files) to depend on such a non-canonical data structure?

corps:
> Hmm, while this is generic, it might not be appropriate for things we want to be canonical, like "removed". Other tags seem useful - for example, you could keep a history of links and unlinks to this data block here - but would you expect other behaviors (for example, showing the file version history or forbidding previously removed files) to depend on such a non-canonical data structure?

I can see your point for keeping 'removed' on the main model then.  I was hoping to still use it store alternative archiver settings (aka flag a data_block as requiring s3 vs tape archiving, which is the exception and not the rule).

One of the other benefits of metadata remaining separate from the main model is to keep the datablock cheaper to load.  File queries that fetch many entries and perform eager loads will naturally select all columns by design.  While it's possible very simply to restrict columns on selects, it becomes increasingly cumbersome to restrict in all load cases.

'Canonical' I suppose here in means, it would be necessary for normal functioning of the model.  The lack of metadata never deprives capability.

corps:
@graft @coleshaw How about this approach for tracking removed state?

metis/lib/models/data_block.rb
```diff
@@ -34,6 +34,14 @@ def set_file_data(file_path, copy=false)
       end
     end
 
+    def removed?
```
coleshaw, line 37: Given the conversation around keeping `removed` as a separate column, should the changes in this file be reverted?

coleshaw:
> @graft @coleshaw How about this approach for tracking removed state?

Sorry, based off of our Slack conversation, I merged in #48 -- I didn't realize you had updated this and pinged us again. 

Since the `removed` flag could be used for audit purposes, I feel like it's better to have it explicitly as a table column instead of calculating it from the metadata table. While it can certainly be calculated from the model, I would worry if someone went in via the console or directly into the database and didn't realize that removed data was in a linked table instead of `data_blocks` itself.

metis/db/migrations/021_add_data_block_metadata.rb
```diff
@@ -0,0 +1,30 @@
...
+      String :value, null: false
+      DateTime :created_at, null: false
+      DateTime :updated_at, null: false
+    end
+
+    alter_table(:data_block_metadata) do
+      add_index [:key, :data_block_id], name: "data_block_metadata_key_data_block_id_idx", unique: true
+    end
+
+    alter_table(:data_block) do
```
coleshaw, line 17: should be `data_blocks`?

metis/db/migrations/021_add_data_block_metadata.rb
```diff
@@ -0,0 +1,30 @@
...
+      add_column :removed_at, DateTime, null: true
+    end
+  end
+
+  down do
+    drop_table(:data_block_metadata)
+    drop_index "data_block_metadata_key_data_block_id_idx"
+
+    alter_table(:data_block) do
```
coleshaw, line 26: same as above

PR #51, opened 2020-08-07
Adding waiver warning in search and browser

corps:
Quick hack in for https://www.notion.so/ucsfdatascience/dff110a355f44962a8c7c21b754e0256?v=1931f65e395242bd964f71640382a19e&p=dd8065a5ba434e86a2e0905138620d47

1.  In search, if any patient row contains a Waiver status, show warning.
2.  In browser for patient, if they have Waiver status, show warning.
3.  Disable unselection for consent value from columns.

### DIFF
--- a/timur/lib/client/jsx/components/browser/browser.jsx
+++ b/timur/lib/client/jsx/components/browser/browser.jsx
@@ -36,6 +36,7 @@ import {
   selectTemplate, selectDocument, selectRevision
 } from '../../selectors/magma';
 import { selectUserProjectRole } from '../../selectors/user_selector';
+import {showMessages} from "../../actions/message_actions";
 
 class Browser extends React.Component{
   constructor(props){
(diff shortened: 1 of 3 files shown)


graft:
I'd avoid using showMessages here, whose behavior can be confusing to people - many people don't realize that the message can be dismissed if you click the check. Furthermore the message persists when you change panes, which means this message could persist if you navigate to a different record, leading to some certain confusion (especially since the message does not include the record identifier).

This is hardly the space to hack in a new messaging system, but in this case I think we at least need something that vanishes when you navigate to a different record.

corps:
> I'd avoid using showMessages here, whose behavior can be confusing to people - many people don't realize that the message can be dismissed if you click the check. Furthermore the message persists when you change panes, which means this message could persist if you navigate to a different record, leading to some certain confusion (especially since the message does not include the record identifier).
> 
> This is hardly the space to hack in a new messaging system, but in this case I think we at least need something that vanishes when you navigate to a different record.

Ok sure.  I can style an alternative then.

corps:
Closing in favor of new up to date branch.

PR #53, opened 2020-08-10
Graft update docs

graft:
A bunch of documentation updates, mostly to add basic data downloading examples here. I made some amendments to the Browse documentation on Timur but have not yet added Search documentation. I also didn't document /update_model yet.


coleshaw:
I like the updated Magma docs -- looks good! I've updated against master to grab an update that should fix the flaky test.

PR #56, opened 2020-08-11
only attempt to archive blocks that have not been removed

coleshaw:
This PR addresses mountetna/metis#96. This PR changes the Archiver behavior to ignore `removed` data blocks when checksumming and archiving.

### DIFF
--- a/metis/lib/commands.rb
+++ b/metis/lib/commands.rb
@@ -71,11 +71,11 @@ class Metis
     usage 'Checksum and archive files.'
 
     def execute
-      needs_hash = Metis::DataBlock.where(md5_hash: Metis::DataBlock::TEMP_MATCH).order(:updated_at).all[0..10]
+      needs_hash = Metis::DataBlock.where(md5_hash: Metis::DataBlock::TEMP_MATCH, removed: false).order(:updated_at).all[0..10]
       puts "Found #{needs_hash.count} data blocks to be checksummed."
       needs_hash.each(&:compute_hash!)
 
-      needs_archive = Metis::DataBlock.exclude(md5_hash: Metis::DataBlock::TEMP_MATCH).where(archive_id: nil).order(:updated_at).all[0..10]
+      needs_archive = Metis::DataBlock.exclude(md5_hash: Metis::DataBlock::TEMP_MATCH).where(archive_id: nil, removed: false).order(:updated_at).all[0..10]
       puts "Found #{needs_archive.count} files to be archived."
       needs_archive.each do |data_block|
         begin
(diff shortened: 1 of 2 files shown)


graft:
Looks good to me.

I'm a bit suspicious that something else must be going on as well, since it seems improbable that these files weren't hashed before the data block was removed; it would require them being uploaded less than a minute before running archive orphan blocks - is that likely?

coleshaw:
I don't think they were uploaded just before removing the orphan blocks, certainly they were attempted to upload beforehand if you look at the `created_at` timestamps. I agree that there must be something else going on.

The only way both of those files wouldn't be archived for the last month+ is if `has_data?` returned `false`, right? Both were uploaded as part of a bulk upload operation, so maybe the uploads were started but ran into issues. 

For the first file, I've pinged Arjun to see if he recalls anything weird happening. It looks like he hasn't re-uploaded the real data set for that sample, so I'm guessing something died during the upload process and he may not have noticed.

For the second file, it looks like I uploaded it. The bad block looks like it was the last one attempted on 7/14, and then I'm assuming I got my VPN connection cut off. The same file was the first one I re-attempted on 7/15 and it completed successfully. 

The interesting thing for both DataBlocks is that there is no corresponding Metis::File record for them, so it seems like the Upload#finish! created the Block but not the File? For example, if the bad DataBlock ID was 3000, then the Files will go `Metis::File{id: 500, data_block_id: 2999}`, then `Metis::Fild{id: 501, data_block_id: 3001}` -- so somehow [the DataBlock is created](https://github.com/mountetna/monoetna/blob/master/metis/lib/models/upload.rb#L55) but [the File is not](https://github.com/mountetna/monoetna/blob/master/metis/lib/models/upload.rb#L57)?

It would be interesting if Harry could find the actual data_blocks in the tapes and note how many bytes they are. If they are 0 bytes or corrupted, that might give us a clue.

graft:
A question is: are the blocks uploaded to an existing file? If the latter
you would see a data block created that points to an existing file with no
new file made, as you saw.

This may also relate to request-killing, a block is made but the partial is
not set before it dies.

coleshaw:
Hm, good question. Let me keep investigating. I'll merge this for now, but I think the mystery is not related to the changes in this PR.

coleshaw:
So I don't think the blocks are uploading to an existing file. Wouldn't that result in an orphaned data block for the previous data_block associated with the existing Metis::File? Unless the previous data pointed to the zero-hash block, I guess.

For Arjun's file, there is no Metis::File in the given folder with that file_name. So no record of a pre-existing Metis::File.

For my file, the only Metis::File record with the same file_name and folder path was created after the DataBlock (per `created_at`), so it doesn't seem likely that I was uploading to an existing Metis::File.

coleshaw:
If Harry can recover those files, I think that would be a good next step for forensic analysis, otherwise I'm not sure we're going to glean much more information from the database. 

There are no Metis::Upload entries that indicate an interrupted upload, either.

PR #57, opened 2020-08-12
delete data block archives on Glacier

coleshaw:
This PR won't remove the archives associated with the orphaned data blocks that I removed earlier this week, but should take care of cleaning up the archives in the future.

When a data_block is removed, the corresponding archive on Glacier should also be scheduled for removal.

### DIFF
--- a/metis/lib/archiver.rb
+++ b/metis/lib/archiver.rb
@@ -31,7 +31,7 @@ class Metis
 
     def delete(data_block)
       raise ArgumentError, "No vault defined!" unless @config
-      raise ArgumentError, "Data block not archived!" unless data_block.archive_id
+      return unless data_block.archive_id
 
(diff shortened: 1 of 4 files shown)


metis/lib/archiver.rb
```diff
@@ -29,6 +29,14 @@ def archive(data_block)
       data_block.save
     end
 
+    def delete(data_block)
+      raise ArgumentError, "No vault defined!" unless @config
+      raise ArgumentError, "Data block not archived!" unless data_block.archive_id
```
corps, line 34: Hmm, is this useful?
I think it's actually better to  return false on nil here.  It would allow for easier "sweeps" across datasets without additional logic on the consuming side.
The contract would be "if there is any reason to think this can be deleted from an archive, it will be deleted."

PR #58, opened 2020-08-12
Zc/waiver notifications

corps:
As per @graft 's feedback, took a more disciplined approach:

1.  Show notifications on search and browse pages when patient results with Waiver status set come up.
2.  Dismiss notifications automatically on new searches or when navigating away
3.  Use a new notification react lib to provide shiny notifications
4.  Hook those shiny notifications up behind a feature flag as an alternative to `showMessages` in general
5.  Generalize notifications lib for all etna apps.


bin/seed_janus
```diff
@@ -0,0 +1,30 @@
+#! /usr/bin/env bash
```
corps, line 1: Just a convenience script for quickly adding a useful developer administrator to janus db.  Used after cleaning my janus db from some other testing.

etna/packages/etna-js/hooks/useFeatureFlag.js
```diff
@@ -0,0 +1,11 @@
+import {useState} from 'react'
+import {useReduxState} from "./useReduxState";
+
+export function useFeatureFlag(flagName) {
+  const user = useReduxState(({ user }) => user);
+
+  return useState(() => {
+    if (user && user.flags && user.flags[flagName]) return true;
```
corps, line 8: When we want janus tokens to support feature flags, we can just parse them into this attribute of the user object.

etna/packages/etna-js/hooks/useFeatureFlag.js
```diff
@@ -0,0 +1,11 @@
...
+    return location.search.includes(`${flagName}=1`);
```
corps, line 9: Or, users can 'force' themselves into features with the url.

etna/packages/etna-js/hooks/useReduxState.js
```diff
@@ -0,0 +1,44 @@
+import {useContext, useState, useEffect} from 'react';
+import {ReactReduxContext} from "react-redux";
+
+const ident = (state) => state;
+
+export function useReduxState(selectorFunction = ident) {
```
corps, line 6: Again, React 'hooks' alternative to the `connect` function.  Greatly reduces boiler plat, instead of having to do things like

```
function MyComponent({ a, b, c, d, e, f }) {}
connect(MyComponent, (state) => { a: 1, b: 2, c: 3, d: 4, e: 5 })
```

Another benefit of using hooks here is that `connect` only allows components to withdraw state from the redux, which means the component then has to pass those values around all over the place.  Hooks, on the other hand, can use other hooks, including this one, meaning you can peel back seperate parts of the system that engage with the react state and don't have to require the host component itself pass around a ton of parameters and configure itself correctly with `connect`.

etna/packages/etna-js/utils/janus.jsx
```diff
@@ -0,0 +1,28 @@
+const ROLES = {a: 'administrator', e: 'editor', v: 'viewer'};
+
+const parsePermissions = (perms) => {
```
corps, line 3: Extracted to generalize between `timur` and `metis`.

timur/lib/client/jsx/components/messages.jsx
```diff
@@ -42,39 +47,92 @@ class Messages extends React.Component{
...
+        container: 'top-center',
+        type: 'warning',
+      })
+    })
+  }, [messages])
+
+  return null;
+}
+
+export function unMarkdownNotification(md) {
```
corps, line 107: Might eventually be worth having server pass down notifications / errors that do not require markdown.  From what I can tell, the markdown usage basically boils down to

1.  A title with `##`
2.  Maybe a list with `*`
3.  Bold or italics.

An alternative might be to just include these components as data elements of a json payload, than requiring the client to parse markdown.

That said, this bridge here works just fine for now.

timur/lib/client/jsx/hooks/useRequestDocuments.jsx
```diff
@@ -0,0 +1,38 @@
+import {useCallback} from 'react';
+import {useNotifications} from "etna-js/components/Notifications";
+import {useActionInvoker} from "etna-js/hooks/useActionInvoker";
+import {requestDocuments} from "../actions/magma_actions";
+import {documentOnWaiver, isWaiverPatientModel} from "../utils/patients";
+
+export function useRequestDocuments() {
```
corps, line 7: "Wraps" the requestDocuments action with additional logic that is aware of the component context in which it is being used.  In the future, I think we should move away from using actions to dispatch side effects, and use React hooks instead.  But we'll get there bit by bit.

PR #59, opened 2020-08-14
Cs/rename folders across buckets

coleshaw:
This PR adds the ability to rename a folder into a different bucket (same project), using the `folder_rename` route. It does not extend that ability to the Metis UI yet -- we can add that functionality later, if needed. Since our main use case currently is an internal scripting tool, this is just an API change.

Much of the changes are related to building off of the `Revision` and validations that were used for the updated `File#copy` and `File#bulk_copy` routes, so that the validation schemes are shared.

To test in a running instance, you have to make a request outside of the UI, via a tool like `curl` or Postman. You need to include two new parameters in the payload, `new_folder_path` and `new_bucket_name`. For example, if you have a bucket called `published` with a folder `stories` in the project 'legends', you would POST to `https://<Metis host>/legends/folder/rename/published/stories` a payload of

```
{"new_folder_path":"story_ideas", "new_bucket_name": "drafts"}
```

And that would move `stories` plus all children folders and files to the `drafts` bucket while also renaming the folder to `story_ideas`. So the Metis "view" URLs would go from `/legends/published/stories` => `/legends/drafts/story_ideas`.

Renaming folders in the Metis UI should behave the same as before.

### DIFF
--- a/metis/lib/folder_rename_revision.rb
+++ b/metis/lib/folder_rename_revision.rb
@@ -1,0 +1,26 @@
+require_relative 'folder_revision'
+
+class Metis
+  class FolderRenameRevision < FolderRevision
+    def revise!
+      raise 'Invalid revision, cannot revise!' unless valid?
+      raise 'Cannot revise without a user' unless @user
+
+      # Have to fetch the actual source folder. What gets
+      #   set in the revision's source.folder attribute
+      #   is the parent folder, which
+      #   we don't need at this point.
+      source_folder = Metis::Folder.from_path(
+        @source.bucket, full_folder_path(@source)).last
+
+      if @source.mpath.bucket_name == @dest.mpath.bucket_name
+        source_folder.rename!(@dest.folder, @dest.mpath.file_name)
+      else
+        source_folder.update_bucket_and_rename!(
+          @dest.folder, @dest.mpath.file_name, @dest.bucket)
+      end
+
+      return source_folder
+    end
+  end
+end
--- a/metis/lib/folder_revision.rb
+++ b/metis/lib/folder_revision.rb
@@ -1,0 +1,98 @@
+require_relative 'revision'
+
+class Metis
+  class FolderRevision < Revision
+    def validate
+      @errors = []
+      validate_source(@source)
+      validate_dest(@dest)
+    end
+
+    def bucket_names
+      # This is kind of weird, but we need the ability to grab
+      #   all relevant bucket names, even before validation of
+      #   the CopyRevision.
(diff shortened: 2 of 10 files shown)


metis/lib/folder_rename_revision.rb
```diff
@@ -0,0 +1,26 @@
...
+
+      # Have to fetch the actual source folder. What gets
+      #   set in the revision's source.folder attribute
+      #   is the parent folder, which
+      #   we don't need at this point.
+      source_folder = Metis::Folder.from_path(
+        @source.bucket, full_folder_path(@source)).last
+
+      if @source.mpath.bucket_name == @dest.mpath.bucket_name
+        source_folder.rename!(@dest.folder, @dest.mpath.file_name)
```
corps, line 17: Remind me what

metis/lib/folder_revision.rb
```diff
@@ -0,0 +1,106 @@
...
+      if target_folder&.read_only?
+        @errors.push("Folder \"#{mpath_w_objs.mpath.path}\" is read-only")
+        errors_found = true
+      end
+
+      return errors_found
+    end
+
+    def validate_dest_is_not_file(mpath_w_objs)
+      # Because we don't want users overwriting files with a folder.
```
corps, line 73: Super minor nit we can't address here easily, but we need to be aware that this doesn't address race conditions in which case a new file is being created in a separate parallel  database transaction.  We have a unique key on folder and file names, but not cross between them.  So it's still possible to create a file by the same name as an existing folder given the right timing.

metis/lib/models/folder.rb
```diff
@@ -97,6 +97,28 @@ def rename!(new_folder, new_folder_name)
...
+      }
+      folders.each { |folder|
+        folder.update_bucket!(new_bucket)
+      }
+
+      refresh
+    end
+
+    def update_bucket_and_rename!(new_folder, new_folder_name, new_bucket)
+      # rename first to avoid any potential collisions with folder names
```
corps, line 115: Does, this logic work?  What if the rename! conflicts locally but not in the remote bucket?  

I want to rename my_bucket/a/b to other_bucket/c/d, but there happens to be a my_bucket/c/d.  Won't this fail although it does not have to?

corps: I think you cannot break this down into 2 discrete steps.  I think the validation only works if the bucket and renaming happen together at the same time.

coleshaw: Thanks, great catch!

PR #60, opened 2020-08-17
add a list-all-folders route for a bucket

coleshaw:
For the "Move folder" script, I think having a route that returns a list of all folders in a bucket would be helpful. This PR adds such a route. Magma does not store the actual Metis routes for any given file or image attribute, so the only way to determine the Metis names of folders that need to be moved is a regex-type match on a Magma-provided parameter (i.e. a pool name or a patient name). However, this requires the script to iterate through all folders on Metis. From a client-perspective, iteratively fetching that data can be very time consuming because the current routes only provide a list of direct child folders.

This PR adds a simple route that returns all folders in a specified bucket in a flat array. While this single route can be computationally intensive on the server (because each folder runs a recursive database query to generate its `folder_path` attribute), I feel like overall this provides a better consumer experience. The move script will be shorter-running and be able to more quickly find the folders that need to be moved.


metis/lib/server/controllers/folder_controller.rb
```diff
@@ -23,6 +23,18 @@ def list
...
   end
 
+  def list_all_folders
+    bucket = require_bucket
+
+    folders = Metis::Folder.where(
+      bucket: bucket
+    ).all.map do |fold|
+      fold.to_hash
+    end
```
graft, line 33: It would be pretty easy to do this in a cheaper way that avoids repeated recursive queries to the db. Since in your original query you actually have all of the folder records loaded, along with their folder_id, there is no need to make return queries to the database. Just group_by folder_id, then you should be able to do a traverse of the tree in O(n) and list all of the folder paths in the bucket.

corps: :+1:  ^^

corps: Also does this potentially need a paging operation?  I can imagine a bucket easily having 10+ items.

coleshaw: :+1: Let me test out grouping and see if that helps. Pagination may make sense.

PR #61, opened 2020-08-18
Add flags to user model/token

graft:
A little bit that just adds a 'flags' attribute to the user model in Janus, which holds an array of words; this is reported out in the user's token as a ;-separated list 'flags'.

There is also a super-user API /flag_user to update this attribute. There isn't any decent way to actually see which users have which flags, I realize, but there is also no decent way to see a list of users in the system, period, so this seems a minor wart in comparison.


corps:
Nice dude!  So this just auto populates into the token hash right?  Does it make sense to add a etna/user.rb utility like `has_flag?`?  

This will have a ton of uses.

coleshaw:
This looks good to me. If we start using it more, I can definitely see how incorporating it into the `etna` gem would be useful, as well as some sort of interface for checking flags. It seems slightly inconvenient that the only way to check now would be to SSH into the box and make a database connection, but it also seems easy to just overwrite your current value or `nil` it out.

PR #62, opened 2020-08-19
Zc/polyphemus and magma client scaffolding

corps:
A lot of little thoughts going on in this PR.

1.  New handrolled magma client in etna gem.  This client differs in that it has explicit domain models and methods in order to make development easier.  For instance, these explicit domain models add better autocompletion sense, simplifies documentation (no digging around server files or creating entirely separate api docs, etc), and allows adding auxilary helpful domain logic (to directed graph, for instance).

But why?  The real reason is that I see this as a first step to a larger goal: code generated clients.  [Swagger](https://swagger.io/tools/swagger-codegen/) is a standard HTTP first api definition language that is language agnostic.  The goal would be that we get to a point that we define our api schemas in swagger.yml, then use code gens to roll out

1.  automatic js clients complying with the inputs & outputs & paths
2.  automatic spec verification that ensures in specs that server inputs and outputs correspond to the schema
3.  automatic ruby clients
4.  automatic python clients
5.  ????

Some logic would not be codegened (the raw http client behavior, for instance, and some domain helpers), but the vast majority of domain models, endpoints, and functions would come "for free".  This is an alternative to having, say, endpoints that server up dynamic schemas because it doesn't require runtime unfolding of the schema by a client.  I prefer the codegen approach because the codegen code itself acts as better documentation, improves autocompletion, and doesn't require one to figure out how to bootstrap the process (for instance the urls and hosts can be baked into the client without first doing a bunch of exploring or metamagic just to generate those values).

Of course, codegen I believe is secondary to first setting up the pattern that we like.  Feel free to critique the way the domain models or clients behave.  Once we're comfortable with a pattern, I'll work on a ruby codegen tool to generate the same logic, but from a schema file.

2.  Introduction of rollbar.  I need to handout accounts, but basically, the idea is to capture exceptions that happen server side and send them to a non vpn sass service for easier inspection and for integration with, say, slack for notification and response.  Rollbar automatically hooks into rack for free, and I've also added it as an option to the Etna::Logger.  We'll want to continue improving where this gets used if we decide to move forward.

3.  Polyphemus basic scaffolding for now.  Can run commands, which is how I'm testing the magma client.


polyphemus/docker/app/puma.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+export TIMUR_ENV=development
```
coleshaw, line 3: nit: should export `POLYPHEMUS_ENV`?

polyphemus/docker/app/Dockerfile.development
```diff
@@ -0,0 +1,54 @@
+FROM ruby:2.5.7
+WORKDIR /app
+
+ENV RUBY_VERSION 2.5.7
+ENV BUNDLER_VERSION 2.1.4
+ENV DOCKERIZE_VERSION 0.5.0
+ENV DOCKERIZE_URL https://github.com/jwilder/dockerize/releases/download/v${DOCKERIZE_VERSION}/dockerize-linux-amd64-v${DOCKERIZE_VERSION}.tar.gz
+ENV NODE_VERSION 12.18.1
+ENV SERVICE_NAME timur
```
coleshaw, line 9: nit: `polyphemus` instead of `timur`?

polyphemus/docker-compose.yml
```diff
@@ -0,0 +1,53 @@
...
+    - "rtemis.development.local:172.16.238.10"
+  networks:
+    edge_net:
+    default:
+
+services:
+  polyphemus_app:
+    <<: *polyphemus_base
+    ports:
+      - 3002:3000
```
coleshaw, line 52: I think Timur is already on `3002`, so this may cause a conflict? I think this could safely be `3004`.

polyphemus/config.yml
```diff
@@ -0,0 +1,7 @@
+:development:
```
coleshaw, line 1: This file should be in `.gitignore`?

etna/lib/etna/logger.rb
```diff
@@ -19,6 +18,8 @@ def log_error(e)
       e.backtrace.each do |trace|
         error(trace)
       end
+
+      Rollbar.error(e)
```
coleshaw, line 22: What does this do if Rollbar isn't configured with a key? Like for local development, or staging or something?

corps: It's a simple no op
https://github.com/rollbar/rollbar-gem/blob/master/lib/rollbar/logger_proxy.rb#L25

coleshaw:
Looks good to me as a base.

PR #63, opened 2020-08-19
Restricted column itself is no longer restricted, but not editable by…

corps:
… restricted users.

### DIFF
--- a/magma/spec/update_spec.rb
+++ b/magma/spec/update_spec.rb
@@ -862,6 +862,25 @@ describe UpdateController do
       expect(victim.country).to eq('nemea')
     end
 
+    it 'prevents updates to the restricted column by a restricted user' do
+      victim = create(:victim, name: 'Outis Koutsonadis', country: 'nemea')
+
+      update(
+          {
+              victim: {
+                  'Outis Koutsonadis': {
+                      restricted: false
+                  }
+              }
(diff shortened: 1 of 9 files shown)


corps:
This works by modifying the censor, which is responsible for filtering edits, to always censor access to restricted, but removing the ability to set the restricted flag on the restricted attribute, which would prevent filtering on restricted models from working.

magma/spec/update_spec.rb
```diff
@@ -881,5 +900,23 @@ def update(revisions, user_type=:editor)
...
+                  'Outis Koutsonadis': {
+                      restricted: false
+                  }
+              }
+          },
+          :editor
+      )
+      expect(last_response.status).to eq(200)
+
+      expect { victim.refresh }.to change { victim.restricted }.to(false)
```
coleshaw, line 919: nit: By default isn't `victim.restricted` set to `false` on `create`, since it wasn't specified? -- should this be testing the opposite (setting it to `true` on update), or should the create() call set it to `true`?

corps: It is set to `nil` actually.  If I start from true, then the call itself would fail for not being able to find the victim, rather than not being able to update its state. The test still works from the perspective of setting it to a value from the nil state.

PR #64, opened 2020-08-20
add relative require to get load project working

coleshaw:
Small fix to Magma's `load_project` command. I was trying to update a project and getting this exception:

```
	8: from ./bin/magma:11:in `<main>'
	7: from /app/vendor/bundle/2.5.7/ruby/2.5.0/gems/etna-0.1.13/lib/etna/application.rb:75:in `run_command'
	6: from /app/lib/magma/commands.rb:12:in `execute'
	5: from /app/lib/magma/commands.rb:12:in `each'
	4: from /app/lib/magma/commands.rb:18:in `block in execute'
	3: from /app/lib/magma/commands.rb:18:in `each'
	2: from /app/lib/magma/commands.rb:19:in `block (2 levels) in execute'
	1: from /app/lib/magma/commands.rb:49:in `load_attribute'
/app/lib/magma/commands.rb:63:in `options': uninitialized constant Magma::Attribute (NameError)
```

This PR just imports `Magma::Attribute` into the method, so it can be available.

### DIFF
--- a/magma/lib/magma/commands.rb
+++ b/magma/lib/magma/commands.rb
@@ -59,6 +59,7 @@ class Magma
     end
 
     def options
+      require_relative './attribute'
       @options ||= Magma::Attribute.options - [:loader] + [:created_at, :updated_at, :attribute_name]
     end
   end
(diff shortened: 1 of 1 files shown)


corps:
This did work for me awhile back, so I'm guessing it worked before due to an implicit require somewhere that is no longer the case.

PR #66, opened 2020-08-20
Zc/cascade restricted

corps:
Ok, so here's the base script to cascade the consent status to restricted values.
I've noted the places we should insert the metis logic.

Notably, recursive handling of restrict isn't yet implemented; this PR assumes it will be.  That's my next task, but I figured getting this setup would help unblock @coleshaw .

@coleshaw -- I recommend merging your metis work into this branch, bumping the etna version in the gem file (note that I also have etna gem changes, though), and getting your code to work in this environment.  Right now I'm pointing to local file etna.

I want to quickly move back over to the docker build branch and get that merged because it includes the ability to make changes to the etna gem and have them deployable without needing to bump etna gem version (via github commit references).  Since we're making a lot of etna changes lately, I believe getting this into master will save us a lot of work.


etna/lib/etna/directed_graph.rb
```diff
@@ -19,17 +19,38 @@ def add_connection(parent, child)
   end
 
   def descendants(parent)
-    result = []
     seen = Set.new
-    queue = @children[parent].keys
+
+    seen.add(parent)
```
corps, line 24: I didn't end up actually using this directed_graph implementation with the simplified approach to restricted recursion, but I thought it was still useful so I left the code.  If you feel this is distracting and not useful, I'll happily remove.

polyphemus/config.yml.template
```diff
@@ -1,7 +1,11 @@
 :development:
   :magma:
-    :host: https://magma.ucsf.edu
+    :host: https://magma.development.local
+  :mvir1:
+    :token: my-token # create a long lived token with janus
```
corps, line 5: I imagine configuring a mvir1 project specific long lived token here for now until we implement proper impersonation via s2s authentication.

polyphemus/lib/commands.rb
```diff
@@ -13,17 +13,129 @@ def execute
...
+  module WithEtnaClients
+    def project
+      raise "project must be implemented in subclasses!"
+    end
+
+    def token
+      Polyphemus.instance.config(project)[:token]
+    end
+
+    def magma_client
```
corps, line 25: @coleshaw I moved the magma_client logic out of the application object into here, since these clients will have per project tokens.

polyphemus/spec/cascade_mvir_restricted_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+        }.to_json })
+
+    stub_request(:post, 'https://magma.test/update')
+        .to_return do |request|
+
+      body = StringIO.new(request.body)
+      content_length = body.read.length
+      body.rewind
+
+      tempfile = Rack::Multipart::Parser::TEMPFILE_FACTORY
```
corps, line 35: I reaaallly wish that multipart parsing were simpler for testing.  Webmock straight up errors and warns it does not support it.  So I had to hack this bit in.

polyphemus/config.yml.template
```diff
@@ -1,7 +1,11 @@
...
 
-:production:
+:test:
   :magma:
-    :host: https://magma.ucsf.edu
+    :host: https://magma.test
+  :mvir1:
+    :token: mvir1-token
```
graft, line 11: NB: I'm pretty un-excited about writing project names into config files any more and would much rather put this into a database column as soon as possible.

PR #68, opened 2020-08-24
small fixes to make seed_databases working

coleshaw:
With the updated docker build, I was having issues getting `seed_databases` working. It seemed like the `create` Make command was changed (??), so seems like `up` is the replacement. I was getting this exception when running the command:

```
make: Entering directory 'monoetna/docker'
make: *** No rule to make target 'create'.  Stop.
make: Leaving directory 'monoetna/docker'
```

Also the actual seeding was failing when `up` was called beforehand, since the apps would be connected to the database and I would get an exception like:

```
dropdb: database removal failed: ERROR:  database "timur_development" is being accessed by other users
DETAIL:  There is 1 other session using the database.
++ error /root/bin/seed_databases 35
++ echo 'Whoops!  Looks like /root/bin/seed_databases:35 failed.'
Whoops!  Looks like /root/bin/seed_databases:35 failed.
++ echo 'Please try rerunning /root/bin/seed_databases again.'
Please try rerunning /root/bin/seed_databases again.
```

But not bringing up all the docker containers before seeding seems to fix the competing connection issue.

### DIFF
--- a/bin/download_metis_dir
+++ b/bin/download_metis_dir
@@ -38,7 +38,7 @@ if [[ -e /dest ]]; then
     /app/bin/metis_client $metisPath get ./ /dest/
 else
   # Ensure images are created
-  make -C docker up
+  make -C docker build
   docker-compose -p monoetna run -e TOKEN="$TOKEN" -v $destPath:/dest -v $DIR:/root/bin metis_app /root/bin/download_metis_dir "$metisPath"
 fi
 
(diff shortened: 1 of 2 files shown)


bin/download_metis_dir
```diff
@@ -38,7 +38,7 @@ if [[ -e /dest ]]; then
     /app/bin/metis_client $metisPath get ./ /dest/
 else
   # Ensure images are created
-  make -C docker create
+  make -C docker up
```
corps, line 41: Oh!  Good catch.  This can now just be `build`.  I feel stupid but I renamed `create` -> `build` for consistency because I already had some build projects, too.

corps:
I'll adjust this real quick, nice catch!

PR #69, opened 2020-08-24
add call to timur create_routes in entrypoint

coleshaw:
The Timur UI requires a call to `./bin/timur create_routes` to update the JS routes used in the front-end. This puts that call into the base `entrypoint.sh`.

Without this line, the Timur UI loads with the list of projects, but then fails to a blank screen if you try to navigate to a specific project.

### DIFF
--- a/docker/etna-base/entrypoints/build.sh
+++ b/docker/etna-base/entrypoints/build.sh
@@ -70,3 +70,7 @@ fi
 if [ -n "$RUN_NPM_INSTALL" ]; then
   if [ -e ../etna ]; then npm link ../etna/packages/etna-js; fi
 fi
+
+if [ "$APP_NAME" = "timur" ]; then
+  ./bin/timur create_routes
+fi


docker/etna-base/entrypoints/build.sh
```diff
@@ -70,3 +70,7 @@ fi
 if [ -n "$RUN_NPM_INSTALL" ]; then
   if [ -e ../etna ]; then npm link ../etna/packages/etna-js; fi
 fi
+
+if [ "$APP_NAME" = "timur" ]; then
```
corps, line 74: Hmm, so this actually already does exist in `timur/build/create-routes` is it not being run?

coleshaw: Hm, I don't see a folder `timur/build` or a script at `timur/build/create-routes` in the repo, and if I grep for `create_routes` or try a `find . -name create-routes` I get no results other than the actual task itself ... sorry, am I just not seeing it?

corps:
I just addressed this here, my apologies: https://github.com/mountetna/monoetna/commit/cfccc4d2634037f50d6dd5aefadca94b215e238d

PR #70, opened 2020-08-24
Recursive restriction

graft:
This sets up recursive restriction on models, so that if a given record is restricted, all its children behave as restricted also, for both retrieval and update.

### DIFF
--- a/magma/lib/magma/query/predicate/model.rb
+++ b/magma/lib/magma/query/predicate/model.rb
@@ -31,11 +31,19 @@ class Magma
       @model = model
       @filters = []
 
-      if question.restrict? && @model.has_attribute?(:restricted)
-        # the model can be restricted, and we should withhold restricted data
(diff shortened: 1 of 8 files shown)


magma/lib/magma/query/predicate/model.rb
```diff
@@ -31,11 +31,19 @@ def initialize(question, model, *query_args)
...
-        query_args.unshift(
-          [ 'restricted', '::false' ]
-        )
+      if question.restrict?
+      # the model can be restricted, and we should withhold restricted data
+        restriction_model = model
+        ancestral_path = []
+        while restriction_model do
+          if restriction_model.has_attribute?(:restricted)
+            query_args.unshift(
```
corps, line 40: Should there be a break here?  We discussed that if you have multiple restricted in your heritage, that the nearest one takes presedence.

PR #71, opened 2020-08-25
extract metis helper

coleshaw:
Because some of the MetisWaiverHelper methods are generic and potentially useful in other contexts, I've extracted them into a separate adapter class in `etna`.

Also made the waiver helper a separate class instead of a module, per feedback from @graft .

Expected behavior of this PR is the same as before -- polyphemus command should work the same.

Includes changes from #78  (so I could use Timur to test state changes).


corps:
I guess my only nit is the name `ProjectClientAdapter` is super confusing.  I usually imagine `Adapters` as transforming some behavior or input to another, but this feels more like... Metis client ++?  I ponder if just putting this into Clients::Metis as additional helper methods makes some sense?

coleshaw:
I know the name kind of sucks, but couldn't think of what would be more appropriate. I agree it's more like a client + convenience methods. I can add those to the base client if that makes the most sense, but wasn't sure if that would get too big, if we keep adding convenience methods.

corps:
> I know the name kind of sucks, but couldn't think of what would be more appropriate. I agree it's more like a client + convenience methods. I can add those to the base client if that makes the most sense, but wasn't sure if that would get too big, if we keep adding convenience methods.

Yeah.  Well, let's just move forward with this for now and see how it unfolds.

polyphemus/lib/metis/waiver_helper.rb
```diff
@@ -0,0 +1,143 @@
...
+  def patient_name_regex(patient_name)
+    /.*\/#{patient_name}-[\w-]+$/
+  end
+
+  def pool_regex(pool_name)
+    /.*\/#{pool_name}$/
+  end
+
+  def restrict_patient_data(patient_name)
+    # Move a folder from release_bucket to restrict_bucket
```
graft, line 21: These four methods are nearly identical and should probably be refactored into a single method. This whole class could also pretty easily be abstracted to accept some generic model_name and bucket_name arguments and become a move_folder_helper class rather than a waiver_helper class.

graft:
I think I agree with @corps that the ProjectClientAdapter class is of ambiguous utility; most of its methods are very thin wrappers around Etna::Clients::Metis, and so it might be best if most of that functionality went in there, perhaps as options on existing methods.

It seems like there should be three layers here:
1) At the lowest level is Etna::Clients::Metis, which does low-level API stuff with Metis.
2) Wrapped around this is some sort of bucket transfer layer, which orchestrates a complex action (moving a folder from one named bucket to another named bucket) on Metis via Etna::Clients::Metis.
3) Finally, the waiver-specific logic (including the mvir1 classnames and so on) and setup of (1) and (2) is in some kind of Mvir1Waiver class; the command could just boot up this class, feed it the config, and call `#run`. Even better if there is NO mvir1-specific naming or code here and all of that stuff is just in config variables.

This sets us up to have a bunch of generic tools like (2) that we could reuse between project-specific scripts like the waiver script, and it allows those scripts to be in an isolated class that we can run however we wish after injecting a simple config hash into it.

polyphemus/spec/examples.txt
```diff
@@ -1,16 +1,16 @@
...
-./spec/metis_waiver_helper_spec.rb[1:5]       | passed | 0.0268 seconds  |
-./spec/metis_waiver_helper_spec.rb[1:6]       | passed | 0.00497 seconds |
-./spec/metis_waiver_helper_spec.rb[1:7]       | passed | 0.01476 seconds |
-./spec/metis_waiver_helper_spec.rb[1:8]       | passed | 0.00275 seconds |
-./spec/metis_waiver_helper_spec.rb[1:9]       | passed | 0.01422 seconds |
-./spec/metis_waiver_helper_spec.rb[1:10]      | passed | 0.0052 seconds  |
+./spec/cascade_mvir_restricted_spec.rb[1:1:1] | passed | 0.08119 seconds |
+./spec/cascade_mvir_restricted_spec.rb[1:1:2] | passed | 0.05934 seconds |
+./spec/cascade_mvir_restricted_spec.rb[1:2:1] | passed | 0.04292 seconds |
+./spec/cascade_mvir_restricted_spec.rb[1:2:2] | passed | 0.03913 seconds |
```
graft, line 6: This file should be in .gitignore.

polyphemus/lib/commands.rb
```diff
@@ -271,7 +71,7 @@ def execute
           restrict!(patient, delete_metis_files: true)
         elsif should_be_restricted
           restrict!(patient)
-        elsif ! should_be_restricted && patient['restricted']
```
coleshaw, line 274: Note: I think if we want every patient to run every time, we need to also remove this check on the existing `restricted` status for the patient.

PR #72, opened 2020-08-25
clarify ::false vs. ::untrue for restricted

graft:
Missed getting this in by a hair's breadth before the recursive restriction stuff got merged, this just clarifies '::false' vs. '::untrue' (the latter including 'null', the former not) and indicates that unrestricted records are ones where restricted is '::untrue'.

### DIFF
--- a/magma/lib/magma/query/predicate/boolean.rb
+++ b/magma/lib/magma/query/predicate/boolean.rb
@@ -14,6 +14,13 @@ class Magma
     verb '::false' do
       child TrueClass
       constraint do
+        basic_constraint(@column_name, false)
+      end
+    end
+
+    verb '::untrue' do
+      child TrueClass
+      constraint do
         not_constraint(@column_name, true)
       end
     end
--- a/magma/lib/magma/query/predicate/model.rb
+++ b/magma/lib/magma/query/predicate/model.rb
@@ -38,7 +38,7 @@ class Magma
         while restriction_model do
           if restriction_model.has_attribute?(:restricted)
             query_args.unshift(
-              ancestral_path + [ 'restricted', '::false' ]
+              ancestral_path + [ 'restricted', '::untrue' ]
             )
           end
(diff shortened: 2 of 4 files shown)


magma/lib/magma/query/predicate/boolean.rb
```diff
@@ -12,6 +12,13 @@ class BooleanPredicate < Magma::ColumnPredicate
     end
 
     verb '::false' do
+      child TrueClass
+      constraint do
+        basic_constraint(@column_name, false)
```
coleshaw, line 17: This seems like it might break some current things ... are there plots or manifests on Timur that rely on the behavior where `nil` records would return from a `::false` query?

coleshaw:
Just a small comment  -- when we get around to clarifying the Timur search for boolean attributes, I think we will have to consider how to best communicate the differences between `::false`, `::untrue`, `nil`, and `::true` in both the filters and the displayed text. I think the current text is either "yes" or "no" (two states instead of three states) and shows as a checkbox in the form -- would that become three states (yes, no, unset) and radio buttons instead of checkboxes? Not super important, just something I think we should keep in mind from a UX perspective.

coleshaw:
I guess radio button isn't the right analogy in this case, though ... you only know it's unset if there is a radio button group with multiple radio buttons. With just one for an attribute, it would still be ambiguous...

graft:
I agree we need something like that, probably for every control, see #189.

coleshaw:
Did a scan of the public manifests and plots on Timur and didn't see any usages of boolean attributes, so this is probably safe to merge. If anyone reports issues with their private stuff it should be an easy fix to point out.

graft:
Yeah i didn't find any instances of this in the codebase; there aren't too many boolean attributes in ipi, and I couldn't come up with any actual uses of it in plots or so on, so I think it should be okay to introduce this.

PR #73, opened 2020-08-26
Improved Docker Build Speeds

corps:
In my never ending quest to improve our testing and build speeds, I found two big clues:

1.  I was using the docker-compose file to run etna tests, but I don't need to.  Doing so actually results in extra image builds for docker-compose, when in reality we already have the etna base image that we can use directly.

2.  Caching was not ideal because of the `ARG` lines in the docker file appearing -before- the bundle and npm installs.  These args are only used in the `development.sh` entrypoint, so putting them after bundle and npm installs means those two lines can be skipped (used from cache) more often.

3.  Downloading of previously built images for caching was not parallel, even though it can and should for faster setup.  Images are not pulled individually now, but are pulled en-masse together with xargs.


corps:
I also fixed an issue where master was not actually pushing up to date images for all projects -- after merging this to master, it will correctly build images and push them, resulting in future builds pulling them and caching better as well.

PR #74, opened 2020-08-28
Etna logger does not automatically log warnings, polyphemus always tr…

corps:
…ies to move patient files

This was actually simpler than expected.

1.  Always try to restrict patient files for all restricted patients.  Even if they have already been restricted.  This means @graft you can set the restricted bit freely, this script will still handle the files.  This seems like a lot of extra work but it's actually not; we cache the entire bucket's folder set in one query anyways, which is a fixed cost.

2.  Reduce rollbar logger messages by not implicitly logging every warn.

### DIFF
--- a/polyphemus/lib/commands.rb
+++ b/polyphemus/lib/commands.rb
@@ -1,5 +1,6 @@
 require 'date'
 require 'logger'
+require 'rollbar'
 
 class Polyphemus
   class Help < Etna::Command
@@ -268,7 +269,7 @@ class Polyphemus
           # Magma does not support delete right now, so we restrict the models, but we can unlink and
           # delete the files in metis
           restrict!(patient, delete_metis_files: true)
-        elsif should_be_restricted && ! patient['restricted']
+        elsif should_be_restricted
           restrict!(patient)
         elsif ! should_be_restricted && patient['restricted']
           unrestrict!(patient)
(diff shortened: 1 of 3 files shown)


polyphemus/lib/commands.rb
```diff
@@ -268,7 +269,7 @@ def execute
           # Magma does not support delete right now, so we restrict the models, but we can unlink and
           # delete the files in metis
           restrict!(patient, delete_metis_files: true)
-        elsif should_be_restricted && ! patient['restricted']
```
corps, line 271: So now we query every patient, and based on their consent status start the file moving proces every time.

polyphemus/lib/commands.rb
```diff
@@ -326,19 +322,20 @@ def restrict!(patient, delete_metis_files: false)
         restrict_patient_data(name)
       end
 
-
-      update_request = Etna::Clients::Magma::UpdateRequest.new(project_name: project)
-      update_request.update_revision('patient', name, restricted: true)
-      magma_client.update(update_request)
+      unless patient['restricted']
```
corps, line 325: We just only apply the restricted update and log if it was strictly necessary.

polyphemus/lib/commands.rb
```diff
@@ -268,7 +269,7 @@ def execute
           # Magma does not support delete right now, so we restrict the models, but we can unlink and
           # delete the files in metis
           restrict!(patient, delete_metis_files: true)
-        elsif should_be_restricted && ! patient['restricted']
+        elsif should_be_restricted
           restrict!(patient)
         elsif ! should_be_restricted && patient['restricted']
```
coleshaw, line 274: Sorry, old PR -- but should this check also not check `patient['restricted']`, and just always rely on `should_be_restricted`?

PR #78, opened 2020-08-31
timur fixes

corps:
Working on the test (writing the first deep test of browser.jsx) but in case this is blocking you @coleshaw it's a simple change to fix the bug.


docker/etna-base/entrypoints/development.sh
```diff
@@ -19,7 +19,9 @@ if [ -z "$SKIP_BUILD" ]; then
 
   if [ -z "$SKIP_RUBY_SETUP" ]; then
     if [ -z "$SKIP_DB_WAIT" ]; then
-      dockerize -wait tcp://${APP_NAME}_db:5432 -timeout 60s
+      #Sometimes during startup the cpu can be slammed for all the servies, so we need to wait
```
corps, line 22: This prevents the issue where sometimes on startup the app_fe contains exit with 1 due to timeout (when starting everything at once this can happen)

timur/lib/client/jsx/components/browser/browser.jsx
```diff
@@ -189,12 +189,12 @@ function useEditActions(setMode, browserState) {
 
   function postEdits() {
     setMode('submit');
-    sendRevisions(
+    invoke(sendRevisions(
```
corps, line 192: @coleshaw This is all you need to get local dev working agian.

coleshaw: Nice! Thanks!

coleshaw: I can't comment outside of the changes...but does `view` need to be passed from `selectOrShowTab` to `showTab`?

```
  function selectOrShowTab(view) {
    if (currentTabName) showTab(view);
    else selectDefaultTab(view);
  }

  function showTab(view) {
    requestTabDocuments(view.tabs[currentTabName]);
    setMode('browse');
  }
```

The current data isn't showing up correctly on page load... (note that adding the `view` parameter doesn't fix things, but does make them better).

coleshaw: Okay, so the above comment about `view` may not be required ... but I'm not getting data loaded correctly somehow. I think there's something else going on in addition to the `invoke` call.

coleshaw: :+1: rebuilt all my docker stuff and it works now. Thanks!

etna/packages/etna-js/spec/helpers.js
```diff
@@ -12,11 +12,24 @@ export const stubUrl = ({
...
   status = 200,
   headers = {},
-  host = 'http://localhost'
+  host = 'http://localhost',
+  url,
 }) => {
   let nocked;
 
-  console.log('Stubbing for', {request, path, verb, response}, '\nactive')
+  if (url) {
```
corps, line 20: Adding the ability to pass in url instead of host and path separately.  Some utility functions (timur's path functions) return full urls, so this is more convenient as a parameter.

timur/jest.config.js
```diff
@@ -1,17 +1,21 @@
...
-    CONFIG: {
-      project_name: 'labors'
-    }
+    TIMUR_CONFIG: {
+      project_name: 'labors',
+      magma_host: 'https://magma.test',
+    },
   },
-  testURL: 'http://www.fake.com',
+  testURL: 'http://localhost',
```
corps, line 9: I'm swapping out all www.fake.com with localhost because previously, our tests actually _Did_ call www.fake.com (a real website).  Pointing to localhost ensures that even if someone the netDisable nock function isn't enabled, it won't make real external network calls with random payloads.

timur/lib/client/jsx/components/browser/browser.jsx
```diff
@@ -102,7 +101,7 @@ export default function Browser({ model_name, record_name, tab_name }) {
     } else if (!tab_name) {
       selectDefaultTab(view);
     } else {
-      showTab();
+      showTab(view);
```
corps, line 104: You were correct -- the view needs to be passed it in.  It worked before by using the view closure from the state, which is out of date when this function is invoked as part of some other callback that has already updated the state.

timur/test/javascript/helpers.js
```diff
@@ -25,8 +25,6 @@ export const mockStore = configureMockStore([
   thunk
 ]);
 
-export const setConfig = (config) => global.TIMUR_CONFIG = config;
```
corps, line 28: Added this to the jest.config.js just so that each test "gets this for free" without needing to  explicitly set it.  We have no tests that dynamically need interesting configs; they all basically just assume "some" static values.

PR #79, opened 2020-09-02
Zc/add project

corps:
Tested locally, works like a charm.


graft:
Haven't tested locally but looks great. I like the idempotence, although the other actions are not idempotent (and maybe cannot be?) which is a bit weird.

Following this we should convert all our projects to db only and ditch model based config. Also maybe think about resource projects visible to all users.

corps:
> Haven't tested locally but looks great. I like the idempotence, although the other actions are not idempotent (and maybe cannot be?) which is a bit weird.
> 
> Following this we should convert all our projects to db only and ditch model based config. Also maybe think about resource projects visible to all users.

Rgr rgr

PR #80, opened 2020-09-03
Metis find

coleshaw:
This PR adds a `find` route to Metis, letting you search within a bucket for files or folders. Currently it supports searching by `updated_at`, `created_at`, and `file_name` / `folder_name` properties.

For `updated_at` and `created_at`, any `Time`-related predicates can be used, like `<` and `>`.

For `name`, the route will extend that to `file_name` or `folder_name` as appropriate, with default behavior being to search across both files and folders. You can search with equality (`=`) or a regex / SQL-like string (`=~`).

You can also limit the type of results returned for each parameter by supplying a `type` value.

The route also supports glob-type searches on the `name`, like `processed/**/*.fastq` or `processed/*-HS1-*`

The route supports pagination through `limit` and `offset` params.


metis/lib/query_builder.rb
```diff
@@ -0,0 +1,66 @@
...
+      @params = params
+    end
+
+    def build
+      @params.each do |param|
+        next unless param.key?(:attribute) && param.key?(:predicate) && param.key?(:value)
+
+        # Modify query behavior based on "type" of attribute
+        case param[:attribute]
+        when 'created_at', 'updated_at'
```
coleshaw, line 17: Are there other things that would be useful to search on?

corps:
Oh nice!  This is exciting to see, thank for digging into this dude.

metis/lib/folder_path_calculator.rb
```diff
@@ -0,0 +1,33 @@
...
+      # No parent folder, so just return this (root) folder's folder_name
+      return folder.folder_name if !folder.folder_id
+
+      # Use the cached path value for the parent folder if it exists
+      return "#{@path_cache[folder.folder_id.to_s.to_sym]}/#{folder.folder_name}" if
+        @path_cache.has_key?(folder.folder_id.to_s.to_sym)
+
+      # Find the path for the parent folder, recursively
+      parent_folder = @folders_by_id[folder.folder_id].first
+      "#{get_folder_path(parent_folder)}/#{folder.folder_name}"
```
corps, line 30: Minor nit; rather than recursion, what about a loop?  Recursion in languages like ruby make for ugly stack traces on error.  This is super super minor nit.

metis/lib/query_builder.rb
corps: What about "globs" searches that would allow parent folder filtering too?
For instance, my ide has two types of name searching

`abc.def` => matches any filename like "vabc.def" or "abc.defg"
`make/abc.def` matches any filename like "abc.def" but also within any directory structure containing "make"

I imagine wanting to, from the top of a data bucket, search for like "processed/HS103" to list all processed records in any structure for HS103

coleshaw: Ah, that would be neat. Let me look into that.

coleshaw:
I think this does globbing like `processed/*HS1-*` or `processed/**/*.fastq`

corps (approved):

PR #82, opened 2020-09-04
Timur attribute ordering

coleshaw:
This PR addresses mountetna/timur#277. This enforces the same attribute ordering that we're using in the model map page, extracts that code and puts it into a shared utility that the Browser Pane uses ... @graft , this shouldn't override any manually defined views, right, but it will sort any attributes that are passed in from that view? Perhaps that is not desired before and the sorting needs to happen higher up in the component tree...I will do some more testing and play around with it.

This works currently for default views. Not sure how it interacts with manually defined views, and I need to test that further.

### DIFF
--- a/timur/lib/client/jsx/utils/attributes.js
+++ b/timur/lib/client/jsx/utils/attributes.js
@@ -27,7 +27,6 @@ const order = (attribute, identifierFirst) => {
 };
 
 export const sortAttributes = (attributes, identifierFirst = false) => {
-  console.log('before sorting', attributes);
   const sortedAttributes = Object.values(attributes).sort(
     (a, b) =>
       order(a, identifierFirst) - order(b, identifierFirst) ||
@@ -41,6 +40,6 @@ export const sortAttributes = (attributes, identifierFirst = false) => {
       index_order: attribute.index_order ? attribute.index_order : index
     };
   });
-  console.log('done sorting', finalAttributes);
+
   return finalAttributes;
 };
--- a/timur/lib/client/jsx/utils/attributes.js
+++ b/timur/lib/client/jsx/utils/attributes.js
@@ -1,17 +1,27 @@
-const order = ({attribute_type}) =>
-  ({
-    parent: 1,
-    identifier: 2,
-    collection: 3,
-    table: 4,
-    child: 5,
-    link: 6,
-    file: 7,
-    image: 8
-  }[attribute_type] || 9);
+const order = (attribute, identifierFirst) => {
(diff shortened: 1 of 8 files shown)


timur/lib/client/jsx/utils/attributes.js
```diff
@@ -0,0 +1,17 @@
...
+    identifier: 2,
+    collection: 3,
+    table: 4,
+    child: 5,
+    link: 6,
+    file: 7,
+    image: 8
+  }[attribute_type] || 9);
+
+export const sortAttributes = (attributes) =>
```
corps, line 13: Could we use this function [here](https://github.com/mountetna/monoetna/blob/master/timur/lib/client/jsx/selectors/search.js#L78) as well, for even mooore consistency?

corps (approved):

graft:
Hmm, I am a bit confused by this since the ViewAttribute includes an explicit `index_order`, but this is apparently unused in BrowserPane, which just seems to use Object.keys(attributes), i.e., no order at all. Calling `sortAttributes` instead therefore doesn't seem like it could break anything, but the existing behavior doesn't seem to make use of the available `index_order` in the ViewAttribute model, as it ought to, and now `sortAttributes` takes the place of such ordering.

coleshaw:
> Hmm, I am a bit confused by this since the ViewAttribute includes an explicit `index_order`, but this is apparently unused in BrowserPane, which just seems to use Object.keys(attributes), i.e., no order at all. Calling `sortAttributes` instead therefore doesn't seem like it could break anything, but the existing behavior doesn't seem to make use of the available `index_order` in the ViewAttribute model, as it ought to, and now `sortAttributes` takes the place of such ordering.

Hm, so that's interesting. On a default view, with no custom view defined, I don't see any `index_order` on the list of pane attributes.  I also don't see how to get `index_order` [from the template](https://github.com/mountetna/timur/blob/master/lib/client/jsx/selectors/tab_selector.jsx#L62), since that information isn't present in the template. So somehow `sortAttributes` or equivalent would have to know if a defined View were being used and stick to the order defined in the view, or set a default ordering otherwise (otherwise it may re-order something defined by ViewAttribute). Did I understand that right?

Let me see if I can move the method call to the selector, maybe that makes more sense.

coleshaw:
I think the ViewAttribute `index_order` is implied in the order of the Hash keys returned by the API....so it just happens to work out because of how the backend returns things when the view is defined. But BrowserPane doesn't reinforce or check that the order matches `index_order`. I've updated BrowserPane to respect the `index_order`.

PR #83, opened 2020-09-08
Zc/chef docker test

corps:
The goal of this PR is to enable a full docker deployment of the polyphemus app.  That includes achieving a few sub goals:

1.  Adding the web server / database components of polyphemus app.  Originally we skipped that as it was just running some background jobs.
2.  Adding github action workflow to push production and staging taged images when pushed to those branches respectively.
3.  Merging master also pushes to staging by default.
4.  A new cookbook, `mountetna`, meant as a future replacement / compliment of existing chef cookbooks, with resources for deploying docker etna-apps.
5.  a kitchen.yml file and associated tools that allows quickly testing the new cookbook using local docker.  A bit faster to iterate than the default VM (vagrant) approach.
6.  Automatic deployment via `watchtower`, updates anytime new images appear, and notifies slack.

Merging this should have no direct impact to the current deployment.  I intend to take and copy the cookbooks into cbc-chef once we're cool (they're in monoetna right now mostly for convenience) and we can test a docker based deployment.

This is not really higher priority than comet tail, but it's some of the side work I've been doing.  It's quite close to us being to roll out docker deployments!


corps:
This pr is big because of the vendor import of `chef/docker` cookbook.  Feel free to ignore all those files.

PR #85, opened 2020-09-10
Adding commands that enable copying models between magma instances

corps:
The goal of this PR is to enable the ability to shallow and deep copy models between magma instances.

It works!  But, the target magma instance needs the magma changes to validation in this PR to work, so if you check out you should copy TO your development instance  FROM production to test.

example:

```
make -C polyphemus bash
SKIP_RUBY_SETUP="" /entrypoints/development.sh # This pulls in the local etna gem changes.  bash command normally does not.

export SOURCE_TOKEN=<your production janus token>
export TARGET_TOKEN=<your dev janus token>
export TOKEN="$TARGET_TOKEN"
./bin/polyphemus api_add_project mvir1_copy
./bin/polyphemus api_copy_model https://magma.ucsf.edu https://magma.development.local mvir1 mvir1_copy project # copy recursively from project, the root model.  copies everything!

./bin/polyphemus api_copy_model_shallow https://magma.ucsf.edu https://magma.development.local ipi mvir1_copy rna_seq # copy just the rna_seq model and its linked models from ipi into your mvir1_copy!
```

The Workflow class is extensible; for instance, we may want to map model or attribute names differently during the copy, so we can easily create subclasses to handle that.

I might want to change the command names and add "magma" to them to clarify.


magma/lib/magma.rb
```diff
@@ -118,17 +118,6 @@ def validate_attributes(model)
     model.attributes.each do |attribute_name, attribute|
       # Check that attribute has a column in the model's table
       raise Magma::ValidationError, "Missing column for #{model}##{attribute_name}." if attribute.missing_column?
-
-      next unless attribute.respond_to?(:link_model)
```
corps, line 122: This validation is masively problematic.  For one, this validation does not exist in the actual add or modify attribute endpoints, so it doesn't actually prevent one from creating links with reciprocals.  In fact, quite the opposite, it prevents the server from starting up once a link has been created.  The net effect is that it doesn't prevent bad links, it prevents the server from working if you add one side of the link at a time.

I literally can't think of a good reason to keep this, since it has thus far wasted at least 3 hours of my time across 2 seperate production instances of this flipping, and adding it back in here would just increase the complexity of clients, who must add link attributes together in a single transaction or risk breaking magma until manual intervention.

coleshaw: Does this shift the breaking behavior? -- if you only have one side of a link, does Magma start up, and only break when you try to access the link?

etna/lib/etna/clients/magma/workflows.rb
```diff
@@ -0,0 +1,146 @@
...
+          add_attribute.type = source_attribute.attribute_type
+          add_attribute.description = source_attribute.desc
+          add_attribute.display_name = source_attribute.display_name
+          add_attribute.format_hint = source_attribute.format_hint
+          add_attribute.hidden = source_attribute.hidden
+          add_attribute.link_model_name = source_attribute.link_model_name
+          add_attribute.read_only = source_attribute.read_only
+          add_attribute.unique = source_attribute.unique
+          add_attribute.validation = source_attribute.validation
+          add_attribute.restricted = source_attribute.restricted
```
coleshaw, line 80: I think I've also seen attribute `options` -- is that something that we would want copied here, too?

coleshaw:
This is a great idea. Keeping the models in sync with production is definitely a challenge.

magma/lib/magma.rb
graft: While irritating, this is an important requirement for the graph. Without reciprocal linking it is possible to filter Y on an attribute of X but not X on an attribute of Y (or X >  Y > Z). If we don't have this requirement then we will end up with broken graphs.

As i think i noted in ali's original add_attribute PR, the api is broken for link attributes and we need a new link_model command which can handle the reciprocal link requirement.

corps: > Does this shift the breaking behavior? -- if you only have one side of a link, does Magma start up, and only break when you try to access the link?

Nope.  All functionality works even with only one side of the linking.  There isn't any problem functionally with just one side of the link.

corps: > While irritating, this is an important requirement for the graph. Without reciprocal linking it is possible to filter Y on an attribute of X but not X on an attribute of Y (or X > Y > Z). If we don't have this requirement then we will end up with broken graphs.
> 
> As i think i noted in ali's original add_attribute PR, the api is broken for link attributes and we need a new link_model command which can handle the reciprocal link requirement.

What about a logger warning then?  As it stands the entire server goes down and stays down until you manually edit entries and tables in the database, which is pretty painful.  If we remove the validation, the server can a tleast come up and fixing the link is easy (just use the api command to add the next link).

Also adding the validation to the add attribute endpoint would get fairly complicated; you'd have to be able to specify options for two separate attributes in one command.  

In truth, I think it's perfectly reasonable to say that you can add one side of a link at a time with separate commands, minus validation.  There is at worst a conceptual downside but no functional downside.

etna/lib/etna/clients/magma/workflows.rb
corps: Good catch.

magma/lib/magma.rb
graft: There is a functional downside, which is that you can't search the graph in the other direction. Allowing people to make one-way links means we will end up with one-way links in real data graphs which won't be searchable in both directions.

The solution to this problem is to fix add_attribute to not allow adding link attributes, and add another command to create links between models which can handle the reciprocal logic (similar to the way add_model works). Removing the reciprocal link requirement is not an option.

graft: Also, I should note: it is possible to add two attributes at once in a single transaction and have this work; as long as the reciprocal link requirement is satisfied the server will come up correctly. Pretty unsafe since it is also possible to create an invalid state currently.

corps: > There is a functional downside, which is that you can't search the graph in the other direction. Allowing people to make one-way links means we will end up with one-way links in real data graphs which won't be searchable in both directions.

It sounds like the reciprocal link requirement is a modeling, but not system requirement.  This is similar to requirements like, "you should spell things correctly" and "you should denormalize things correctly".  

I'd like to compare a one-way link to a mispelled attribute for a moment.  One, a mispelled attribute is probably by all accounts more likely and common, and causes more user distress than one-way links.  Secondly, a one-way link is actually easier to fix than a mispelled attribute; we can always use a command to add the reciprocal linking trivially, it's a pure add change.  In other words, there is not much value in aggressively preventing this state, because fixing it is still no-cost, and it has no operational impact whatsoever.

Considering your proposal.
 Adding a link is nearly the same as adding any particular attribute in ever way, but now we'd be differentiating it by call convention, attributes, and potential failure cases.  So for instance, every magma client must now support an additional add link endpoint in addition to the existing add attribute endpoint.  The copy workflows need additional logic and parameter massaging unique to links (right now, I can easily copy attribute model fields from one to the other without special handling differentiated based on attribute type.  now I will need additional separate code paths that both must be correct and maintained).  The backend must also now mostly have duplicated endpoints with similar functionality and behavior.  If add_attributes and add_links ever further deviate in their behavior or calling convention, even more work must be done on clients to account for this.  That's fine when there is a functional difference that adds value, but I'm unconvinced this validation really does.

We could allow the linking to remain in add_attribute, but push the validation there and require that, within a single transaction, add_attributes add up to reciprocal link.  But that also increases worker and client side complexity for little gain.  Batching the correct attributes together and in the correct order would involve all clients "solving" for the reciprocal link by scanning and searching the current server state to identify if the reciprocal link requirement has already been met, or if it needs to find a potential recirpocal link attribute from the source model that needs to be special case batched (but not applied on its own like other attributes).  The server does this, but only in the binary sense: "does there exist a reciprocal link?"  But clients must solve for the harder problem: "Does there exist recipocal links, _and if not what attributes should be copied and to where in order to add the correct reciprocal of this link in particular?_".   

I feel that if the main concern is that in the future, we want to ensure correct modeling using some widely distributed tool, we move this concern _to that tool_, in much the same way we'd move other modeling concerns: checking for spelling (not something we should enforce in our backend), encouraging re-use of existing models (not validating the uniqueness of each model put into our system).  Furthermore, as I noted, making a non-reciprocal link is still easy to fix after the fact, with minimal downside in the mean time.  Of all the modeling problems that could go wrong, it's the lowest hanging fruit, whereas the cost of pushing the validation forward actually does impact all tools complexity.

etna/lib/etna/clients/magma/workflows.rb
corps: Ah, so `options` and `match` are read only from the json_template; they are actually both derived from the `validation` input which we set.

PR #87, opened 2020-09-17
Metis client find

coleshaw:
This PR adds support for the `find` route on Metis, to the Etna gem's Metis client. There is also an example in Polyphemus commands on how to use it.


corps (approved):
Cool!

PR #88, opened 2020-09-22
Link flow wsp to records

coleshaw:
This PR adds a Polyphemus command to link Flow WSP files in the IPI `integral_data` bucket to Magma patient records.

This command assumes that the WSP file names are well-formed (i.e. `IPIPROJ001.wsp`), so it will only work after all of this cleaning is done. For example, currently there are WSP files with non-standard names, like `IPIPROJ001_timur.wsp`, which would break this script.


etna/lib/etna/clients/magma/workflows/crud_workflow.rb
```diff
@@ -11,7 +11,7 @@ def initialize(args)
         end
 
         def lookup_record(model_name, record_id)
-          if (cached = @cache.dig(model_name, record_id))
+          if (cached = (@cache ||= {}).dig(model_name, record_id))
```
coleshaw, line 14: I was getting an error about `nil Class` for `@cache`, so maybe an issue with doing a lookup before an update?

corps: Ah, doh!  Yeah, makes sense.  I think I added the cache mechanism as a last minute optimization and missed this.

polyphemus/lib/ipi/flow_wsp_file_linker.rb
```diff
@@ -0,0 +1,58 @@
...
+    super(**{bucket_name: 'integral_data', model_name: 'patient'}.update(opts))
+  end
+
+  def find_matches
+    {}.tap do |all_matches|
+      metis_client.find(
+        Etna::Clients::Metis::FindRequest.new(
+          project_name: project_name,
+          bucket_name: bucket_name,
+          params: [Etna::Clients::Metis::FindParam.new(
```
corps, line 15: Oh nice, using the metis command makes this much faster then!

corps: I was thinking in the near future we could add "cursors" (last updated at scanned) to make all these etl-like jobs run faster, too.  They'd only have to scan files that have updated since the last attempt.

corps (approved):
I like!  Cool!

polyphemus/lib/ipi/flow_wsp_file_linker.rb
coleshaw: Oh yeah, that would be great!

PR #89, opened 2020-09-23
ETLs

corps:
Plan is to add two ETL classes,
`MetisFileETL` => continuously processes files, keeping track of the last updated_at, ensuring future scans are fast.
`MagmaRecordETL` => similar but for magma records.

Also added some more conveniences around the app's clis:

1.  Added `magma` `janus` `metis` and `polyphmus` scripts to the `bin/` that invoke docker and run it more conveniently (no need to run make -C magma bash + ./bin/magma, it just becomes ./bin/magma`
2.  Added the ability to auto generate bash completion scripts for etna commands, and extend that autocompletion easily.


etna/lib/etna/application.rb
```diff
@@ -13,9 +13,16 @@ def self.included(other)
...
   def self.find(klass)
-    Kernel.const_get(
-      klass.name.split('::').first
-    ).instance
+    namespace = klass.name.split('::').first
+    if (namespace_klass = Kernel.const_get(namespace)) && (namespace_klass.respond_to? :instance)
+      return namespace_klass.instance
+    end
+
+    if (app_name = ENV['APP_NAME']) && (app_klass = Kernel.const_get(app_name.capitalize)) && (app_klass.respond_to? :instance)
```
corps, line 21: These changes make it possible to use Etna::Command subclasses that are not specific to any given app namespace by looking up the app class via `APP_NAME` environment variable.

etna/lib/etna/application.rb
```diff
@@ -80,7 +87,7 @@ def run_command(config, cmd = :help, *args)
     cmd = cmd.to_sym
     if commands.key?(cmd)
       commands[cmd].setup(config)
-      commands[cmd].execute(*args)
+      commands[cmd].execute(*commands[cmd].fill_in_missing_params(args))
```
corps, line 90: Some etna commands take a lot of parameters.   With this change, if you forget to fill out all parameters, it will prompt you for the missing params one at a time.  Makes it super easy to use commands from the CLI.

bin/janus
```diff
@@ -0,0 +1 @@
+run-command
```
corps, line 1: Check it!  I make ./bin/janus and ./bin/magma etc files that can be used outside of `make -C metis|janus|magma bash`.  It's just a convenient wrapper.

etna/lib/etna/client.rb
```diff
@@ -96,7 +96,12 @@ def body_request(type, endpoint, params = {}, &block)
 
     def query_request(type, endpoint, params = {}, &block)
       uri = request_uri(endpoint)
-      uri.query = URI.encode_www_form(params)
+
+      if uri.query
```
corps, line 100: Necessary as part of adding upload and download to metis clients in the etna gem.  Because the upload url alraedy encodes hmac params, this logic ensures that any query that is part of the endpoint url is appended to, not removed, by other given params.

etna/lib/etna/client.rb
```diff
@@ -24,7 +24,7 @@ def route_path(route, params)
 
     def multipart_post(endpoint, content, &block)
       uri = request_uri(endpoint)
-      multipart = Net::HTTP::Post::Multipart.new uri.path, content
+      multipart = Net::HTTP::Post::Multipart.new uri.request_uri, content
```
corps, line 27: Necessary to keep, not drop, any query params that exist on the POST (which happens for upload_blob)

etna/lib/etna/clients/metis/client.rb
```diff
@@ -47,6 +48,45 @@ def find(find_request)
           @etna_client.bucket_find(find_request.to_h))
       end
 
+      def download_file(file_or_url = File.new, &block)
```
corps, line 51: Now support upload / download endspoints in the etna metis client!  Makes automation a lot easier.
In the future, I think we can consolidate and make the metis_client just depend on the etna gem and use this client directly.

Advantage is that we can re-use this logic in automation, and consolidate on a single implementation that gets lots of active testing.

etna/lib/etna/clients/metis/models.rb
```diff
@@ -64,7 +64,7 @@ def to_h
         end
       end
 
-      class FindRequest < Struct.new(:project_name, :bucket_name, :params, keyword_init: true)
+      class FindRequest < Struct.new(:project_name, :bucket_name, :limit, :offset, :params, keyword_init: true)
```
corps, line 67: Support paging in the find model.

etna/lib/etna/clients/metis/workflows/metis_download_workflow.rb
```diff
@@ -0,0 +1,37 @@
...
+  module Clients
+    class Metis
+      class MetisDownloadWorkflow < Struct.new(:metis_client, :project_name, :bucket_name, :max_attempts, keyword_init: true)
+
+        def initialize(args)
+          super({max_attempts: 3}.update(args))
+        end
+
+        # TODO: Might be possible to use range headers to select and resume downloads on failure in the future.
+        def do_download(dest_file, metis_file, &block)
```
corps, line 16: The workflow encapsulates the metis client and file writing logic all together.

etna/lib/etna/clients/metis/workflows/metis_upload_workflow.rb
```diff
@@ -0,0 +1,137 @@
...
+  module Clients
+    class Metis
+      class MetisUploadWorkflow < Struct.new(:metis_client, :metis_uid, :project_name, :bucket_name, :max_attempts, keyword_init: true)
+
+        def initialize(args)
+          super({max_attempts: 3, metis_uid: SecureRandom.hex}.update(args))
+        end
+
+        def do_upload(source_file, dest_path, &block)
+          upload = Upload.new(source_file: source_file)
```
corps, line 17: This workflow encapsulates the entire upload workflow, including retry & creating the containing directory if necessary.

corps: Most of this log is adapted from metis_client.

etna/lib/etna/command.rb
```diff
@@ -5,11 +5,72 @@ class Command
...
+
+    def fill_in_missing_params(args)
+      req_params = method(:execute).parameters.select { |type, name| type == :req }
+      args + req_params[(args.length)..(req_params.length)].map do |type, name|
+        puts "#{name}?"
+        STDIN.gets.chomp
+      end
+    end
+
+    def completions
```
corps, line 38: Support for generation bash completions!

magma/lib/magma/storage.rb
```diff
@@ -19,9 +19,6 @@ def download_url project_name, path
     def upload_url(project_name, path, params={})
     end
 
-    def setup_uploader(model, name, type)
```
corps, line 22: This, and the uploader classes, are no longer referenced or used by any active code paths.

metis/lib/query_builder.rb
```diff
@@ -4,44 +4,51 @@
...
     def build
+      result = @base_query
       @params.each do |param|
         next unless param.key?(:attribute) && param.key?(:predicate) && param.key?(:value)
 
         # Modify query behavior based on "type" of attribute
         case param[:attribute]
         when 'created_at', 'updated_at'
-          @base_query = @base_query.where{|o|
+          result = result.where{|o|
```
corps, line 22: Don't mutate the base_query variable -- use a local variable instance to ensure that successive calls to build don't continuously mutate all previous query builds.

metis/lib/query_builder.rb
```diff
@@ -4,44 +4,51 @@
...
-            @base_query = @base_query.where(
+            result = result.where(
               [[Sequel[id_param], globber.folder_path_ids]]).where(
               Sequel.like(model_name_attribute, globber.sql_search_string))
           end
         end
       end
-      @base_query
+
+      result = result.order(:updated_at, :id)
```
corps, line 48: Push down limit and offsets into the actual querying, improving performance for large folders or searches that could be paginated client size for better peformance.

Strictly necessary for the ETL.

metis/lib/server/controllers/bucket_controller.rb
```diff
@@ -77,29 +77,25 @@ def find
...
+      limit: limit,
+      offset: offset,
     )
 
     results = query.execute
 
     files = results[:files]
     folders = results[:folders]
 
-    limit = limit ? limit : [files.length, folders.length].max
```
corps, line 88: limit and offset no longer applied at this leve, but in the query object instead.

polyphemus/db/migrations/002_create_cursors.rb
```diff
@@ -0,0 +1,14 @@
+Sequel.migration do
```
corps, line 1: Cursors keep the state of ETL jobs so that only newly updated files are actually searched and processed.

polyphemus/lib/commands.rb
```diff
@@ -177,14 +177,66 @@ def add_string_attribute(attributes, name)
     end
   end
 
+  class CopyMetisFilesCommand < Etna::Command
```
corps, line 180: New command that copies files from one metis environment into another, filtered by a glob.  I'm using this to test against production flojo files in my development environment.  I intend to delete all data after finishing testing.  Still very useful for testing things like linking data locally without risking hurting production.

polyphemus/lib/commands.rb
```diff
@@ -177,14 +177,66 @@ def add_string_attribute(attributes, name)
...
+    end
+  end
+
   module CopyCommand
     def self.included(mod)
       mod.instance_eval do
         include WithEtnaClientsByEnvironment
         include WithLogger
 
-        usage "#{name.snake_case.split(/::/).last} <source_env> <target_env> <source_project> <target_project> <model_name>"
```
corps, line 186: This sort of usage is no longer necessary, it is autogenerated via meta programming with params.

polyphemus/lib/etls/stub_metis_file_etl.rb
```diff
@@ -0,0 +1,10 @@
+require_relative '../metis_file_etl'
+
+class Polyphemus::StubMetisFileEtl < Polyphemus::MetisFileEtl
+  def initialize
```
corps, line 4: To replace with other etls (such as the new bulk rna linker, etc).  This is a test etl.

polyphemus/spec/metis_file_etl_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+    def process(cursor, files)
+      @process_calls ||= []
+      @process_calls << [cursor.updated_at.dup, cursor.value.dup, files.dup]
+    end
+  end
+
+  # To re-record this test, add RERECORD=1 and METIS_TOKEN=xxxxx to your environment before running.
+  it 'should run end to end' do
+    VCR.use_cassette('metis_file_etl.e2e') do
+      metis_client = Etna::Clients::Metis.new(host: 'https://metis.development.local', token: ENV['METIS_TOKEN'])
```
corps, line 18: VCR allows me to run a local test against live servers, record the interactions to a file, and then use that file for future tests (which don't require the live servers anymore).  Allows very simple to setup integration testing.

metis/lib/models/file.rb
```diff
@@ -168,6 +168,7 @@ def to_hash(request: nil, file_path: nil)
       file_path ||= self.file_path
 
       {
+        id: self.id,
```
coleshaw, line 171: What is the use case for sending database Ids back to the user?  I thought we didn't really want consumers relying on that for anything...

polyphemus/spec/metis_file_etl_spec.rb
coleshaw: Will you be committing the cassettes to the repo?

corps: Yes, I will.

metis/lib/models/file.rb
corps: Because updated_at is not monotonically increasing (it's possible to have many records share the same updated_at), the id is needed to differentiate and order records in case of a tie on updated_at.

bin/janus
graft: Hmm, for me there is always a considerable amount of overhead (~10s) + output when I run make bash, so I usually just keep it open while running specs or whatnot. Is this avoidable?

metis/lib/models/file.rb
coleshaw: I thought that the pagination and ordering in QueryBuilder / Query happens on the Sequel dataset before this method is called, not on a set of File hashes....so this shouldn't be required for that functionality to work? Or do you mean that `id` would be used to differentiate on the consumer side, like an ETL script?

bin/janus
coleshaw: I feel like before we changed to a shared image, it was super fast to bring up a bash terminal because it didn't re-run the `build.sh` script every time?

metis/lib/models/file.rb
graft: The path itself might suffice to provide that ordering and avoid exposing the id. On the other hand maybe it is nice to start to use the id on occasion and avoid unnecessary folder lookups.

polyphemus/db/migrations/002_create_cursors.rb
```diff
@@ -0,0 +1,14 @@
+Sequel.migration do 
+  up do
+    create_table(:cursors) do
+      primary_key(:id)
+      String :name, unique: true, null: false
+      column :value, :json, null: false
+      DateTime :updated_at, null: false
```
graft, line 7: Hmm, might these be less generic than "name", "value" - like "job_name", "completion_status" or some such?

polyphemus/spec/metis_file_etl_spec.rb
graft: While it sounds dumb, we should guard against the possibility of PHI ending up in these cassettes, which might happen if we are pulling live data. Not sure how to do this other than vigilance.

coleshaw: Should we set this test to a not-real-bucket but a test project and test buckets, as part of that? We would all have to synch on the structure and contents of that...

metis/lib/models/file.rb
corps: You are right though I can use the file path instead, I'll remove again.

polyphemus/db/migrations/002_create_cursors.rb
corps: For name, the name is unique only to the cursor but not the job.  As written, it is possible have multiple cursors per job.  An example for this is a job that processes multiple buckets.  Each bucket would receive different cursors, but a single job class could process each of those separate cursors.  For instance, if we have a common job that applies to all projects.

For value, yeah it's a little ambiguous what name to give it; in general it's just the json payload associated with the cursor.  The idea is that cursors for different types of jobs (such as magma, redcap, metis, or some other source) may have an arbitrary bag of attributes needed to correctly scan.

corps:
Adding a few more abstractions / tests here, wrapping up soon.

bin/janus
corps: You can still keep the persistent console open, nothing to prevent that.  This interface is just "in addition".  It's most useful for when I want to quickly run a janus one-off but I don't need that persistent console.

Yeah, I could probably do some work to reduce the amount of rebuilding the system does.  The shared image work reduced building in some cases, but the etna gem rebuild is triggered aggressively, I think, due to it not filtering files when it checks for changes.  I can sneak that change in soon.

polyphemus/spec/metis_file_etl_spec.rb
corps: In this case, interestingly enough, no file contents are actually pulled -- only the file metadata (path, download url, updated_at).  Check the cassette file to double check.

In general I agree that we should create some tools to create useful test data to reduce this risk.  This test, as written, can actually be run against any source of data anyways, it's expectations aren't particularly tied to any dataset.

bin/janus
corps: Actually, I now see what may be causing the excessive rebuilds.  It's tracking every node_modules change as needing a rebuild of the base image, but because that directory is mounted, every restarted updates it anyways.  I can filter that in this PR to make that issue go away.

polyphemus/spec/metis_file_etl_spec.rb
corps: What about potentially encrypting the cassettes with a secret?  CI and local users can decrypt using some common secret, but the committed file is opaque.

docker/build_image
```diff
@@ -81,15 +81,15 @@ function addToDigest() {
 }
 function recursiveDigest() {
   if [[ -n "$digestOnly" ]]; then
-     find "$1" -type f -exec md5sum {} \;
+     find "$1" -type f ! -path '*/node_modules/*' -exec md5sum {} \;
```
corps, line 84: This fixes base image builds by filtering node_modules from trigger re-builds.  These are mounted anyways so no point in rebuilding the base image here.

etna/lib/etna/clients/magma/workflows/record_synchronization_workflow.rb
```diff
@@ -0,0 +1,57 @@
+require 'ostruct'
+
+module Etna
+  module Clients
+    class Magma
+      class RecordSynchronizationWorkflow < Struct.new(:target_client, :source_client, :project_name, keyword_init: true)
```
corps, line 6: Starting place for copying magma records between environments.  Kind of simple at the moment, doesn't handle tables correctly currently.  I'd like to pair this with a smart fuzzier to increase safety against PHI in a future PR.

magma/lib/magma/censor.rb
```diff
@@ -23,6 +23,19 @@ def censored_reasons(model, record_set)
         restrict: true
       ).answer.map(&:last)
 
+      existing_identifiers = Magma::Question.new(
```
corps, line 26: This fixes a bug I discovered where a non priveledged user could not create new magma records since it would trigger the restricted check in the censor.  Matching spec found.

The idea is to identify restricted records as records that both don't show up from a restricted query AND already exist in the database.

magma/lib/magma/query/predicate/model.rb
```diff
@@ -129,8 +129,15 @@ def select
       [ column_name.as(identity) ]
     end
 
-    def column_name
-      Sequel[alias_name][@model.identity.column_name]
+    def column_name(attribute = @model.identity)
```
corps, line 132: These improved methods make selecting by an attribute on the model easier, which is necessary for changes made to `question.rb`

magma/lib/magma/query/question.rb
```diff
@@ -168,72 +189,115 @@ def count_query
       end
 
       query.select(
-        *@start_predicate.select
+          *order_by_column_names.zip(order_by_aliases).map { |c, a| c.as(a) }
       )
     end
 
     # get page bounds for this question using @options[:page] and @options[:page_size]
     def bounds_query
```
corps, line 197: bounds query now respects the 'order' parameter(s)

magma/lib/magma/query/question.rb
```diff
@@ -168,72 +189,115 @@ def count_query
...
         if bounds[index + 1]
-          bound[:upper] = bounds[index + 1][@start_predicate.identity]
+          bound[:upper] = bounds[index + 1].map { |c| row[c] }
         end
 
         bound
       end
     end
 
+    def apply_multi_stage_ordering_bounds(query, upper: nil, lower: nil)
```
corps, line 235: Now that bounds is a tuple of potential multiple values (like an ordering on updated_at, id), we need a more correct way of selecting values between multi-column intervals.  This is not as simple as selecting where(a >= 1 & b >= 2), since there is no guarantee the total ordering of a and b coincide.  Instead, we need to select where for each element of the tuple, either being  greater or equal to  the lower bound && either each additional tuple element being greater or equal, or the previous tuple elements being strictly greater.  

The logic is kind of noted in the comments, it's a tricky looking algorithm but works.

metis/lib/server/controllers/metis_controller.rb
```diff
@@ -54,20 +53,24 @@ def file_hashes_with_calculated_paths(bucket:, offset:, limit:, files:)
...
 
       # Sorting folders by depth level makes some subsequent calculations simpler,
       #   especially when not paging. Shallow -> deep
-      sorted_folders = all_folders ? sort_folders_by_depth(
-        all_folders: all_folders,
-        limit: limit,
-        offset: offset
-      ) : target_folders
+      if all_folders
+        sorted_folders = sort_folders_by_depth(
```
corps, line 64: A bit messy, but basically on a paged query first sort by the order_by given, but on a full query also sort by depth.

polyphemus/spec/metis_file_etl_spec.rb
coleshaw: Would that process be automated for whenever we re-record the cassette? And automated decryption for local testing?

magma/lib/magma/censor.rb
coleshaw: Nice find!

magma/spec/censor_spec.rb
```diff
@@ -51,6 +51,23 @@ def to_record_set(model, revisions)
     )
   end
 
+  it 'new record revisions not censored for users with restricted permissions' do
```
coleshaw, line 54: Nit: the name of this test seems confusing ... because the assertion is that the revision is censored? Should the user have an `A` permissions and no censored reasons?

polyphemus/polyphemus.completion
```diff
@@ -0,0 +1,84 @@
+#!/usr/bin/env bash
```
coleshaw, line 1: In general, are these `.completion` files auto-generated, or do we manually update? If we add a new command or change arguments, for example?

coleshaw:
A meta-question about MetisFileEtl ... if we run something like the BulkRnaFileLinker as an ETL on a schedule, would that mean the source of truth for linking files to Magma records is the folder hierarchy in Metis, and not necessarily Magma itself? So if someone wants to make a change for some reason (like they find out a file is assigned to the wrong patient due to a typo, or a file is invalid because the sample had too few cells), they need to make the change in Metis and in Timur (so the ETL picks up the corrected file in Metis and assigns to the right record, and then deletes the old reference via Timur)? Making the change in Timur only might cause the change to be overwritten on the next ETL run? Just want to make sure I understand, since we may have to teach users a new, with-ETL workflow if it is different than the current one.

magma/lib/magma/censor.rb
graft: Hmm, perhaps a bad result of having most of the update_spec tests run with the permissions `editor: { perm: 'E:labors' }`. There appear to be serious problems here, even with @corps's fix, there are many failing tests (besides the censor ones) if you change the above perm to `e:labors`. I'm investigating...

magma/spec/censor_spec.rb
corps: It's a bit misnomer, the spec ensures that the censor still enforces the attribute restriction but not the record one (which was the only way I could get it to ensure that it's not because the user skipped the entire restriction from `restrict?`).  I'll update the name and add a comment.  

In other words, the test is set up this way to verify the underlying assumption: the user IS being restricted, but NOT because of the record level one.

graft:
I agree, an excellent question that cuts to a really fundamental issue about where source-of-truth is. The same issue also applies for Redcap, which is source-of-truth for a lot of stuff. It's not clear what should happen here, and I think the larger issue bears a bunch of discussion.

I've had the vague notion that Magma ought to be source-of-truth, but since Magma's main function seems to be to agglomerate and organize, it may turn out that source-of-truth, which is often to be really found in obscure documents, people's memories, etc., doesn't actually belong in Magma, and all we can hope for there is to represent a clean, valid, well-organized version of truth-as-we-know-it, but that ultimately good data has to flow in from somewhere else.

To put it another way, in order for data in Magma to be correct, it has to receive correct inputs. This means the source-of-truth always has to exist outside of Magma first, before the truth can be propagated into magma.

However, while this is true I don't think it necessarily means that *maintenance* of Magma has to come externally, and if it does then all of the careful validation and organizing we do in Magma is for naught, and redundant. For this reason I'm uncomfortable having continuous dumps in Metis that are meant to be well-formed, since it means we're pushing the task of determining what is well-formed elsewhere, in addition to Magma.

This seems bad, because it requires us to maintain two (or more) different sets of validations, and because it means we are trying to control what is in Magma not using the Magma API, but a subset of the API available via whatever ETL process is in play, which may not be able to accomplish everything (in the example above, the ETL process may not be able to "sync" with magma and update references as @coleshaw suggests).

There are a bunch of different ways we might solve this, the main opponency I think is whether we rely on a well-formed Metis hierarchy and a reliable, continuous ETL process, or whether we treat Metis as an ad-hoc tool for loading (i.e., doesn't run continuously) and use it to bounce badly-formed data off Magma's basalt walls.

Obviously, since this comment is so long, it bears a lot more discussion before we figure this out.

corps:
Not to pick a side, but some interesting observations about the implications of various approaches:

1.  IFF Magma is source of truth, then magma also needs revision history.  It would be too fragile if our source of truth's adhoc edits were easily lost to a tool that ran and aggressively updated things (we have a bunch of these floating around, even aside from etls).  Right now, we get away with aggressive bulk modifications and no revision history _because_ the source of truth is external from magma, and magma can be refreshed.  This makes magma the *cache and indexing system* of multiple input sources.

2.  IFF Metis and external systems are source of truth, Magma needs a notion of "local overrides".  A local override is a change made in Timur which has greater precedence than the original source of truth.  For instance, editing an attribute to a magma record generally isn't easy to back propogate to redcap, so in order to preserve that change in the face of the future, we need some way to undertand that change as having precedence.  In general, Magma local CRUD has less value in this world, changes should generally propogate from the source of truth itself.  This increases the value of Redcap and possibly reduces our responsibilities as a data library.

I think one question that helps us understand which world we should tilt is this:
Is data library more valuable as a means of organizing external data for the sake of READS primarily (#2) or correcting and making adhoc fixes to external data after the fact (#1)?  These aren't strictly opposed views, but which direction you lean changes which workflow should be forefront in our arhictecture.

corps:
My theory is that data already has sources of truth, and that if magma is to be a source of truth, we need to convince people to look in magma and NOT redcap or local files to answer a question.

the only worse than heterogenous sources of truth is overlapping sources of truth.

dtm2451:
Naively perhaps, I'd also throw my hat in for No2 [of here](https://github.com/mountetna/monoetna/pull/89/#issuecomment-702284507).

Having metis (+ the metisETL logic layer?) be the primary source of truth is also helpful as this is what the Data Scientists would probably prefer to interact with.  One major goal here is freeing up yall's time, right?

PR #90, opened 2020-09-23
Allow delete on table attribute

graft:
This PR allows us to delete tables via /update when we post a new table.

Previously I had established a new syntax for pushing tables: essentially, you create new entries in the table model. You may link these to an existing record in one of two ways: either by specifying the parent in each table record, e.g.:
```
update(
  'prize' => { '::temp1' => { labor: 'The Golden Apples of the Hesperides', name: 'apple of joy', worth: 3000 }, '::temp2' => { labor: 'The Golden Apples of the Hesperides', name: 'apple of joy', worth: 2000 } }
)
```
or by including a table attribute entry in the parent record, e.g.:
```
update(
  'labor' => {
    'The Golden Apples of the Hesperides' => { prize: [ '::temp1', '::temp2' ] },
  },
  'prize' => { '::temp1' => { name: 'apple of joy', worth: 3000 }, '::temp2' => { name: 'apple of joy', worth: 2000 } }
)
```

Previously these were equivalent, but now the two have different behavior. In the first case we will append to the existing table attribute (that is, the entries remain). In the second case, old entries will first be removed, then new entries will be added.

This is apparently a simple change, but required a bunch of background work which constitutes most of this PR (the actual table deletion code took about 1 minute to write). Viz:

1) Axe Magma::Revision and move most of its functionality into Magma::RecordEntry. Mostly this means the majority of parsing and validation happens within Magma::Loader, which is responsible for creating each Magma::RecordEntry. This removes a lot of ugly coupling and unnecessary manipulation of the revision format as it is passed around between classes.

2) Move the Censor functionality inside Magma::Loader. This does not run via Magma::Validation, unfortunately, because the Censor wants to bulk-query for restricted records, while the validator runs on individual records.

3) Create an "attribute hooks" functionality in the loader; previously, there was a file hook that ran within the Update controller. In fact this wasn't running in the right place since it would run *before* Magma::Loader#dispatch_record_set ran validations, and therefore could ask Metis to link files for an update that was doomed to fail. Now, instead, the hooks run within dispatch_record_set after the validations and censor have run. There are three types of hook introduced: `#load_hook`, `#bulk_load_hook` and `self.type_bulk_load_hook`, which allow some data-passing between them; this lets the attribute collect data across the entire update before it does its thing (if so desired). There are now hooks defined for `file` (the metis bulk copy stuff) and `table` (deletion).

4) Repair the temp_id system, much broken as a result of (1). I think I got it right?

Also incidentally fixes @mountetna/magma#202


coleshaw:
Hm, I can't set file attribute paths now -- get a vague "Server error":

```
magma_app_1      | WARN:2020-09-23T11:33:35+00:00 xca0c8 ["Server error."]
```

Will try to debug further so I can provide more detail ...

Also, just to clarify, the expectation for @corps 's flojo loader work is to use the second method outlined above to delete previous entries (specifying the table attribute entry in the revision)?

corps:
> Hm, I can't set file attribute paths now -- get a vague "Server error":
> 
> ```
> magma_app_1      | WARN:2020-09-23T11:33:35+00:00 xca0c8 ["Server error."]
> ```
> 
> Will try to debug further so I can provide more detail ...
> 
> Also, just to clarify, the expectation for @corps 's flojo loader work is to use the second method outlined above to delete previous entries (specifying the table attribute entry in the revision)?

To be honest, my approach with the flojo loader is that it won't be written in or on top of any magma logic itself.  It will simply be an elt that runs outside of magma using only magma apis.

corps:
Yeah I think this is the correct direction for sure.

magma/lib/magma/attributes/file.rb
```diff
@@ -68,10 +47,100 @@ def query_to_tsv(file)
       file ? file[:url] : nil
     end
 
-    def entry(value, loader)
+    def entry(file, loader)
+      value = case file[:path]
+      when '::blank'
```
corps, line 52: Does this work?  There's a return below so these expressions don't do anything I believe.

coleshaw:
So the issue is that the revisions now passed in (`attribute_copy_revisions`) are not in the expected format for Metis. They come in as RecordEntry format:

```
magma_app_1      | {:revisions=>[["metis://ipi/TestBucket/test3.txt", "metis://ipi/magma/patient-IPIGYN001-gross_specimen.txt"]]}
```

But they need to reformatted into something like:

```
{:revisions=>[{ :source=>"metis://ipi/TestBucket/test3.txt", :dest=>"metis://ipi/magma/patient-IPIGYN001-gross_specimen.txt"}]}
``

So I've added a commit with the following change to `attributes/file.rb`, around line 88:

```
    def self.type_bulk_load_hook(loader, project_name, attribute_copy_revisions)
      revisions = attribute_copy_revisions.values.map(&:to_a).flatten(1).map{|rev|
        {source: rev[0], dest: rev[1]}}
```
And that seemed to work. Also updated two specs related to this, so they'll catch it in the future.

corps:
> So the issue is that the revisions now passed in (`attribute_copy_revisions`) are not in the expected format for Metis. They come in as RecordEntry format:
> 
> ```
> magma_app_1      | {:revisions=>[["metis://ipi/TestBucket/test3.txt", "metis://ipi/magma/patient-IPIGYN001-gross_specimen.txt"]]}
> ```
> 
> But they need to reformatted into something like:
> 
> ```
> {:revisions=>[{ :source=>"metis://ipi/TestBucket/test3.txt", :dest=>"metis://ipi/magma/patient-IPIGYN001-gross_specimen.txt"}]}
> ``
> 
> So I've added a commit with the following change to `attributes/file.rb`, around line 88:
> ```
> 
> ```
> def self.type_bulk_load_hook(loader, project_name, attribute_copy_revisions)
>   revisions = attribute_copy_revisions.values.map(&:to_a).flatten(1).map{|rev|
>     {source: rev[0], dest: rev[1]}}
> ```
> 
> ```
> And that seemed to work. Also updated two specs related to this, so they'll catch it in the future.
> ```

I think this may be related to my above comment; there is a return statement that may be blocking the right values from being returned from `entry` method?

magma/lib/magma/attributes/file.rb
coleshaw: This seems to work, when I check in the database, for `::blank`, `::temp`, and `null`. `value` gets set to the right hash value depending on the case (and does not update the database for `::temp` but does return an upload URL).

corps: OHHHHH now I'm reading this right.  The value is being set to the expression of the case.

graft:
Ergh, meant to repair that source: dest: business but the test didn't catch it and I forgot - thanks for the fix!

coleshaw (approved):
:+1: Updating with master, looks good to me if tests pass afterwards.

PR #91, opened 2020-09-23
IPI new flow model

coleshaw:
This PR adds a Polyphemus class / command that does several things:

* Creates a new model for Flow, child of Sample
* For all existing Sample records, migrates any Flow data (file, quality flag, or notes for each stain) to the new Flow model and attaches it to a Flow record linked to the Sample
* Hides the Flow-related attributes on the Sample model.

It felt like a "migration" to me, so I've named it as such. If something else would be more appropriate or it feels confusing, let me know.


polyphemus/lib/ipi/migrations/01_flow_model_migration.rb
```diff
@@ -0,0 +1,161 @@
+require 'etna'
+
+class IpiAddFlowModelMigration
```
corps, line 3: Seems like this migration would blow up in any environment that did not have the ipi model already configured?  Thus this migration would break, for instance, bringing up a new development environment.

I'd feel more safe with this being an optional script that can be run manually, or if it can check some precondition before running (remember, this could even be run just as the magma environment is booting, and no api is even available yet).

corps:
OH.  Disregard my comment.  This isn't in the `db/migrations` directory so it won't be run automatically on polyphemus migrate command, which is what my fear was.  Ok, this is fine.

corps (approved):

PR #93, opened 2020-09-23
accept ranges header on downloads

graft:
Very simple, just adds "Accept-Ranges: bytes" header to responses from download_url. The result is you can, e.g., scrub back and forth in a video without having to download the whole thing. But also, maybe later, scrub back and forth in an fcs file.


coleshaw (approved):
That is some pretty awesome magic from x_sendfile. Works well for me, and looks like we do offer video streaming now!

PR #94, opened 2020-09-24
Add view update endpoint to timur

graft:
This is a piece of code I have had lying around for a while which lets you update a Timur view using a piece of JSON, here I am just exposing this via API, which should be sufficient basis to later build a basic view editing interface. Also much easier than hacking things into the database from the console.

### DIFF
--- a/timur/lib/server.rb
+++ b/timur/lib/server.rb
@@ -16,6 +16,8 @@ class Timur
       erb_view(:client)
     end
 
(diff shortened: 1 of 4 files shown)


coleshaw (approved):
Works well for me, :+1: . Super minor comment: I think it might help if we supply one of the existing `view.json` files to Anton as an example. I know the test has the expected JSON format, but I think it helps to be able to match the JSON format to what the view looks like in the browser, to understand what the different view parts are.

graft:
Yeah i have such an example I can give him.

PR #95, opened 2020-09-24
Cs/make file attrs blank

coleshaw:
This PR adds a polyphemus command and Etna workflow to help mark all `nil` File and Image attributes as `::blank`, for all records in a model -- so basically we know that any record without a file assigned in a File / Image attribute won't have one in the future.

For now, this will only be used for IPI Flow model, but could conceivably also be used for Patient or any other model that we know won't have new Files / Images added. Only really useful at the end of a project, when we're "closing it out".


corps (approved):

graft:
Hmm, so the usage would be, once I know my dataset is complete, I just blank the remaining file attributes?

coleshaw:
Yes. That's the request for the IPI Flow model, once we finish this last push to clean stuff up.

PR #96, opened 2020-09-24
Metis#find provides down load urls in response

corps:
Some adjustments so that the find endpoint can still efficiently return download_urls

### DIFF
--- a/metis/lib/models/file.rb
+++ b/metis/lib/models/file.rb
@@ -164,12 +164,14 @@ class Metis
       read_only
     end
 
-    def to_hash(request: nil, with_path: true)
+    def to_hash(request: nil, file_path: nil)
+      file_path ||= self.file_path
+
       {
         file_name: file_name,
         project_name: project_name,
         bucket_name: bucket.name,
-        file_path: with_path ? file_path : nil,
+        file_path: file_path,
         updated_at: updated_at.iso8601,
         created_at: created_at.iso8601,
         author: author,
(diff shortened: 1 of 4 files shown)


graft:
Nitty pick, indirectly related to this change - it makes me itchy to have methods relating to a single controller collecting on Metis::Controller, which in my mind should only contain a very few pan-controller methods - could we move them inside Metis::Query?

corps:
> Nitty pick, indirectly related to this change - it makes me itchy to have methods relating to a single controller collecting on Metis::Controller, which in my mind should only contain a very few pan-controller methods - could we move them inside Metis::Query?

Hmm.  I don't disagree, but it can't go specifically in Metis::Query because it needs a reference to the request due to the way the hmac url generation works.

graft:
> Hmm. I don't disagree, but it can't go specifically in Metis::Query because it needs a reference to the request due to the way the hmac url generation works.

Argh, what a deep rabbit hole that turns out to be, ending in Etna::Server.route_path, where the best part of the joke is:
```
      def route_path(request,name,params={})                                                        
        route = routes.find do |route|                                                              
          route.name.to_s == name.to_s                                                              
        end                                                                                         
        return route ? route.path(params) : nil                                                     
      end 
```

It doesn't even use the request! The only place it gets used is download_url uses request.host, which should probably be hard-coded anyway rather than taken from the request, which seems somewhat dodgy.

corps:
I tried pulling the logic out of the metis controller, but ended up having failing tests because both folder controller and bucket controller depend on the logic.  At best I can extract it to a module and include that in both classes.  Thoughts?

graft:
These days i tend to prefer the service object to the injected module and the method-from-nowhere. Looks like all of these are just extending the interface of FolderPathCalculator and could be folded into that, which might be instantiated directly in the controller.

My other habit in this situation, maybe only mildly better, is putting e.g. folder-relates helpers as class methods on Metis::Folder.

PR #98, opened 2020-09-28
Update attributes by csv workflow

coleshaw:
This PR adds a workflow for updating Magma attributes from a CSV, as well as a Polyphemus command to invoke it. It only works for simple attribute types (string, integer, boolean, float). We might be able to support some more complex types if we moved to a different input file format, but I think for the majority of use cases where we would want bulk update via an input file, CSV should work well.

My plan is to use this to update the `stain_version` and `reference_patients` attributes for the IPI patient model, based on the CSV that Jessica and Lenny have put together.


corps (approved):
Nice dude.  I like the tests and the workflow.  This will be super useful.

graft:
This is definitely an important workflow as we would like to facilitate the ability of any-old-person to formulate bulk updates.

I might favor a simpler input format, limiting the input to a table for a single model with attributes in columns. This doesn't allow the flexible entry you've defined here like the /update endpoint has, but it would be much easier for someone to format and probably would cover the vast majority of bulk-update cases.

coleshaw:
> This is definitely an important workflow as we would like to facilitate the ability of any-old-person to formulate bulk updates.
> 
> I might favor a simpler input format, limiting the input to a table for a single model with attributes in columns. This doesn't allow the flexible entry you've defined here like the /update endpoint has, but it would be much easier for someone to format and probably would cover the vast majority of bulk-update cases.

I thought about doing that as well. One of my concerns was that, wouldn't you be forced to put in the current value for an attribute or column, even if you didn't want to update that attribute for a given record, otherwise the logic would push an empty string / 0 revision? So essentially every row would have to have the same number of columns and values ... you couldn't selectively update "reference_patient" and "notes" for IPIGYN001 but only "notes" for IPIGYN002? If that's not a significant use case, then a simpler format definitely makes sense (copying all those values of "patient" is easy but kind of a weird requirement).

graft:
Hmm, you are right, that is ambiguous. There is also a general null/blank distinction that is missing here (could you bulk-set null this way?).

I still think i would favor simplicity over explicitness here, one can always formulate multiple csvs.

coleshaw:
Sounds good, I'll simplify the input format.

I think we could handle `nil`, but I think we would have to update the `etna-client` `magma-client` to send a JSON request instead of `multipart/form-data` (as an option). I had done this several commits earlier, but it looked messy ... I can put that in as an option, if we want to use that feature. It might be confusing to the user, depending on how Excel exports empty cells (blank or empty string could be confusing, since it wouldn't be obvious in the Excel interface -- more obvious in a text editor).

corps:
My assumption is that most use cases for this endpoint, you are setting data that is somewhat final (else the row wouldn't be in the csv).  I think defaulting to blank and not allowing null is less work and most likely the intention in 99% of cases.

"Null" in our data library often means that we haven't yet set the data, and in the case we are setting data.  That is what blank is, I assume.

graft:
> I think defaulting to blank and not allowing null is less work and most likely the intention in 99% of cases.

This seems like the right idea to me as well. On the other hand I can imagine the desire to revoke (i.e., null) a bunch of bad data without overwriting it, for example, if you have discovered that some data is incorrect and wish to remove it to stop propagating incorrect data, but you don't yet know what the correct data should be. Probably there should be a separate workflow to accomplish this kind of revocation; I think it makes sense to have this be a "fill" workflow, with positive data + blanks, and have another "empty" one, which only nulls stuff out.

graft:
A separate issue: the /update endpoint allows* you to create new records by setting a new identifier into a link attribute (e.g. { samples: [ 'NEW_SAMPLE_ID' ] }). Will this sort of thing be allowed through this interface? /update also lets you link files, which this might also allow with a metis path? I think we have been assuming this only deals with data attributes, but I think as-is many of these might simply be passed through - perhaps we should be explicit about this.

*This is probably a pretty dangerous behavior and should at least be put behind some kind of option guard, and maybe prevented from creating orphans, which it can if you set a non-existent parent - but anyway it exists right now.

coleshaw:
When I tested files, the data doesn't go through cleanly because it's a JSON string -- like `nil`, I think that would be easier done via a JSON payload rather than a form. 

Regarding `link` attributes, I'm not sure the API behaves that way anymore, where you can create ones on an update call. This is what I get when I try -- perhaps the new temp id logic prevents that from happening?

```
User developer@ucsf.edu calling update#action with params {:revisions=>"{:patient=>{:IPIADR001=>{:reference_patient=>\"IPIREF491\"}}}", :project_name=>"ipi"}
magma_app_1      | ERROR:2020-09-29T10:43:19+00:00 y3b9tn Caught unspecified error
magma_app_1      | ERROR:2020-09-29T10:43:19+00:00 y3b9tn undefined method `real_id' for nil:NilClass
magma_app_1      | ERROR:2020-09-29T10:43:19+00:00 y3b9tn /app/lib/magma/loader/temp_id.rb:14:in `real_id'
magma_app_1      | ERROR:2020-09-29T10:43:19+00:00 y3b9tn /app/lib/magma/loader/record_entry.rb:102:in `block in temp_entry'
magma_app_1      | ERROR:2020-09-29T10:43:19+00:00 y3b9tn /app/lib/magma/loader/record_entry.rb:99:in `each'
magma_app_1      | ERROR:2020-09-29T10:43:19+00:00 y3b9tn /app/lib/magma/loader/record_entry.rb:99:in `map'
magma_app_1      | ERROR:2020-09-29T10:43:19+00:00 y3b9tn /app/lib/magma/loader/record_entry.rb:99:in `temp_entry'
magma_app_1      | ERROR:2020-09-29T10:43:19+00:00 y3b9tn /app/lib/magma/loader.rb:269:in `map'
magma_app_1      | ERROR:2020-09-29T10:43:19+00:00 y3b9tn /app/lib/magma/loader.rb:269:in `block in update_temp_ids'
magma_app_1      | ERROR:2020-09-29T10:43:19+00:00 y3b9tn /app/lib/magma/loader.rb:264:in `each'
magma_app_1      | ERROR:2020-09-29T10:43:19+00:00 y3b9tn /app/lib/magma/loader.rb:264:in `update_temp_ids'
magma_app_1      | ERROR:2020-09-29T10:43:19+00:00 y3b9tn /app/lib/magma/loader.rb:100:in `dispatch_record_set'
magma_app_1      | ERROR:2020-09-29T10:43:19+00:00 y3b9tn /app/lib/magma/server/update.rb:27:in `load_revisions'
magma_app_1      | ERROR:2020-09-29T10:43:19+00:00 y3b9tn /app/lib/magma/server/update.rb:8:in `action'
```

coleshaw:
Oh, I misread what you meant for the `link` attributes. Not quite the exact same as what you described, but it is possible to replicate the behavior with a CSV. You can't just add a blank record, but if you attach information it does work.

For example, the script complains if you try uploading a CSV like this to Sample:

```
record_name
IPIBLAD001.N1
```

because there are no revisions outside of the record_name.

But, if you include some other information, you could create orphan records:

```
record_name,notes
IPIBLAD001.N2,testing
```

graft:
>Regarding link attributes, I'm not sure the API behaves that way anymore, where you can create ones on an update call. This is what I get when I try -- perhaps the new temp id logic prevents that from happening?

Blargh. I think there must be a regression, I will investigate.

PR #99, opened 2020-09-29
Janus client

coleshaw:
This PR adds two things:
* A new route to Janus for a super user to refresh their token. My thought is that when a super user creates a new project (via some workflow), it would be nice to return an updated token to them so they can continue work without having to navigate to Janus itself. So this `refresh_token` route is a "nice to have".
* A simple Janus client in `etna` gem. This should be helpful in supporting any create project workflow that we come up with.


janus/lib/server.rb
```diff
@@ -23,6 +23,13 @@ class Server < Etna::Server
 
     get '/project/:project_name', action: 'admin#project', auth: { user: { can_edit?: :project_name } }
 
+    get '/refresh_token', action: 'user#refresh_token', auth: { user: { active?: true } }
+
+    # Once we figure out a long-term token strategy, this should probably get
+    #   consolidated with /refresh_token so we don't just keep creating
+    #   small token-related views.
+    get '/viewer_token', action: 'user#viewer_token', auth: { user: { is_superuser?: true } }
```
coleshaw, line 31: I've kept these endpoints separate for now, since I think the `viewer_token` will only be used by us for testing. If other users also need it, I can merge this route into the `refresh_token`.

coleshaw:
@graft , I've opened up `refresh_token` to all users.

@corps , I've added a viewer-only token and route.

PR #100, opened 2020-09-29
fix creation via link

graft:
A fix for the issue @coleshaw noticed earlier regarding creating new links.

While this create-if-missing behavior is consistent with existing behavior in other attributes, I'm not sure this is actually such a great idea, since it is pretty error-prone (i.e., it is easy to create & link with a typo) and has few safeguards.


coleshaw (approved):
Works for me. As you say, dangerous for typos, especially because you can't delete records easily, but that's a more general problem to address than just for links.

PR #101, opened 2020-10-01
add back updated_at

graft:
Bug-fix for the issue @corps noted, that updated_at is no longer set on updates. A regression introduced by the shift away from using Sequel::Model#update for updates, which had used the Sequel timestamp plugin to handle this.


coleshaw (approved):
Works great for me.

PR #102, opened 2020-10-01
fix censoring for tables

graft:
Another bug-fix coming off @corps's repair of the censor, affecting tables. I also flipped the restricted bit on most of the specs by default, so now they're being tested as a non-privileged user, which I think is better as the restriction is always being exercised.


coleshaw (approved):
Looks good to me!

PR #103, opened 2020-10-02
minor fixes from mac setup

corps:


docker/docker-compose.etna-app.shared.yml
```diff
@@ -20,14 +20,14 @@ x-base-etna-service: &base-etna-service
   # Base configuration used by most etna services.
   image: etna-base
   volumes: &base-etna-service-volumes
-    - .:/app:cached
+    - .:/app:delegated
```
coleshaw, line 23: This seems to be an impactful change --- so if we want to test a change to app code, or etna code (as below), does that mean we have to make edits in the container now, not outside on the host?

corps: No, not quite.  It's a bit unintuitive at first but think about it:

delegated: the container’s view is authoritative (permit delays before updates on the container appear in the host)

That is only saying that if the container is editing a file, that edit is authoritative (and thus do not pay any sycnhronization cost).  The only places we edit files from the container are from node module, bundler installs, and webpack, and we never edit the outputs of those processes from the host.  Right now mac docker is so slow prceisely because these three processes must pay synchronization costs to build files that are never edited from the host in the first place.

The delay from host -> docker is so small it is not noticeable for individual file edits at the speed of human typing.  It's more noticeable for machine run "download 5000 files as fast as possible".

coleshaw: Ah, okay cool, thanks for clarifying! I assume this is what led to issues with npm.

coleshaw (approved):
I don't have a Mac to test on...so assuming this works. Everything still appears to work the same on Linux, so looks good to me.

dtm2451:
- [x] Looks like the files we'd changed to get my setup to work are included here.

Should that count as my review, or do we want to risk failure in order to have this be my setup moving forward?  I do like the idea of having everything run faster!

docker/build_image
```diff
@@ -89,7 +89,7 @@ function recursiveDigest() {
 # Searches recursively for latest updated file and keeps this variable tracking that file.
 newestFile=$0
 function findNewest() {
-    lastModifiedFile=$(find $1 -f ! -path '*/node_modules/*' -printf "%T+ %p\n" | sort | tail -1 | cut -d ' ' -f2)
+    lastModifiedFile=$(find $1 -type f ! -path '*/node_modules/*' -printf "%T+ %p\n" | sort | tail -1 | cut -d ' ' -f2)
```
dtm2451, line 92: just checking that this should still be `-f`?

corps: It should be `-type f`.  It still worked fine before but gave a noisy warning.

PR #104, opened 2020-10-05
Make attr from csv more robust

coleshaw:
While uploading a CSV for the stain_version and reference_patient, I ran into a couple of issues with the formatting. Adding some whitespace / nil checks in the update_attributes workflow to provide better error messaging and validation before attempting to update Magma records.


corps (approved):
Yeah, this makes sense.

PR #105, opened 2020-10-05
Flowjo ETL + Etl DSL / Script Manager

corps:
This PR's main goal is to get the IPI flowjo loader mapped to a new DSL based etl methodology.  You can see that in the `ipi+patient+flowjo.rb` file, the dsl makes this implementation very simple.  The user only needs to configure stain mappings and call processing functions.  The script can, however, overwrite any particular default behavior inline with the script, add conditions, and add any other logic we make available.  The idea is that we can have main 'workflows' that in the simple case are just configuration and calling a process function, but also include the ability to extend with conditionals or overwrites on a case by case basis. 

Now, this PR also includes a ton of supporting utilities:

1.  Ability to define heirarchical commands.
2.  A ton of workflow / etl improvements.
3.  A new binary 'etl_tool' that will form the basis of a CLI tool that data scientists will use to create, upload, and test new etls.
4.  Removal of the Help command -- it is now automatically provided for all command executors / applications

Remaining work:
1.  Need to adjust the etl to match the changes to the flow modeling with @coleshaw 
2.  Need to flesh out the CLI tool and add the runner into etna gem.  The idea is that the etna gem will pack a script that can download the polyphemus docker image and run the etl_tool inside of it (with dependencies packaged inside the polyphemus docker image).  TODO:  I'm saving adding this in separate PR so as to reduce review complexity.
3.  Deploy it.


polyphemus/lib/commands.rb
```diff
@@ -469,14 +497,15 @@ def execute
       folders = metis_client.find(
           Etna::Clients::Metis::FindRequest.new(
             project_name: 'ipi',
-            bucket_name: 'integral_data',
+            bucket_name: 'data_integrity',
```
corps, line 500: I'll need to revert this, was digging around for files.

docker/etna-base/entrypoints/build.sh
```diff
@@ -70,3 +70,6 @@ fi
 if [ -n "$RUN_NPM_INSTALL" ]; then
   if [ -e ../etna ]; then npm link ../etna/packages/etna-js; fi
 fi
+
+echo 'for file in /app/*.completion; do source $file || true; done' >> /root/.bashrc
```
corps, line 74: Completions are now autoloaded when running bash inside of containers.

etna/lib/etna/application.rb
```diff
@@ -4,13 +4,17 @@
...
 require 'singleton'
 require 'rollbar'
 
 module Etna::Application
   def self.included(other)
     other.include Singleton
+    other.include Etna::CommandExecutor
     @@application = other
+
+    other.const_set(:GenerateCompletionScript, Class.new(Etna::GenerateCompletionScript))
```
corps, line 17: Automatically define the GenerateCompletionScript for all applications.

etna/lib/etna/application.rb
```diff
@@ -84,26 +88,13 @@ def find_descendents(klass)
...
-  def run_command(config, cmd = :help, *args)
-    cmd = cmd.to_sym
-    if commands.key?(cmd)
-      commands[cmd].setup(config)
-      commands[cmd].execute(*commands[cmd].fill_in_missing_params(args))
-    else
-      commands[:help].execute
-    end
+  def run_command(config, *args)
+    cmd, cmd_args = find_command(*args)
```
corps, line 92: find_command is the new interface for routing command line arguments to the correct inner command.

etna/lib/etna/clients/magma/client.rb
```diff
@@ -55,6 +52,15 @@ def update(update_request = UpdateRequest.new)
         UpdateResponse.new(json)
       end
 
+      def update_json(update_request = UpdateRequest.new)
```
corps, line 55: Version of update that uses json, not form data.

etna/lib/etna/clients/magma/models.rb
```diff
@@ -34,6 +34,15 @@ def update_revision(model_name, record_name, attrs)
           record = revision[record_name] ||= {}
           record.update(attrs)
         end
+
+        def append_table(parent_model_name, parent_record_name, model_name, attrs, attribute_name = model_name)
```
corps, line 38: Magma update request utility for creating table entries.

etna/lib/etna/clients/magma/workflows/crud_workflow.rb
```diff
@@ -25,6 +25,26 @@ def lookup_record(model_name, record_id)
           result
         end
 
+        def page_records(model_name, request = Etna::Clients::Magma::RetrievalRequest.new(
```
corps, line 28: New utility function in crud workflow that pages entries of a model.

etna/lib/etna/clients/magma/workflows/ensure_containing_record_workflow.rb
```diff
@@ -0,0 +1,44 @@
+require 'ostruct'
+require_relative './crud_workflow'
+
+module Etna
+  module Clients
+    class Magma
+      class EnsureContainingRecordWorkflow < Struct.new(:magma_crud, :models,  keyword_init: true)
```
corps, line 7: Might be useful for @graft .  Extracts out the logic of creating parent records to enable deep model creation from identifiers.

etna/lib/etna/clients/magma/workflows/record_synchronization_workflow.rb
```diff
@@ -30,17 +30,23 @@ def copy_model(model_name, copied_models = Set.new)
...
 
-          model = source_client.retrieve(RetrievalRequest.new(project_name: project_name, model_name: model_name, record_names: 'all')).models.model(model_name)
-          return if model.nil?
-
-          crud.update_records do |update_request|
-            model.documents.document_keys.each do |identifier|
-              record = model.documents.document(identifier)
-              record = record.each.select { |k, v| v != nil && !v.is_a?(Array) }.to_h
-              update_request.update_revision(model_name, identifier, record)
+          crud.page_records(model_name) do |documents|
```
corps, line 38: Now does this in pages, which works much better.

etna/lib/etna/clients/magma/workflows/record_synchronization_workflow.rb
```diff
@@ -30,17 +30,23 @@ def copy_model(model_name, copied_models = Set.new)
...
+            begin
+              crud.update_records do |update_request|
+                documents.document_keys.each do |identifier|
+                  record = documents.document(identifier)
+                  record = record.each.select { |k, v| v != nil && !v.is_a?(Array) }.to_h
+                  update_request.update_revision(model_name, identifier, record)
+                end
+              end
+            rescue => e
+              raise unless ignore_update_errors
```
corps, line 49: When using this workflow, there is no guarantee that existing records match validations; the ability to optionally ignore update errors allowed me to test my etl easier locally by copying the records that pass validations and ignore those that don't.

polyphemus/bin/etl_tools
```diff
@@ -0,0 +1,62 @@
...
+
+    runner.run(record, magma_client: magma_client, commit: self.commit)
+  end
+
+  def commit
+    false
+  end
+end
+
+class Polyphemus
```
corps, line 30: I plan to add two more commands to simplify this process:
1.  Generate a new etl script with the right name + template text.
2.  Create a PR to monoetna given an etl script.

coleshaw (approved):
Looks like there is a simple conflict, but otherwise looks good to me. I'll probably have some questions about how to integrate the command stuff, later.

PR #107, opened 2020-10-08
Create project workflow

coleshaw:
This PR adds a create-project workflow and some sample commands in Etna for testing. @corps is working on the packaging up of `etna` gem into a deployable cli tool, so the final structure of what the commands look like will change.

I've taken a simplified version of the Magma project JSON and combined that with the API format to create an input project JSON format. You can see an example file [here](https://github.com/mountetna/monoetna/blob/bfd806f437a0289313dade7bc42d3b50d4916027/etna/spec/fixtures/create_project/create_project_fixture_valid.json).

There is a command to do basic validation of the JSON format to check that linked models exist, values are not blank / nil, the types are valid, etc. I'm sure it is still breakable, but hopefully the current validations catch most of the problems, and we can patch other ones as they are found.

If the JSON appears valid, the workflow then does the following:
1) Creates the project in Janus.
2) Adds the user as an editor to the project.
3) Elevates the user's permissions to administrator.
4) Creates the project in Magma.
5) Adds the models to Magma.
6) Adds the attributes for each model to Magma.
7) Adds the project record to Magma.

After creating the project, the user can visit Janus to update their token, and then view the project in Metis / Janus / Timur and create buckets, add records, etc.

If the project JSON has issues, they are reported back to the user in detail, so they can fix the issues before re-attempting. I've included a "validate project" command that people can use during project development to do a quick check on their JSON without attempting to create the project.

The workflow does assume the user is a superuser (otherwise cannot create projects!).

In follow-on PRs, I plan to add in commands / workflows (with validations) for add_model, add_attribute, update_attribute, and rename_attribute so those are available as part of the same tool.


etna/lib/etna/clients/magma/client.rb
```diff
@@ -7,10 +7,12 @@
 module Etna
   module Clients
     class Magma
+      attr_reader :host
       def initialize(host:, token:, persistent: true)
         raise 'Magma client configuration is missing host.' unless host
         raise 'Magma client configuration is missing token.' unless token
         @etna_client = ::Etna::Client.new(host, token, routes_available: false, persistent: persistent)
+        @host = host
```
coleshaw, line 15: Saving and exposing the host value on Magma client so that the create-project workflow can instantiate a new Magma client with an updated token that contains the new project's permissions.

etna/lib/etna/clients/magma/workflows/create_project_from_json_workflow.rb
```diff
@@ -0,0 +1,159 @@
...
+        end
+
+        def create_magma_project
+          magma_client.update_model(Etna::Clients::Magma::UpdateModelRequest.new(
+            project_name: project_name,
+            actions: [Etna::Clients::Magma::AddProjectAction.new]))
+        end
+
+        def create_magma_models
+          # Technically ensure_magma_tree should create models as needed,
```
coleshaw, line 72: When just running `ensure_magma_tree` without creating models beforehand, I ran into issues with `/update_model` where the Magma service would alternate between exceptions about not finding the project or not being able to retrieve models on the project, even though the Magma service had restarted and the project existed in the database. I suspect maybe due to the Sequel cache? However, by creating the models explicitly first, everything worked as expected.

janus/lib/server.rb
```diff
@@ -23,6 +23,8 @@ class Server < Etna::Server
 
     get '/project/:project_name', action: 'admin#project', auth: { user: { can_edit?: :project_name } }
 
+    get '/whoami', action: 'user#whoami', auth: { user: { active?: true } }
```
coleshaw, line 26: Adding this route to Janus because we need to extract the user information from just the token in order to add them to the new project.

etna/config.yml.template
```diff
@@ -0,0 +1,16 @@
+---
```
coleshaw, line 1: Since these just point to the hosts, in this case it might be okay to include the `config.yml` in the repo? I guess it depends on how we do the packaging for installation.

graft: Seems like a bad habit. I would like to see these configs become database records soon in any case.

janus/lib/server.rb
graft: Why not just parse the token?

coleshaw: Would you need to pass around a config file (or database record or whatever) with the Janus public key and the hash algorithm to do that? So if either of those changes, we'd have to push those changes to the clients?

etna/config.yml.template
coleshaw: Then it seems like installing with docker and a db for clients would be the way to go...

janus/lib/server.rb
coleshaw: Or, I guess that's an argument for keeping the commands on a polyphemus server and the installed etna gem just passes calls along to polyphemus. For the moment I was staying away from that workflow (and sticking to etna gem makes HTTP calls directly), since there currently isn't an API on top of polyphemus, and seems like for straightforward workflows, polyphemus would become an unnecessary potential point of failure.

corps: The public key is technically safe to pass along to our clients, although we'd need to include all possible public keys (prod, staging, dev) and users would need to update the gem when the key changed (for instance, revoking currently requires changing out the cert).

What about, instead of a new endpoint, we just -include- basic user info in say the login info if you are already authenticated?  I think it is reasonable to overload the login endpoint this way.

etna/config.yml.template
corps: Hmm.  I disagree.  What advantages are there to configuring hosts in a database?
While I do believe many parts of our config should be in a db, I think hostnames should be static, since it is generally not possible to dynamically update hostnames without breakage (we don't control clients, we'd be forced to ask every client user to update their ... database?)

corps: Also, in general, any state we keep for clients I think should just be in flat files.  If we need to address a user issue, asking them to edit a file is a faaaar more friendly interface for them than asking them to edit a local db and learn sql.  config.yml works great because _we already have the tooling and code_ for loading this file when running commands. 

That's my thoughts in general.

corps: I don't disagree that some of the config.yml stuff could become db records though.  Hosts I think are different, and in general, I think it makes sense for client gems to have a config.yml themselves.

janus/lib/server.rb
graft: The token is not encrypted, it is just base64-encoded. If you want you can just parse it using split & Base64.decode64. Timur et al parse data out of the token if you want an example.

etna/config.yml.template
graft: >I don't disagree that some of the config.yml stuff could become db records though. Hosts I think are different, and in general, I think it makes sense for client gems to have a config.yml themselves.

I agree with this, the previous comment was just shower talk. I'd still avoid checking in the actual config.yml however. My rule of thumb is if "ucsf.edu" is in some string, it should not be checked in.

corps: That, is a good point.
Perhaps we need some process to 'build' the config.yml that is provided to clients.

janus/lib/server.rb
coleshaw: Oh, I didn't realize that...makes things a lot simpler. From looking at the code in `Etna::SignService`, I thought the token was encrypted. Thanks!

etna/lib/etna/clients/magma/workflows/json_models.rb
```diff
@@ -0,0 +1,359 @@
...
+            Etna::Clients::Magma::AttributeType::FLOAT,
+            Etna::Clients::Magma::AttributeType::IMAGE,
+            Etna::Clients::Magma::AttributeType::INTEGER,
+            Etna::Clients::Magma::AttributeType::LINK,
+            Etna::Clients::Magma::AttributeType::MATCH,
+            Etna::Clients::Magma::AttributeType::MATRIX,
+            Etna::Clients::Magma::AttributeType::TABLE,
+          ]
+
+          @valid_validation_types = [
```
corps, line 295: So by the way, you can actually generate this programmatically, any place that we're using enums like this, by creating a module and adding it to the enums like this:

```
module Enum
  def self.included(cls)
    cls.instance_eval do
      def entries
        self.constants.map { |c| self.const_get(c) }.select { |v| v.instance_of?(self) }
      end
    end
  end
end
```

Then just  `include Enum` and use `Etna::Clients::Magma::AttributeType.entries.select { |c| criteriahere }`

etna/lib/helpers.rb
```diff
@@ -0,0 +1,57 @@
+require_relative './etna/clients'
```
corps, line 1: If we're adding this from polyphemus, are we also removing it from polyphemus and just using this shared implementation?  Didn't see that in the PR but maybe I'm blind.

coleshaw: This is one of those files where I just added it to make local testing inside of `etna` easier. Depending on how you decide to package it up for installation, feel free to remove this and I can plug the workflow back in where it makes sense (polyphemus or other location).

coleshaw: And no, you didn't miss it, I didn't remove it from Polyphemus since I wasn't intending this to be a shared implementation moving forward...

janus/lib/server.rb
graft: This is a good primer on JWT: https://jwt.io/introduction/ - basically, there are three parts, base64 encoded - the first two are a JSON header and payload, the last is a signature of the first two parts made with the Janus (or whatever) private key, which may be verified with the public key. However, we may decode the payload without verifying it, the signature just tells us that Janus approved the payload.

PR #108, opened 2020-10-08
better attribute_name format checking

graft:
A tiny but crucial one - if you happen to add an attribute where attribute_name contains a space, Magma will approve the update, make the column, update the attributes table, then fail to come up after reboot because it can't find a column named "your_attribute_name ". The amendment requires going into postgres to update the attributes table, egad.

Here this is repaired with some more stringent format-checking on attribute names.


coleshaw:
Oh this is great. I will take a closer look later. I found several other name validations too, that can break Magma, that I will include with the create project stuff.

magma/lib/magma/attribute.rb
```diff
@@ -196,8 +196,8 @@ def validate_validation_json
     end
 
     def validate_attribute_name_format
-      return if attribute_name == attribute_name&.snake_case
-      errors.add(:attribute_name, "must be snake_case")
+      return if attribute_name =~ /\A[a-z][a-z0-9]*(_[a-z0-9]+)*\Z/
+      errors.add(:attribute_name, "must be snake_case with no spaces")
```
coleshaw, line 200: nit: should the error message clarify that it should also not start with a number?

coleshaw (approved):
looks good to me.

coleshaw:
@graft , do you have additional updates for this PR?

magma/lib/magma/attribute.rb
graft: Hmm, i might take that as implied by "snake_case".

PR #109, opened 2020-10-08
Use gsed if available

corps:
Don't have mac, so can't confirm this easily, but the fix is to force scripts to use gnu sed if it is available on a system instead of bsd or mac variants.

Check if `which gsed` shows a path or not.  I think? homebrew includes gsed somewhere (similar to gfind in findutils)


Gvaihir (approved):
If these changes to `gutils` start to pile up suggest to include Mac set up to README including `brew install gnu-sed` and other with respective changes to `$PATH`

dtm2451:
I don't seem to have gsed. `which gsed` gives `gsed not found`.

PR #111, opened 2020-10-09
log unauthorized users

graft:
Microscopic logging change to allow logging of unauthorized emails, for the rather-too-frequent issues where the response from shibboleth disagrees with the user on what their email address is.


coleshaw (approved):
Looks good to me.

PR #112, opened 2020-10-13
Encrypt vcr test results

corps:
To re-record or add new vcrs, you'll need CI_SECRET which I will email to you.  If you don't have CI_SECRET, however, the tests just are marked as 'pending' and don't fail (so it doesn't cause noise when you are just testing local changes).


etna/lib/etna/spec/vcr.rb
```diff
@@ -0,0 +1,88 @@
...
+  VCR.configure do |c|
+    c.hook_into :webmock
+    c.cassette_serializers
+    c.cassette_library_dir = ::File.join(spec_helper_dir, 'fixtures', 'cassettes')
+    c.allow_http_connections_when_no_cassette = true
+
+    c.register_request_matcher :try_body do |request_1, request_2|
+      if request_1.headers['Content-Type'].first =~ /application\/json/
+        if request_2.headers['Content-Type'].first =~ /application\/json/
+          JSON.parse(request_1.body) == JSON.parse(request_2.body)
```
coleshaw, line 17: Hm....so this doesn't cause an error in CI, but when I test locally, I'm getting the following exception:

```
Failures:

  1) Polyphemus::MetisFileEtl should process metis files, and support reset
     Failure/Error: metis_client = Etna::Clients::Metis.new(host: 'https://metis.development.local', token: ENV['TOKEN'] || 'test-token', persistent: false)
     
     JSON::ParserError:
       765: unexpected token at ''
     # /etna/lib/etna/spec/vcr.rb:19:in `block (2 levels) in setup_base_vcr'
```

This is caused when trying to match on the `options` request to set up the Metis client, because the body is an empty string.

```
[1] pry(main)> request_1
=> #<struct VCR::Request
 method=:options,
 uri="https://metis.development.local/",
 body="",
 headers=
  {"Content-Type"=>["application/json"],
   "Accept"=>["application/json, text/*"],
   "Authorization"=>["Etna test-token"],
   "Accept-Encoding"=>["gzip;q=1.0,deflate;q=0.6,identity;q=0.3"],
   "User-Agent"=>["Ruby"]}>
[3] pry(main)> request_2
=> #<struct VCR::Request
 method=:options,
 uri="https://metis.development.local/",
 body="",
 headers=
  {"Content-Type"=>["application/json"],
   "Accept"=>["application/json, text/*"],
   "Authorization"=>["<AUTHORIZATION>"],
   "Accept-Encoding"=>["gzip;q=1.0,deflate;q=0.6,identity;q=0.3"],
   "User-Agent"=>["Ruby"]}>
[4] pry(main)> JSON.parse(request_1.body)
JSON::ParserError: 765: unexpected token at ''
from /usr/local/lib/ruby/2.5.0/json/common.rb:156:in `parse'
```

Should this stick to [the previous way of checking the matching requests](https://github.com/mountetna/monoetna/blob/21a4c8e56c94bbc5a58e7ee96354fc77d9e90b48/polyphemus/spec/spec_helper.rb#L41-L46)?

coleshaw:
I don't think the secret is being injected correctly in CI...if you check the [logs for the Polyphemus tests](https://github.com/mountetna/monoetna/pull/112/checks?check_run_id=1247666871#step:9:321), they show:

```
  1) Polyphemus::MagmaRecordEtl should process magma records, and support reset
     # CI_SECRET must be set to run this test
     Failure/Error: documents = self.magma_client.retrieve(retrieve_request).models.model(retrieve_request.model_name).documents

     RuntimeError:
       CI_SECRET must be set to run this test

   2) Polyphemus::MetisFileEtl should process metis files, and support reset
     # CI_SECRET must be set to run this test
     Failure/Error: metis_client = Etna::Clients::Metis.new(host: 'https://metis.development.local', token: ENV['TOKEN'] || 'test-token', persistent: false)

     RuntimeError:
       CI_SECRET must be set to run this test
```

etna/lib/etna/spec/vcr.rb
corps: Yeah, I'll revert to that logic.

corps:
Thanks for checking @coleshaw  -- I have pushed up another fix and will ensure the test suite actually runs them in CI.

PR #113, opened 2020-10-13
Add model workflow

coleshaw:
This PR adds a workflow / test command around adding a model + attributes to an existing project. It accepts an input JSON file and validates the attribute values against the project.

I've left the model_name out of the JSON and in the command-line arguments ... I wasn't sure if there might be some use case where a project would want to share a model JSON (or a model template JSON?) with another project, but want to give it a different model_name.


etna/lib/etna/clients/magma/workflows/json_models.rb
```diff
@@ -330,6 +357,10 @@ def validate_add_attribute_data
             check_key("model #{model.name}, attribute #{name}, validation", @raw['validation'], 'value')
             check_type("model #{model.name}, attribute #{name}, validation", @raw['validation'], 'type', @valid_validation_types)
           end
+
+          if link_model_name
+            @errors << "Attribute name #{name} in model #{model.name} should match the link_model_name, \"#{link_model_name}\"." unless link_model_name == name
+          end
         end
 
         def to_magma_model(builder)
```
corps, line 366: Is this logic shared / shareable with the copy workflow?  My only real concern is how much duplication / fracturing of logic might be occurring between these workflows.

corps (approved):

etna/lib/etna/clients/magma/workflows/json_models.rb
coleshaw: Yeah, that's a good point. All the core models that validate the JSON are common between the two workflows, and my plan is to keep the same core models shared with the attribute workflow.

coleshaw: Sorry, I misread your question... I'm not sure these are completely compatible with the copy workflow. It may be possible to merge the models cleanly or shift some of the logic, but there were two reasons I kept this logic separate:

*  The JSON input I'm expecting is not the same as the Magma output, to simplify things for someone typing these out by hand. This is mostly around links / parents, so someone doesn't have to remember to put attribute entries under two models.
* I'm trying to do validation client-side, so users have a way to test a project or model JSON locally during their brainstorming and development process with minimal risk of us having to clean something up manually for them.

The approach I tried to take was "take user-written JSON, clean it up + validate it, then reformat into something that can be used by the ModelSynch workflow".

It may make sense to put some of the naming validations into the synch workflow, but Magma would provide the same validation exceptions when executing an update (i.e. space in an attribute name). However, I think some of the validations would be messy to merge. For example, the Synch workflow tries to create a model if it doesn't currently exist (`ensure_model`), but if someone has a typo in their project JSON file, I don't really want the project created with a wrong model that then has to be cleaned up manually.

I feel like it's still useful to give users the ability to validate a model or project without having to submit to Magma...but it does feel like there is some duplication of logic.

corps: Ok, sure.  I was most wondering if you had logic similar to https://github.com/mountetna/monoetna/blob/master/etna/lib/etna/clients/magma/workflows/model_synchronization_workflow.rb#L107 that shapes from magma output to magma input, but maybe I'm reading into it wrong.

coleshaw: Yeah, that's probably the most similar method and does shape the input, though from a JSON source (not a nice Magma Attribute model class source). I'll see if I can consolidate the logic cleanly.

coleshaw: I've added a `copy` method to `Attribute` to consolidate the logic, though because there isn't an exact match between the two copy actions, I also had to add some methods to the other classes.

Just as a note, the logic in the synch workflow is copy from `Attribute` -> `AddAttributeAction` (Magma output -> Magma action input). The logic in the add-attribute workflow is copy from `JsonAttribute` -> `Attribute` (JSON -> Magma output).

PR #115, opened 2020-10-15
Etna gem configuration

coleshaw:
This PR just adds a simple API on top of polyphemus to return the hostnames for each service, from it's own `config.yml` file.

Also some small refactors around `EnvironmentScoped` and output messages for Etna workflows.

One odd change -- for staging (and dev environments), it's basically necessary to ignore SSL cert verification (for dev yes, we can install the root CA locally, but I'd prefer not to have users do any extra steps to be able to test against staging), otherwise none of the commands will run. I've included a flag in the Etna::Client for this as well.

To recap what we talked about in Slack regarding the work to build on top of this, the workflow for users should be something like:

0. the user installs a ruby environment if necessary
1. the user `gem installs etna`
2. the user attempts to use some etna command - it checks to see if `~/etna.yml` exists
3. User types `etna configure ucsf.edu`
4. It points the user to janus.ucsf.edu if there is no TOKEN
	1. Hardcode the link in a text prompt
5. With a token, the client automatically contacts polyphemus.ucsf.edu for the production configuration and writes ~/etna.yml
6. the commands run as expected


etna/lib/etna/clients/polyphemus/workflows/set_configuration_workflow.rb
```diff
@@ -0,0 +1,61 @@
...
+require 'yaml'
+require 'ostruct'
+
+module Etna
+  module Clients
+    class Polyphemus
+      # Note!  These workflows are not perfectly atomic, nor perfectly synchronized due to nature of the backend.
+      # These primitives are best effort locally synchronized, but cannot defend the backend or simultaneous
+      # system updates.
+      class SetConfigurationWorkflow < Struct.new(:host, :environment, :config_file, keyword_init: true)
```
coleshaw, line 23: This isn't a complete workflow .... so will need to be finalized.

polyphemus/config.yml.template
```diff
@@ -14,7 +14,21 @@
     :user: developer
     :password: password
     :search_path: [ public ]
-
+  :auth_redirect: https://janus.development.local
```
coleshaw, line 17: Note these changes to the `config.yml` will have to be propagated to your dev environments.

coleshaw: And to Chef / the deployed roles...

polyphemus/lib/polyphemus/controllers/configuration_controller.rb
```diff
@@ -0,0 +1,14 @@
+require_relative 'controller'
+
+class ConfigurationController < Polyphemus::Controller
+  def action
+    return_data = {
+      magma: Polyphemus.instance.config(:magma)&.dig(:host),
```
coleshaw, line 6: This will also let us add `metis_client` configuration values, which would make the user setup more minimal for Metis client.

etna/lib/etna/clients/magma/workflows/add_model_from_json_workflow.rb
```diff
@@ -126,9 +126,11 @@ def update_magma_attributes
         end
 
         def add!
+          puts "Adding the new model to Magma."
```
corps, line 129: I've been thinking that at some point we may want to make it easier to inject the application logger into workflows.  But I don't have a solution right now and I appreciate the point of this change. TODO

polyphemus/lib/server.rb
```diff
@@ -1,9 +1,22 @@
...
 class Polyphemus
   class Server < Etna::Server
     def initialize
       super
       # application.load_models
       application.setup_db
     end
+
+    # Return the app host configuration values
+    get '/configuration', as: :configuration, action: 'configuration#action', auth: { user: { is_superuser?: true } }
```
coleshaw, line 16: We can relax this if needed for third-party users? I figure all DS will need to be superusers to create projects, so this auth requirement is sufficient for now.

etna/lib/etna/clients/polyphemus/workflows/set_configuration_workflow.rb
```diff
@@ -0,0 +1,61 @@
...
+        def fetch_configuration
+          polyphemus_client.configuration(Etna::Clients::Polyphemus::ConfigurationRequest.new).raw
+        end
+
+        def update_configuration_file
+          if !File.exist?(config_file)
+            File.open(config_file, 'a+')
+          end
+
+          etna_config = YAML.load_file(config_file) || {}
```
corps, line 47: nice!

etna/lib/etna/environment_scoped.rb
```diff
@@ -0,0 +1,19 @@
+class EnvironmentScoped < Module
```
corps, line 1: This makes me think we ought to move over this logic from polyphemus and have polyphemus import it from etna.  But we can do that in future PR.

corps: DERP NEVERMIND

polyphemus/Gemfile.lock
```diff
@@ -1,7 +1,5 @@
-GIT
-  remote: https://github.com/mountetna/monoetna.git
-  revision: 6f7c04889f55ef2d38e147a4abbb2d0d2a44dfc2
-  branch: refs/artifacts/gem-etna/c8091f573c914e6db2f4c3d09eed812782a5670e
+PATH
```
corps, line 1: This will get fixed in the github action that runs after merge to master.  
TODO:  This magic will go away once I complete the docker deployment work, since we can always just us the path reference to the gem for applications.

corps (approved):

polyphemus/lib/server.rb
corps: True.  To be honest... I'm not sure that the configuration endpoint needs any auth.  None of it is really secret per say?  If it's just hostnames.

polyphemus/config.yml.template
corps: Rgr rgr.  I can take take care of adding this to roles.
Also, when you update the config.yml.template now, it will notify you the file has been updated and ask if you want to copy it over and replace the existing one, but in the future it could be a yml merge instead.

PR #116, opened 2020-10-16
support attribute_group on attributes

graft:
Basic attribute groups support - in the end I decided to keep it simple and not have tags or weird format-based group hierarchies, the group is just a single snake_case word.

It's unclear to me how general the utility of this thing will be, we'll have to see. There is annotation penalty to be paid in maintaining this (that is, it must be filled in by someone), and it's not clear that this is the best place to put this work, but it's a starting point.

### DIFF
--- a/etna/lib/etna/clients/magma/models.rb
+++ b/etna/lib/etna/clients/magma/models.rb
@@ -445,6 +445,7 @@ module Etna
           dest.format_hint = source.format_hint
           dest.hidden = source.hidden
           dest.link_model_name = source.link_model_name
+          dest.read_only = source.read_only
           dest.attribute_group = source.attribute_group
           dest.unique = source.unique
           dest.validation = source.validation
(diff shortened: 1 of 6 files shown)


etna/lib/etna/clients/magma/models.rb
```diff
@@ -438,7 +445,7 @@ def self.copy(source, dest)
           dest.format_hint = source.format_hint
           dest.hidden = source.hidden
           dest.link_model_name = source.link_model_name
-          dest.read_only = source.read_only
```
coleshaw, line 441: Is there a reason to remove the read_only flag while copying an attribute?

coleshaw:
I think the simple fix for the failing tests is to add the following method to `etna/lib/etna/clients/magma/workflows/json_models`, `JsonAttribute` (around line 300):

```
        def attribute_group
          @raw['attribute_group']&.strip
        end
```

etna/lib/etna/clients/magma/models.rb
graft: Nice catch! Just a foolish error...

coleshaw (approved):
Looks good to me. I should have the json validation stuff refactored early next week, so that'll clean up the messiness.

PR #118, opened 2020-10-20
Awesome Etna CLI

corps:
1.  Adds etna bin to the etna gem output, so it is available on user's path installation fo gem
2.  Adds flag support in etna commands
3.  Wires up etna big configuration logic
4.  Hooking up the polyphemus / etl tools.
5.  Adds project management tools from @coleshaw 's branch
6.  Idempotency for janus create project
7.  Super cool autompletions

PR #119, opened 2020-10-20
Attribute actions workflow

coleshaw:
This PR wound up kind of big, but does the following:

* Removes the JSON models in favor of the `magma/models`. This also results in separate `json_validator` and `json_converter` classes to handle manipulation of the user-input JSON into the right Magma format and client-side validation of the content. Should be cleaner and help avoid the confusion surfaced in #116 of what client models exist.
* Updates the janus, metis, and magma clients with the `ignore_ssl` flag, so that the gem can more easily be used against staging.
* Adds in an `AttributeActions` workflow and command to let users submit a set of attribute actions.
* Provides a set of annotated JSON templates as part of the gem (not sure this works yet....will test more once the gem is installable).

There is still integration work to be done with #118 to get this all working in a gem (so all the stuff in `etna/lib/helpers.rb` and `etna/lib/commands.rb` will change, but wanted to get this out for review.


etna/bin/etna
```diff
@@ -9,4 +9,6 @@ require_relative '../lib/commands'
 
 config = YAML.load(File.read(File.expand_path("../../config.yml",__FILE__)))
 
+# config = YAML.load(File.read(File.expand_path("~/etna.yml",__FILE__)))
```
coleshaw, line 12: This will change...leaving the cruft here for local testing.

etna/bin/etna
```diff
@@ -1,12 +1,11 @@
 #!/usr/bin/env ruby
 
-require 'bundler'
```
coleshaw, line 3: Keeping bundler means I can only run the gem in a directory with a `Gemfile` ...? Doesn't seem necessary for the basic project commands.

etna/etna.gemspec
```diff
@@ -6,10 +6,11 @@ Gem::Specification.new do |spec|
   spec.email             = 'Saurabh.Asthana@ucsf.edu'
   spec.homepage          = 'http://github.com/mountetna/etna'
   spec.author            = 'Saurabh Asthana'
-  spec.files             = Dir['lib/**/*.rb']
+  spec.files             = Dir['lib/**/*.rb', 'lib/etna/templates/*.json']
```
coleshaw, line 9: Also need to include the JSON templates in the gem.

etna/bin/etna
```diff
@@ -1,12 +1,11 @@
...
-
 require 'yaml'
 require_relative '../lib/etna'
 require_relative '../lib/commands'
 
 config = YAML.load(File.read(File.expand_path("../../config.yml",__FILE__)))
 
+# config = YAML.load(File.read(File.expand_path("~/etna.yml",__FILE__)))
+
 EtnaApp.instance.run_command(config, *ARGV)
```
graft, line 11: Might this be namespaced inside `Etna` instead of being its own root class?

etna/bin/etna
corps: Correct, also because the gem installation will have our dependencies installed in the system and not the bundle directory.
This line is most useful, however, for development purposes since it allows us to run without needed to perform the gem install each time.  Maybe we can conditionalize it somehow?

coleshaw: Ah, I see. What would you conditionalize on, like an environment variable?

etna/bin/etna
graft: I'm also confused about the difference between this class and Etna::Application.

coleshaw: Etna::Application is a module....so don't we need to include that in a concrete class in order to get an instance of the class and access the methods included with Etna::Application? Sorry, I'm probably misunderstanding something...

graft: Not sure what the right thing to do here is - the main motive behind having Etna::Application be a module was just to avoid having to declare `class Magma < Etna::Application`, so that you could write `class Magma; class Something; end; end`. The module itself is supposed to collect the application-specific behavior that does not have to do with the http server, which amounts to the signing service, the configuration, and the database.

Once you've created a singleton, like `class Magma; include Etna::Application; end`, you can access the singleton using either `Magma.instance` or `Etna::Application.instance` - seems like the latter would suffice for your usage?

coleshaw: If I understood what you're suggesting, you mean something like the below, instead:

```
Etna::Application.instance.run_command(config, *ARGV)
```

Assuming that `lib/etna.rb` is changed to a singleton:

```
require 'singleton'

module Etna
  class EtnaApp
    include Etna::Application
  end
end
```

? That doesn't work... I get: `./bin/etna:11:in `<main>': undefined method `run_command' for nil:NilClass (NoMethodError)
`. Seems like the call here has to be the singleton class (`Etna::EtnaApp.instance`), not the module name (`Etna::Application`)?

graft: Hmmm.... since run_command is responsible for configuring the application itself (normally invoked in like `bin/magma` or `bin/janus`) Etna::Application.instance is not defined yet. However, once you have called `Magma.instance.run_command()` or whatnot, the singleton will be bound to Etna::Application.instance, and thereafter `Etna::Application.instance` ought to work as advertised.

graft: I guess the problem here is you want to execute in some generic "etna" context, not a specific app context? Perhaps the real issue is we should make configuration independent of Etna::Application; it also might not be a good idea to mix server-side and client-side configuration contexts.

etna/lib/commands.rb
```diff
@@ -4,96 +4,200 @@
...
+      include WithLogger
+      usage 'attribute_actions_template'
+
+      def execute()
+        spec = Gem::Specification.find_by_name("etna")
+        gem_root = spec.gem_dir
+        FileUtils.cp(
+          "#{gem_root}/lib/etna/templates/attribute_actions_template.json",
+          'attribute_actions_template.json')
+        puts "A sample model JSON template has been provided at `attribute_actions_template.json`."
```
graft, line 84: Should this say 'attribute actions template'?

etna/bin/etna
coleshaw: Correct, this is the etna gem context, not a server-side app context...@corps 's work also touches on some of the command environment context stuff. I'll see if I can pull out some of the config stuff just for clients.

etna/bin/etna
corps: Yeah.  We can add an environment variable in the Dockerfile that won't effect local ge installations.

etna/bin/etna
corps: I don't see much value of changing this.  

I think you can do a lot of work, that is strictly no better than doing no work.  What if we just change the name to `EtnaBin`?  Like, whatever.  _there is no etna server to be confused with_.  There will only ever be an etna client gem bin.

Pulling or refactoring here seems like a rabbit hole (you'll probably break my PR and the subcommand changes), and I'm unconvinced of value.

corps: Also, thinking about this as client config vs server config seems meaningless. Where config is loaded is determined by entrypoint of process (the `config.ru` or the `/bin/blah` files), which ensures that, for instance, the wrong config file isn't consumed by accident.

There is also currently no use case for having a process have both a client and server configration loaded, so separating them again is questionable.

There is the common desire, for client and server tools, to access a global configuration loaded from a file that differs from environment.  And Etna::Application can be seen as a generic global container (for commands, for config, etc, etc).  I'd compare it to a dependency injection "root object" which is a decent enough pattern in both cases.

graft: Hmm, I think I disagree with everything you said, with varying degrees of vehemence.

First, while you're right, of course, that we could call the class `EtnaBin` or even `Xyzzy`, and the code would run the same, I hope I don't have to convince you of the value of namespacing for code organization. It's nice not to have stray classes around and at least have the possibility of inspecting stuff by how it is namespaced. Just a good habit, not required.

However, I feel certain that we should separate the client and server configs in the code and that the classes that deal with server-side config and client-side config should be separate, or we will suffer greatly on the road ahead. For example, Etna::Application includes Etna::SignService, which is concerned with things like signing routes with mac secrets, encrypting things with private keys, and the like, none of which the client has any cause to deal with at the moment. Shoehorning these things into a single class will lead to confusion. There are likely many other such config concerns which only exist in a server or client context and should not be handled by the same class.

The most obvious and dangerous confusion is exactly the "environment", which has a completely different meaning in the client and the server context. To the server, the "environment" is one of several mutually exclusive modes it may run in, depending on which its execution state differs. To the client, the "environment" is merely one of several identical services it may connect to, which as far as it is concerned should behave identically other than being identified differently. To have these two mutually-exclusive but oddly parallel ideas of "environment" handled by the same configuration objects seems like an exceedingly bad idea to me.

corps: > Hmm, I think I disagree with everything you said, with varying degrees of vehemence.
> 
> First, while you're right, of course, that we could call the class `EtnaBin` or even `Xyzzy`, and the code would run the same, I hope I don't have to convince you of the value of namespacing for code organization. It's nice not to have stray classes around and at least have the possibility of inspecting stuff by how it is namespaced. Just a good habit, not required.

At this point, everything inside of the etna namespace is shareable library code.  It is generalized components we intend to share to specialized parts.  For that reason, keeping the Bin outside of the namespace makes a lot of sense; EtnaBin is not shareable generic code, it's specialized for a specific application / purpose. 

> 
> However, I feel certain that we should separate the client and server configs in the code and that the classes that deal with server-side config and client-side config should be separate, or we will suffer greatly on the road ahead. For example, Etna::Application includes Etna::SignService, which is concerned with things like signing routes with mac secrets, encrypting things with private keys, and the like, none of which the client has any cause to deal with at the moment. Shoehorning these things into a single class will lead to confusion. There are likely many other such config concerns which only exist in a server or client context and should not be handled by the same class.

I mean, the same argument could be made of namespaces in general.  Does it make sense to shoehorn so much functionality into the Etna:: namespace?  Maybe we should have a Etna::Server::Secrets namespaces.  Maybe we should refactor all applications to only include exactly the modules they need.

There are times when doing that kind of splitting is useful.  But one reason that something like Rails is so successful is that having a framework that brings functionality en masse with the lowest overhead is _the best way to be productive for small-mid size apps with small development teams_.   Modularization has large overhead costs that are easy to hand wave as "good practice".

Point is -- keeping it all in application.rb means we don't have to solve other problems (like injecting various dependencies across modular components, and testing applications that include some but not all application functionality, etc, etc).  It allows us to build many related parts that can assume commonality by default, even if there are seeming natural boundaries between them (for now).


> 
> The most obvious and dangerous confusion is exactly the "environment", which has a completely different meaning in the client and the server context. To the server, the "environment" is one of several mutually exclusive modes it may run in, depending on which its execution state differs. To the client, the "environment" is merely one of several identical services it may connect to, which as far as it is concerned should behave identically other than being identified differently. To have these two mutually-exclusive but oddly parallel ideas of "environment" handled by the same configuration objects seems like an exceedingly bad idea to me.

I think it's the opposite, actually.  An environment is a consistent concept across clients and servers that they share.  The staging environment, for instance, is a series of agreed protocols, certificates, hosts, ports, and behaviors for a group of running services. 


I think you may be simply confusing the concept of an environment.  It is _not execution environment_.  It is _configuration environment_, which is a shared concept between client and server.  For most of our workflows, a user wants to say "I want to copy the data from staging to production."  And they expect the clients and servers to agree on what the concepts of "staging" an "production" are.  It's a critical concept to nearly all our client and server workflows.

corps: Also worth noting that the `config` method as is, elegantly handles a 'default' config environment along with random access of any other config environment

https://github.com/mountetna/monoetna/blob/master/etna/lib/etna/application.rb#L73

I guess if I could make any argument for not making the change it would be this:  the current structure works really well.  It is simple, it works for servers, it works for polyphemus, it's working in my branch for the etna gem.  Any change runs counter to working structure, and I don't see any benefit to more refactoring.

coleshaw: >  The most obvious and dangerous confusion is exactly the "environment", which has a completely different meaning in the client and the server context. To the server, the "environment" is one of several mutually exclusive modes it may run in, depending on which its execution state differs. To the client, the "environment" is merely one of several identical services it may connect to, which as far as it is concerned should behave identically other than being identified differently. To have these two mutually-exclusive but oddly parallel ideas of "environment" handled by the same configuration objects seems like an exceedingly bad idea to me.

So I kind of get what you are both saying around this, though I also see trying to fix this as potentially a rabbit hole and could lead to breakage. Let me make sure I understand the concerns around server-client confusion.

At this point, I think we inherently assume or know that in the gem case "environment" is the environment we're operating **against** (i.e. the gem is not in "production mode" but rather making requests against the production server). However, are you concerned about us adding new configurations in the future, that wouldn't fit neatly into this idea of "the environment we're operating against", specifically on the gem side? For example...let's say we want the etna gem to start logging information for auditing or debugging purposes. So we need the etna gem to have a non-production / non-staging / non-development related configuration for the log file location. Where would you put that in the YAML file? That breaks the way we're parsing and reading the configuration file, because we're doing it by "environment", but the log should be written to regardless of what service tier / environment you're running against?

coleshaw: I feel like part of the issue is that the current config file format actually assumes that the environment is the root key in the config -- is having multiple environments defined in a single config useful, outside of the local dev / test workflow? I believe our deployments only have a single environment defined in their config files, anyways.

Would an alternate suggestion be that we shouldn't specify the "environment" label in the config at all, but change the configs to be single-environment configs? Then a single config class would be agnostic to server or client usage.

We should be able to copy in a `config.yml.test` in Docker for running tests locally and in CI, and otherwise they would just look something like:

```
:db:
  :blah:
:magma:
  :host:
:metis:
  :host:
```

etna/bin/etna
coleshaw: @corps, I've set it to only require if `ETNA_GEM_DEVELOPMENT` is set in the environment.

corps:
Merged this work into https://github.com/mountetna/monoetna/pull/118
https://github.com/mountetna/monoetna/pull/118/commits/36070fb28588c7edf0b8a883d02500fcd3737424

Leaving this PR for now but I intend to merge this work in with above PR together.

coleshaw:
Closing this now that #118 has been merged in.

PR #120, opened 2020-10-21
Graft/fix tsv download

corps:
I added the ability to paginate by multiple columns, but unfortunately we didn't have coverage over this one particular function.  Now we do.  

Fix is to ensure that each bounds maps to the order_by_aliases, which is dynamic and transforms bounds into the appropriate structure for consumption by bounds functions.


coleshaw (approved):
Looks good to me, though don't have enough local data to do a thorough test.

PR #122, opened 2020-10-23
Fix orphaning

coleshaw:
This PR adds logic to allow orphaning of records and re-assigning parents of records. Also adds a set of tests around the expected behavior. Besides the general parent / link <-> child assignment behavior, some important use cases to note are:

* When adding new children record_names from a parent record, if they do not exist, Magma will create them.
* Updating child or collection record_names will find and orphan "old" records that would implicitly be removed from the parent.
* Can set the parent attribute value to `nil` from a child record (i.e. orphan the child directly).

This PR does not enforce strict graph validations of any sort, so you can still attach additional child records to orphaned parents.


magma/lib/magma/loader.rb
```diff
@@ -69,15 +69,172 @@ def to_payload
...
+      explicit_link_revision_exists?(
+        revisions,
+        parent_model.model_name,
+        parent_record_name,
+        parent_attribute,
+        child_record_name)
+    end
+
+    def push_implicit_link_revisions(revisions)
+      # When updating link or parent attributes from the top-down,
```
graft, line 136: The main benefit of doing this here seems to be to collect inquiries across the update, otherwise this might even fit inside `revision_to_links`, if it could be aware of the existing records; but since we don't want to query for those one-at-a-time, this method becomes necessary.

Another possibility I considered was, the loader first actually loads the current data for all items in the revision. This data can then be made available as required to revision_to_links, which would have no trouble computing these implicit links in place. You could also calculate which parts of the revision were redundant (that is, not an update) much more easily, and probably thereby avoid setting `updated_at` unnecessarily.

I think what you have written is the most efficient route, but I thought the idea of always preloading is interesting - updates shouldn't be extremely common, so the overhead might be worth it.

coleshaw: Would pre-loading be less efficient if the validation turns out to reject the updates?

PR #123, opened 2020-10-26
Bin to add etna gem locally during development

corps:
Just a convenience bin that uninstalls and reinstalls the etna gem from sources.

In general it is better to test the etna gem _outside the docker container_ so you don't have issues with the home directory and so that we can test its behaviors in a vanilla ruby environment.


coleshaw (approved):
:+1:

PR #125, opened 2020-10-27
Hide orphaned data by default

graft:
Following #122 it is possible to orphan data again. This PR adds the default behavior to Magma that orphaned data will be excluded from query results. This works both for the /retrieve API and the /query API, and both can be released with `show_orphans: true`.

I do not think we should merge this as-is, I would like to add a UI element to support the 'show_orphans' flag on the Search page.


magma/spec/query_spec.rb
```diff
@@ -107,15 +144,10 @@ def update(revisions, user_type=:editor)
...
-      poison = create(:prize, labor: hydra, name: 'poison', worth: 5)
-      poop = create(:prize, labor: stables, name: 'poop', worth: 0)
-      iou = create(:prize, labor: stables, name: 'iou', worth: 2)
-      skin = create(:prize, labor: lion, name: 'skin', worth: 6)
+      poison = create(:prize, labor: @hydra, name: 'poison', worth: 5)
+      poop = create(:prize, labor: @stables, name: 'poop', worth: 0)
+      iou = create(:prize, labor: @stables, name: 'iou', worth: 2)
+      skin = create(:prize, labor: @lion, name: 'skin', worth: 6)
 
       query(['labor', '::all', 'prize', '::count' ])
```
graft, line 153: During development, this was the only test that failed for actual reasons, not simply because the test data was now disconnected. At first I added the disconnected data filtering to ModelPredicate, the same way as the `restricted` filtering operates. However, this test began to fail if disconnected data was filtered, because it would no longer return the entry `[ 'Ceryneian Hind', 0 ]`. It turned out that this was because the filtering added two non-disconnected requirements to the same table, by transforming the original query to look like this:
`[ 'labor', [ '::has', 'project' ], '::all', 'prize', [ '::has', 'labor' ], [ 'labor', '::has', 'project' ], '::count' ]`

Somehow these two requirements intersect and drop out the row (somehow, I imagine, due to left outer-joining an empty row to itself).

My solution here is to impose the requirement only on the first clause; the presumption is that if that is connected, then anything joined to that must also be okay (that is, not disconnected). This logic might fall apart for link attributes (which could be disconnected but attached via a link to some connected data), though, so orphaned data might sneak into the query results anyway.

I'm more disturbed by the failure of the left outer join, and the removal of empty rows of data, which is a pretty crucial ability for querying. While I am papering over the problem here by creating a better filter, there is nothing preventing someone from making the query above, and it is not obvious why it should result in a count == 0 row being removed. There are some real deep issues lurking behind the relational algebra of this query logic that must eventually be confronted.

coleshaw:
This seems to work well, nice!

One side-effect seems to be that I can no longer go to orphaned records directly, using the URL. The only way to re-attach them is to know the record ID and add it back via the parent. So being able to surface the detached nodes in the Search UI seems very important. I think that re-attaching workflow is a little confusing (I think ideally you could go directly to the URL for the record and re-attach via the child), but perhaps that extra effort is a good thing in that it forces people to think a little more about detaching and re-attaching.

On the flip side, not being able to view and edit detached records may hobble some of the IPI workflow. I think people rely on being able to see the detached nodes in the Search box in the header (not the Search page) to do investigative work on which records are valid or not. So hiding the detached records there may make that difficult to do. Unless you are planning to also add a "show orphans" toggle there, too?

graft:
Egad! Good catch. I guess it makes sense to pass `show_orphans: true` in the Browse view. Children or collections have to be connected anyway to show up, so it can't hurt. I think i would leave disconnected records invisible in identifier search, the workflow for finding them can just move to Search?

coleshaw:
Hm, that's a good question. I think there are two use cases around the detached records.

1) For external partners, I assume they are finding all the records via direct Magma query, or at the very least using the Search UI and downloading a TSV. I can't imagine they are individually combing through things. So to them, it may not matter if the identifier search shows or hides the detached records? Not sure, definitely an assumption on my part.

2) For internal folks, I feel like moving the workflow only to the Search UI and leaving it out of identifier search might be a hindrance. It is super convenient to have that search box and just type in the patient name and see all the related records across multiple models. Going through Search UI lets you discover all the detached records at once (depending on how we set up the filter?), but only for a single model, and then you have to re-search once you click out to a record, since there is no "back to saved search results". That seems like lots of extra work. Generally I think people using the identifier search know the record name they want, and having the links constructed for them in that search box is very helpful. I guess they could manually reconstruct the URL from the known record name, but that seems error-prone.

Assuming the partners don't use the identifier search, I would lean towards keeping `show_orphans: true` in the identifier search and the Browse tab, but off by default in the Search UI with a toggle to enable it...

graft:
Okay, I updated Magma to allow "or" in query filters; this is tentative and needs quite a bit of real-world testing, but it appears to work okay; I successfully used it to employ a disconnected data-only retrieve option.

The former filter syntax was like `[ 'model_name', [ filter ], [ filter ], rest of query ... ]`; this syntax remains. However, the "filter" is now handled by a new Filter predicate, which has three verbs. One simply matches the old syntax and allows filtering using attributes of the model (i.e., it passes the filter arguments to the Record predicate), e.g. `[ 'model_name', [ '::has', 'attribute_name'], ... ]`.
The other two implement `::or` and `::and`, each followed by a list of filter arguments, e.g. `['model_name', [ '::or', [ '::has', 'attribute_name' ], ['::has', 'attribute_name2'] ], ... ]`
The filter arguments here are also Filter predicate and may contain nested '::or' and '::and' filters.

coleshaw:
Do you plan to update this PR with a UI toggle for search, to include disconnected nodes, as well as let different levels of authorized users see the disconnected nodes in the search box? Or future PR?

graft:
Yes, I'll post that PR later today.

coleshaw:
Looks good:

* I disconnected something and cannot find it in Timur. :heavy_check_mark: 
* I can hand-construct a Magma query to find it using `show_disconnected`. :heavy_check_mark: 
* It's pretty neat that you can get disconnected items further up the tree. For example, this query: `{"query": ["rna_seq", ["::has", "sample"], "::all", "::identifier"], "show_disconnected": true, "project_name":"ipi"}` compared to this `{"query": ["rna_seq", ["::has", "sample"], "::all", "::identifier"], "project_name":"ipi"}`. The former one returns `rna_seq` records where they are disconnected further up the tree (at patient level), whereas the latter returns all `rna_seq` records with a direct parent...but you can't tell if they are disconnected further up or not.
* I think "::or" works, but yeah more real-world testing is required...

Questions / issues:
* I think there is an editing bug introduced with the new view stuff, when a plot or attribute fails. :frowning_face:  I will file a more detailed bug report in Timur...

coleshaw (approved):
I think this looks good, but important to have the UI-side PR merged in before deploy, otherwise there will be no way to edit the disconnected records in Timur and folks might be a bit surprised when data appears to disappear.

PR #127, opened 2020-10-29
Updates to flow integral data linker

coleshaw:
This PR adds an HMAC-signed bulk-copy request to the metis client, and reconfigures the IPI flow-to-integral-dataset command in Polyphemus to use that method instead of the vanilla `copy_files`. The command also now makes a copy request for every file, because unfortunately the FCS file set is very "branchy" and causes the Metis proxy connection to Apache to time out.

### DIFF
--- a/etna/lib/etna/clients/metis/client.rb
+++ b/etna/lib/etna/clients/metis/client.rb
@@ -9,7 +9,7 @@ require_relative './models'
 module Etna
   module Clients
     class Metis
-      attr_reader :token, :host
+      attr_reader :token
       def initialize(host:, token:, persistent: true, ignore_ssl: false)
         raise 'Metis client configuration is missing host.' unless host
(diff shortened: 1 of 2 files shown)


graft:
Hmm so to be clear this is a client meant to run server to server (and thus everyone knows the mac secret), where polyphemus is signing a request as Metis?

coleshaw:
Yeah, this could definitely be handled better. Currently, the `bulk_copy` endpoint requires an hmac signature even if the user is authenticated via a token [due to this bucket owner check](https://github.com/mountetna/monoetna/blob/master/metis/lib/server/controllers/file_controller.rb#L141). If you don't provide a hmac-signature, you get a `NilClass` exception.

I can patch that to account for non-hmac requests between Metis-owned buckets (which is what this workflow is).

PR #128, opened 2020-10-30
Updates to linking flow files from integral_data

coleshaw:
This PR updates the flow FCS and WSP file linking workflows to use the new `environment` handling.

At some point these should probably be transitioned to the ETL workflow, too.


polyphemus/lib/ipi/migrations/01_flow_model_migration.rb
```diff
@@ -174,6 +174,10 @@ def format_attribute_value(suffix, value)
     # So even though we don't know the original Metis
     #   path or file name, we can still create a copy of the
     #   Magma copy of the file.
+    return {
```
coleshaw, line 177: This is not strictly necessary since we aren't copying old data into the new `flow` model.

PR #129, opened 2020-10-30
proposed updates to population loader?

coleshaw:
I added the `flow` model and updated `population` as a child of `flow`. That means the population loader is now ready to be updated!

I think these are the required changes -- I'm not very familiar with the loader, and would definitely appreciate some reviews. I'm especially unsure about the changes to `file_name_regex_dsl.rb` and how strict we want to validate that names match a specific format...

Is the best way to test this just run against a sample and poke @graft to look at the resulting table data?

Basically the changes should be:

* Remove loading the MFI tables.
* Each population update has to point to a `flow` record instead of a `sample` record.
* The `flow` record name is constructed like `IPIADR001.T1.flow.dc`.
* Tube name values from the WSP should all be in the format `IPIADR001_T1_flow_dc` now.


coleshaw:
So this is off and running -- seems to work. I'm feeding it the list of samples that passed gating checks.

If we think the workflow may be different moving forward in an ETL scenario where it's scanning for updates, may have to modify this to match.

corps:
> So this is off and running -- seems to work. I'm feeding it the list of samples that passed gating checks.
> 



Nice dude, thanks for the updates!  I was planning to help but it looks like you had this under control.

Not sure what this means

> If we think the workflow may be different moving forward in an ETL scenario where it's scanning for updates, may have to modify this to match.

I believe the plan is to run this continuously in polyphemus as an etl, the only reason I hadn't done so was waiting for the model changes you are now finishing.  Are there any concerns with doing so?

coleshaw:
So at least for now, I'm not sure what the best solution is (and hence why I'm running manually) -- because there are now a set of "good", processed WSP files with flow records, and a set of "unknown" WSP files (that haven't gone through the pipeline) with no flow records. I'd prefer to keep any flow records as sourced from processed WSP files, so I modified the loader to not create flow records and check for existence of a flow record before trying to update the population table.

ETL might work if we could say, set the date to only process WSP files uploaded since yesterday? But we don't know when the file was linked to Magma, only when the patient was updated, though we could detect the creation date in Metis?

corps:
> So at least for now, I'm not sure what the best solution is (and hence why I'm running manually) -- because there are now a set of "good", processed WSP files with flow records, and a set of "unknown" WSP files (that haven't gone through the pipeline) with no flow records. I'd prefer to keep any flow records as sourced from processed WSP files, so I modified the loader to not create flow records and check for existence of a flow record before trying to update the population table.
> 
> ETL might work if we could say, set the date to only process WSP files uploaded since yesterday? But we don't know when the file was linked to Magma, only when the patient was updated, though we could detect the creation date in Metis?

Gotcha, that makes more sense, thanks for the explanation.
What about we delay making this an ETL until all data is processed with flow records.  Then we just run manually as you are and update only when necessary.  Once all the data is processed and "good", we can go ahead and turn the ETL on for continued maintenance?

_The main benefit of an etl is continued process without manual intervention on our part_.  But if we already are needing to do that right now, we can just wait.

coleshaw:
Can I ask -- were you thinking of the ETL process start as "Magma records have been updated" or "WSP and FCS are dropped into Metis integral_data bucket"? The latter would sidestep the concern about unprocessed WSP files still in Magma, I think, but would also have to encompass the "link file to Magma" logic.

I'm still trying to figure out some WSP file issues for the processed ones, that may address whether it's a good idea for this ETL to create flow records or not...

coleshaw:
So an observation when running this manually....I ran into this exception 9 times:

```
	26: from /app/bin/etl_tools:64:in `<main>'
	25: from /app/vendor/bundle/2.5.7/ruby/2.5.0/bundler/gems/monoetna-86879d2b9731/lib/etna/application.rb:101:in `run_command'
	24: from /app/bin/etl_tools:24:in `execute'
	23: from /app/lib/etls/magma/magma_etl_script_runner.rb:26:in `run'
	22: from /app/lib/etls/etl_script_runner.rb:33:in `run_script'
	21: from /app/lib/etls/etl_script_runner.rb:33:in `open'
	20: from /app/lib/etls/etl_script_runner.rb:35:in `block in run_script'
	19: from /app/lib/etls/etl_script_runner.rb:35:in `eval'
	18: from /app/lib/etls/magma/scripts/ipi+patient+flowjo.rb:16:in `__pry__'
	17: from /app/lib/data_processing/flow_jo_dsl.rb:206:in `process_all_populations'
	16: from /app/lib/data_processing/flow_jo_dsl.rb:197:in `patient_flow_record_names'
	15: from /app/vendor/bundle/2.5.7/ruby/2.5.0/bundler/gems/monoetna-86879d2b9731/lib/etna/clients/magma/client.rb:44:in `query'
	14: from /app/vendor/bundle/2.5.7/ruby/2.5.0/bundler/gems/monoetna-86879d2b9731/lib/etna/client.rb:34:in `post'
	13: from /app/vendor/bundle/2.5.7/ruby/2.5.0/bundler/gems/monoetna-86879d2b9731/lib/etna/client.rb:96:in `body_request'
	12: from /app/vendor/bundle/2.5.7/ruby/2.5.0/bundler/gems/monoetna-86879d2b9731/lib/etna/client.rb:145:in `request'
	11: from /app/vendor/bundle/2.5.7/ruby/2.5.0/gems/net-http-persistent-4.0.0/lib/net/http/persistent.rb:885:in `request'
	10: from /app/vendor/bundle/2.5.7/ruby/2.5.0/gems/net-http-persistent-4.0.0/lib/net/http/persistent.rb:606:in `connection_for'
	 9: from /app/vendor/bundle/2.5.7/ruby/2.5.0/gems/net-http-persistent-4.0.0/lib/net/http/persistent.rb:891:in `block in request'
	 8: from /usr/local/lib/ruby/2.5.0/net/http.rb:1467:in `request'
	 7: from /usr/local/lib/ruby/2.5.0/net/http.rb:1494:in `transport_request'
	 6: from /usr/local/lib/ruby/2.5.0/net/http.rb:1494:in `catch'
	 5: from /usr/local/lib/ruby/2.5.0/net/http.rb:1497:in `block in transport_request'
	 4: from /usr/local/lib/ruby/2.5.0/net/http/response.rb:29:in `read_new'
	 3: from /usr/local/lib/ruby/2.5.0/net/http/response.rb:40:in `read_status_line'
	 2: from /usr/local/lib/ruby/2.5.0/net/protocol.rb:167:in `readline'
	 1: from /usr/local/lib/ruby/2.5.0/net/protocol.rb:157:in `readuntil'
/usr/local/lib/ruby/2.5.0/net/protocol.rb:189:in `rbuf_fill': end of file reached (EOFError)
```

However, re-running with the specific WSP files that failed with that error worked, so I'm inclined to believe it was a corrupt download of the WSP (Lenny can open the WSPs just fine in FlowJo). 

I'm wondering if this happened due to a corrupt download or something. How can we identify and catch / prevent that when this is an ETL?

corps:
Good question.  We can add some integrity headers to metis and adjust the metis client to check a hash against those integrity headers to ensure this kind of issue is caught.  what do you think?

corps:
@graft Is there already some sort of built in integrity or something we could use?  I dont' think we store the shas I guess.

coleshaw:
Metis does store the md5 hashes for each file, which is passed back on in the `find` results, but because the Magma record data is in a private bucket, I'm not sure that would be visible to this ETL script, and there isn't currently a way to "find" a publicly-available version of the file, either. I think the issue is the same as what we brainstormed for the waiver-data and cleaning up disconnected data blocks.

Sleuthing on the web shows that the `rbuf_fill` error may happen when the connection gets killed by the server. I'm re-running this using `persistent: false` on the Magma client, to see if the issue re-occurs.

polyphemus/bin/etl_tools
```diff
@@ -10,9 +10,10 @@ require 'yaml'
 config = YAML.load(File.read(File.expand_path("../../config.yml", __FILE__)))
 
 class RunOrTest < Etna::Command
+  attr_reader :runner
   def execute(script_file, record_id)
     runner = Polyphemus::MagmaEtlScriptRunner.new(script_file)
-    magma_client = Etna::Clients::Magma.new(token: ENV['TOKEN'], host: ENV['MAGMA_HOST'])
+    magma_client = Etna::Clients::Magma.new(token: ENV['TOKEN'], host: ENV['MAGMA_HOST'], persistent: false)
```
coleshaw, line 16: So this makes things run slower ... but it seems to prevent the server from timing out long-lived connections which lead to issues like `/usr/local/lib/ruby/2.5.0/net/protocol.rb:189:in `rbuf_fill': end of file reached (EOFError)`.

corps: Yeah.  I have some suspicions about the ruby persistent connection / server side implementations unfortunately.  I've seen many unusual behaviors and thus I had begun turning it off.

graft:
You are asking this: 
```
    query = [ "patient", ["::identifier", "::equals", patient_record['ipi_number']],
              "::first", "sample", "::all", "flow", "::all", "::identifier"]
```
A query this trivial ought not to time out at all, I suspect there is some issue with the nesting (::all ::all) which causes the query to be slow; I suspect the same thing is behind the sometimes-failure of the "fingerprint" plot with certain indications. I've been thinking about actually forbidding multiple '::all' clauses in a single query (so the result is always either a flat table or a single item).

Barring any of that, you could get the same result with maybe less effort like this:

```
query = [ 'flow', [ 'sample', 'patient', '::identifier', '::equals', patient_record['ipi_number'] ], '::all', '::identifier ]
```

coleshaw:
Oh nice, thanks, I'll give that query a shot.

I don't think it's specifically a single query that times out, but rather that a persistent connection over 400+ samples gets killed by the server at some point...like, the server only lets you stay connected for X seconds before it forces you to re-establish a new connection?

graft:
Hmm, I note we are still using Net::HTTP::Persistent in Etna::Client, which is super, super old and probably no longer required. For example `metis_client` just uses Net::HTTP.new, which I *think* gets reused (at least it is reusing the object, dunno if it creates a new HTTP connection every time), and doesn't seem to have any timeout issues across large metis pushes.

graft:
Also regarding the unavailability of Metis md5s for magma data, I have been thinking about how limiting not being able to ask Metis about details for stuff in magma buckets is. You could also get `updated_at` and many other useful things from Metis if you were only able to ask. In fact you DO have the metis_path for the file from Magma. However, Metis both (1) lacks a file metadata route and (2) requires magma to sign any access into its bucket.

Some possibilities:
1) Don't be so hard-nosed and let people with read permission read freely from magma's bucket.
2) Provide a way for magma to give signed query URLs for the metis bucket
3) Extend the download controller to serve metadata for the file with an optional parameter, so anyone with the URL from magma can ask for file metadata from Metis.

corps:
> Hmm, I note we are still using Net::HTTP::Persistent in Etna::Client, which is super, super old and probably no longer required. For example `metis_client` just uses Net::HTTP.new, which I _think_ gets reused (at least it is reusing the object, dunno if it creates a new HTTP connection every time), and doesn't seem to have any timeout issues across large metis pushes.

Oh nice, interesting.  I can take a look at migrating this.

corps:
> Oh nice, thanks, I'll give that query a shot.
> 
> I don't think it's specifically a single query that times out, but rather that a persistent connection over 400+ samples gets killed by the server at some point...like, the server only lets you stay connected for X seconds before it forces you to re-establish a new connection?

This might be a setting we can adjust either in apache or rack.   I'm not sure but maybe?  worth investigating?  But also I think to graft's point, upgrading away from the old persistent connection and just using Http.new might be generally the best option.

coleshaw:
Yeah, I had forgotten about using the old library. I think that's a reasonable first step to seeing if the behavior goes away.

coleshaw:
I'm going to merge this in because it's what I used with batch 1 of the flow files...but I'm not sure it'll be what we want with an ETL. So we'll probably want to revisit this at some point.

PR #130, opened 2020-10-30
Account for no hmac in the environment in Metis bulk_copy route

coleshaw:
So the reason this never came up during tests was because `Etna::TestAuth` automatically added an `hmac` element to the request environment, even if no signature or headers were supplied. With `Etna::Auth` though, `hmac` is not added to the request environment unless a signature and headers are provided. Hence in deployed environments I ran into `#<NoMethodError: undefined method 'id' for nil:NilClass>` when trying to call `bulk_copy`.

This PR updates `Etna::TestAuth` to more closely mimic the behavior of `Etna::Auth` in that it only adds a valid hmac element to the request environment when a signature is provided.

This also updates the `Etna::Route` to account for possible `nil` `hmac` elements in the environment, which can now happen with `TestAuth`. This is mainly a fix for testing, since it should have correctly failed in deployed environments when no `hmac` was provided but one was expected?

And the PR updates Metis's `bulk_copy` route to squash and remove empty owners when hmac is not provided.

### DIFF
--- a/etna/lib/etna/route.rb
+++ b/etna/lib/etna/route.rb
@@ -118,7 +118,7 @@ module Etna
 
     def hmac_authorized?(request)
       # either there is no hmac requirement, or we have a valid hmac
-      !@auth[:hmac] || request.env['etna.hmac'].valid?
+      !@auth[:hmac] || request.env['etna.hmac']&.valid?
     end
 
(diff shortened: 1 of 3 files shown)


corps (approved):
Nice catch!

PR #131, opened 2020-10-31
Add Project Models Workflow Refactor

corps:
Sneak peak.
Still have a few changes and additional tests.

But the commands work!
Will comment and annotate reasoning / changes soon.


etna/bin/etna
```diff
@@ -18,8 +18,7 @@ end
 EtnaApp.instance.run_command(config, *ARGV) do |cmd, args|
   if cmd.class.included_modules.include?(RequireConfirmation)
     puts "Attempting to run #{cmd.program_name} in environment #{EtnaApp.instance.environment} with args #{args}"
-    puts "Confirm? y/n"
-    STDIN.gets.chomp == 'y'
+    StrongConfirmation.confirm
```
corps, line 21: Implementing a StrongConfirmation that requires a bit more work before executing dangerous commands.

etna/etna.gemspec
```diff
@@ -16,4 +16,5 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'net-http-persistent'
   spec.add_dependency 'multipart-post'
   spec.add_dependency 'rollbar'
+  spec.post_install_message = "\n\n\033[1;31mAdd 'source ~/etna.completion' to your .bashrc to get etna command line completions!\033[0m\n\n"
```
corps, line 19: Now, after installing with `gem install etna` it will provide a post install message instructing users to source the completion.

etna/lib/commands.rb
```diff
@@ -24,10 +24,12 @@ def dispatch_to_subcommand(cmd = 'help', *args, environment: nil, **kwds)
   def environment
     if @environment
       @environment
-    elsif @config && @config.keys.length == 1
+    elsif @config && @config.is_a?(Hash) && @config.keys.length == 1
       @config.keys.last.to_sym
+    elsif @config && @config.is_a?(Hash) && @config.keys.length > 1
+      raise "You have multiple environments configured, please specify your environment by adding --environment #{@config.keys.join("|")}"
```
corps, line 30: If you have multiple environments, require that it be set with `--environment` on any command that requires it.

etna/lib/commands.rb
```diff
@@ -24,10 +24,12 @@ def dispatch_to_subcommand(cmd = 'help', *args, environment: nil, **kwds)
...
     else
-      :production
+      raise "You do not have a successfully configured environment, please run #{program_name} config set https://polyphemus.ucsf.edu"
```
corps, line 32: Better feedback about when the user tries to run co,mands before an environment has been configured.

etna/lib/commands.rb
```diff
@@ -38,6 +40,20 @@ def set_environment(env)
   class Config
     include Etna::CommandExecutor
 
+    class Show < Etna::Command
+
+      boolean_flags << '--all'
+
+      def execute(all: false)
```
corps, line 47: Little utility to show the user's current config.  Figure it'd be useful for debugging.

etna/lib/commands.rb
```diff
@@ -114,38 +105,111 @@ class Create < Etna::Command
         include WithEtnaClients
         include WithLogger
 
-        def execute(filepath)
-          create_project_workflow = Etna::Clients::Magma::CreateProjectFromJsonWorkflow.new(
+        def execute(project_name, project_name_full)
+          create_args = {project_name: project_name, project_name_full: project_name_full}
+          Etna::Clients::Magma::ProjectValidator.new(**create_args).validate!("#{program_name} args invalid!")
```
corps, line 110: Create project now uses the split validator / workflows fior that purpose.

etna/lib/commands.rb
```diff
@@ -114,38 +105,111 @@ class Create < Etna::Command
...
+        string_flags << '--file'
+        string_flags << '--target-model'
 
-          validator = Etna::Clients::Magma::ProjectValidator.new(project)
-          validator.validate
+        def execute(project_name, execute: false, target_model: 'project', file: "#{project_name}_models_#{target_model}_tree.csv")
+          reset
 
-          return puts "Project JSON is well-formatted!" if validator.valid?
+          unless File.exists?(file)
```
corps, line 136: When the csv file does not exist, pull down the project's data into a csv.

etna/lib/commands.rb
```diff
@@ -114,38 +105,111 @@ class Create < Etna::Command
...
+            puts
+            puts "Done!  You can start editing the file #{file} now, and I will report validation errors here."
+          end
 
-          puts "Project JSON has #{validator.errors.length} errors:"
-          validator.errors.each do |error|
-            puts "  * " + error.gsub("\n", "\n\t")
+          load_models_from_csv(file)
+
+          while true
```
corps, line 146: Polling watch loop that validates the file and allows executing the file with confirmation once it passes validation.

etna/lib/etna/application.rb
```diff
@@ -76,6 +76,13 @@ def config(type, env = environment)
     @config[env][type]
   end
 
+  def env_config(env = environment)
```
corps, line 79: Used by the config show command  to attempt to get all of a single environment's config.

etna/lib/etna/clients/magma/formatting/models_csv.rb
```diff
@@ -0,0 +1,250 @@
...
+            restricted: [:restricted, COLUMN_AS_BOOLEAN],
+            read_only: [:read_only, COLUMN_AS_BOOLEAN],
+            options: [:validation, -> (s) { {"type" => "Array", "value" => s.split(',').map(&:chomp)} }],
+            match: [:validation, -> (s) { {"type" => "Regexp", "value" => Regexp.new(s).source} }],
+            attribute_group: :attribute_group,
+            hidden: [:hidden, COLUMN_AS_BOOLEAN],
+            unique: [:unique, COLUMN_AS_BOOLEAN],
+        }
+
+        def self.apply_csv_row(models = Models.new, row = {}, &err_block)
```
corps, line 41: A reducer function that accumulates csv row changes into the given models.  Forms the backbone of the csv parser.

etna/lib/etna/clients/magma/formatting/models_csv.rb
```diff
@@ -0,0 +1,250 @@
...
+          name.split('_').map(&:capitalize).join(' ')
+        end
+
+        def self.get_col_or_nil(row, col)
+          c = row[col]&.chomp
+          return nil if c&.empty?
+          c
+        end
+
+        def self.each_tsv_row(models = Models.new, model_keys = models.model_keys.sort, &block)
```
corps, line 190: Output function, yields rows of column values for each model, given by the ordering of model_keys.

etna/lib/etna/clients/magma/workflows/model_synchronization_workflow.rb
```diff
@@ -113,8 +115,37 @@ def ensure_model_attribute(model_name, attribute_name)
         # This method, and it's partner prepare_parent, should never call into any re-entrant or potentially
         # cyclical method, like ensure_model_tree.
         def ensure_model(model_name)
+          ensure_model_create(model_name)
+          ensure_model_identifier_configured(model_name)
+        end
+
+        def ensure_model_identifier_configured(model_name)
```
corps, line 122: Borrowing from @coleshaw 's work, allows synchronizing the properties of identifier attributes via updates after creation.

etna/lib/helpers.rb
```diff
@@ -30,6 +30,9 @@ def magma_client
     @magma_client ||= Etna::Clients::Magma.new(
         token: token,
         ignore_ssl: EtnaApp.instance.config(:ignore_ssl),
+        # Persistent connections cause problem with magma restarts, until we can fix that we should force them
```
corps, line 33: I noticed that using persistent connections to magma while executing model updates has a small chance of failing.  I think what happens is that the persistent connection delays the restart of update_model actions, which can cause future commands to execute before the sequel model has updated.  Seems to go away without persistent connections.

This only effects the etna bin.  In the future we should investigate other solutions.

magma/lib/magma/actions/add_project.rb
```diff
@@ -2,11 +2,22 @@ class Magma
...
     def perform
-      begin
-        Magma.instance.db.run "CREATE SCHEMA #{@project_name}"
-      rescue Sequel::Error => e
-        raise unless e.message.start_with?('PG::DuplicateSchema: ERROR:')
-      end
+      # Unfortunately we cannot just run a create and catch the error due to the way ruby sequel poorly handles graceful
+      # recovery of error inside transaction, which this is run inside of implicitly from parent caller.
+      Magma.instance.db.run <<-SQL
+DO $$
```
corps, line 8: I also found that relying on catching the sequel here for idempotence doesn't actually work -- it results in errors and does not continue to adding the project model as it should.  This has to do with the way sequel handles errors in transactions (it basically refuses to gracefully recover until the transaction is restarted, but the transaction is owned by the update_model controller and notthis level of code).

Solution was to use some PSQL here to check for the existence of the schema before executing the create.

corps:
Also, net minus 600 lines from this change set means consolidation of logic!

etna/etna.gemspec
coleshaw: Nice!

etna/lib/helpers.rb
coleshaw: Interesting find -- nice investigative work for this and the add_project fix.

etna/lib/etna/clients/magma/formatting/models_csv.rb
```diff
@@ -0,0 +1,250 @@
...
+            description: :desc,
+            display_name: :display_name,
+            format_hint: :format_hint,
+            restricted: [:restricted, COLUMN_AS_BOOLEAN],
+            read_only: [:read_only, COLUMN_AS_BOOLEAN],
+            options: [:validation, -> (s) { {"type" => "Array", "value" => s.split(',').map(&:chomp)} }],
+            match: [:validation, -> (s) { {"type" => "Regexp", "value" => Regexp.new(s).source} }],
+            attribute_group: :attribute_group,
+            hidden: [:hidden, COLUMN_AS_BOOLEAN],
+            unique: [:unique, COLUMN_AS_BOOLEAN],
```
coleshaw, line 38: How were you thinking the workflow would be for the four attribute actions -- still as separate commands? Would it be worth it to add a column here for `new_attribute_name`, and then it seems like all four of those could also be rolled into the CSV model? `add_attribute`, `update_attribute`, and `add_link` seem like they should already work with this CSV synch workfow.

etna/lib/etna/clients/magma/formatting/models_csv.rb
coleshaw, line 190: nit: should this now be `each_csv_row` to match the model?

coleshaw:
Switching to a CSV format makes a lot of sense and occurred to me after my lab meeting last week, too -- so this looks really neat!

One high-level question: is there a way to mitigate the risk of CSV divergence? Since the tool does model synch, seems like people might treat their CSVs as the "source of truth" or canonical model, since the tool synchs one-way from CSV to Magma, instead of Magma being the canonical source of the project model. One place I see a risk is when multiple people edit a model independently of each other. For example, Friday morning I changed some columns on the IPI Sample model to `hidden: true`. Later in the day @graft changed other columns to hidden for different reasons. So if Saurabh had an out-of-date local version of the IPI Project template, he might have accidentally undid my changes and would have had no way to realize this even if he checked his changes afterwards.

It's unclear to me if that's just something we have to live with, how large a risk that is, or if there's something we can do to make that less likely to happen. Like, should we be keeping track of project revision dates and letting people know that their local version is out of date when they try to synch? Might get kind of complicated though...

etna/lib/etna/clients/magma/formatting/models_csv.rb
corps: Yeah, that makes a lot of sense.  I think that can be a natural extension of this and it would simplify the workflow even further.

etna/lib/etna/clients/magma/formatting/models_csv.rb
corps: Yes

corps:
> Switching to a CSV format makes a lot of sense and occurred to me after my lab meeting last week, too -- so this looks really neat!
> 
> One high-level question: is there a way to mitigate the risk of CSV divergence? Since the tool does model synch, seems like people might treat their CSVs as the "source of truth" or canonical model, since the tool synchs one-way from CSV to Magma, instead of Magma being the canonical source of the project model. One place I see a risk is when multiple people edit a model independently of each other. For example, Friday morning I changed some columns on the IPI Sample model to `hidden: true`. Later in the day @graft changed other columns to hidden for different reasons. So if Saurabh had an out-of-date local version of the IPI Project template, he might have accidentally undid my changes and would have had no way to realize this even if he checked his changes afterwards.
> 
> It's unclear to me if that's just something we have to live with, how large a risk that is, or if there's something we can do to make that less likely to happen. Like, should we be keeping track of project revision dates and letting people know that their local version is out of date when they try to synch? Might get kind of complicated though...

Yes, that's an excellent concern.  For syncing just new columns, it would be fine because the logic currently only adds, doesn't remove.  For updating existing models, it will matter much more.

One solution is to add versioning per model, and require that the user have the latest version before executing changes.  We'd need to make some server backend changes but it wouldn't be too much.

corps:
My thought process is this: landing this version of the create models is safe sense it only adds new models, and it leaves the actions logic in place for now.

In a followup PR, I can extend the csv model to support updates and add the backend support for versioning.

PR #133, opened 2020-11-05
Metis rename file across buckets

coleshaw:
This PR allows the Metis waiver_data functionality to handle folders that exist in both the source and target buckets. Current behavior of Metis throws an exception when you try to rename onto an existing folder, which is a problem if waiver data moves a partially-uploaded folder.

This PR does several things:

* Adds the ability for Metis `file_rename` route to handle moving files to a different bucket. The current behavior does not clobber existing files (the unix `mv` command does), which I thought was safer since we don't have any UI toggle or prompt to warn the user that they may clobber an existing file. This functionality only exists at the API level and is not exposed in the UI.
* Updates the "rename with regex" convenience function in the `etna` Metis client to call a new method if the target folder exists. This new method moves sub-folders and files into the target folder, and then deletes the source folders. So to the user it appears like the old "rename folder" behavior but instead individually moves files / folders over.
* Also pins the version of Puma in each app's Gemfile. I ran into the same issue as Anton after bringing the apps back up, in that the Puma version automatically bumped and introduced new dependencies.

There may still be an exception thrown if a file is duplicated in both folders (since the rename won't overwrite), but that shouldn't happen with the waiver_data use case unless someone uploads a file multiple times.

This requires a deployment to both Metis and Polyphemus.

### DIFF
--- a/metis/lib/file_rename_revision.rb
+++ b/metis/lib/file_rename_revision.rb
@@ -1,0 +1,72 @@
+require_relative 'revision'
+
+class Metis
+  class FileRenameRevision < Revision
+    def validate
+      @errors = []
+      validate_source(@source)
+      validate_dest(@dest)
+    end
+
+    def bucket_names
+      # This is kind of weird, but we need the ability to grab
+      #   all relevant bucket names, even before validation of
+      #   the CopyRevision.
+      # So, if @dest doesn't exist, we don't return it
+      source_bucket_names = super
+      if @dest.mpath.valid?
+        return source_bucket_names.push(@dest.mpath.bucket_name)
+      end
+      return source_bucket_names
+    end
+
+    def mpaths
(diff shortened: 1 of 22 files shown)


metis/lib/file_rename_revision.rb
```diff
@@ -0,0 +1,72 @@
...
+class Metis
+  class FileRenameRevision < Revision
+    def validate
+      @errors = []
+      validate_source(@source)
+      validate_dest(@dest)
+    end
+
+    def bucket_names
+      # This is kind of weird, but we need the ability to grab
```
graft, line 12: i find this method difficult to follow  - might you instead push @source and @dest into @buckets in the initializer, then have bucket_names just iterate across @buckets in `Revision`? This would save having to update this method whenever you subclass...

PR #135, opened 2020-11-05
Etna gem documentation + bug fixes

coleshaw:
This PR fixes several small issues I ran into while testing the `etna` gem locally with the new CSV workflow, and also adds documentation for using the gem.

It includes:

* Instructions on how to configure the gem
* Examples of the CSV workflow
* Examples of updating attributes by CSV
* Details the constraints on model and attribute names


etna/lib/commands.rb
```diff
@@ -149,16 +150,16 @@ def execute(project_name, execute: false, target_model: 'project', file: "#{proj
...
-                    puts "Executing #{update_action.raw}..."
+                    puts "Executing #{update_action.to_h}..."
                   end
                   puts "Success!"
                 end
 
                 return
               else
-                puts "To commit, run \033[1;31m#{program_name} #{command_name} --file #{file} --target-model #{target_model} --execute\033[0m"
+                puts "To commit, run \033[1;31m#{program_name} #{project_name} --file #{file} --target-model #{target_model} --execute\033[0m"
```
coleshaw, line 162: This was duplicating the `add` command in the instructions. I think `program_name` already includes `add`.

etna/lib/etna/client.rb
```diff
@@ -147,7 +147,10 @@ def request(uri, data)
             yield response
           end
         else
-          Net::HTTP.start(uri.host, uri.port, use_ssl: true) do |http|
+          verify_mode = @ignore_ssl ?
+            OpenSSL::SSL::VERIFY_NONE :
+            OpenSSL::SSL::VERIFY_PEER
+          Net::HTTP.start(uri.host, uri.port, use_ssl: true, verify_mode: verify_mode) do |http|
```
coleshaw, line 153: We also need to pass through the `ignore_ssl` flag for non-persistent connections, otherwise development and staging environments don't work.

etna/lib/etna/clients/magma/workflows/add_project_models_workflow.rb
```diff
@@ -21,7 +21,7 @@ def prepare_models_from_csv(csv_io, &err_block)
           models = csv_lines.inject(Models.new) do |acc, n|
             line_no += 1
             ModelsCsv.apply_csv_row(acc, n) do |err|
-              err_block.call("Error detected on line #{line_no}: #{err}")
+              err_block.call("Error detected on line #{line_no + 1}: #{err}")
```
coleshaw, line 24: Due to the CSV headers adding an extra line, I think we should add `1` to the line number reported to the user. It's a bit more intuitive.

etna/lib/commands.rb
corps: I already have this fixed in my latest branch actually, but in a different way, can we revert?  Or I can just remove when I merge y branch in.

etna/lib/etna/client.rb
corps: Nice catch!

etna/lib/commands.rb
corps: Actually don't worry about it, I'll just address on my merge.

corps:
Thank you for spending time on this, I appreciate.

etna/lib/commands.rb
coleshaw: Okay, that sounds good.

PR #136, opened 2020-11-06
Update / Create Attributes From CSV

corps:
* Improves the CSV workflow by supporting updates and renames from the CSV as well.  
* Includes some bug fixes to CSV workflow
* Displays useful visual diff of changes implied by the CSV before executing it.
* Inclusion of a version based optimistic locking mechanism on model updates, which I decided against using in csv.  Options are to remove it or leave it for future use.  It is optional and would not effect any current use cases as is.
* Deletes csv files after update completes successfully.

Feel free to clone the branch and play with it yourselves.

TODO:
Add more tests, fix existing tests.


corps:
I'm still adding some fixes and testing right, should have some further improvements before end of day.

corps:
I want to clean up how the output looks (it's a mess right now), add tests, and probably adjust the CLI (now that we've collapsed actions and csv into just one command, it prbably can be simplified more).

BUT, it's all working, and fully testable.  @graft @coleshaw

corps:
@coleshaw @graft 
This has now fully been tested and the code reflects closer to final version.
I need to merge master in, address conflicts, and polish.
Let me know if there are any issues you'd like to see before moving to get this in our team's hands on monday.

myspecialthing.csv
```diff
@@ -0,0 +1,374 @@
+comments,model_name,identifier,parent_model_name,parent_link_type,attribute_name,new_attribute_name,attribute_type,link_model_name,description,display_name,format_hint,restricted,read_only,options,match,attribute_group,hidden,unique
```
coleshaw, line 1: Do we need this in the repo, or was this only for testing?

corps: Good call, removing.

PR #137, opened 2020-11-06
Removing bundler env variable path for Polyphemus binary

coleshaw:
I ran into this today when updating polyphemus on production -- seems like setting this Bundler environment variable breaks polyphemus production / staging because they are not currently deployed in Docker (?). So I manually unset this on the server but am also submitting this PR .... the other apps don't appear to set this environment variable, so I'm wondering if it will be safe to remove both for local development in Docker and server deploys outside of Docker?

### DIFF
--- a/polyphemus/bin/polyphemus
+++ b/polyphemus/bin/polyphemus
@@ -1,7 +1,7 @@
 #!/usr/bin/env ruby
 
 require 'bundler'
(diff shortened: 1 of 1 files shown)


corps:
It's not necessary, I think this is an accidental addition.  Thank you for removing.

corps (approved):

PR #138, opened 2020-11-09
Timur upload file

coleshaw:
This PR adds a button to the Timur UI for file and image attributes that lets users upload a file as part of a revision. This file is uploaded directly to Metis, and then the final location is sent to Magma once the upload is complete.

This PR also makes the following ancillary changes to make the upload stuff easier:
* A minor amount of refactoring in this PR to put the upload meter and controls as shared components in `etna-js`. 
* Changing the `TIMUR_CONFIG` environment variable in the Timur config to `CONFIG` so it works with the shared upload code.
* Magma will take the file extension from an `original_filename` argument if provided, since temporary URLs don't have extensions, but we don't want to lose that information.

I'll annotate the PR. Once it merges, I can load these into staging for testing -- this requires updates to both Magma and Timur, and Metis updates are just for refactoring purposes.


etna/packages/etna-js/upload/workers/uploader.jsx
```diff
@@ -102,6 +102,11 @@ export function Upload({
     current_byte_position = 0,
     next_blob_hash = ZERO_HASH,
     file_size = file.size,
+    metis_uid,
```
coleshaw, line 105: Need to add these values to the `Upload` class, so that the right `file_attribute` and `image_attribute` entries can be found in Timur, since there can be multiple uploads on a single page running at any given time.

etna/packages/etna-js/upload/workers/uploader.jsx
```diff
@@ -447,7 +457,8 @@ export class Uploader {
         blob_data,
         next_blob_size: newBlobSize,
         next_blob_hash: newBlobHash,
-        current_byte_position
+        current_byte_position,
+        metis_uid
```
coleshaw, line 461: Because we're starting the upload with a Magma HMAC-signed URL, we need to provide the Metis UID (since we're not uploading from Metis directly). This has to be passed in from Timur.

magma/lib/magma/attributes/file.rb
```diff
@@ -144,8 +144,10 @@ def self.type_bulk_load_hook(loader, project_name, attribute_copy_revisions)
 
     private
 
-    def filename(record_name, path)
+    def filename(record_name, path, original_filename=nil)
       ext = path ? ::File.extname(path) : ''
+      original_ext = original_filename ? ::File.extname(original_filename) : ''
+      ext = original_ext if ext.empty?
```
coleshaw, line 150: Uses the original file name's extension if available. Otherwise temporary upload URLs have no knowledge of the right extensions and would just appear as `.dat`, which is confusing for users.

timur/config.yml.template
```diff
@@ -17,6 +17,7 @@ default: &default
   :auth_redirect: https://janus.development.local
   :token_algo: RS256
   :token_name: JANUS_DEV_TOKEN
+  :metis_uid_name: METIS_DEV_UID
```
coleshaw, line 20: Note this change to Timur config. Will have to update this on Chef as well.

graft (approved):
OK, following the latest update this WFM (after painfully creating a magma bucket with a tmp/ folder), very nice. I especially like how the upload persists across navigation (i.e., if you click away from browsing a record you are uploading into, then click back, the upload shows up as still in progress).

coleshaw:
Just a note, once PR #141 is merged in, running `project create` via the `etna` gem will also create the `magma` bucket for an existing project, so you won't have to manually (and painfully) create it!

PR #139, opened 2020-11-10
Add `magmaR` to the repo

dtm2451:
Add the magma <=> R client, R package `magmaR`, which I have been working on for a bit now.


magmaR/R/retrieve.R
```diff
@@ -0,0 +1,168 @@
...
+    if (json.params.only) {
+        return(jsonParams)
+    }
+    
+    ### Perform '\retrieve'-al
+    
+    h <- RCurl::basicTextGatherer()
+    
+    RCurl::curlPerform(
+        url = "https://magma.ucsf.edu/retrieve",
```
coleshaw, line 151: It might be useful for testing if this value could be configurable in a non-committed / `.gitignore` config file.

magmaR/R/query.R
```diff
@@ -0,0 +1,99 @@
...
+    if (json.params.only) {
+        return(jsonParams)
+    }
+    
+    ### Perform '\query'
+    
+    h <- RCurl::basicTextGatherer()
+    
+    RCurl::curlPerform(
+        url = "https://magma.ucsf.edu/query",
```
coleshaw, line 86: It might be useful for testing if this value could be configurable in a non-committed / .gitignore config file.

coleshaw:
This looks nice, I like the documentation!

Question: Is there a way in R to specify the dependencies, so it works "out of the box" if someone installs it? I think this depends on the user having `RCurl` and `rjson` installed in their R instance ... I didn't see a way to specify that, but maybe I missed it.

graft:
https://github.com/mountetna/rtemis/blob/master/install-packages.R is how i did this in rtemis. Maybe there is some better lib for this by now.

graft:
Speedy work! Since there is no independent magmaR repo it might do to put this in the /etna tree? Like etna/packages/magmaR

magmaR/R/query.R
dtm2451: Another option: I already have it so that "hidden" options are passable to the `retrieve()` and `query()` functions for troubleshooting, via `...`. They're documented with a 'for troubleshooting only' descriptor, but that could be updated to 'for advanced users', and we could add an input for adjusting the url to the underlying `.retrieve()` and `.query()` functions.  Perhaps `url.base` and it would only modify the "https://magma.ucsf.edu" portion.

Then (useful for DSs) a reinstall of the package wouldn't be required for switching between production vs staging vs dev.

Tokens would still need to match up, so I think this change is safe?  

I'm not sure on the details of how your original suggestion would work, but we could do that too in order to not need to be explicit in every function call in testing. What I'd suggest we target with such an approach would be the default option of the `url.base` input that I'll add in.

dtm2451:
> Is there a way in R to specify the dependencies, so it works "out of the box" if someone installs it?

Yes, yes!  For package, this is handled with via the DESCRIPTION file.  This line in particular is doing that:
https://github.com/mountetna/monoetna/blob/9caa4adce8d39c0ebc885002a0239d7a502ba759/magmaR/DESCRIPTION#L8

`Imports:`, `Suggests:`, and `Depends:` are three field options for this goal which work in slightly different ways, all of which I used for dittoSeq.  More detail: https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Package-Dependencies.

Eventually, `test_that` will be added to `Suggests:` in order to tell the R check suite / CRAN that the package is needed for a full package check, but that the package is not actually needed for installing magmaR.

dtm2451:
> https://github.com/mountetna/rtemis/blob/master/install-packages.R is how i did this in rtemis. Maybe there is some better lib for this by now.

Indeed there is for a full package.  See above.

dtm2451:
> Speedy work! Since there is no independent magmaR repo it might do to put this in the /etna tree? Like etna/packages/magmaR

Sure, ¯\_(ツ)_/¯, you tell me where's best!

magmaR/R/query.R
coleshaw: Oh, I see, like a default value that someone who calls the method can overwrite? Yeah, I think hidden options or default option would work, too, and may make more sense in an R context -- having to uninstall / reinstall seems like a bad workflow. Logically I think most of the work will be against this production Magma URL, but it feels weird to have it hardcoded or to force someone to edit source code if they want to hit a different endpoint.

coleshaw:
>  Imports: RCurl, jsonlite

Did you figure out a way to get `jsonlite` to work with the Magma endpoints? Or is that for the Dash / visualization part? Do you need to add an import for `rjson`, if there isn't a way to massage `jsonlite` output?

magmaR/R/query.R
dtm2451: These:
https://github.com/mountetna/monoetna/blob/ff2dbfab9bc1cfae75032183cd127a6776f138ff/magmaR/R/retrieve.R#L115
https://github.com/mountetna/monoetna/blob/ff2dbfab9bc1cfae75032183cd127a6776f138ff/magmaR/R/query.R#L61

Agree that users editing source code is not ideal. I hadn't thought towards how best to add in non-production capability until this discussion so just to be clear, it's not that I was leaning that direction.

Anyway, *should* work now via giving `retrieve/query(..., url.base = "staging_url")`, but would you perhaps be able to check for staging / dev?  I think I'm only set up for dev, but I don't actually have that running right now.

Code for testing would be:
```
# Install current:
remotes::install_github("mountetna/monoetna@magmaR", subdir = "magmaR")

# RESTART your R session (in RStudio, it's in the menu bar Session > Restart R)
library(magmaR)
retrieve(
    projectName = "ipi",
    modelName = "sample",
    recordNames = "all",
    attributeNames = "tissue_type",
    url.base = *staging/dev*,
    token = *appropriate_token*)
```

coleshaw: Thanks for the R code, very helpful!

I think making the url configurable with `.base` looks like it works. But this also leads down a rabbit hole I think we should put off until later. The issue will be that staging / dev use self-signed SSL certificates, so you get an error like:

```
Error in function (type, msg, asError = TRUE)  : 
  SSL certificate problem: unable to get local issuer certificate
```

But I think most folks will only need production, and it may not be worth trying to fix at this point.

dtm2451:
> > Imports: RCurl, jsonlite
> 
> Did you figure out a way to get `jsonlite` to work with the Magma endpoints? Or is that for the Dash / visualization part? Do you need to add an import for `rjson`, if there isn't a way to massage `jsonlite` output?

I did, yea. It was mostly a matter of setting `toJSON(..., auto_unbox = TRUE)` plus a teeny bit more.  Further justifying the `jsonlite` choice: `jsonlite` also offers lots of other points of control in the `toJSON()` function that `rjson`'s does not.

coleshaw:
Nice find! `auto_unbox` ... would not have recognized that from the name of the flag, but that does make sense.

magmaR/R/query.R
graft: The dev cert uses a custom CA, it is not "self-signed" like staging; we should just convert staging to use the same custom CA.

coleshaw: Oh that's a good idea. That should be simple to do -- monoetna has a helper script to generate certs off of that CA.

graft:
Just a note - there are a bunch of places where you reference ucsf.edu - I would generally like to consider this as a bad smell; if the code is free of the string "ucsf.edu" then it means that there must exist a proper configuration path for specifying instance-specific path names.

graft:
@dtm2451 I would move this code to /etna/packages/magmaR (this will cause it to get pushed out into the mountetna/etna repo where it will be distributed with the other etna clients) and call it a day on this PR.

dtm2451:
Moving the package per @graft's suggestion. See https://github.com/mountetna/etna/pull/64

coleshaw:
@dtm2451 , since you have some demo-able work today, is that a clean stopping point for this PR, where you would feel comfortable merging this in? Then any future work to add in tests or feature updates can be in new, separate PRs -- that way each PR can be relatively small and more manageable to review.

dtm2451:
@coleshaw it's close. Probably today or tomorrow.

I would like to rewrite some of the documentation first. And probably remove the `retrieveAttributeCounts()` function too because I think that one was really just quantifying my tsv-parsing bug.

dtm2451:
Ready now from my perspective!

coleshaw (approved):

.gitignore
```diff
@@ -60,3 +60,13 @@ build-iPhoneSimulator/
...
+
+*.DS_Store
+
+# For magmaR dev via RStudio.
+etna/packages/magmaR/.Rproj.user
+etna/packages/magmaR/.Rhistory
+etna/packages/magmaR/.RData
+etna/packages/magmaR/.Ruserdata
+etna/packages/magmaR*.tar.gz
+.Rproj.user
```
dtm2451, line 72: Just a note that although we don't technically need this line in the .gitignore, there's a current bug in RStudio where it will continuously try to add it if we don't put the line.  Doesn't hurt us any, so why not just have it.

dtm2451:
So wait... should I just like, merge this in?  I know I have the power to do so, but I want to confirm lol.

coleshaw:
I think you can go ahead and merge this in. Changes you make in the future should go into a different branch and different PR.

PR #141, opened 2020-11-12
create metis bucket for magma on project creation

coleshaw:
This PR adds an action for Magma to create it's own bucket in Metis whenever a project is created.  It would be nice to get this in before the data scientists start using the etna gem to create projects, so that we don't have to create the `magma` bucket for them.

Also changes the temp file path for Metis upload URLs to use `tmp-` instead of `tmp/` so that we don't have to create an extra folder.

Updates the Metis `create_bucket` restrictions so that if `bucket_name` == `owner`, then the request has to have a valid HMAC signature from that new `owner`.


magma/lib/magma/actions/add_project.rb
```diff
@@ -37,6 +38,60 @@ def validate_project_name
...
+      )
+
+      params = {
+        owner: 'magma',
+        description: 'For magma use only',
+        access: 'administrator'
+      }
+
+      # Now populate the standard headers
+      hmac_params = {
```
graft, line 71: Hmm, looking at the metis bucket#create endpoint it does not appear to make any checks whatsoever that the create request is signed by the assigned owner (i.e., i can create a bucket with owner: 'magma' without Magma signing the request). Dunno if we should resolve this by fixing the endpoint or by removing this hmac param here.

graft: Thinking on this a bit more, there are a couple of other checks that we suggested at one point that don't exist on `bucket#create` - checking that someone else doesn't make the `magma` bucket and set its owner to something other than `magma`. Probably the right thing to do here is to change `bucket#create` so it won't allow you to create a bucket with `@params[:bucket_name] == @params[:owner]` unless it is signed by `@params[:owner]`

magma/lib/magma/actions/base_action.rb
```diff
@@ -2,9 +2,10 @@
 
 class Magma
   class BaseAction
-    def initialize(project_name, action_params = {})
+    def initialize(project_name, user, action_params = {})
       @project_name = project_name
```
graft, line 6: You already made this extensive revision, but - since `user` is not required for most actions, would it be easier to simply merge it into `action_params` for the add_attribute action?

PR #144, opened 2020-11-16
Show type errors on add_attribute action

graft:
Just a little one to add some error-handling to the add_attribute action, which I often have trouble with because of missing parameters.

### DIFF
--- a/magma/lib/magma/actions/add_attribute.rb
+++ b/magma/lib/magma/actions/add_attribute.rb
@@ -32,12 +32,22 @@ class Magma
     def validations
       [
         :validate_model,
+        :validate_type,
         :validate_attribute_name_unique,
         :validate_restricted_attribute,
         :validate_options,
         :validate_not_link,
         :validate_attribute,
       ]
+    end
+
+    def validate_type
+      return if Magma::Attribute.attribute_types.include?(@action_params[:type])
+
+      @errors << Magma::ActionError.new(
+        message: 'Type is invalid',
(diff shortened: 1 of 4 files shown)


magma/lib/magma/actions/add_attribute.rb
```diff
@@ -32,6 +33,15 @@ def validations
       ]
     end
 
+    def validate_type
+      return if Magma::Attribute.attribute_types.include?(@action_params[:type])
+
+      @errors << Magma::ActionError.new(
+        message: 'Type is invalid',
+        source: @action_params.slice(:action_name, :model_name)
```
corps, line 41: `:model_name` => `:type`

magma/lib/magma/model.rb
```diff
@@ -1,9 +1,7 @@
 class Magma
   class Model < Sequel::Model
     class << self
-      Magma::Attribute.descendants.
-        reject { |attribute| attribute == Magma::ForeignKeyAttribute }.
-        each do |attribute|
+      Magma::Attribute.type_attributes.each do |attribute|
```
corps, line 4: I'm curious; what is this change?

graft: I noticed there is no method collecting the actual list of magma types; because of a weird inheritance pattern all the descendants of Magma::Attribute are type classes except ForeignKeyAttribute, which is a base class for some others. The only place this logic was codified was here, so i just isolated it in a method. At some point maybe ForeignKey could be refactored into a module for consistency.

PR #145, opened 2020-11-16
R support in etna

coleshaw:
This PR installs the latest `r-base` for R 4 package into the `etna` Docker image and container. (currently version `4.0.3`). I can run simple R scripts when I go into the container.

Also, this stubs out where we'd need to add R commands for running tests automatically as part of CI. Those will probably have to be tweaked for syntax and whatnot.


etna/.github/workflows/r-tests.yml
```diff
@@ -0,0 +1,11 @@
+name: Run R tests
```
corps, line 1: If we add the R test stuff into the `release-test` and `test` subsections we shouldn't need this?  The base workflow runs the these actions automatically.  Although adding it here will cause the tests to also run in the etna repo itself.

coleshaw: Cool, thanks, that's good to know. I think it's still good to add have the tests running in the separate repos (I ran into an issue last week where the CI results were different for Timur between `monoetna` and `timur`), so I'll try to get these working in `etna`, too.

corps: Do you know what the discrepancy is?  In general, if we identify those issues, we should try to get both repos to agree on test behavior.  I'd hate for, say, one of the repos to be running the tests wrong and misreport success / failure.  I can take a look if you have an example.

coleshaw: Yeah, I think it was related to the issues you've run into with `npm link` and getting `etna-js` into Timur. There were dependencies in `etna-js` that were found / installed just fine in the docker container CI (`monoetna`), but weren't present in the standalone Timur repo and also thus not found during actual deployments, causing webpack to crash. I think it's related to [this code for etna install in the containers](https://github.com/mountetna/monoetna/blob/master/docker/etna-base/entrypoints/build.sh#L32)?

coleshaw: https://github.com/mountetna/timur/runs/1397864450?check_suite_focus=true

corps: Ah shit, yea.  So this is a known bug in npm.  Basically, in normal circumstances, npm will cascade dependencies of dependencies on install.  _Except for linked packages_ which are bizarrely only installed in a shallow way.  (https://github.com/npm/npm/issues/17287#issuecomment-400833982) This is going to not be fixed for awhile, and the only solution is we must guarantee that all dependencies of etna-js are _also_ installed in those subprojects, until we finally move to full docker (in which case the linking will no longer be necessary).

Ack, figured this would bite us.

etna/Makefile
```diff
@@ -7,10 +7,12 @@ export BUILD_REQS:=../docker/etna-base $(BUILD_REQS)
...
 	@ docker-compose run -e RUN_NPM_INSTALL=1 -e SKIP_RUBY_SETUP=1 --rm etna_app npm test
+	# We will have to add in a command here to test the R stuff, too.
 
 bash::
 	@ docker-compose run -e SKIP_RUBY_SETUP=1 --rm etna_app bash
 
 release-test::
 	docker run --rm -e APP_NAME=etna -e RELEASE_TEST=1 -e SKIP_DB_WAIT=1 -e CI_SECRET=$${CI_SECRET} -e IS_CI=$${IS_CI} --network monoetna_default $(fullTag) /entrypoints/development.sh bundle exec rspec
 	docker run --rm -e APP_NAME=etna -e RELEASE_TEST=1 -e SKIP_RUBY_SETUP=1 -e RUN_NPM_INSTALL=1 -e SKIP_DB_WAIT=1 --network monoetna_default $(fullTag) /entrypoints/development.sh npm test
+	# We will have to add in a command here to test the R stuff, too.
```
dtm2451, line 18: Seems like this PR altogether adds R installation into the docker image & preps for having R tests via GitHub actions, both of which are great.  I'm just commenting for my own understanding...

Would the stub test lines here be replaced with just like a line of R code that will work only if R is installed?  Or would the goal be to point to the magmaR test suite?

coleshaw: Yup, exactly! These commented out lines would basically be replaced with a shell command (or however R tests are invoked?) that would run the actual magmaR test suite -- the idea is that on every commit, GitHub will automatically run these set of tests (Ruby tests, JavaScript tests, and R tests) and notify us of any failures. Each PR should add or modify tests as appropriate, to make sure that things don't break. And we don't merge code unless all tests pass and you see the green checkmark on the PR.

PR #146, opened 2020-11-17
update metis to nest data blocks two levels using leading hash charac…

coleshaw:
…ters

This PR re-configures Metis to use the first two characters of each data_block to nest them in subdirectories. This should help prevent too much data in a single directory, which could lead to a problem as we upload more data to Metis.

This is also a proposed script that we can use to re-organize the existing data blocks on Metis, when we deploy this:

```
for file in * 
do                                
    if [ -f "$file" ]; then

        [ "$file" == "reorg_metis.sh" ] && continue

        STOREDIR=${file:0:1}/${file:1:1}

        if [ -d ${STOREDIR} ]         
        then
            mv ${file} ${STOREDIR}  
        else                      
            mkdir -p ${STOREDIR}    
            mv ${file} ${STOREDIR} 
        fi
    fi                        
done
```

I will check with Harry to see how this affects backups.


graft:
Nooice! We might have to schedule some downtime to do this...

coleshaw:
Yeah, migrating the blocks might take a little while. I can check when might work for folks and do it morning my time.

PR #147, opened 2020-11-18
do not allow _# in project names

coleshaw:
Ran into this issue while testing this morning -- creating a project with an `_` in its name actually breaks Magma.

```
$ etna administrate project create cole_1 'Cole test 1' --environment development
<blah blah stack trace>

/etna-0.1.16/lib/etna/client.rb:130:in `status_check!': [{"message":"Unexpected error","source":null,"reason":"PG::InvalidSchemaName: ERROR:  schema \"cole1\" does not exist\nLINE 1: CREATE TABLE \"cole1\".\"projects\" (\"id\" integer GENERATED BY D...\n                     ^\n"}] (Etna::Error)
```

The Magma logs also show it trying to create a table with schema `cole1` which doesn't exist. In fact, for some reason it doesn't even create a schema for `cole_1` or `cole1`, though it's supposed to create a schema for `cole_1` based on the project name. I think it has to do with the transition between snake_case and camel_case, like the issue with model names from #110.

But no `_` works fine:

```
$ etna administrate project create cole2 'Cole test 2' --environment development
Creating Janus project.
Done! Adding you as an administrator on the project.
Add more users? Y/n
n
Done with setting up the project in Janus!
Creating the project in Magma.
Done! Adding your new project record.
All complete!
You need to visit Janus to refresh your token.
You can now log into any app to manage your data.
```

So this PR just makes the project name more restrictive in Janus and Magma, to just lowercase letters + numbers, with no leading number.

### DIFF
--- a/janus/Gemfile
+++ b/janus/Gemfile
@@ -6,7 +6,7 @@ gem 'rack'
 gem 'rack-throttle'
 gem 'pg'
 gem 'sequel'
-gem 'etna', git: 'https://github.com/mountetna/monoetna.git', branch: 'refs/artifacts/gem-etna/de61c828a97a0f29d1076c1d7882e7bc1d75728c'
+gem 'etna', git: 'https://github.com/mountetna/monoetna.git', branch: 'refs/artifacts/gem-etna/62cf1216a0f46d0b31a9b0e1e80ba583199bd7ec'
 gem 'jwt'
 gem 'puma', '5.0.2'
 
--- a/janus/Gemfile.lock
+++ b/janus/Gemfile.lock
@@ -1,7 +1,7 @@
 GIT
   remote: https://github.com/mountetna/monoetna.git
-  revision: d4df448e7c828bd3e370ced905d31ecc2fd98944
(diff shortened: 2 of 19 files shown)


graft:
Hmm, there shouldn't be an issue here - the schema name is snake_case for postgres, and you can create a schema with a _ in it just fine. Perhaps somewhere along the way cole_1 => Cole1 => cole1 is happening, as you suggest. My strong, strong preference would be not to sacrifice the _, which is extremely valuable as our only separator in these strings; if we can't resolve the case conversion path issue, we might require a stronger snake_case where _ must be followed by [a-z]?

coleshaw:
Okay, this now works with `_` but no numbers directly afterwards:

```
$ etna administrate project create cole_x5 'Cole test x5' --environment development
Creating Janus project.
Done! Adding you as an administrator on the project.
Add more users? Y/n
n
Done with setting up the project in Janus!
Creating the project in Magma.
Done! Adding your new project record.
All complete!
You need to visit Janus to refresh your token.
You can now log into any app to manage your data.
```

corps (approved):

PR #149, opened 2020-11-19
various magmaR updates

dtm2451:
... that were talked about in standup yesterday or Monday:

- Adds a `.MAGMAR_URL` global variable, with paired .get_URL() internal function, so that a user can set the package to work with staging or dev via single line, e.g. `.MAGMAR_URL <- "http://magma.development.local"`, at the top of their code. Default = production url.
- Also changes `.TOKEN` to `.MAGMAR_TOKEN` to match ^^ and be all that more likely not to interfere with a user's data.
- Adds a `retrieveModels()` function
- Renames the `retrieveAttributeNames()` function `retrieveAttributes()` to be consistent with similar functions.
- Adds details to the `filter` input documentation for `retrieve()` (an example followed by the link to the online /retrieve doc.)
- Removes "Are you signed into vpn?" from the error given when a user's request is not authorized. (New error: "You are unauthorized. If you think this is a mistake, run `rm(.MAGMAR_TOKEN)` or update your 'token' input, then retry.")
- Abstracts the RCurl code that was previously duplicated between `.retrieve()` and `.query()` into a single `.perform_curl()` function.

Also shares the demo code (updated today) that I had shown in our standup: https://github.com/mountetna/monoetna/blob/db/magmaR-update-base-url/etna/packages/magmaR/magmaR_demo.R


dtm2451:
Ready from my perspective.  Should someone review before I merge it in? @graft @corps @coleshaw ?

corps (approved):
:+1:

PR #154, opened 2020-11-25
Addressing config behavior

corps:
Changes the way config works in the etna bin based on my experience with Ravi.  This PR may generate some discussion and be controversial but I think it's important to consider ahead of rolling out for more of our users.

There were three main issues when guiding Ravi through config setup
1.  Ravi needed to swap out tokens in order to call `etna config set` for staging and production.  Because the config endpoint authenticates, it requires users to remember to swap tokens just to get setup.  This was confusing and non obvious since the authentication failing obviously has no way to inform a user that they need to go back to janus and swap tokens for one in the environment of the configuration setup.
2.  Remembering to use `--ignore-ssl` for staging but not production.  This one is difficult to explain the why for, but again, the error messages around the http client failing are not super useful, and it wasn't clear why one needed this and other didn't.  It's kind of magic that has to be memorized.
3.  Took some to explain 'when' to use the config commands.  I explained he shouldn't need to use them most of the time after we successfully set it up, but he 'might' need to run them again for both staging and production if there's a change or upgrade.

I felt like this was a pretty bad experience, so I spoke with @coleshaw and we came up with alternative

1.  Package production and staging config defaults.    When upgrading the etna gem, production and staging config defaults come 'for free', so for most users config set is never needed.
2.  Add the concept of enabled environments.  By default, staging is not enabled and must be enabled with a single command.  This command doesn't require memorizing much, and only ever needs to be run once.  External app users never need to run this at all.
3.   Keep config set for other environments (development, dogfood, etc) as is.

I feel this will greatly reduce the overhead spent trying to keep user's configurations working overall.

### DIFF
--- a/etna/bin/etna
+++ b/etna/bin/etna
@@ -9,10 +9,59 @@ require 'yaml'
 require_relative '../lib/etna'
 require_relative '../lib/commands'
 
-config = if File.exists?(EtnaApp.config_file_path)
-  YAML.load(File.read(EtnaApp.config_file_path))
-else
-  {}
+# Base config enabled
+config = {
+    production: {
+        docker: {
+            default_tag: 'latest',
+        },
+        magma: {
+            host: 'https://magma.ucsf.edu',
+        },
+        metis: {
+            host: 'https://metis.ucsf.edu',
+        },
+        janus: {
+            host: 'https://janus.ucsf.edu',
+        },
+        timur: {
+            host: 'https://timur.ucsf.edu',
+        },
+        polyphemus: {
+            host: 'https://polyphemus.ucsf.edu',
+        },
+        auth_redirect: 'https://janus.ucsf.edu',
+        ignore_ssl: false,
+    },
+    staging: {
+        docker: {
+            default_tag: 'latest',
+        },
+        magma: {
+            host: 'https://magma-stage.ucsf.edu',
+        },
+        metis: {
+            host: 'https://metis-stage.ucsf.edu',
+        },
+        janus: {
+            host: 'https://janus-stage.ucsf.edu',
+        },
+        timur: {
+            host: 'https://timur-stage.ucsf.edu',
+        },
+        polyphemus: {
+            host: 'https://polyphemus-stage.ucsf.edu',
+        },
+        auth_redirect: 'https://janus-stage.ucsf.edu',
+        ignore_ssl: true,
+    },
+    local: {
+        enabled_environments: [ :production ],
+    }
+}
+
+if File.exists?(EtnaApp.config_file_path)
+  config.update(YAML.load(File.read(EtnaApp.config_file_path)))
 end
 
 EtnaApp.instance.run_command(config, *ARGV)
--- a/etna/etna.completion
+++ b/etna/etna.completion
@@ -512,12 +512,100 @@ all_flag_completion_names="$all_flag_completion_names  "
 string_flag_completion_names="$string_flag_completion_names  "
 while [[ "$#" != "0" ]]; do
 if [[ "$#" == "1" ]];  then
-all_completion_names="help set show"
+all_completion_names="disable enable help set show"
 all_completion_names="$all_completion_names $all_flag_completion_names"
 if [[ -z "$(echo $all_completion_names | xargs)" ]]; then
 return
 fi
 COMPREPLY=($(compgen -W "$all_completion_names" -- "$1"))
+return
+elif [[ "$1" == "disable" ]]; then
+shift
+if [[ "$#" == "1" ]];  then
+all_completion_names="__environment__"
+if [[ -z "$(echo $all_completion_names | xargs)" ]]; then
+return
+fi
+COMPREPLY=($(compgen -W "$all_completion_names" -- "$1"))
+return
+fi
+shift
+all_flag_completion_names="$all_flag_completion_names  "
+string_flag_completion_names="$string_flag_completion_names  "
+while [[ "$#" != "0" ]]; do
+if [[ "$#" == "1" ]];  then
+all_completion_names=""
+all_completion_names="$all_completion_names $all_flag_completion_names"
+if [[ -z "$(echo $all_completion_names | xargs)" ]]; then
+return
+fi
+COMPREPLY=($(compgen -W "$all_completion_names" -- "$1"))
+return
+elif [[ -z "$(echo $all_flag_completion_names | xargs)" ]]; then
+return
+elif [[ "$all_flag_completion_names" =~ $1\  ]]; then
+all_flag_completion_names="${all_flag_completion_names//$1\ /}"
+a=$1
+shift
+if [[ "$string_flag_completion_names" =~ $a\  ]]; then
+if [[ "$#" == "1" ]];  then
+a="${a//--/}"
+a="${a//-/_}"
+i="_completions_for_$a"
+all_completion_names="${!i}"
+COMPREPLY=($(compgen -W "$all_completion_names" -- "$1"))
+return
+fi
+shift
+fi
+else
+return
+fi
+done
+return
+elif [[ "$1" == "enable" ]]; then
+shift
+if [[ "$#" == "1" ]];  then
+all_completion_names="__environment__"
+if [[ -z "$(echo $all_completion_names | xargs)" ]]; then
+return
+fi
+COMPREPLY=($(compgen -W "$all_completion_names" -- "$1"))
+return
+fi
+shift
+all_flag_completion_names="$all_flag_completion_names  "
+string_flag_completion_names="$string_flag_completion_names  "
+while [[ "$#" != "0" ]]; do
+if [[ "$#" == "1" ]];  then
+all_completion_names=""
+all_completion_names="$all_completion_names $all_flag_completion_names"
+if [[ -z "$(echo $all_completion_names | xargs)" ]]; then
+return
+fi
+COMPREPLY=($(compgen -W "$all_completion_names" -- "$1"))
+return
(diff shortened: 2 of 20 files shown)


corps:
Other potential alternative if packaging the configuration directly is a non starter: what if we post the base `etna.yml` file into slack and ask all users to download and install it in their home directories?  That would achieve a similar effect of preventing people from generally having to wrangle the config set stuff.

graft:
The alternative i was going to suggest was an etna-ucsf gem under @UCSF-DSCOLAB; you could include the etna.yml, make other custom installs and also require etna gem, a single install that is auto-configured without defaults in etna.

coleshaw:
With an `etna` gem in the DSCOLAB organization, does that make it non-installable via `gem install`? Like, you would have to add a source repository / use a GH URL instead of the name of the gem? Does that add some extra, non-standard steps (as opposed to `gem install etna`? If so, posting the file in Slack almost seems easier...

coleshaw:
Or, do you mean a private "ruby gems" organization, where folks can `gem install etna-ucsf` from (if they offer something like that)?

graft:
Hmm, i think it could be both a public repo and published to rubygems if
there are no secrets (which there shouldn't be). The gem command won't
install from github anyway (although bundler would).

corps:
I don't see much advantage in creating new repos and increasing complexity of development, so here's my current proposal:

1.  Add new gemspec to the etna repo, which is already part of monoetna, called dscolab-etna.
2.  Gemspec is identical to etna, but includes the base etna.yml configured for dscolab.  Explicitly package entire source code dependencies to simplify end user experience (don't want to require users to have to learn to manage shared gem versions aka DLL hell).
3.  Etna github action publishes both etna gem and dscolab-etna when either version bumps.

It's low effort, and clearly distinguishes the gem with preference for dscolab configuration from the 'general' gem with non preference.  This can be expanded to other organizations or use cases fairly simply.

corps:
I guess I'm just confused by how far we want to spend to create this strong distinguishment of etna "the general library currently only used by dscolab" and "etna-dscolab, the main reason we're building etna".  

Splitting by repos increase lateral costs for development; there is even more syncing from monoetna, it becomes more painful  to develop or test against source code, it requires more process for release, and debugging involves unwinding the intermediate abstraction (the gem with one version that wraps the gem with another version).  So I am strongly against that if we can avoid it, especially when for the near future we will likely be iterating heavily on this tool to support multiple projects.

If we can't reach consensus, I'm also ok with simply "doing nothing", and I'll continue just slacking the etna.yml file to people and asking them to install by hand.  :shrug:

coleshaw:
Yeah, I agree that adding a separate repo (whether in this org or another one) for a wrapper gem seems like it is going to add a non-insignificant amount of overhead trying to keep those two things in sync and may cause issues / confusion with debugging. I think flat file in Slack or turning off authentication on the Polyphemus config endpoints seem like more manageable options, other than just a flat file in the gem.

graft:
I also agree the extra repo seems like a bad idea, i think i favor the
noauth config endpoint on polyphemus at this point.

corps:
Ok.  In that case, could we make the ssl configuration implicit?

IE: production _requires_ ssl.  No choice in the matter.

Reasoning: giving the option to ignore-ssl to a user requires them to essentially _know something about the state of the certificates in the environments they are trying to use_.  

In reality users won't know this or care about, and will likely copy paste commands to get this stuff put together.  Ignoring ssl in production _works_ as far as the user sees, but it it is equivalent to silently using plaintext http to move potentially PI onto servers.  I can't think of any possible case where this is desirable.

Broadly, I think, we need to decide about staging: do we want to guarantee staging's security and actually use real certificates for all services, not self signed ones?  Do we want to require the user to acquire install and update our self signed root certificates themselves based on the 3 operating systems (sounds like a use case for the etna gem to have pre-built configuration and base certificates to do that itself)?  Or do we simply decide that staging is fundamentally a testing ground for not-real data, and ignoring ssl (== sending plain text over HTTP, essentially) is reasonable under the assumption that no one ever uploads PI to it?

This whole, "etna is a generic framework for anyone in any environment" and "etna is a way for our users to interact with our DL" have conflicting requirements.

corps:
That's basically my last concern.  If we can figure out our policy about ssl, and remove the user step there, we'll have a decent approach.

coleshaw:
If we get rid of the auth config for polyphemus configuration, couldn't we just run the `etna config blahblah` commands on gem installation, so the user doesn't have to do anything?

I definitely think our policy should be no PHI in staging...you have to be on the VPN anyways to access it, and there really isn't a reason for someone to upload PHI to it?

corps:
That's a good point about staging, it's already over VPN.

So we can just enforce production = ssl, staging = ignore ssl, eveyrthing else requires configuration? (most users won't use anything else anyways)

Ok cool, this becomes alot simpler.  I might also extend the config set command so you can atleast specify multiple environments together and just 'bulk' setup (copy paste one command) since we no longer need to swap out tokens anyways.

graft:
None of the services can or should ever run without SSL (except maybe
in a dev environment), because the token would be compromised, and
none of the services run on http in any environment that i'm aware of
(even in dev), so I'm unclear on why this would ever happen. However,
ignoring certificate errors is probably also bad, in that in theory
someone with a lot of patience could substitute a compromised cert and
do some bad things in staging (or production if they --ignore-ssl
there for some reason? seems an unlikely combination of events).

Here are the options i see:
1) Keep --ignore-ssl and use it for staging. Ugly and insecure.
2) Remove --ignore-ssl but require users to add our root CA to their
system/browser stores. Onerous indeed across many systems and
browsers.
3) Get staging some real SSL certificates, use SSL everywhere.

corps:
> However,
ignoring certificate errors is probably also bad, in that in theory
someone with a lot of patience could substitute a compromised cert and
do some bad things in staging (or production if they --ignore-ssl
there for some reason? seems an unlikely combination of events).

Less unlikely than you think.  You do not need to compromise the remote server.  All you need is a wifi interceptor in a backpack and sit next to a victim.  You intercept the HTTPS request, and send your own response back.  If you don't verify the certificate, _anyone's claim to being our servers is as good as the next_.

That said, if we can simply safely assume that staging is in VPN anyways, with which modern implementations already encrypt, we might be ok to simply drop the flag for anything other than development, where it truly doesn't matter (development services run on loopback interface and thus could not be intercepted).

coleshaw:
> Here are the options i see:
1) Keep --ignore-ssl and use it for staging. Ugly and insecure.
2) Remove --ignore-ssl but require users to add our root CA to their
system/browser stores. Onerous indeed across many systems and
browsers.
3) Get staging some real SSL certificates, use SSL everywhere.

I think #1 and #3 are the least-work-for-user options. If we want to go with 3, I can request them, we'll just have to watch the expiration dates.

coleshaw:
> That said, if we can simply safely assume that staging is in VPN anyways, with which modern implementations already encrypt, we might be ok to simply drop the flag for anything other than development,

A little confused by this :arrow_up: . If we assume VPN, don't we still need the `--ignore-ssl` flag on staging with the current, expired certs? The only way we can drop the `--ignore-ssl` flag is if we get valid SSL certs?

corps:
> > That said, if we can simply safely assume that staging is in VPN anyways, with which modern implementations already encrypt, we might be ok to simply drop the flag for anything other than development,
> 
> A little confused by this arrow_up . If we assume VPN, don't we still need the `--ignore-ssl` flag on staging with the current, expired certs? The only way we can drop the `--ignore-ssl` flag is if we get valid SSL certs?

If we assume VPN, `staging` can simply ignore ssl without configuration from the user.  It becomes a safe default behavior.  This is because, whilst without verifying SSL it would be possible to MITM the HTTP connection, VPN connections themselves use certificate verification on the outer connection, meaning that an attacker still has a layer to compromise.  Verifying http ssl inside of a VPN isn't as strictly necessary.  In other words, the flag can disappear and we can safely assume the default of ignore, _for staging_.

coleshaw:
Ah, got it. Thanks for clarifying! :+1:

coleshaw:
It is also relatively simple to request SSL certs from IT, and I can do that if we want to have valid certs on all the servers.

corps:
> It is also relatively simple to request SSL certs from IT, and I can do that if we want to have valid certs on all the servers.
> […](#)
> On Nov 30, 2020, at 19:26, Zach Collins ***@***.***> wrote: ﻿ However, ignoring certificate errors is probably also bad, in that in theory someone with a lot of patience could substitute a compromised cert and do some bad things in staging (or production if they --ignore-ssl there for some reason? seems an unlikely combination of events). Less unlikely than you think. You do not need to compromise the remote server. All you need is a wifi interceptor in a backpack and sit next to a victim. You intercept the HTTPS request, and send your own response back. If you don't verify the certificate, anyone's claim to being our servers is as good as the next. That said, if we can simply safely assume that staging is in VPN anyways, with which modern implementations already encrypt, we might be ok to simply drop the flag for anything other than development, where it truly doesn't matter (development services run on loopback interface and thus could not be intercepted). — You are receiving this because you were mentioned. Reply to this email directly, view it on GitHub, or unsubscribe.

Yeah.  I mean either way, having certs is probably strictly speaking the -safest- route forward for all purposes.  I just wasn't sure if there was any bigger reason we hadn't rolled that out (I guess because rotating certs is a PITA, but).

coleshaw:
I'll put in SSL cert requests. It's kind of sad, I sent this comment via e-mail yesterday... took awhile to show up:

> It is also relatively simple to request SSL certs from IT, and I can do that if we want to have valid certs on all the servers.

PR #155, opened 2020-11-26
Add unit testing to magmaR

dtm2451:
Got it working! Holiday time now =)

## How it works

All testing materials are in the `magmaR/tests` folder.  Unit tests, `magmaR/tests/testhat/`, obtain a token and target url from an `.Renviron` file (not tracked, but easy to create! see `magmaR/clear_stubs_for_true_test.R`), then upon first run of each individual test, the `vcr` package saves stubs to the `magmaR/tests/fixtures` folder.  Any instances of the token are scrubbed from the stubs and replaced with dummy text.  Subsequent testing, currently only works *on the same system*, then does not require a token or internet connection. (This can be made non-system/clone-specific, but would require tracking the stubs via GitHub. I don't do that currently because the stubs contain real data. See below.)

### Some concerns that I have:

1. The stubs end up being rather large, ~45-50mb. There are some tests which do target large-ish amounts of data, so this could be addressed, but I'm not sure how much we could shrink that down without creating a mock data used specifically for testing purposes.
2. Tracking stubs would expose real data via GitHub. To prevent this, stubs are not currently tracked. Ideally they should be.
3. Tests only work for production currently because there is no guarantee of ipi project consistency from production to staging / dev.

### To Do ideas, likely for a future PR:

- Create a mock project that can exist in all of production/staging/development
- Make tests target this mock project so they 1) target smaller data chunks, 2) don't expose real data in stubs, and 3) can work for dev & staging too
- Add magmaR testing to GitHub actions!

### Changes made in this PR:
- Updates `.perform_curl()`, `.retrieve()`, and `.query()` to switch the curl interface from `RCurl` to `crul`. For compatibility with `webmockr`/`vcr` stubbing, but it seems like a better interface anyway!
- Creates mechanism for clearing the stubs and setting a token for testing: Run the `clear_stubs_for_true_test.R` script.
- Adds unit tests for all functions.

### Bug Fixes:
- Corrects `retrieveMetadata()`'s id mapping from meta_model to target_model
- Ensures no column names (=attributes names) should be modified within data merges.


dtm2451:
Sharing because of the joy this brings, at least to me =p.
image

graft:
Your tests should not use ipi - same goes for examples in the vignette i think. My habit has been to make toy projects for examples (e.g. spec/labors in magma). This might be insufficient for some really complex functions but for basic client interactions in magma there is no need to use real data and plenty of reason to avoid it (notably, we should avoid a path to introducing PHI into a public repo). With a few well-crafted examples you should be able to test your client behavior without having to use real data. This should give you nice skinny stubs you can check on (which is required before tests can run in CI).

graft:
Should add: nice work, setting up a test harness is not easy!

dtm2451:
Thanks and I definitely agree that tests and the vignette should not be based on real data. Swtching all tests and vignette examples to a mock project will be a focus next week. Seems obvious now, but I'd not thought about how testing of other pieces of the pie would have led to creation of some mock projects already. Such projects should make your suggestion / my `To Do ideas, likely for a future PR` section quite doable 'now'...

Perhaps, in our next stand up or with a 1:1 zoom, you or someone could give an overview of the structure of labors?  I can start the migration after.  Also, I think if we can put together a model map for that project, I'd put it in the magmaR vignette && we would want to put it in the magma docs too so that the /retrieve and /query examples are easier to follow.

dtm2451:
I've now swapped the tests to run off of the `example` project which is entirely mock / simulated data.

Once that project is re-populated, hopefully today, any objection to me checking in the stubs, @graft?

Once those are checked in, this PR should be good to go!

dtm2451:
Ready to go / for review from my perspective! =)

Just to update:

## How it works
All testing materials are in the magmaR/tests folder. Unit tests, magmaR/tests/testhat/, obtain a token and target url from an .Renviron file (not tracked, but easy to create! see magmaR/clear_stubs_for_true_test.R), then upon first run of each individual test, the vcr package saves stubs to the magmaR/tests/fixtures folder. Any instances of the token are scrubbed from the stubs and replaced with dummy text. Subsequent testing then does not require a token or internet connection because all curl requests turn to the stubs.

## To Do for a future PR:
- Add magmaR testing to our GitHub actions checks

etna/packages/magmaR/R/retrieveMatrix.R
```diff
@@ -91,16 +91,26 @@ retrieveMatrix <- function(
...
+    # Extract matrix data as a list of columns
     data_cols <- lapply(
         recordNames,
         function(x) {
             (json$models[[modelName]]$documents[[x]])[[attributeNames]]
         })
-    data <- as.matrix(do.call(data.frame, data_cols))
+    
+    # Identify any empty records
+    empty <- vapply(data_cols, FUN = length, FUN.VALUE = integer(1)) == 0
```
coleshaw, line 102: Just curious, is there ever a case where you would want the empty entries there for analysis purposes (like to know you are missing data), and so this filtering would be helpful behind a flag, with a default value? Or basically empty entries are never really useful, so baking in this data-cleaning step is super useful?

metis/bin/metis_client
```diff
@@ -743,13 +743,11 @@ EOT
 
           if current_attempt_number <= max_number_attempts
             puts "UploadError -- retrying the upload"
-            MetisShell.log_error(e)
```
coleshaw, line 746: Can you please revert this to include the `MetisShell.log_error(e)` calls, as in the main branch? We're using this method to provide better error messages to folks uploading data, so we don't want to lose it :smiley:

coleshaw (changes requested):
Everything looks good to me, except please revert the PR changes to `metis/bin/metis_client` so that it matches the current code on the main branch. I'd like to keep the error handling in the Metis Client, as folks are using that feature to check for upload errors. After that's fixed, I think it's good to merge! Great job, the tests look great, and I like the auto-scrubbing of the token from the fixtures!

graft (approved):

graft:
Looks great, ready to merge after rebasing in the changes cole wants. Will tests run in CI?

metis/bin/metis_client
dtm2451: 🤦 the one time I didn't scroll all the way through the files-changed list...

Good catch.  Reverted.

This commit, https://github.com/mountetna/monoetna/commit/6698472b6664035341debaa1fbc3dfd645381666,  had appeared in a different way than I'm used to, and with attribution to me. So I'd reverted it thinking it was a problem, but clearly it isn't lol so, "*re-reverted" to add these changes back in!

etna/packages/magmaR/R/retrieveMatrix.R
dtm2451: Hmmm.  I could definitely see returning the ids of empty records altogether being useful.  I think the way these records come into existence is likely either 1) data gets added for other attributes of the model, but not the target one being retrieved, or 2) a record is "removed".  So perhaps useful for QC?

In this function, the trimming is needed in order to have a matrix as output -- length 0 entries break the rectangular shape of the potential matrix -- and I think we want to keep this output structure.

I do report identifiers of all these records in a warning a few lines below.

But it wouldn't be super easy to scrape the full set of ids together from the warnings... Perhaps collection of such records could be a separate endpoint added in a future PR?

metis/bin/metis_client
coleshaw: No worries! Thanks, this looks great!

coleshaw (approved):

etna/packages/magmaR/R/retrieveMatrix.R
coleshaw: Oh, I see, so you have to keep the shape. Yeah, certainly don't worry about it for this PR, maybe if someone needs it they can also use lower-level functions (using this as an example), in their own R code. Or as we collect feedback, we can see if building in such a feature would be useful or not to bake in.

dtm2451:
@graft the testing isn't integrated into our Github actions pipeline yet, but I don't believe that would be hard to set up!  Tests could be run with this code:

```
devtools::test(pkg = "path/to/magmaR/base/directory")
```

PR #156, opened 2020-11-29
Update views

graft:
Introduces a new view model to Timur, replacing the old trio of view models (view_tab, view_pane, view_attribute). The new view model has a single blob that holds a view JSON. There is currently no validation of the format on this blob in the browse#update endpoint; I had the notion that this would be done when the blob is written (by @Gvaihir's view editor), and thus the entire control of the format would rest in the hands of the javascript client. I'm not sure this is a good idea and could use some feedback on the notion - should there be server-side validation of the view format, since the server doesn't in fact know anything about how the view is rendered?

A few other things are happening here, along with the introduction of the new model - mostly, a separation of the notion of an "attribute" in a Magma record and the notion of a single "item" in the view. Now, an item in the view might *display* a Magma attribute, but the item could also have some other type that displays something else (e.g. a plot). Right now this is pretty basic, with only 'plot', 'markdown', and 'magma' types, but now there is room to add other object types into the view, e.g.: a manifest (runs the manifest and shows the resulting consignment in the view), a folder on Metis, 'markdown' but strings stored in the view json instead of in magma, etc.

This separation results in the disappearance of the dreaded `interleaveAttributes`, probably one of my least-favorite pieces of code in this entire codebase. :volcano: 

Along with this comes a few styling updates to the view panes to make them a little more clearly defined. I also snuck in a few other UI things: clicking dismisses the identifier search, collections flow into columns.


coleshaw:
Sorry, didn't notice this PR last week. I'm running into some bugs when testing this locally, will try to narrow them down and report inline.

corps:
> There is currently no validation of the format on this blob in the browse#update endpoint; I had the notion that this would be done when the blob is written (by @Gvaihir's view editor), and thus the entire control of the format would rest in the hands of the javascript client. I'm not sure this is a good idea and could use some feedback on the notion - should there be server-side validation of the view format, since the server doesn't in fact know anything about how the view is rendered?

I find it reasonable to do the validation client side in this case for a few reasons.
One, it is liightly less critical to be correct than raw data.  For instance, server validations on model and data are super important if the lack of them can cause long running integrity issues, so the extra layer is warranted.
Messing up the format of a view json will cause problems, but I imagine they would be easier for us to address and likely easier to notice (almost immediately).

Validating on the client side is great because _it is a better user experience_.  Unlike the data or modeling, where we have many api consumers, I suspect that the view editing will primarily be edited in the browser.  In general, having faster, more responsive validation in the browser will lead to a nicer experience for the user than relying on server side validation.

Given that, I find client side only validation in this case to be a pretty smart optimization of time and work.

magma/lib/magma/attribute.rb
```diff
@@ -96,8 +96,7 @@ def json_template
         attribute_name: attribute_name,
         model_name: self.is_a?(Magma::Link) ? link_model.model_name : nil,
         link_model_name: self.is_a?(Magma::Link) ? link_model.model_name : nil,
-        type: database_type.respond_to?(:name) ? database_type.name : database_type,
-        attribute_class: attribute_class_name,
```
corps, line 100: I checked our clients, and this seems safe.  All our workflow rely on `attribute_type` and not either of these two.

timur/lib/client/jsx/components/browser/view_items/magma_item.jsx
```diff
@@ -0,0 +1,23 @@
...
+
+const isRevised = (revision, record, att_name) =>
+  att_name in revision && record[att_name] != revision[att_name];
+
+export const MagmaName = ({item, template, mode, revision, record }) => {
+  let { attribute_name, title } = item;
+  let attribute = template.attributes[attribute_name];
+
+  return <div className={ `item_name ${ (mode == 'edit' && isRevised(revision,record,attribute_name)) ? 'revised' : '' }` } title={attribute.desc}>
+    {attribute.display_name || title || attribute_name}
```
coleshaw, line 14: When I navigate to a patient (and the `overview` view), I get a blank screen with this console error thrown around here:

```
No such attribute clinical in template for patient timur.bundle.js:38394:13

TypeError: attribute is undefined
    MagmaName timur.bundle.js:46781
```

Is it because the "tab" headers are being treated as attributes?

timur/lib/client/jsx/components/attributes/attribute_viewer.jsx
```diff
@@ -18,73 +18,49 @@ import MarkdownAttribute from './markdown_attribute';
...
-      case 'BoxPlotAttribute':
-        return <BoxPlotAttribute {...this.props} />;
-      case 'TextAttribute':
-        return <TextAttribute {...this.props} />;
-      case 'MarkdownAttribute':
-        return <MarkdownAttribute {...this.props} />;
-      case 'MetricsAttribute':
-        return <MetricsAttribute {...this.props} />;
+  if (!attribute) {
+    console.log(`No such attribute ${attribute_name} in template for ${model_name}`);
```
corps, line 26: Nit; make it a console.error so it's easier to notice?  Especially because test runners by default filter logs, but not errors, from output.

timur/lib/client/jsx/components/browser/view_items/plot_item.jsx
```diff
@@ -0,0 +1,41 @@
+// Framework libraries.
```
coleshaw, line 1: Hm, this component no longer gets the right props, for me? When I open an `rna_seq` page with plots on it, I get several errors in the console and on the server:

Screenshot from 2020-12-07 14-20-02

```
ERROR:2020-12-07T19:17:17+00:00 niuslr Caught unspecified error
timur_app_1      | ERROR:2020-12-07T19:17:17+00:00 niuslr PG::InvalidTextRepresentation: ERROR:  invalid input syntax for integer: "null"
timur_app_1      | LINE 1: SELECT * FROM "plots" WHERE "id" = 'null'
timur_app_1      |                                            ^
timur_app_1      | 
```

```
172.23.0.10 - - [07/Dec/2020:19:17:17 +0000] "GET /api/plots/ipi/fetch HTTP/1.1" 200 48299
timur_app_1      | WARN:2020-12-07T19:17:25+00:00 User calling browse#view with params {:project_name=>"ipi", :model_name=>"rna_seq"}
timur_app_1      | WARN:2020-12-07T19:17:25+00:00 User calling manifests#fetch with params {:project_name=>"ipi"}
timur_app_1      | WARN:2020-12-07T19:17:25+00:00 User calling plots#fetch with params {:project_name=>"ipi"}
timur_app_1      | ERROR:2020-12-07T19:17:25+00:00 962qqu Caught unspecified error
timur_app_1      | ERROR:2020-12-07T19:17:25+00:00 962qqu undefined method `to_hash' for nil:NilClass
timur_app_1      | ERROR:2020-12-07T19:17:25+00:00 962qqu /app/lib/server/controllers/browse_controller.rb:6:in `view'
```

timur/lib/client/jsx/components/browser/view_items/magma_item.jsx
coleshaw: Running into the same issue with `flow` records...I think the translation from the three different `view_` tables to `views` is not quite doing what it is expected to do?

coleshaw:
So....probably related to the other errors I'm running into, but the browser's "Back" button no longer seems to work for me, in navigating the app, once I hit a blank page? At the very least, I have to click it several times to get any visible change.

timur/lib/client/jsx/components/model_viewer.jsx
```diff
@@ -34,9 +34,8 @@ const ModelViewer = ({
     Cell: ({value, original}) => (
       <AttributeViewer
         template={template}
-        document={original}
-        value={value}
```
corps, line 38: Intentional to drop the value arg?

corps (commented):
Minor feedback, testing more locally.

timur/db/migrations/006_add_views.rb
```diff
@@ -0,0 +1,82 @@
...
+      DateTime :updated_at, null: false
+
+      index [:model_name, :project_name], name: :index_views_on_project_and_model_name, unique: true
+    end
+
+
+    db = Timur.instance.db
+
+    panes = db[:view_panes].all
+    attributes = db[:view_attributes].all
```
corps, line 42: I guess, the critical question is; do all these view_attributes line up to entries in the model.template in all cases?  Seeing similar issues to @coleshaw

graft:
> > There is currently no validation of the format on this blob in the browse#update endpoint; I had the notion that this would be done when the blob is written (by @Gvaihir's view editor), and thus the entire control of the format would rest in the hands of the javascript client. I'm not sure this is a good idea and could use some feedback on the notion - should there be server-side validation of the view format, since the server doesn't in fact know anything about how the view is rendered?
> 
> I find it reasonable to do the validation client side in this case for a few reasons.
> One, it is liightly less critical to be correct than raw data. For instance, server validations on model and data are super important if the lack of them can cause long running integrity issues, so the extra layer is warranted.
> Messing up the format of a view json will cause problems, but I imagine they would be easier for us to address and likely easier to notice (almost immediately).
> 
> Validating on the client side is great because _it is a better user experience_. Unlike the data or modeling, where we have many api consumers, I suspect that the view editing will primarily be edited in the browser. In general, having faster, more responsive validation in the browser will lead to a nicer experience for the user than relying on server side validation.
> 
> Given that, I find client side only validation in this case to be a pretty smart optimization of time and work.

I think one thing I am worried about is writing migrations, or otherwise dealing with format changes. It seems wrong to run any such migration client-side, which might lead to a proliferation of formats and client-side validators?

timur/lib/client/jsx/components/model_viewer.jsx
graft: Yes, it now gets computed inside AttributeViewer, since we are passing in record + attribute name already, it is redundant (and confusing) to also pass in the value, so to simplify the arguments here I moved the inference inside the AttributeViewer.

timur/lib/client/jsx/components/browser/view_items/magma_item.jsx
graft: Hmm, the translation works, it just failed to fully account for a non-existent attribute in the template - in the default development ipi view, there is a spurious attribute "clinical" which should not exist (visible as an extra empty space at the bottom of the first pane the timur.ucsf.edu/ipi patient view). I had noted this before rebuilding the view table and removed it, thus missed this in my review.

timur/lib/client/jsx/components/browser/view_items/plot_item.jsx
graft: I'll investigate this further, however, the plots in my default dev instance are mostly non-functional and seem to all be missing a plot configuration (although they have the scripts). I had to copy the plot config from production ipi to test this out.

coleshaw: I'm not sure my local plots work either, but I expected the requests to at least be for a plot Id, not `null` or `fetch`...?

graft: Sigh, it appears this is the result of bad data. For example, here is the to_hash of the original sample/overview ViewTab record in my timur db:
```{:name=>"overview", :title=>nil, :description=>nil, :panes=>[{:name=>"default", :title=>"Summary", :items=>[{:name=>"patient", :title=>nil, :type=>"magma", :attribute_name=>"patient"}, {:name=>"headshot", :title=>nil, :type=>"magma", :attribute_name=>"headshot"}, {:name=>"tumor_type", :title=>nil, :type=>"magma", :attribute_name=>"tumor_type"}, {:name=>"processed", :title=>nil, :type=>"magma", :attribute_name=>"processed"}, {:name=>"description", :title=>nil, :type=>"magma", :attribute_name=>"description"}, {:name=>"notes", :title=>nil, :type=>"plot", :plot_id=>nil}]}, {:name=>"qc", :title=>"Quality Control", :items=>[{:name=>"qc", :title=>"Quality Control", :type=>"plot", :plot_id=>8}]}]}```

As you can see there is a spurious field "notes" with plot_id "nil" - this is an incorrect piece of data and could never have rendered correctly; this appears to be propagating to the new view. On the other hand you can see the "qc" attribute has a plot_id (with a correctly-configured plot 8), and renders correctly for me.

corps:
Maybe the correct course of action then is to make the views resilient here.  Like the page should render even if a view configuration includes attributes that do not exist.

If we don't do that, we'll further make modifying models painful (renaming columns).

graft:
> Maybe the correct course of action then is to make the views resilient here. Like the page should render even if a view configuration includes attributes that do not exist.
> 
> If we don't do that, we'll further make modifying models painful (renaming columns).

I think it should do that now, following the change in 0a2651c - of course I can't render a plot which isn't defined, but the rest of the page ought to render, even with undefined attributes. Maybe it should obnoxiously surface undefined attributes in the UI?

graft:
Ook, some changes relating to the "default view", when no view has been built yet. It will build off the template. If there are attribute groups defined, it will create tabs for each group (moving the parent attribute and any ungrouped attributes into the default "overview" tab). If there are no attribute groups, it builds a simple view based on all attributes, sorting the parent and any links upwards, and shunting any table or matrix attributes to another tab.

There are also some changes relating to tab loading resulting from the changes to the "default view". I think I have avoided any traps and navigation and tab-based loading seems to still work ok.

coleshaw:
rna_seq_plate#overview still doesn't work for me, do I need to do something manually to make those re-appear?

rna_seq_plate#metrics causes a Server Error, but I suspect that might be because I don't have real data or have a bad manifest somewhere...

Also, there is this weird heading on top of the Plate tab:

Screenshot from 2020-12-11 07-20-49

Is there something I need to do to clean that up?

coleshaw:
Oh, I think the `eisenberg_gene` header in the plate tab is a broken plot. Must be old, because I don't see that in production.

coleshaw (approved):

PR #158, opened 2020-12-01
Tentative workflow for copying around users more easily.

corps:
Just to generate some discussion and in case it is useful, maybe it's not. 

This afternoon I was trying to replicate production projects into staging and get the users setup there so that our data scientists could easily start editing and sharing their tentative modeling. 

It's actually a major PITA to add users en masse.  It's difficult to copy each column bit by bit and set things like affiliation and role correctly as well.  It's also super error prone.

I create a simple workflow to help me.  Maybe we don't include this in our public gem, but it seemed useful enough to keep around in case I needed to do something like this again in the future.


coleshaw:
Looks good to me. Seems like it might be overkill (since most users don't interact with staging), but I can see the utility for bootstrapping projects for the data scientists.

PR #160, opened 2020-12-01
Magma FileCollection attribute

coleshaw:
This PR adds a `FileCollection` attribute type to Magma, where it accepts a list of Files. The user can add an arbitrary number of files. In Timur, these show up like a collection, but show the original file name instead of the `model-record-attribute` format like the `File` or `Image` attributes. This is to avoid confusion when deleting or adding files, since using the Magma generated name would make it very confusing as to what each file actually is.

Currently the Timur UI only allows a user to link files from Metis using a Metis path. Handling multiple uploads for a single attribute is a bit more complicated, and I'd like to save that for a future PR.

Looks like the below in Timur:

Screenshot from 2020-12-01 15-26-54
Screenshot from 2020-12-01 15-27-03


magma/lib/magma/attributes/file_collection_attribute.rb
```diff
@@ -0,0 +1,155 @@
...
+      #   It becomes impossible to do the magma-file-1 -> original-magma-file-0-data-block
+      #   revision.
+      # Thus, we must enforce these requirements:
+      #   1) Any existing files must appear in ascending index order.
+      #       [file[0], file[1], file[7], file[11]] is okay.
+      #       [file[11], file[1], file[0], file[7]] is NOT okay.
+      #   2) New files / Metis paths must appear at the end of the sequence.
+      #       [file[1], file[8], {path: "metis://foo/bar/bim.txt"}] is okay.
+      #       [{path: "metis://bad/revision/order.txt"}, file[1], file[8]] is NOT okay.
+      #   3) Any ::temp files must be at the end of the array, so the filename indices stay sequential.
```
coleshaw, line 68: Not sure keeping the `::temp` files at the end is strictly necessary, but I probably won't figure that out until implementing the upload for FileCollection.

magma/lib/magma/query/predicate/file_collection.rb
```diff
@@ -0,0 +1,53 @@
...
+
+    verb '::original_filename' do
+      child String
+
+      extract do |table, identity|
+        table.first[column_name] ? table.first[column_name].map { |f| f["original_filename"] } : nil
+      end
+    end
+
+    verb '::all' do
```
graft, line 34: What is this verb for? Seems basically the same as '::url' above?

graft: nm, now i get it - i might make this `verb nil` (that is, match this if there are no following args) to avoid confusion with the model predicate's `::all`

graft: nm x2 i now see that i introduced this confusion in `file` already :cry:

magma/lib/magma/attributes/file_collection_attribute.rb
```diff
@@ -0,0 +1,155 @@
...
+      # Note: We'll also have to consider that "new files" may be ::temp or ::blank.
+
+      return "Not all files have the required :path key." unless files.all? { |file| file&.dig(:path) }
+
+      return "Existing files are not in ascending index order." unless existing_files_in_sequence?(record_name, files)
+
+      return "New files must be at the end of the revision." unless new_files_at_end?(record_name, files)
+
+      return "Temp files must be at the end of the revision." unless temp_files_at_end?(record_name, files)
+
```
graft, line 83: I'm not sure I understand the need for all of these ordering requirements for loading, can you clarify?

coleshaw: I'll do my best -- might be easier to hop on a call and chat about it. I think the first two (existing file ordering / new-files last) are hard requirements, for the following reasons.

On the Metis side of things, it processes a set of copy commands sequentially, in the order they are sent. If we start with a set of newly added files, we provide Metis paths like so:

```
[{path: "metis/file-0.txt"}, {path: "metis/file-1.txt"}]
```

Leads to a copy request like:

```
[{source: "metis/file-0.txt", dest: "magma/file-0.txt"}, {source: "metis/file-1.txt", dest: "magma/file-1.txt"}]
```

Would pass validation in Metis (assuming all files exist), and then Metis would save into the database something like:

```
File: {id: 1, folder: "magma", name: "file-0.txt", data_block: 123}
File: {id: 2, folder: "magma", name: "file-1.txt", data_block: 456}
```

Let's say you come back and make a revision where the existing-order constraint is violated, like below:

```
[{path: "magma/file-1.txt"}, {path: "magma/file-0.txt"}]
```

Note that the paths are now, by necessity, pointing to files in the Magma bucket, because that is all we have reference to. This leads to a copy request like:

```
[{source: "magma/file-1.txt", dest: "magma/file-0.txt"}, {source: "magma/file-0.txt", dest: "magma/file-1.txt"}]
```

Metis then goes through and tries to update them sequentially. So it executes the first copy:

```
File: {id: 1, folder: "magma", name: "file-0.txt", data_block: 456} <- updated first
File: {id: 2, folder: "magma", name: "file-1.txt", data_block: 456} 
```

But now when it tries to execute the second copy command, it points the file `magma/file-1.txt` to the data block pointed to by `magma/file-0.txt`, which is now`456` instead of the desired / original `123`, because the reference to data block `123` in the Magma bucket has been lost. Something similar would happen if you interspersed new files, like:

```
[{path: "metis/new-file-3.txt"}, {path: "magma/file-0.txt"}]
```

In order to make sure the data blocks are preserved for all copy requests, you have to go in ascending index order to not overwrite the data block references.

We could make Metis do a single bulk update to the database, but there isn't a "Magma::Loader" equivalent for Metis right now.

Note that there are no controls in Timur to re-order items in a ListInput, so the issue would only appear with scripts talking directly to the Magma API.

coleshaw: I think the `::temp` requirement might be needed, though it may not strictly be necessary from a functional standpoint. I was thinking that you ideally want the indices to be sequential:

```
[{name: "model-record-attribute-0.txt"}, {name: "model-record-attribute-1.txt"}, {::temp}]
```

instead of 

```
[{name: "model-record-attribute-0.txt"}, {::temp}, {name: "model-record-attribute-2.txt"}]
```

Which might confuse people, who would be like "where is file -1?" (since ::temp doesn't get saved to the database).

But maybe that's not a big deal, especially since we're showing the original filename in Timur, not the generated name with the index.

magma/lib/magma/query/predicate/file_collection.rb
coleshaw: And it is different than `::url` because it includes all the data, like `original_filename`.

graft:
Nah, seems fine. :shipit:

PR #161, opened 2020-12-02
remove auth constraint on polyphemus configuration endpoint

coleshaw:
Small PR to set no-auth requirement on the polyphemus `/configuration` endpoint. Should be easier for users to configure the gem with this.

### DIFF
--- a/magma/spec/retrieve_spec.rb
+++ b/magma/spec/retrieve_spec.rb
@@ -333,7 +333,7 @@ describe RetrieveController do
       )
 
       header, *table = CSV.parse(last_response.body, col_sep: "\t")
-      expect(table).to eq([ [ "The Twelve Labors of Hercules", labors.map(&:identifier).join(', ') ] ])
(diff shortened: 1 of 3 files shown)


magma/spec/retrieve_spec.rb
```diff
@@ -333,7 +333,7 @@ def retrieve(post, user_type=:viewer)
       )
 
       header, *table = CSV.parse(last_response.body, col_sep: "\t")
-      expect(table).to eq([ [ "The Twelve Labors of Hercules", labors.map(&:identifier).join(', ') ] ])
+      expect(table).to match_array([ [ "The Twelve Labors of Hercules", labors.map(&:identifier).join(', ') ] ])
```
coleshaw, line 336: I see this spec failing intermittently because of the table order -- so changing the spec to use `match_array` instead of `eq` -- we don't expect the TSV data to come back in a specific order, do we?

PR #163, opened 2020-12-04
Fix link validation, refactoring csv handling, fixing add users in cr…

corps:
…eate project flow

Minor validation fix as per experience with Anton.
Fixes for the create project workflow (adding' admins will work now).
Refactor CSV into shareable class.  No major change in behavior (although actually improves some output incidentally).

This is a smaller piece of a few changes I have coming RE update attributes workflow and linking.


coleshaw (approved):
Seems like the Janus spec failure is because of the order of `Users` that are returned (unrelated to the PR)...should probably make that more robust. But otherwise looks good to me!

PR #164, opened 2020-12-07
add stderr messages to metis client

coleshaw:
This PR adds `stderr` messages to Metis Client. It is useful for folks who feed commands to the client instead of running it interactively, so they can log `stdout` and `stderr` messages to more easily detect failed uploads in a large sea of uploads, where it's impractical to parse through a log file manually.


coleshaw:
Going to merge this in so that we can gather more debugging info as folks continue uploading data...

PR #165, opened 2020-12-07
Metis logging by day

coleshaw:
This is a small PR that adds an `Etna::Logger` option to rotate logs by time period (daily, weekly, monthly), and also configures daily logging in Metis instead of size / # of files like the other applications. 

Note that the Ruby Logger doesn't actually rotate out logs when used with this configuration (delete ones older than X days), so I'll set that up with `logrotate` after this is merged and deployed. Any system software could be used to rotate the logs, but we seem to have `logrotate` already installed on the server.

### DIFF
--- a/metis/lib/models/data_block.rb
+++ b/metis/lib/models/data_block.rb
@@ -29,6 +29,29 @@ class Metis
           location
         )
       else
+        # We're running into an issue with Metis uploads where the upload
+        #   cannot be renamed when creating the data_block.
+        #   > Errno::EBUSY: Device or resource busy @ rb_file_s_rename -
+        #   > File "/var/www/metis/lib/models/data_block.rb", line 32, in rename
+        #   > File "/var/www/metis/lib/models/data_block.rb", line 19, in create_from
+        #   > File "/var/www/metis/lib/models/upload.rb", line 55, in finish!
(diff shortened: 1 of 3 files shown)


graft:
Can you clarify the logrotate requirement?

coleshaw:
It seems like the Ruby logger doesn't delete old files when you only specify frequency, so eventually log files would just eat up disk space. `logrotate` is already on the server, so it should be easy to set up and just keep say, the last 7 days of logs.

metis/lib/models/data_block.rb
```diff
@@ -29,6 +29,29 @@ def set_file_data(file_path, copy=false)
...
+        #   > # open for a short time after close() has returned. Files accounted as open
+        #   > # cannot be moved.
+        #   > # Default: false
+        # So it seems like files may be kept open for some amount
+        #   of time after the final blob is appended, which will
+        #   prevent it from being moved.
+        # Adding a delay (super hacky?!?) will hopefully give the
+        #   file system time to close the file. Hopefully this does not
+        #   significantly slow down uploads.
+        sleep(0.01)
```
coleshaw, line 54: @graft , throwing this change in here -- seems kind of silly, but maybe this the best solution since the file closed behavior may just be part of normal operation for beegfs...? I can't find any documentation on how long closing a file may take though, so kind of a shot in the dark.

PR #167, opened 2020-12-08
Graft/update root view

graft:
A small one to repair the root view, to resolve the confusion between short and long names, as noted yesterday (sorry for the dup @coleshaw). Also unifies the Metis and Timur home page components. Incidentally, working with etna-js in dev is such a breezy experience, the whole setup is really excellent work. The one snag I ran into was that there is as of yet no global stylesheet, so there is some style duplication/divergence here.


etna/.npm-packages/lib/node_modules/etna-js
```diff
@@ -0,0 +1 @@
+/etna/packages/etna-js
```
coleshaw, line 1: Hm, this file isn't tracked for me. Do we need to add it to the repo, or is it created by one of the docker build scripts?

etna/packages/etna-js/package.json
```diff
@@ -39,6 +39,7 @@
     "redux-mock-store": "^1.5.4",
     "redux-thunk": "^2.3.0",
     "regenerator-runtime": "^0.13.5",
+    "reselect": "^3.0.1",
```
coleshaw, line 42: In one of the weird quirks of how `etna-js` works, you'll need to add this same dependency to Metis and Timur's `package.json`. I get this error in Metis when I load the root view:

```
Uncaught Error: Cannot find module "reselect"
```

And blank page -- nothing loads.

coleshaw: Here is a comment thread about the issue...we've been bitten by it before:

https://github.com/mountetna/monoetna/pull/145#discussion_r524530334

coleshaw (changes requested):
Just need to update Timur and Metis `package.json` files with `reselect`, otherwise it looks good and works well for me.

etna/.npm-packages/lib/node_modules/etna-js
graft: Hmm, I didn't add this, so I assume some build process must be creating it.

corps: Huh.  I hadn't seen that before, to be honest.  I don't know how this got creatd.

PR #168, opened 2020-12-09
Adds `/update` functionality to magmaR

dtm2451:
Creates functions to power using magma's /update function to send data to magma from within R.

Functions added:
- `updateValues()` is essentially a simple wrapper for /update, mainly just differing in that the `revisions` input structure is nested, named, lists rather than nested JSON hashes.
- `updateMatrix()` a matrix-focused version which 1) takes in matrix data as a matrix, dataframe, or file location, 2) performs validation that matrix rownames are valid 'options' of the target attribute, 3) checks and reports how many column names represent new versus old record IDs, 4) asks the user before proceeding to... 5) convert the matrix to the needed nested list structure and make the /update curl request.


coleshaw:
Is it possible to do VCR captures of the update and updateMatrix requests, too, and add tests for those? I think that might be helpful moving forward. On the Magma side, you should be able to re-upload the same data and overwrite what is there.

dtm2451:
The update stuff should work with vcr too, just fine.  I just haven't written those unit tests yet!

And yea, my plan was to work off slight modifications of https://github.com/mountetna/monoetna/blob/ed4abd520e751c0411940a52036343437ab250c0/etna/packages/magmaR/R/update.R#L115 for unit tests.

dtm2451:
Unit tests are added now.
Ready for review!

## Ideas to do in future:

- ensure that `updateMatrix()`'s handling of attributes without 'options' matches with magma. (Currently, a warning message is output, but if magma would simply reject the /update, it might as well be an error which stops the processing altogether.)
- Build formatting and/or summaries of projected changes in to the `updateValues()` function.

coleshaw (approved):
Looks nice!

PR #169, opened 2020-12-10
Dictionary support

coleshaw:
This PR adds some dictionary support around the existing implementation. Namely:

* Adds dictionary support to the Etna Magma client, since I think that will be necessary for copying / creating dictionaries on models, as well as the REDCap converter workflows.
* Adds an `add_dictionary` action, for an existing model. This will be necessary to add dictionaries to existing models in existing projects.
* Allows the `add_model` action to include `dictionary` params. This can be useful for new models instead of adding in dictionaries after the fact...I think this will be useful for future projects (?).


magma/lib/magma/actions/add_model.rb
```diff
@@ -23,15 +23,21 @@ def has_parent?
...
 
-      project.load_model(
-          project_name: @project_name,
-          model_name: @action_params[:model_name]
-      )
+      # Here dictionary needs to be a JSON string, otherwise
+      #   Sequel will try to find a column for each dictionary key.
+      params[:dictionary] = JSON.generate(@action_params[:dictionary]) if @action_params[:dictionary]
+
+      Magma.instance.db[:models].insert(params)
```
graft, line 35: might set `params = { ... dictionary: @action_params[:dictionary] }` and then do like `Magma.instance.db[:models].insert(params.merge(dictionary: params[:dictionary].to_json))` to avoid the rewrite.

graft:
I have no idea how these tests are suddenly failing... doesn't look like you touched any of the relevant code. Some sort of wave/particle probability field effect.

coleshaw:
Oh, I know, it's proof the dictionary kind of works, I think. I've narrowed it down to the test that actually performs `AddDictionary` on the `Labors::Monster` model ... now I'm stuck trying to figure out why the teardown doesn't fix things cleanly.

PR #170, opened 2020-12-11
Simple file linking support

corps:
The shortest, sweetest, simplest route I could come up with for now.

More interest solution that decomposes identifiers of parents and can multimodel load is on my other branch for future work.

WIP Still testing and fixing, this is just for reference.


etna/lib/etna/clients/magma/workflows/update_attributes_from_csv_workflow.rb
```diff
@@ -53,7 +53,11 @@ def models
 
       class RowBase
         def stripped_value(attribute_value)
-          attribute_value ? attribute_value.strip : attribute_value
+          attribute_value = attribute_value&.strip
+          if attribute_value && @workflow.json_values
+            attribute_value = JSON.parse(attribute_value)
+          end
+          attribute_value
```
coleshaw, line 60: Maybe here (possibly elsewhere?), I think you'll need to detect a `file` or `file_collection` attribute from the model template and set the revision value to actually be:

```
{ path: <the input metis path from the file,
 original_filename: <the filename extracted from the metis path>
}
```
And an array of those objects for `file_collection`.

corps: right, so this will be handled by the fact that when generating the file linking csv, columns will all be JSON string values, and encoded in that way.

coleshaw: Ah, I see. Nice!

corps:
Tested this locally and it seems to work just fine.  Rawk

coleshaw:
I have two very stupid questions about how to use this (trying to test...):

1) When I try to call `create_file_linking_csv`, I just get a CSV with a header row, but no data? Is this command acceptable: `$ etna administrate models attributes create_file_linking_csv ipi TestBucket flojo_file txt --environment development` -- that should find all `.txt` files in the `TestBucket` bucket in my dev environment? My output CSV is:

```
$ cat ipi_flojo_file.csv 
identifier,flojo_file
```

So I feel like I'm doing something wrong... and there are files that I expect to be found by the default regex.

2) In lieu of generating a file linking CSV, I'm trying to write my own by hand. But I run into a quoting error -- how are the JSON strings for the `{path: <metis path>, original_filename: <filename>}` supposed to be formatted?

```
$ etna administrate models attributes update_from_csv ipi patient test_update_attrs.csv --json-values --environment development
<stacktrace>
/.rbenv/versions/2.5.7/lib/ruby/2.5.0/csv.rb:1911:in `block (2 levels) in shift': Illegal quoting in line 2. (CSV::MalformedCSVError)
```

For a file like

```
record_name,flojo_file
valid-record-name,{"path": "metis://ipi/TestBucket/docker.png", "original_filename": "docker.png"}
```

I've tried `"`, `""`, `\"` , but no luck. Again, feel like I'm missing something...?

corps:
> I have two very stupid questions about how to use this (trying to test...):
> 
>     1. When I try to call `create_file_linking_csv`, I just get a CSV with a header row, but no data? Is this command acceptable: `$ etna administrate models attributes create_file_linking_csv ipi TestBucket flojo_file txt --environment development` -- that should find all `.txt` files in the `TestBucket` bucket in my dev environment? My output CSV is:
> 
> 
> ```
> $ cat ipi_flojo_file.csv 
> identifier,flojo_file
> ```
> 
> So I feel like I'm doing something wrong... and there are files that I expect to be found by the default regex.

The default regex doesn't do any recursive searching which I can change.  I don't think there's anything wrong -- I just need to improve the default regex.  It definitely does search using the endpoint and it definitely does generate csvs.

> 
>     1. In lieu of generating a file linking CSV, I'm trying to write my own by hand. But I run into a quoting error -- how are the JSON strings for the `{path: <metis path>, original_filename: <filename>}` supposed to be formatted?
> 
> 
> ```
> $ etna administrate models attributes update_from_csv ipi patient test_update_attrs.csv --json-values --environment development
> <stacktrace>
> /.rbenv/versions/2.5.7/lib/ruby/2.5.0/csv.rb:1911:in `block (2 levels) in shift': Illegal quoting in line 2. (CSV::MalformedCSVError)
> ```
> 
> For a file like
> 
> ```
> record_name,flojo_file
> valid-record-name,{"path": "metis://ipi/TestBucket/docker.png", "original_filename": "docker.png"}
> ```
> 
> I've tried `"`, `""`, `\"` , but no luck. Again, feel like I'm missing something...?

Yeah, quoting JSON values in the csv sucks.  My bigger branch had a more complicated solution to this, but for here, what you need is

`"{""a"": ""b""}"`

coleshaw:
> Yeah, quoting JSON values in the csv sucks. My bigger branch had a more complicated solution to this, but for here, what you need is

> "{""a"": ""b""}"

Oh duh didn't try that. Thanks!

Yeah, I was originally thinking just take the Metis path in the CSV cell, that way you can construct the JSON blob in the code without worrying about writing out a JSON string in a CSV...

coleshaw (approved):
File linking works via update_from_csv!

corps:
Just as a note @coleshaw I am slipping in the /dev/stdout fix for windows in here as well.

coleshaw (approved):

PR #171, opened 2020-12-11
Add vignettes to magmaR

dtm2451:
In the `vignettes/` folder
One for installation/explanation/exploration/download - Download.Rmd
Separate one for the more rare use case of data upload - Upload.Rmd

Upload.Rmd is not finished yet, but the parts that are in place (right now, that's everything up until the last 20 or so lines of Upload.Rmd, and all of Download.Rmd) are final enough to be worth review.


dtm2451:
Okay, I think I'll probably make minor updates endlessly with every read through... but it's all in a good state now!  Certainly ready for review.

I added stubbing too, so the vignettes won't cause issues if we use `R CMD Check` when with work magmaR into CI =)

graft:
In general this looks great (although I could only use your build to review). A few things I noticed:

- In "Uploading" you mention mountetna.github.io to avoid duplication, but in "Downloading" you do not and describe magma at length.
- You don't describe retrieveAttributes or how to look at templates in general in these vignettes, dunno if that is intentional and this is described elsewhere

dtm2451:
I've changed how I refer to the Mount Etna docs in the Download vignette now to say 'we'll explain some here, but know that finer detail / the main documentation is here'.

I added an extra reference to the `retrieveAttributes()` function within the `retrieve()` section (it was already given in the 'Helper functions' section). Do you think more is needed?  The function mainly gives the names of attributes and I think what's in the Helper function section does make that clear?  Note that each function also has is own documentation, accessible via `?function_name`, so it's common to give overviews in R vignettes that don't dive in fully.

For templates, I give a little bit more there as well.  I added some exploration of how to explore the output in a fairly condensed way using the `str()` function.  But lemme know if you we thinking instead that more text is needed about what a templates actually holds?

dtm2451:
Next step, I'll put together a README.md about how to set up a dev environment & knit the vignettes / run the tests without needing RStudio.  Then you should be able to render the vignettes 'live'.

dtm2451:
@graft Hopefully the README.md file allows you to build the vignettes to check / edit-check cycle these vignettes for real!

While writing that file, I remembered we had discussed shoving the rendered vignettes somewhere that's accessible from https://mountetna.github.io/. That's definitely needed during the current stage where vignettes are not included via the development-style install instructions. But this PR is pretty old now so maybe we can push first and decide where / how to automate that after?

graft (approved):

etna/packages/magmaR/README.md
```diff
@@ -0,0 +1,109 @@
...
+    R CMD INSTALL --no-multiarch --with-keep.source magmaR
+    ```
+
+- **To "knit" & view the vignettes:**
+  1. Run `usethis::edit_r_environ(scope = "project")` (which is included in the 
+`clear_stubs_for_true_test.R` file), and set it as:
+```
+TOKEN="YOUR-staging-token"
+URL="https://magma-stage.ucsf.edu"
+```
```
graft, line 68: Hmm, why does this run against staging?

PR #173, opened 2020-12-14
Update timur search

graft:
This PR latterly adds the "show_disconnected" filter to the Search page on Timur, but also updates ReactTable to v7, which is a "headless" table which uses hooks, and makes various style cleanups to the search page.

Regarding the "show_disconnected" checkbox I added it to the 'filters' modal if you are a project editor. Perhaps not the best place to hide such a thing, but I think it should be relatively obscure.


coleshaw:
The styling looks nice, and the search bar / search page work well for me with different roles. :+1:

coleshaw:
Sorry, found a bug (?) -- Download TSV in the Search page doesn't seem to honor the "Show only disconnected records" checkbox. If I run a search with the box checked, the results on the page look correct (only disconnected records). However, the TSV still downloads with the connected records.

PR #174, opened 2020-12-15
Additional Metis Client commands

coleshaw:
Based off of the recent uploading, this PR adds in some additional commands to the Metis Client. All of these work with commands fed in from a text file as well:

1) `cp` of files. You should be able to use relative paths (i.e. `cp file.txt ../` or `cp ../file.txt .` as well as just copy within a directory. To copy to / from a different bucket, you need to use a Metis path for the appropriate file, in the form of `metis://<project>/<bucket>/<file_path>`
2) `rm` of files and folders. I think this would be useful for startup. I'm hesitant to implement `rm -r`, so people still have to think a little bit and removing a whole hierarchy is still a bit onerous, but at least there is now a way to do it in the client instead of going to the UI, and it could be scripted to make it easier. If we don't think `rm -rf` is too dangerous, we can add that later...
3) `mkdir` (but really `mkdir -p`) for folders. This seems like it will make Arjun's life easier, since the folder hierarchy on the storage server isn't easy to match to the desired target on Metis sometimes.

### DIFF
--- a/etna/spec/clients/magma/workflows/update_attributes_from_csv_workflow_spec.rb
+++ b/etna/spec/clients/magma/workflows/update_attributes_from_csv_workflow_spec.rb
@@ -330,7 +330,7 @@ describe Etna::Clients::Magma::UpdateAttributesFromCsvWorkflowSingleModel do
         project_name: PROJECT,
         filepath: './spec/fixtures/magma/nonexistent_input.csv'
       )
-    }.to raise_error(StandardError, 'Single Model invokation must include keyword :model_name.')
+    }.to raise_error(RuntimeError, 'Single Model invocation must include keyword :model_name.')
 
     expect(WebMock).not_to have_requested(:post, /#{MAGMA_HOST}\/update/)
   end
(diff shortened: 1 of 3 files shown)


graft:
If i rm a file that does not exist, there is no warning or error, it simply does nothing.

Also maybe this is a general issue with error handling in the client but when there is an error (like cp'ing a non-existent file) it shows the same error twice, then prints a stack trace, e.g.:
```
ipi:test> cp gene_exp.tsvf test2.tsv
That source file does not exist test/gene_exp.tsvf
That source file does not exist test/gene_exp.tsvf
bin/metis_client:618:in `run'
bin/metis_client:1085:in `run_command'
bin/metis_client:1035:in `run'
bin/metis_client:1185:in `start'
bin/metis_client:1189:in `<main>'
```

graft:
Also just nothing: I don't think any of the commands have real globbing support, so probably can't do it right now, but it would be nice to be able to glob with rm.

coleshaw:
Thanks, I've added in an error message for `rm` and also removed the redundancy. That redundant printing of the error message was because I added `stderr` support to try and help debug the intermittent upload error with Arjun, didn't realize it would print `stderr` into the interactive console, too. 

I wonder if globbing could be achieved with the server-side `find`. Will think about it to include in a future PR.

graft (approved):

PR #175, opened 2020-12-16
Fixing tricky exit bug in etna specs

corps:
Womp womp, big dumb mistake on my part.

This spec had an incidental `exit` when TOKEN wasn't setup by default, which causes the  result of the rspec always completing with status 0.


coleshaw (approved):
I assume the failing test is due to the merge with master overwriting the cassette ... nice find, that seems pretty subtle!

PR #176, opened 2020-12-16
Table upload command for etna gem

coleshaw:
Don't know if this would be useful in the `etna` gem, but here is the command I used to upload table data for COMET and XNEU1. I didn't put this out earlier because of the effort to re-do the entire upload-attribute-by-csv workflow, but in case this is useful in the interim, wanted to make it available.


corps:
Nice!   I like.

Yeah, figuring out the exact set of tooling for data entry is still something on our plate.  My last attempt to release a better version was a reminder to me that specification is still flux.  Something like this can be helpful until we have a greater handle.

Nit -- should we also bump gem version so it gets published, or just defer for now?

corps (approved):

coleshaw:
Yes, thanks for the reminder, I'll bump the gem version!

coleshaw:
@corps , silly question, is there some trick to getting the etna dependencies lined up correctly when bumping the version? Do I need to push a `ref/artifacts` branch manually for the given commit?

corps:
Nope, just run `./bin/update-package-references`

coleshaw:
Great, thanks!

PR #177, opened 2020-12-17
Single project record

graft:
This PR adds in the missing validation we noted this week, where several root project records can be created. This should keep it down to a single entry, but allow creation of an entry if no project record yet exists. The single project record's identifier can still be updated as usual via `/update`.


corps:
Reviewing!

magma/lib/magma/loader/record_entry.rb
```diff
@@ -135,7 +135,7 @@ def record_exists?
 
     def check_document_validity
       @complaints = []
-      @loader.validator.validate(@model, @record) do |complaint|
+      @loader.validator.validate(@model, @record_name, @record) do |complaint|
```
corps, line 138: This seems in reverse with `validate` below.

Is the order model, document, record_name or model, record_name, document?  Seems contradicting.

magma/lib/magma/validation/attribute.rb
```diff
@@ -6,7 +6,7 @@ def initialize(model, validator)
         @attribute_validations = {}
       end
 
-      def validate(document)
+      def validate(record_name, document)
```
corps, line 9: This also seems to contradict the above order.  Is the inner model validation's param ordering document, record_name or the other way around?

Maybe consider switching these to named args: `def validate(record_name:, document:)`

magma/lib/magma/validation/attribute.rb
```diff
@@ -58,16 +58,24 @@ def validation_object
         private
 
         def format_error(value)
-          if @attribute.format_hint
-            "On #{@attribute.name}, '#{value}' should be like '#{@attribute.format_hint}'."
+          if link_attribute.format_hint
```
corps, line 61: Is this intended change?

corps (approved):
Only one major concern around argument ordering.

magma/lib/magma/validation/attribute.rb
graft: I think this is correct? The error comes from link_validate, which actually just runs the validation for the link model identifier. If there is a format hint for this validation, it is likely on that `link_attribute`, and not on our `@attribute` (which is probably like a `parent` or something).

PR #179, opened 2020-12-17
Metis client can ls file directly

coleshaw:
Per mountetna/metis#97, this add `ls` for files in metis_client. If a file exists, `ls` will return the full path (including bucket) to the file. Otherwise it throws an error as before.


graft:
Seems great, minor nit, there are two different errors saying "Could not list folder" or "invalid folder" if there is no resource found - it might instead say "No such file or folder"?

PR #180, opened 2020-12-18
Timur Chrome CSS

coleshaw:
This small change fixes the CSS for collection attributes in Chrome when there is only one item. Instead of the one item being centered in the div, it will be pulled to the left so it lines up with other elements. This mimics the behavior seen in FireFox.


graft:
hmm a bit odd that the first item is always inline-block even if there are many items, but it seems to work ok... :+1:

PR #182, opened 2020-12-23
REDCap ETL

coleshaw:
This PR moves over @graft 's REDCap loading code into the Polyphemus ETL framework. This PR does several things:

1) Translates the current REDCap code into a pseudo-ETL type process. Each project is its own Ruby file, stored in a private repository mountetna/redcap-projects.
2) Adds a simple job API to polyphemus to trigger the ETL loading process. The REDCap job currently streams data back to the client with job updates (should change in the future once we start scheduling jobs).
3) Adds a polyphemus command to run the data loading manually or from the command-line.
4) Adds some high-level tests around REDCap loading...something to build off of.
5) `etna` gem command to submit a REDCap loading job to the polyphemus job API.
6) Clones the private REDCap projects repository as part of the monoetna docker build process.

Using the gem, a user would run a command like:

```
$ export TOKEN=<janus token>
$ etna administrate models load_from_redcap __project_name__ __redcap_tokens__ --models __model_names__
NOTE: This is a **preview** of what the data loading will look like. Use the --commit flag to load records into Magma.
Attempting to load model X.
Processing form Y.
blah blah blah
```


corps:
This is, amazing work dude.  I'm thoroughly impressed, and I'm excited to take a deeper look next year!
This is a huge step forward for a major piece of our data loading process.

corps:
@coleshaw 
Depending on your timeline, I could probably assist with the streaming component of this work by getting a redis instance deployed for us.  After docker is up and running, installing / testing / deploying a redis server would be trivial.  There are many great ruby libraries that can turn redis into a websocket pub/sub service very easily.

polyphemus/lib/etls/redcap/redcap_etl_script_runner.rb
```diff
@@ -0,0 +1,83 @@
...
+      @redcap_tokens = redcap_tokens
+
+      raise "REDCap host must use https://" if redcap_host.start_with?("http://")
+      raise "Magma host must use https://" if magma_host.start_with?("http://")
+
+      @redcap_host = redcap_host
+      @magma_host = magma_host
+      @dateshift_salt = dateshift_salt
+
+      warn "Warning: no dateshift_salt provided, will use a randomly generated one." unless dateshift_salt
```
corps, line 36: Is this something that we might want to fail with error on instead, so that users aren't surprised by dates if something isn't configured right?  Just not sure either way.

polyphemus/lib/polyphemus/controllers/job_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+    $stdout = out
+    yield
+    return out
+  ensure
+    $stdout = STDOUT
+  end
+
+end
+
+class JobController < Polyphemus::Controller
```
corps, line 27: Nice, I like this.  With this approach, it would mean we can put the run redcap command in etna instead of polyphemus due to the lessened gem requirements?

polyphemus/spec/etls/redcap/redcap_etl_script_runner_spec.rb
```diff
@@ -0,0 +1,168 @@
+describe Polyphemus::RedcapEtlScriptRunner do
```
corps, line 1: Nice tests!

polyphemus/lib/etls/redcap/redcap_etl_script_runner.rb
coleshaw: That's a good point, I'm not sure what the expected behavior is during analysis (since the date differences should still be the same with different salts). However, it does seem confusing if dates changed between loading events (due to a salt changing), so throwing an error makes sense. Thanks!

coleshaw:
> @coleshaw
> Depending on your timeline, I could probably assist with the streaming component of this work by getting a redis instance deployed for us. After docker is up and running, installing / testing / deploying a redis server would be trivial. There are many great ruby libraries that can turn redis into a websocket pub/sub service very easily.

Thanks, I think Redis will be the right way to go, especially once we have a more formal job scheduler behind the Polyphemus controller. For now I'm not blocked on it -- I have a working solution that streams the process output back through Puma / Rack (not super confident in how robust it is or how it will handle errors, but I can see the output coming back live via curl). So I can work on the rest of the items for now.

polyphemus/lib/etls/redcap/redcap_etl_script_runner.rb
graft: Yeah a consistent salt is vital. I would eventually like to do this in magma's date_time attribute via a loading hook, to avoid this issue. Really this is a valuable secret guarding PHI and should be tightly controlled. In the meanwhile ensuring we all use the same salt is required so the dates are shifted consistently (otherwise they are useless).

coleshaw:
@graft @corps , I've cleaned this up and abstracted out the Jobs API on Polyphemus so it's not super specific to just REDCap loading (still will need polishing as we define more job types). You should be able to re-build your docker containers and get the actual project config files locally.

I'm also happy to pair and walk through this for a review, since this turned out to be a large PR.

etna/lib/etna/clients/polyphemus/client.rb
```diff
@@ -26,6 +26,22 @@ def configuration(configuration_request = ConfigurationRequest.new)
...
+        #
+        # polyphemus_client.job(request) do |response|
+        #   response.read_body do |fragment|
+        #     <fragment contains a chunk of text streamed back from the server>
+        #   end
+        # end
+        @etna_client.post(
+          "/#{job_request.project_name}/job",
+          job_request) do |res|
+            yield res
```
corps, line 42: In this case the yield is redundant with

```
        @etna_client.post(
          "/#{job_request.project_name}/job",
          job_request, &block)
```

But not worth waiting on a whole set of tests to change :P

polyphemus/Makefile
```diff
@@ -1,5 +1,15 @@
 include ../make-base/stubs.mk
 
+.projects-mark:
+	git clone git@github.com:mountetna/redcap-projects.git lib/etls/redcap/projects || CLONE_FAILED=1
+	if [ -z "$(CLONE_FAILED)" ]; then \
+		echo "Could not clone REDCap projects -- do you have the right permissions?"; \
+	fi
+	@ touch .projects-mark
```
corps, line 8: Do we want to touch the mark on failure?  We should maybe put this into an else clause.

etna/lib/etna/clients/polyphemus/client.rb
coleshaw: I didn't realize that, thanks for pointing that out!

polyphemus/lib/etls/redcap/redcap_etl_script_runner.rb
```diff
@@ -0,0 +1,98 @@
...
+      }
+    end
+
+    def get_binding
+      binding
+    end
+
+    protected
+
+    def define_model(model_name, &block)
```
corps, line 84: Since this method is used exclusively for testing, does it make sense to give it a name that makes that clear?  `define_dummy_test_model` ?

corps (approved):
I'm generally ok with this, it's pretty straightforward to read and understand even if it's big.  I had some minor nits but nothing that should block a merge.

polyphemus/lib/etls/redcap/redcap_etl_script_runner.rb
coleshaw: It's not only for testing, the project config files use it too:

https://github.com/mountetna/redcap-projects/blob/master/mvir1.rb

You have to define a new class for each Magma model, so it's project specific.

PR #183, opened 2020-12-27
Unify navbar

graft:
This PR provides a "unified" navbar, at least between Timur and Metis, via etna-js. Janus remains an ugly outlier, but it is too difficult to replicate all of this work in ERB; I think we should soon enough port the Janus UI to react to allow easier maintenance and code-sharing.

There are also some new icons here for Timur and Janus. I'm not really happy with the Janus one and would like to try another variation, possibly an ornate key.

The new nav provides the logo behavior (goto root - the apps are responsible for drawing their logos correctly) along with user status and role. Apps can put whatever they want in the middle space. 

I have also included a new quick navigation bar. The styling for this is so-so, but the functionality is pretty nice. Unfortunately navigating to Janus is an ugly dead-end, but you can go back and forth between timur and metis easily this way.


janus/lib/server.rb
```diff
@@ -42,6 +40,8 @@ class Server < Etna::Server
 
     post '/update_key', action: 'user#update_key'
 
+    get '/:project_name', action: 'admin#project', auth: { user: { can_edit?: :project_name } }
+
```
graft, line 44: This (et al) provides consistent routing across apps for getting to a project's main page.

timur/lib/client/jsx/components/timur_nav.jsx
```diff
@@ -2,120 +2,64 @@
...
-      <div
-        id={logo_id}
-        className={Object.keys(exchanges).length > 0 ? 'throb' : null}
-      >
-        <div className="image"></div>
-        <Halo radius={27.5} />
-      </div>
-    </a>
+const Logo = connect(({exchanges}) => ({exchanges}))(({exchanges}) =>
+  <div id='timur-logo'
```
graft, line 29: I ended up using `connect` here because `let exchanges = useReduxState(state=>state.exchanges)` did not behave equivalently for reasons I don't quite understand.

timur/lib/server/controllers/timur_controller.rb
```diff
@@ -17,7 +17,8 @@ def config_json
       {
         project_name: @params[:project_name],
         token_name: Timur.instance.config(:token_name),
-        magma_host: Timur.instance.config(:magma)[:host],
+        magma_host: Timur.instance.config(:magma)&.dig(:host),
+        metis_host: Timur.instance.config(:metis)&.dig(:host),
         metis_uid_name: Timur.instance.config(:metis_uid_name),
         janus_host: Timur.instance.config(:auth_redirect)
```
coleshaw, line 23: Since you're adding a `:janus:` `:host:` explicitly to the `config.yml` file, does it make sense to use that here instead of `auth_redirect`?

etna/packages/etna-js/components/Nav.jsx
```diff
@@ -0,0 +1,75 @@
...
+
+  if (!shown) return <div className='etna-links'>
+    <div className='etna-links-show' onClick={ () => setShown(true) } >
+      <Icon icon='chevron-circle-left'/>
+    </div>
+  </div>;
+
+  return <div className='etna-links'>
+    { apps.map( app => <Link key={app} app={app}/>) }
+    <div className='etna-links-hide' onClick={ () => setShown(false) } >
```
coleshaw, line 61: Without your demo today, it's kind of inscrutable what this new arrow icon symbolizes in the Nav...the UX could be more intuitive. Would it help to include some hover text like `Other Apps` to indicate that it is an actionable button, or that it is hiding some information? Or, an alternative might be a more standardized menu icon, like the hamburger or waffle icon.

coleshaw (approved):

etna/packages/etna-js/components/Nav.jsx
graft: I tried a hamburger there at first (there are no good waffle icons), but I wasn't sure it meant the right thing (the hamburger is usually a config menu). It also didn't quite have the feel of a drawer since the burger is so static. I put it back to burgerform, lemme know what you think.

etna/packages/etna-js/components/Nav.jsx
```diff
@@ -0,0 +1,75 @@
...
+      <Icon icon='bars'/>
+    </div>
+  </div>;
+}
+
+const Nav = ({logo, app, children, user}) => 
+  <div className='etna-nav'>
```
coleshaw, line 68: super-nit: should we use more semantic HTML here? That would involve changing this to a `nav` element instead of the generic `div`.

etna/packages/etna-js/components/Nav.jsx
coleshaw: You're right, the hamburger looks funky because the menu slides out to the left, not down like it usually does. If it slid down instead, it would function more like the StackOverflow / StackExchange dropdown, where they show sister sites in a dropdown, or like the Office365 Waffle that drops down to show apps. Perhaps clicking your username drops down to reveal the other app icons?

I was thinking though, it's probably not a big deal to keep the arrow, too, since we have a small user base and it would be fairly simple to explain or point out, too.

etna/packages/etna-js/components/Nav.css
```diff
@@ -0,0 +1,88 @@
+.etna-nav {
```
corps, line 1: This isn't really a nit but a cool thing to note: the way the css loader works is that it actually name mangles css classes to prevent collisions.  When you do the import with require, it returns a hash that maps the local class names to their true class names for purpose of controlling them in React.

A side effect is that you can totes use generic and less specialized names without fear of collision.  `nav` or even `component` is sufficient.  But you don't need to go through the effort of changing it, just noting.

corps: https://webpack.js.org/loaders/css-loader/

corps: Specifically, check the scoping features under Features

timur/lib/client/jsx/components/timur_nav.jsx
corps: It may be behavior difference for arrays vs objects in change tracking, unfortunately.  I'd have to take a closer look.

timur/lib/client/jsx/components/timur_nav.jsx
```diff
@@ -2,120 +2,64 @@
...
-        {mode !== 'home' && <IdentifierSearch />}
-        <div id="nav">{mode !== 'home' && this.renderTabs()}</div>
-        <Login user={user} />
-      </div>
-    );
-  }
-}
+const TimurNav = ({mode, user}) =>
+  <Nav user={user} logo={Logo} app='timur'>
+    {mode !== 'home' && <IdentifierSearch />}
```
corps, line 59: Just more like an FYI, but you can use `<React.Fragment>` here to compose siblings:

```
{mode !== 'home' && <React.Fragment>
   <IdentifierSearch />
   <ModeBar mode={mode}/>
</React.Fragment>}
```

This is another one of those super minor things that isn't a big deal either way, but just interesting to share as it can be more useful than a continuous chain of conditions in some cases.

corps (approved):

etna/packages/etna-js/components/Nav.jsx
graft: I tried another iteration, moving the burger to the left where a menu would be. I isolated the menu icon so it is less buttony in appearance to avoid confusion. The burger now opens a drawer downwards, which is not great since it covers up some important UI elements. However, it does seem like we'll want to have some settings stuff in a menu here at some point, so I think it seems worth putting in a drawer here.

timur/lib/client/jsx/components/timur_nav.jsx
graft: I've seen React.Fragment in places but haven't really used it yet. Thanks for the tip!

etna/packages/etna-js/components/Nav.jsx
coleshaw: That looks nice! It does cover up some important elements, especially on Metis. If we get feedback that it is an issue (or especially if we add things to the menu and it starts covering up more items), we can always make other UI tweaks.

PR #185, opened 2021-01-07
Implements workflow for materialize magma metadata and associated files.

corps:
Unfortunately, after trying many paths, I couldn't come up with a good sustainable test.

1.  Mocking is difficult and not deep enough a test, because most of the logic is actually the integration and cooperation of magma and metis. 
2.  Using an available data set like ipi works, almost, except that the resulting cassettes are too big.
3.  Trying to use subsets of datas works, but it is far less useful a test overall (want to test things like filtering, attribute masks, circular graphs, etc).

That said, I have manually tested this several times and feel comfortable.  I think some work on a finding a better testing strategy for "heavy integration" type pieces is something to consider (cassettes may not be enough, we may need a way to run services together in CI.  Not impossible but also a PITA).


corps:
Separate from but relevant to this work is a Asepra and Metis client implementation of the `Filesystem` abstraction that would enable long term metis -> aspera piping with no intermediate materialization.

etna/lib/etna/clients/magma/workflows/walk_model_tree_workflow.rb
```diff
@@ -0,0 +1,101 @@
+require 'ostruct'
+require 'digest'
+require 'fileutils'
+require 'tempfile'
+
+module Etna
+  module Clients
+    class Magma
+      class WalkModelTreeWorkflow < Struct.new(:magma_crud, :logger, keyword_init: true)
```
corps, line 9: I suspect this workflow will be useful for other purposes, too.  It's a general magma walker that could be helped to also do, say, magma table joins

polyphemus/lib/commands.rb
```diff
@@ -504,6 +504,225 @@ def execute(env, project_name, file_path, execute: false)
...
+  class PullGNESharedData < Etna::Command
+    include WithEtnaClients
+    include WithLogger
+
+    string_flags << '--out-dir'
+    boolean_flags << '--stub-files'
+
+    MATERIALIZE_ATTRIBUTES = {
+        'patient' => [
+            'created_at',
```
coleshaw, line 516: super-nit: does GNE need the `created_at` and `updated_at` fields for all these models? Those aren't the source timestamps (i.e. REDCap), but rather Magma timestamps.

coleshaw (approved):

polyphemus/lib/commands.rb
corps: Not 100% sure, but this list came from Gabi herself.   We can't currently send the REDCap values in any case since these values are updated automatically as part of magma logic.

At some point I imagine us formalizing the idea that magma is not a source of truth, but a composition of sources of truth, in which case values like these could be pass through to their source and not owned by magma itself.

For now, any value which atleast reflects for the purpose of GNE wether data is 'new' or 'updated' seems useful.

coleshaw: > For now, any value which atleast reflects for the purpose of GNE wether data is 'new' or 'updated' seems useful.
:+1:

PR #188, opened 2021-01-10
Zc/aspera fs

corps:
1.  Implements generic aspera based filesystem.
2.  Shifts some R related stuff into etna-base (the shared layer for all apps).  This actually simplifies deployment; layers are cached and shared anyways, so it doesn't result in a net increase of time or space, but it does allow that shared tooling to exist on all apps in case we need R to run in multiple places in the future.  It also simplifies etna builds, which were not working correctly (updates to etna/Dockerfile are picked up in CI but not in development thanks to the way docker-compose handles building).
3. Tests!  For aspera upload / download.
4. Generic pipe / streaming based interface makes integration with SCP and metis file systems possible in future PR.

This is just one more small step towards some changes coming to assist Bushra / Arjun in doing metis -> aspera uploads.

Another step that (may) need to happen is to possible build onto @coleshaw 's work with jobs to enable upload jobs being scheduled (due to ip whitelisting and speed of streaming from metis being much faster within the VPC).  Will consider next steps.  This PR atleast makes it possible to stream from polyphemus  without using it's limited disk space.


polyphemus/lib/commands.rb
```diff
@@ -689,44 +688,30 @@ class PullGNESharedData < Etna::Command
...
-            size[0] += chunk.length
-          end
-
-          io
-        end
-      end
+      ascli_bin = `which ascli`.chomp
+      ascp_bin = `ascli config ascp`.chomp
+      @filesystem ||= Etna::Filesystem::AsperaCliFilesystem.new(ascli_bin: ascli_bin, ascp_bin: ascp_bin,
+          host: "aspera01p-pxy.gene.com", username: "collinsz99", key_file: "/home/runner/.ssh/id_rsa_gne")
```
graft, line 711: Maybe this stuff should be in a config file?

graft (approved):

coleshaw (approved):

polyphemus/lib/commands.rb
corps: Done, moved this into a config.

PR #189, opened 2021-01-11
WIP: Archimedes

graft:
This PR adds a new Archimedes service, intended to replace the Archimedes endpoint currently attached to Timur. This is a work in progress, but this midpoint seems a good place to pause and solicit wise council.

The new service is written in Python. The service interprets scripts ("manifests") written in a service language (I suggest the name "Vulcan"), and returning a JSON payload of data ("consignment"). This service is currently made use of by Timur, mostly in constructing plots. The principal motivation behind this port is to remove several of the interchanges in calculation required by running the service in Ruby, and to bring the bulk of the data work that the service performs directly into Python. This allows us to take advantage of the considerably greater analytical tooling available in Python compared to Ruby, without requiring expensive serialization + deserialization to push calculations from ruby to an external python service. (N.B., this also obviates the need for a separate Pythia service). We may also rest our main data interchange formats (Matrix and Vector) on the base of Pandas, which defines DataFrame and Series types that, mercifully, are close parallels to the Matrix and Vector classes I wrote in Ruby.

The service is still incomplete, a few remaining pieces need to be put in place to bring it in line with the existing service, which should not be too much work:
1) Most of the methods from the Ruby service have been ported over, with the exception of the magma query methods. These require at least a rudimentary python Magma client. In addition, there is no working bridge to Rtemis. It is my hope to replace this with RPy where necessary (e.g. for the differential expression package voom - Bushra actually wrote an RPy implementation of voom long ago which we may resurrect), but mostly to replace R methods with comparable python methods.
2) Some control flow productions are missing (if/else, map and concat operators).
3) A ruby service providing Etna::Auth using Rack::Proxy must stand in front of the python service.

Future work that I can imagine:
1) Caching the output of calculations for subsequent requests.
2) Queuing scripts for execution by workers rather than running them within the request cycle.

Regarding syntax changes in this new service: I have tried to maintain continuity with the previous syntax, making use of the old test cases as much as possible.  Beyond this I have also attempted to fill out the description in vulcan.md, which adds some new syntax to the language, most notably functions. I have not yet enabled the previous "macro" syntax, and my inclination is to leave it out in favor of the new function syntax. This will cause many existing scripts to break and require editing.

It is tempting to imagine, as I did many times when rewriting this service, that we could melt away the service language and replace it with python scripts. According to my current reasoning, I think we will always find reason to retain a service language. However, in any case, the course of action undertaken here is the least divergent from the current state of affairs.

There are many classes of comment that might be made here (e.g. questioning the basic motivation, poking at the syntax choices), all of which are welcome.


coleshaw:
Since there is a need for a (simple) python client anyways, would it be worth it to make that a standalone `pip` package? That way others could use it independent of Archimedes to at least query Magma, much like magmaR.

archimedes/function.py
```diff
@@ -0,0 +1,42 @@
...
+from scope import Scope
+
+class Function:
+    def __init__(self, parser, arg_defs, production):
+        self.arg_defs = arg_defs
+        self.production = production
+        self.parent_scope = parser.current_scope
+        self.parser = parser
+
+    def getargs(self, args):
```
corps, line 11: Hmm.  Is this necessary?  I feel like we can use python itself to do the heavy lifting of expanding args, and buy ourselves more flexibility, too.

What if calls are formatted like this:
```
{
  args: ["a", "b"],
  kwargs: {"c": "d"}
}
```

In that case, we can simply "invoke" a function `set_args(*payload['args'] or [], **parload['kwargs'] or {})` or even just `call(...)` and allow us to specify arguments using natural python syntax and gain it's full calling convention.  

If we want to convert errors to archimedes errors, we can also broadly catch the `TypeError` and convert that?

But just minor suggestion.

archimedes/function_library.py
```diff
@@ -0,0 +1,26 @@
+from pandas import Series, DataFrame
+import functions_vector
+import functions_matrix
+
+def add_to_library(module):
+    namespace = vars(module)
+    public = (name for name in namespace if name[:1] != "_")
+    for name in getattr(module, "__all__", public):
+        setattr(Library, name, namespace[name])
```
corps, line 9: Should we check for collisions?

corps (commented):
I have my broad doubts that we're going to find a manifest language that is useful enough to justify requiring a non engineer to learn it as opposed to text fields and selectors, especially given the lack of documentation, local REPL, and other factors.  But if you believe this is the right way forward, we can go ahead.

corps:
Broadly, I'm not sure how archimedes fits into our "pipeline" process.

Who calls archimedes?
How does archimedes handle binary inputs and outputs?
How does archimedes integrate with long running asynchronous processes?  Does archimedes have a mechanism to stream status on in progress work?
How are failures from archimedes reported to a user?

graft:
> Broadly, I'm not sure how archimedes fits into our "pipeline" process.
> 
> Who calls archimedes?

Currently archimedes is called by plots on Timur (e.g. on the ipi project page) to assemble the data payload the plot requires. This (feeding calculation data into plots) will probably remain the main use case for archimedes, with more and more elaboration of the mechanism for determining what is being plotted.

> How does archimedes handle binary inputs and outputs?
It does not, yet. This is an open problem mostly, I think, about how to mount data into whatever machine is running archimedes from metis. However I think if we trade only in metis file paths for both input and output we can provide some tools for Archimedes to handle the file i/o to metis under the hood.

> How does archimedes integrate with long running asynchronous processes? Does archimedes have a mechanism to stream status on in progress work?

Currently it expects everything to complete in a request cycle. This will become impossible if we have tasks with 1 hour or more latency, which I think we are interested in supporting. It also contains none of the indispensible caching that I think it requires.

In the long run, I think what we will need is something like this:
1) A script is called with certain input arguments and a timestamp
2) If a cached result exists with those inputs (+ some kind of magma cache value, et al), a reference to the cached result is returned.
3) Otherwise a ticket in a queue is returned.
4) The script is calculated; the result is cached.
5) Subsequent queries return the reference to the cached result.
6) The cached result may be queried for certain portions of it (or all of it).

> How are failures from archimedes reported to a user?

Archimedes returns some JSON payload of errors. Currently they would appear to the user through Timur's janky showMessages() interface. Much work is required in this regard.

archimedes/encoder.py
```diff
@@ -0,0 +1,28 @@
...
+            return {
+                'matrix': {
+                    'row_names': obj.index.tolist(),
+                    'col_names': [col_name for col_name in obj.columns],
+                    'rows': obj.values.tolist()
+                }
+            }
+
+        if isinstance(obj, int64):
+            return obj.item()
```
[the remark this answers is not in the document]
graft: Eventually we should eschew JSON and use a better binary intermediate as output; in the meantime that much precision is not required, this is biology not physics :P

graft:
Closing this as obsolete.

PR #190, opened 2021-01-11
Etna clients require active token

coleshaw:
This PR adds some logic to the Etna clients (and thus the gem + polyphemus) that checks a user's token for expiration. If it's expired, it'll raise an exception and ask the user to set a new token. So hopefully that failure mode becomes less mysterious for our users. There is also a method to check for pending expiration, if we want to do some sort of refresh in the future.

Some *meh* things that shook out from this:

* There is some more duplicate behavior in Metis client. So at some point we should move it into the gem and use the same Etna::Clients as the other client apps.
* I had to update a bunch of tests and test tokens, because now we can't just use simple strings for tokens. So the test token is in several places.
* I updated `Etna::TestAuth` to expect the user permissions to be in a more JWT-like format, `<filler>.<user hash>`. This is to accommodate tests for Metis client when checking for expiration, and makes TestAuth behave a little more like regular Auth.
* Removed the old Magma client / gem, in favor of the shared gem in `etna`.


corps (approved):
Nice!  This is a super good idea.

graft:
Err, unwritten shower comment: it might be worth composing your own token rather than pasting one in from dev; having the structure of the token visible on the test might be easier to read. Since the client cannot verify the signature you can leave that as gibberish.

PR #192, opened 2021-01-12
REDCap update specific records

coleshaw:
This PR adds a parameter to the REDCap loading job (`record_names`) that accepts one of three values:

* (default), `nil`, all records found in REDCap are created / updated in Magma.
* `existing`, only records that currently exist in Magma are updated.
* `[record1, record2, record3, etc.]`, an array of record names that will be updated in Magma.

I've also tested this manually and it seems to filter as expected.


graft:
:crying_cat_face: This is beautiful.

PR #193, opened 2021-01-12
Metis copy across projects

coleshaw:
This PR updates the Metis file copy and bulk_copy endpoints to allow the destination metis path to be in a different project. Also updates the metis_client `cp` command to support this. User only can copy to the new project if they have permissions in that project + bucket.


metis/bin/metis_client
```diff
@@ -257,8 +257,8 @@ class MetisClient
   def copy_file(project_name, file_path, proposed_file_path)
     if proposed_file_path.start_with?('metis://')
       match = metis_path(proposed_file_path)
```
corps, line 259: Hmm, is match actually used now?

metis/lib/models/file.rb
```diff
@@ -111,7 +111,7 @@ def self.link_to_block(params)
       dest_folder_path, dest_file_name = Metis::File.path_parts(params[:dest_file_path])
 
       dest_bucket = Metis::Bucket.find(
-        project_name: params[:project_name],
+        project_name: params[:dest_project_name] || params[:project_name],
```
corps, line 114: Does this elegantly handle invalid project_names?  We get a long of rollbar errr pings for simple invalid parameter issues, if we could have a non 500 response to bad dest project name, that would be ideal.

metis/spec/bucket_spec.rb
```diff
@@ -224,7 +224,9 @@ def app
       json_post('/athena/bucket/update/my_bucket', owner: 'metis', access: 'viewer')
 
       expect(last_response.status).to eq(422)
-      expect(json_body[:error]).to eq('Invalid bucket: "my_bucket"')
+      expect(json_body[:error]).to eq(
+        "Invalid bucket on project athena: \"my_bucket\""
```
corps, line 228: Oh nice, this is pretty helpful.

corps (approved):
Overall ok, but I had some minor comments.

metis/lib/models/file.rb
coleshaw: That should get handled at the controller level already, in the sense that the project is compared to the user's list of permissions. If the user doesn't have access to either of these projects (or the project doesn't exist), the controller returns a 422 about an invalid bucket and the action should never reach this point.

We could check again here against the user's permissions, but it feels redundant.

PR #195, opened 2021-01-13
disable record name filtering for now -- needs fixing for tables

coleshaw:
This PR disables the record filtering feature in the REDCap loader -- the current logic breaks a little on tables, in terms of getting their templates and getting the list of current record names. So disabling the feature for now while I re-think a bit.

:disappointed:  No review really necessary, just an FYI ping for you, @graft

### DIFF
--- a/polyphemus/lib/etls/redcap/lib/loader.rb
+++ b/polyphemus/lib/etls/redcap/lib/loader.rb
@@ -6,11 +6,15 @@ module Redcap
       @records = {}
       @logger = logger
 
+      # Fetch existing records separately, since tables
+      #   don't have identifiers and including
+      #   "record_names": "all", "attribute_names": "identifiers"
+      #   will not return their
+      #   templates.
       @magma_models = magma_client.retrieve(
         Etna::Clients::Magma::RetrievalRequest.new(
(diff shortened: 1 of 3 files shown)


graft:
A thought I had about your previous PR was that it should perhaps be part of the template rather than a command-line option. I am thinking of how for a particular model, only *some* portions of the records should be loaded this way (e.g. biospecimen_date for the assay models), while the rest of the record might be loaded more conventionally. Since a model may have multiple templates loading into it (reflecting disparate data sources converging into that model) which might each load different attributes, or the same attributes with different behavior depending on the input source, this behavior should perhaps be template- rather than model-specific. I think this means you can't specify an actual list of record names to update (instead of "existing"), but I'm not sure that behavior is useful anyway.

coleshaw:
Hm, I might need some clarification on the behavior of these attributes, and which ones besides biospecimen_date would need to behave like this? The assay models aren't currently in the project definitions -- should they be?

For example, biospecimen_date would only create a revision if the record is an existing Magma record? Would new records then require two steps -- first, create the new record in Magma, then re-run the loader so that these attributes (like biospecimen_date) populate?

PR #197, opened 2021-01-14
REDCap prettyprint and strict mode

coleshaw:
This PR makes the following changes to the REDCap loader:

* prettyprints the JSON output for the records
* provides a summary with counts of various model records updated
* enables a "strict" mode for table models, where records in Magma that do not have corresponding REDCap data have their table attributes set to `[]` (so stale data is removed from Magma).


corps (approved):
Yeah, I think this approach will work.  Cool.

polyphemus/lib/etls/redcap/lib/magma_models.rb
```diff
@@ -0,0 +1,66 @@
...
+      #    is a table from its parent's definition of the
+      #    child attribute.
+      models.model(
+        parent_model_name(model_name)
+      ).template.attributes.attribute(
+        model_name.to_s
+      ).attribute_type == Etna::Clients::Magma::AttributeType::TABLE
+    end
+
+    def model_record_names(model_name)
```
corps, line 54: Does make me ponder, though, that eventually we'd consider wanting per-model definition of what constitutes redcap "owned" records.  For instance, imagine assays being split between redcap and other sources.  I think the model class would want to have a hook to say "this is the filter to use when retrieving to find my records".

coleshaw: Hm...is that more true at the attribute level than at the individual record level? You're saying, what if `sc_rna_seq` record 1, `date` was from REDCap but record 2 `date` is from a different source? 

I can see "sc_rna_seq" has some subset of attributes from REDCap and others from some other source, but that's controlled by the config JSON, right?

coleshaw: Having assay records for a single model come from multiple sources would also imply that there are multiple processing pipelines, which seems unlikely?

corps: :shrug:  Ok if it's not likely we don't have to address it.  I fear one day it might occur though because, hey, people do things lol.  But we can also address that on the day that it does occur, it isn't a priority as much as a thought.

coleshaw: If you can imagine it, it'll probably happen. But since we don't have a way to pull that metadata info "per record" now, I think we'll have to punt...

PR #198, opened 2021-01-14
Metis files should always be downloaded, not rendered in the browser …

corps:
…inline.

Files sent from the download endpoint will no longer be rendered, but downloaded in all cases.  Prevents XSS attacks and such.

Tested locally with html file before and after change.


coleshaw (approved):
:+1:

coleshaw:
Will this make doing thumbnail previews in Timur difficult, but that's the tradeoff to make.

corps:
> Will this make doing thumbnail previews in Timur difficult? But that's the tradeoff to make.

It should not.  The content disposition only comes into play when the browser directly visits the url.  You can still reference the url as an html element itself, and that will force it to be rendered as expected.

PR #199, opened 2021-01-15
do not compact out nil values in strict mode

coleshaw:
This PR does the following:

* Re-organizes the redcap test fixtures by form and to include the `record_id` field.
* Leaves `nil` values in strict mode, to include in the Magma update. This might turn out to be undesirable if it runs into fields that aren't "nullable" via the API, or if it leads to blank / null rows in tables...but the latter would make Magma an accurate reflection of REDCap, which I believe is the goal.


corps (approved):
Those are some heavy duty tests for these changes, much appreciated!!

PR #200, opened 2021-01-19
Assign hmac headers for authorized upload

coleshaw:
This PR adds HMAC headers for the user when authorizing an upload. Currently when you upload via Browser or metis_client, the user is set to `metis|Metis`, not the original user that initiated the upload. This change allows for better trackability for each upload.

### DIFF
--- a/metis/lib/models/file.rb
+++ b/metis/lib/models/file.rb
@@ -150,8 +150,7 @@ class Metis
           id: :metis,
           headers: user ? {
             email: user.email,
-            first: user.first,
-            last: user.last
+            name: user.name.strip
           } : {}
         ).url_params
       )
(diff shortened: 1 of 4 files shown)


metis/lib/models/file.rb
```diff
@@ -147,7 +148,11 @@ def self.hmac_url(method, host, path, expiration=0)
           expiration: (Time.now + expiration).iso8601,
           nonce: Metis.instance.sign.uid,
           id: :metis,
-          headers: { }
+          headers: user ? {
+            email: user.email,
+            first: user.first,
+            last: user.last
```
graft, line 154: Could we actually use `user.name` here instead? I've been meaning to deprecate First/Last for years.

PR #201, opened 2021-01-20
Simplify the browser data loading pathway.

corps:
Still testing, but I found a way to unwind a lot of the logic and present a clear loading path for the browser component.  Prevents infinite cycles and is orders of magnitude easier to read.

Also introduced a crude "error state" visually when loading fails.  I'd like to develop it more, but as a first step getting the component to be able to detect and render based on a failed load is better than existing (just shows nothing and loops infinitely).

### DIFF
--- a/docker/build_image
+++ b/docker/build_image
@@ -81,7 +81,7 @@ function addToDigest() {
 }
 function recursiveDigest() {
   if [[ -n "$digestOnly" ]]; then
-     find "$1" -type f ! -path '*/node_modules/*' -exec md5sum {} \;
+     find "$1" -type f ! -path '*/node_modules/*' ! -path '*/.npmrc' ! -path '*/.bundle/config' -exec md5sum {} \;
      echo
   fi
 }
@@ -94,7 +94,7 @@ function findNewest() {
     find=gfind
   fi
 
-  lastModifiedFile=$($find $1 -type f ! -path '*/node_modules/*' -printf "%T+ %p\n" | sort | tail -1 | cut -d ' ' -f2)
+  lastModifiedFile=$($find $1 -type f ! -path '*/node_modules/*' ! -path '*/.npmrc' ! -path '*/.bundle/config' -printf "%T+ %p\n" | sort | tail -1 | cut -d ' ' -f2)
   if [ "$lastModifiedFile" -nt "$newestFile" ]; then
       newestFile="$lastModifiedFile"
    fi
--- a/etna/packages/etna-js/hooks/useAsyncWork.js
+++ b/etna/packages/etna-js/hooks/useAsyncWork.js
@@ -25,7 +25,9 @@ export default function useAsyncWork(f, { cancelWhenChange = undefined }) {
   const [cancellable, setCancellable] = cancelWhenChange ? useState(new Cancellable()) : [null, null];
 
   if (cancellable) useEffect(function () {
-    cancellable.cancel();
+    // Return the cancel as a cleanup operation -- this will get invokved only between the cancelWhenChange has new
+    // values, or the component is dismounted.
+    return () => cancellable.cancel();
   }, cancelWhenChange);
 
   const wrapper = useCallback(_wrapper, [setLoading, f, cancellable, setCancellable]);
--- a/etna/packages/etna-js/hooks/useAsyncWork.js
+++ b/etna/packages/etna-js/hooks/useAsyncWork.js
@@ -20,8 +20,15 @@ import {Cancellable} from "../utils/cancellable";
     that result in calls to this hook may trigger a cancellation of the last known invocation of f.  Any consecutive
     invocations of f cancel any previous invocations.
  */
-export default function useAsyncWork(f, { cancelWhenChange = undefined }) {
+export default function useAsyncWork(f, { cancelWhenChange = undefined, renderedState = undefined }) {
   const [loading, setLoading] = useState(false);
+  const [lastPendingResolve, setPendingResolve] = useState(null);
+  if (lastPendingResolve) {
+    lastPendingResolve(renderedState);
+    setPendingResolve(null);
+  }
+  // const [last, setLoading] = useState(false);
+
   const [cancellable, setCancellable] = cancelWhenChange ? useState(new Cancellable()) : [null, null];
 
   if (cancellable) useEffect(function () {
@@ -31,7 +38,14 @@ export default function useAsyncWork(f, { cancelWhenChange = undefined }) {
   }, cancelWhenChange);
 
   const wrapper = useCallback(_wrapper, [setLoading, f, cancellable, setCancellable]);
-  return [loading, wrapper];
+  const awaitNextRender = useCallback(_awaitNextRender, [setPendingResolve]);
+  return [loading, wrapper, awaitNextRender];
+
+  function _awaitNextRender() {
+    return new Promise((resolve) => {
+      setPendingResolve(() => resolve);
+    })
+  }
 
   function _wrapper() {
     let result = f.apply(this, arguments);
--- a/timur/lib/client/jsx/actions/magma_actions.js
+++ b/timur/lib/client/jsx/actions/magma_actions.js
@@ -118,14 +118,14 @@ export const requestDocuments = (args) => {
         return Promise.reject(e);
       }
 
-      e.response.json().then((response) => {
+      return e.response.json().then((response) => {
         let errStr = response.error
           ? response.error
           : response.errors.map((error) => `* ${error}`);
         errStr = [`### Our request was refused.\n\n${errStr}`];
         dispatch(showMessages(errStr));
+        return Promise.reject(e);
       });
-      return Promise.reject(e);
     };
 
     let get_doc_args = [
--- a/timur/lib/client/jsx/actions/magma_actions.js
+++ b/timur/lib/client/jsx/actions/magma_actions.js
@@ -115,7 +115,7 @@ export const requestDocuments = (args) => {
     let handleRequestError = (e) => {
       if (!e.response) {
         dispatch(showMessages([`Something is amiss. ${e}`]));
-        return;
+        return Promise.reject(e);
       }
 
       e.response.json().then((response) => {
--- a/timur/lib/client/jsx/actions/magma_actions.js
+++ b/timur/lib/client/jsx/actions/magma_actions.js
@@ -125,9 +125,7 @@ export const requestDocuments = (args) => {
         errStr = [`### Our request was refused.\n\n${errStr}`];
         dispatch(showMessages(errStr));
       });
-      return new Promise((resolve, reject) => {
-        reject(e);
-      });
+      return Promise.reject(e);
     };
 
     let get_doc_args = [
--- a/timur/lib/client/jsx/actions/magma_actions.js
+++ b/timur/lib/client/jsx/actions/magma_actions.js
@@ -395,20 +395,22 @@ const cleanFileCollectionRevisions = (revisions, model_template) => {
 // Download a TSV from magma via Timur.
 export const requestTSV = (params) => (dispatch) => getTSVForm(params)
 
-export const requestAnswer = (question, callback) => {
+export const requestAnswer = (question, callback = null) => {
   return (dispatch) => {
     let localSuccess = (response) => {
       if ('error' in response) {
         dispatch(showMessages([`There was a ${response.type} error.`]));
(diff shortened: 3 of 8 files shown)


etna/packages/etna-js/hooks/useAsyncWork.js
```diff
@@ -25,7 +25,9 @@ export default function useAsyncWork(f, { cancelWhenChange = undefined }) {
   const [cancellable, setCancellable] = cancelWhenChange ? useState(new Cancellable()) : [null, null];
 
   if (cancellable) useEffect(function () {
-    cancellable.cancel();
+    // Return the cancel as a cleanup operation -- this will get invokved only between the cancelWhenChange has new
```
corps, line 28: Slight adjustment here.  useEffect can return a function that will be used _cleanup_ when the component dismounts,  We want to cancel any request chains after a component dismounts so that it does not attempt to change the state of a dismounted component.

timur/lib/client/jsx/actions/magma_actions.js
```diff
@@ -125,9 +125,7 @@ export const requestDocuments = (args) => {
         errStr = [`### Our request was refused.\n\n${errStr}`];
         dispatch(showMessages(errStr));
       });
-      return new Promise((resolve, reject) => {
-        reject(e);
-      });
+      return Promise.reject(e);
```
corps, line 128: Shorter, but equivalent syntax.

timur/lib/client/jsx/actions/view_actions.js
```diff
@@ -10,17 +10,17 @@ const addView = (view_name, view) => ({ type: 'ADD_VIEW', view_name, view })
  * Request a view for a given model/record/tab and send requests for additional 
  * data.
  */
-export const requestView = (model_name, success, error) => (dispatch) => {
+export const requestView = (model_name) => (dispatch) => {
```
corps, line 13: This was used in only one place so that change was easy enough.  The callbacks are no longer necessary since `invoke` will both pass through the dispatch for us and give back the return value (a promise).

timur/lib/client/jsx/components/browser/browser.jsx
```diff
@@ -67,21 +77,62 @@ export default function Browser({model_name, record_name, tab_name}) {
...
+  }, [invoke, setLocation]);
 
   // Set at 'skin' on the browser styling.
   let skin = 'browser';
   if (mode === 'browse') skin = 'browser ' + model_name;
   const editMode = useCallback(() => setMode('edit'), [setMode]);
 
-  // On mount
+  // const loading = !view || !template || !record || !tab_name;
+  const [loading, loadDocuments] = useAsyncWork(function* loadDocuments(model_name, view) {
```
corps, line 94: `useAsyncWork` was a hook I introduced for the search page, too.  It basically uses es6 generator expressions to "flatten" promise chains.  It is similar to expanding each yielded promise with a `.then(() =>` except for two main benefits:
1.  It supports cancelling.  Promises do not (natively)
2. It's a flatter syntax that more elegantly handles native js syntax (like loops, ifs, try, etc).

timur/lib/client/jsx/components/browser/browser.jsx
```diff
@@ -67,21 +77,62 @@ export default function Browser({model_name, record_name, tab_name}) {
...
+        attribute_names,
+        exchange_name,
+      });
+    }
+
+    setMode('browse');
+  }, { cancelWhenChange: [] });
+
+  // On mount, check if we have model and record names, and if not, navigate away and reload this component.
   useEffect(() => {
```
corps, line 135: This useEffect is now a one time on mount hook.  It either redirects the browser or starts loading data.  Once.

timur/lib/client/jsx/components/browser/browser.jsx
```diff
@@ -200,64 +250,3 @@ function useEditActions(setMode, browserState) {
   }
 }
 
-function useTabActions(browserState, setMode) {
```
corps, line 203: All of this goes away!!!! :rainbow: :star2: :rabbit: :tophat: :boom:

timur/lib/client/jsx/components/search/search.jsx
```diff
@@ -37,23 +37,9 @@ import {
 import ModelViewer from '../model_viewer';
 import useAsyncWork from "etna-js/hooks/useAsyncWork";
 import SearchQuery from "./search_query";
-import {Loading} from "etna-js/components/Loading";
```
corps, line 40: My editor was showing this stuff isn't actually used anymore (the other loader supercedes it).

corps:
Still fixing one issue, hold up before trying locally.

timur/lib/client/jsx/components/browser/browser.jsx
```diff
@@ -12,6 +12,8 @@
 // Framework libraries.
 import React, {useState, useCallback, useEffect, useMemo} from 'react';
 import {connect} from 'react-redux';
+import 'regenerator-runtime/runtime';
```
graft, line 15: Should this be in `etna-js/hooks/useAsyncWork` since it appears to set a global?

timur/lib/client/jsx/components/browser/browser.jsx
```diff
@@ -67,21 +77,62 @@ export default function Browser({model_name, record_name, tab_name}) {
...
-  // On mount
+  // const loading = !view || !template || !record || !tab_name;
+  const [loading, loadDocuments] = useAsyncWork(function* loadDocuments(model_name, view) {
+    setMode('loading');
+
+    if (!template) yield invoke(requestModel(model_name));
+    if (!view) {
+      // we are told the model and record name, get the view
+      view = yield invoke(requestView(model_name));
+    }
```
graft, line 101: I don't quite understand the difference between the two lines - in line 97 we are calling requestModel but do not set the template from the result (then how does it get set without remounting the component?), but in line 100 we set the view after resolving the promise?

timur/lib/client/jsx/components/browser/browser.jsx
```diff
@@ -67,21 +77,62 @@ export default function Browser({model_name, record_name, tab_name}) {
...
+
+    if (!tab_name) {
+      browseToTab(getDefaultTab(view));
+      return;
```
graft, line 105: This is also a redirect after loading has commenced - it is a bit strange that this happens here while the other redirect happens in useEffect()

timur/lib/client/jsx/components/browser/browser.jsx
corps: Unfortunately, it can't exist inside of the etna-js package.  For reasons that I haven't determined, the global has to be set in  timur-local package.  I suppose I could put it at the timur/App.js level though.

timur/lib/client/jsx/components/browser/browser.jsx
corps: Yeah, this is kind of the weirdness of the redux store.  I didn't adventure to ripping all of that out quite yet.

`view` needs to be set locally, because the updated value is required to continue loading the tab (see the template reference on line 115).  `template` wasn't actually necessary to be updated because the value was ultimately still going to be correct on re-render thanks to the store (not quite a remount though).  In the end I just added setting template = the result of that yield and it still works either way, but maybe less confusingly.

Long term, for us to remove all the redux store stuff and keep data in react state and context values, we'd have to refactor  a lot more.  If we're going to do that refactor, I'd like us to scope and schedule it because I think it will involve introducing several bugs and testing and refining to land it right.

timur/lib/client/jsx/components/browser/browser.jsx
corps: The first browseTo that happens in useEffect only requires loading the project.  This browseTo requires the view to have loaded to determine the default tab, which cannot happen in the useEffect without also increasing complexity of that useEffect logic.

That said I can flatten this into the loader generator and simplify how it looks.

corps: Ok, check again, looks a lot more clean.

corps:
@graft @coleshaw Feel free to test this locally now, it works quite well.

docker/build_image
```diff
@@ -81,7 +81,7 @@ function addToDigest() {
 }
 function recursiveDigest() {
   if [[ -n "$digestOnly" ]]; then
-     find "$1" -type f ! -path '*/node_modules/*' -exec md5sum {} \;
+     find "$1" -type f ! -path '*/node_modules/*' ! -path '*/.npmrc' ! -path '*/.bundle/config' -exec md5sum {} \;
```
corps, line 84: This also reduces unnecessary etna builds.  When the docker stuff is deployed all of this goes away!  The etna-gem will just be a local package reference!

timur/lib/client/jsx/actions/magma_actions.js
```diff
@@ -115,7 +115,7 @@ export const requestDocuments = (args) => {
     let handleRequestError = (e) => {
       if (!e.response) {
         dispatch(showMessages([`Something is amiss. ${e}`]));
-        return;
+        return Promise.reject(e);
```
corps, line 118: Lots of consuming errors without propagating results in weird states.

timur/lib/client/jsx/actions/magma_actions.js
```diff
@@ -397,28 +395,30 @@ const cleanFileCollectionRevisions = (revisions, model_template) => {
...
 
-export const requestAnswer = (question, callback) => {
+export const requestAnswer = (question, callback = null) => {
   return (dispatch) => {
     let localSuccess = (response) => {
       if ('error' in response) {
         dispatch(showMessages([`There was a ${response.type} error.`]));
         console.log(response.error);
-        return;
+        if(!callback) throw new Error(response.error);
```
corps, line 404: Same here.  To preserve the old original behavior in other places in the code, this is conditional on having passed in a callback or expecting the promise to be returned directly.  Not great, but I think to tackle this as a greater refactor we'd need to do some planning.

coleshaw:
Hm, I'm seeing the blank loading page and spinny Timur when there isn't a view defined....will try to investigate further to pinpoint where the issue happens.

corps:
> Hm, I'm seeing the blank loading page and spinny Timur when there isn't a view defined....will try to investigate further to pinpoint where the issue happens.

Maybe timur didn't rebuild, maybe try restarting the service?

coleshaw:
It's pretty consistent, across docker restarts and re-builds. Seems like it's because the view doesn't exist, so the API request fails with a 404 (i.e. GET https://timur.development.local/api/view/mvir1/cytof), which then stops the render or leaves the page stuck in `loading`?

corps:
> It's pretty consistent, across docker restarts and re-builds. Seems like it's because the view doesn't exist, so the API request fails with a 404 (i.e. GET https://timur.development.local/api/view/mvir1/cytof), which then stops the render or leaves the page stuck in `loading`?

Ok, got it.  Thank you, I'll take a deeper look.

corps:
Ah, got the fix!

coleshaw:
Not to be a stinker, but now I get `Failed to load.` due to the 404 -- I should get a default render of attributes instead?

corps:
> Not to be a stinker, but now I get `Failed to load.` due to the 404 -- I should get a default render of attributes instead?

That's not being a stinker, I just am unsure of what the correct behavior.  That seems reasonable, I'll make that change.  Anything seemed better than no notification that something had gone wrong (which is non distinguishable from "still loading").

coleshaw:
Thanks! Yeah, I'm not sure how the default render was generated in the previous logic on the 404...

graft:
The previous logic was, it would ask for a view from the server, if there wasn't one, the server responds with 404. The action then deposits an empty hash in place of the view. The `selectView` method then checks to see if the view is an empty hash; if it is, it uses `defaultView(template)` to generate a default view.

corps:
Goooott it, ok, so yeah the data selectors did some heavy lifting there.  I replicate that by calling the data selectors inline the generator and it works now, thanks!

corps:
@graft @coleshaw Take another look?

I introduced an ability for asynchronous effects to synchronize state with a component.  It can yield a function which provides a promise that awaits for the next component's render, and resolves to the state that it is passed in.  

This allows us to use selectors both outside and inside components, and allows async work to 'await' the dispatched redux logic to update the component's internal state and receive those values (having passed all action + selector logics on the way).

coleshaw (approved):
Looks great!

PR #205, opened 2021-01-22
metis client auto-refreshes token when it nears expiration

coleshaw:
This PR updates metis_client to check if a token is near expiration. If so, it calls Janus to get a new token so that long-lived downloads and uploads can occur without manual intervention.

### DIFF
--- a/metis/bin/metis_client
+++ b/metis/bin/metis_client
@@ -114,7 +114,77 @@ class Upload
   end
 end
 
-class MetisClient
+class Client
+  # This is duplicated code because we need to refactor Metis client
+  #   to use the etna gem clients...
+  def token_expired?
+    # Has the token already expired?
+    token_will_expire?(0)
+  end
+
+  def token_will_expire?(offset=3000)
+    # Will the user's token expire in the given amount of time?
+    epoch_seconds = JSON.parse(Base64.urlsafe_decode64(@token.split('.')[1]))["exp"]
+    expiration = DateTime.strptime(epoch_seconds.to_s, "%s").to_time
+    expiration <= DateTime.now.new_offset.to_time + offset
+  end
+
+  private
+
+  def get(path)
+    request(path, Net::HTTP::Get)
+  end
+
+  def post(path)
+    request(path, Net::HTTP::Post)
+  end
+
+  def delete(path)
+    request(path, Net::HTTP::Delete)
+  end
+
+  def request
+    raise "Needs to be implemented in subclass."
+  end
+
+  def https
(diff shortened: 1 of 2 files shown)


graft:
Looks neat! Does this obviate the need for a long lived token?

coleshaw:
Hmmmm, I would say semi-obviates. The token in your environment still has to be a current token, so you can't have an open terminal that you keep re-using for a long time. But if you start the process with an active token, that singular process won't die in the middle. Starting a new process in that terminal would then require you to still fetch a new token from Janus (assuming the old one expired).

graft:
Hmm, so this fixes interactive mode, but command mode (i.e. doing `metis_client <metis_path> <cmd>`) is still perhaps vulnerable to failure, since it would involve re-initializing the process and the token doesn't get re-exported.

coleshaw:
Ah, great point! Let me poke around a little more.

coleshaw:
So I'm not sure there's a way to re-export the refreshed token into the parent shell. Do you know if any of the users loop through individual commands using command mode? I think Arjun feeds in a file, i.e. `$ metis_client < instructions.txt`, which works with the current method because all the instructions are fed into a single instance of the client. Others use interactive mode.

corps:
> So I'm not sure there's a way to re-export the refreshed token into the parent shell. Do you know if any of the users loop through individual commands using command mode? I think Arjun feeds in a file, i.e. `$ metis_client < instructions.txt`, which works with the current method because all the instructions are fed into a single instance of the client. Others use interactive mode.

So, exporting to the parent process's environment is explicitly impossible due to security without cooperation from the parent process.
What you can do, is have the child, for instance, output to stdout the new token, and the parent would have to watch the stdout and modify itself with the new token.  Otherwise, this isn't possible.

corps:
You could have the command mode maybe write to a file the most "fresh" token on completion.

```
metis_client my_new_token.txt < my_commands # outputs TOKEN=ASDFASDFASDF to the my_new_token.txt file
source my_new_token.txt
```

coleshaw:
So basically an informational message...? :+1:

graft:
I don't think we should try to export out a new token. We could leave this be here, so that long-running single transactions are safe. Between metis_client commands, there is another machine-friendly route, which is to set a public key on your janus account and generate a new token as required in the shell. This might also be done automatically within metis_client, perhaps, if the user has a public key set on janus and a private key configured with metis_client.

coleshaw:
Okay, that makes sense. If there's an alternate route for the command-method, I'll take out the export.

PR #206, opened 2021-01-22
Zc/archimedes dsl

corps:
This is a WIP, I have not finished the last couple of pieces before I am ready to demo, but I did want to share so as to atleast hopefully shed more light on my thinking.

This approach makes archimedes a python library that provides 3 key pieces of functionality:
1.  A set of functions intended to either call out to R or provide pandas access to data utilities
2.  A linter that verifies that workflow cells only use archimedes provided functionality
3.  A job runner, that wraps a given python script into a sandbox via docker.  The job runner itself would be a docker process that has access to all secrets -- like secrets to reading the storage, for instance.  It would also have the docker socket mounted from the host, allow it to start sibling docker processes that run the given script, but in a much more restricted environment: that is, not access to the secrets or filesystem of the runner itself.  In this restricted environment, the worker script interfaces purely through named pipes and http access to our apis.  It can only read and write data that has been backed into the image or mounted explicitly.

Docker has the power to constrain this worker process in various ways -- cpu cycles, temporary disk size, memory, network, the whole deal.

This PR also has a working development docker environment for python, using `poetry` as the python package manager and installing `R` along with other useful tools by default.

Again, this is still WIP but I figured it might be something to chew on mentally.  I'll have a demonstrable demo early next week (just finishing the runner right now). 

Of course, this demo does NOT answer the following questions:
1.  What is the backing storage?  In my demo, the backing storage would just be a static file system, but obviously we would want something more robust and remote.
2.  What is the scheduler & api?  The runner is modular and intended to be completely ignorant of the gateway api and the scheduler parts.  Either of those two can wrap this single python function very simply by any number of means.


archimedes/archimedes/runner.py
```diff
@@ -0,0 +1,93 @@
...
+                ),
+                Mount(
+                    target="/var/run/vulcan/outputs",
+                    source=outputs_dir,
+                    read_only=False
+                ),
+            ]
+
+            output_pipes = [
+                (os.mkfifo(os.path.join(outputs_dir, md5(output_key).hexdigest())), output_key)
```
corps, line 63: This ensures that any inputs and outputs are streamed and do not buffer -- even if the storage is remote, like redis, the inputs and outputs would only occupy small chunks at a time as they make their way over the network.

corps:
Another important goal of this approach:  scripts can be run locally without any support (or even docker).

The script runner provides named pipes as access to the storage, and they wrap the script in docker to provide the security sandbox, but the scripts themselves can be run locally and behave exactly the same, only that they write directly to disk (instead of the named pipes) and obviously have full access to the user's filesystem (security isn't a concern for local runs).

archimedes/archimedes/runner.py
```diff
@@ -0,0 +1,93 @@
...
+                for output_key in output_keys
+            ]
+
+            input_pipes = [
+                (os.mkfifo(os.path.join(outputs_dir, md5(input_key).hexdigest())), input_key)
+                for input_key in output_keys
```
graft, line 69: Here you have a fifo to your input stream, presumably when complete from the calculation data store, which you are mounting into the container for execution - but presumably this process itself is running in a container? Can it re-mount its volume into another running container?

graft:
Pretty neat so far - I actually like the use of files on disk and I think this might end up being the most reasonable way to actually store data, given the need to persist and the potential size off the files.

I think a good plan of action to get this in place (which we should do before we get UMAP stuff started) might be something like:

1) Make a running service available
2) Add a `script_type` flag to plot scripts and have them run their script via this service if `script_type=python`
3) Convert all running plots over to python

The major hole in this regard would seem to be delivering a JSON packet in the correct format (that is, dealing in string/int/float/date/boolean/matrix/vector). To this end the runner ought to be checking the output type of the cell to ensure that it is in the correct format - it doesn't seem like the linter could handle this.

archimedes/archimedes/runner.py
corps: Yes.  Volumes and mounts are shareable  between volumes.  For instance, in staging I share the `/app/public` across multiple containers.  The trick here is to think of mounts in docker less as virtualized file systems syncing between each other, and more like `ln -s` files that all point to the same source node (atleast in linux, where docker is uses linux container technology).

corps:
> Pretty neat so far - I actually like the use of files on disk and I think this might end up being the most reasonable way to actually store data, given the need to persist and the potential size off the files.
> 
> I think a good plan of action to get this in place (which we should do before we get UMAP stuff started) might be something like:
> 
>     1. Make a running service available
> 
>     2. Add a `script_type` flag to plot scripts and have them run their script via this service if `script_type=python`
> 
>     3. Convert all running plots over to python
> 
> 
> The major hole in this regard would seem to be delivering a JSON packet in the correct format (that is, dealing in string/int/float/date/boolean/matrix/vector). To this end the runner ought to be checking the output type of the cell to ensure that it is in the correct format - it doesn't seem like the linter could handle this.

I agree with #1. I have doubts if we should do 2 & 3 first, or after we support the vulcan front end efforts, but we can and should discuss soon in person.

corps:
@graft @coleshaw 

Now includes tests and the stable interface I believe will work between vulcan and this piece.

corps:
When you guys get a chance, mind taking another look?

archimedes/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM archimedes-base
```
corps, line 1: This powers the `release-test` `docker push` commands.  It builds the final true production image, and thus is similar to the docker-compose but without any locally mounted files.

archimedes/Makefile
```diff
@@ -0,0 +1,16 @@
+include ../make-base/stubs.mk
+
+app_name=archimedes
+include ../make-base/docker-compose.mk
+
+app_name_capitalized:=$(shell echo ${app_name} | tr [a-z] [A-Z])
```
corps, line 6: Don't yet have any other python apps, but once we did I may generalize some of this out more.

archimedes/archimedes/__init__.py
```diff
@@ -0,0 +1 @@
+__version__ = "0.1.0"
```
corps, line 1: When we package archimedes and release it as a pypi package, this will be our version.  Currently not prioritizing that, though.

archimedes/archimedes/functions/dataflow/path.py
```diff
@@ -0,0 +1,27 @@
+# Do not let these imports become usable within scripts themselves
+import os as _os
+import os.path as _os_path
+
+
+def input_path(name, inputs_env=_os.environ, inputs_dir=None):
```
corps, line 6: These are just the lowest level primitives.  When we get to supporting pandas, we can have smarting `input_dataframes()` functions, etc, that look nicer.

archimedes/archimedes/runner.py
```diff
@@ -0,0 +1,314 @@
+"""
```
corps, line 1: This is really just a demonstration of a synchronous scheduler, suitable for locally running scripts or running in docker.  
In this version, I imagine that the vulcan web endpoint just calls this runner directly, by invoking the `archimedes:latest` image and running `poetry run archimedes-run --args --here`.  In the future, though, this logic would get folded into whatever scheduler we decided to use full time.

archimedes/pyproject.toml
```diff
@@ -0,0 +1,29 @@
...
+pylint = "^2.6.0"
+black = "^20.8b1"
+docker = "^4.4.1"
+dataclasses-json = "^0.5.2"
+
+[tool.poetry.dev-dependencies]
+pytest = "^5.2"
+
+[tool.poetry.scripts]
+archimedes-checker = 'archimedes.checker:main'
```
corps, line 24: Allows DSL script checker and the runner to be executed as scripts on their own.

coleshaw:
Pretty neat! To make sure I'm understanding the overall goal for this quarter, is the goal that Vulcan stands up and issues commands to a local Archimedes container (and thus Vulcan stores the CWL and interprets that), or is the goal of this iteration to build out a hosted-version of Archimedes that stores the CWL and includes all the individual steps' scripts, but exposes some sort of REST (or other) API to Vulcan?

coleshaw:
Oh, I see, you're saying that follow-on work will be part of the Vulcan - Archimedes Bridge work, a separate card / chunk of stuff...?

corps:
In this quarter, our worker api is synchronous, due to the lack of a proper scheduler.  Which is totally fine for our scope.

I don't think turning archimedes itself into an API is either necessary, nor clearly the long term vision.  When we _do_ add a scheduler, like airflow, it itself may provide its own api or tooling about how things are scheduled and managed.  At that point, we'd change vulcan to interact with this new api, _instead of having to also change our frontend_.   Building archimedes as purely a process that runs synchronously for vulcan means we are committing to less work that will likely have to be scrapped.  It also means the system is less fragile -- starting processes is more reliable than serializing everything and transporting it again over a network to a host name (dns, tcp, serialization overheads).  This _also_ means that things like the storage system can -just be the filesystem of vulcan- for now, with out any extra complexity.

Might these decisions change?  Over time I imagine vulcan will run separately from scheduled jobs.  But we'd need a scheduler, atleast, to understand how this organization would work.  *So rather than committing to building an api and transport for a thing we are not  certain of the requirements for and we will likely throw away*, I really propose the process bridge -- it's smaller, and has less assumptions built in.  We can remain focused on getting the core components working (storage system, the interface between the frontend and vulcan, the archimedes library) and then address the big hole of scheduling separately (that alone could take a few months to get reliably working in a distributed world).

corps:
@graft @coleshaw my goal here is to wait for https://github.com/mountetna/monoetna/pull/213 to be stable in terms of the frontend <-> mock backend, and then merge that branch into this PR and attempt to tie archimedes interface into that API, having a sort of full code demonstration of the workflow (even if we still mock certain parts, like storage, and results right now).

That way we can all review this architecture in a 'skeleton pieces all lined up' sort of way, and decide fi we're ok with proceeding?

corps:
@graft @coleshaw You can see https://github.com/mountetna/monoetna/pull/206/commits/6df80e626f47a986f3f88003e4d1f89711d09c42 how I'm starting to tie together the backend with functional components that will orchestrate in the following manners:

1. inputs posted by frontend
2. controller adds them to "session" (either that is persisted somewhere for the user, or it's just a big state blob maintained on the frontend and passed in on every request) which manages all known user supplied inputs
3.  controller asks orchestrator to find the step(s) in the workflow that can be run next, by examining which step has all inputs met and not outputs met.
4.  controller continues to run each step that can be run, dumping contents into storage location, until there are no more steps that can be run
5.  controller runs status payload, awaiting frontend to provide additional inputs

There is still a bit more to go, but I"m getting quite close.  I think I can have the basic skeleton working e2e by early next week, and we can discuss and move forward with tighter integration.

vulcan/lib/server.rb
```diff
@@ -16,7 +16,7 @@ class Server < Etna::Server
     end
 
     get 'api/:project_name/workflows', action: 'workflows#fetch', as: :workflows_view, auth: { user: { can_view?: :project_name } }
-    get 'api/:project_name/workflows/:workflow_name/:data', action: 'data#fetch', as: :data_view, auth: { user: { can_view?: :project_name } }
+    get 'api/:project_name/data/:cell_hash/:data_filename', action: 'data#fetch', as: :data_view, auth: { user: { can_view?: :project_name } }
     post 'api/:project_name/workflows/:workflow_name/:status', action: 'workflows#submit', as: :submit_view, auth: { user: { can_view?: :project_name } }
```
coleshaw, line 20: For a real API, I think this should actually just become `/api/:project_name/workflows/:workflow_name`...I was trying to get different status JSONs back, but not sure how to mock that out...

vulcan/lib/server/controllers/workflows_controller.rb
```diff
@@ -15,12 +15,11 @@ def fetch
   def submit
     # This is a stub API that will be used for development only.
     # Submit input data and get the specified output status back.
-
     raise Etna::NotFound, "No data for workflow #{@params[:workflow_name]}." unless "umap" == @params[:workflow_name]
 
     success(File.read(File.join(
       File.dirname(__FILE__),
-      "../data/#{@params[:status]}.json")), 'application/json')
+      "../../../data/#{@params[:status]}.json")), 'application/json')
```
coleshaw, line 22: For a real API, I think this should actually just become `/api/:project_name/workflows/:workflow_name`...I was trying to get different status JSONs back, but not sure how to mock that out...

vulcan/lib/server/models/workflow.rb
```diff
@@ -0,0 +1,150 @@
...
+        ),
+        description: description,
+        display_name: display_name,
+    )
+  end
+
+  # Returns a list of lists, describing the dependency between steps, primary inputs, and primary outputs.
+  # Each inner list is a unique path in side of the workflow starting a primary_inputs and terminating
+  # at either a primary_output or a step that is not used as an input the workflow
+  def unique_paths
```
coleshaw, line 29: Will the step name be included in the paths? It's the key in the YAML, but we also need it in the UI to display to the user, so don't want to lose it in the YAML conversion...

vulcan/lib/server.rb
```diff
@@ -16,7 +16,7 @@ class Server < Etna::Server
     end
 
     get 'api/:project_name/workflows', action: 'workflows#fetch', as: :workflows_view, auth: { user: { can_view?: :project_name } }
```
coleshaw, line 18: Does there need to be an additional route to get the steps for a given workflow? Or is that considered "data" somehow, and the UI will be passed the cell_hash for retrieving that?

vulcan/lib/server/models/workflow.rb
coleshaw: Similar question for the outputs ... especially if there are multiple outputs at a given step, I think we'll need to include the output file name in each object in the `out` Array, so the user knows which output they are looking at.

vulcan/lib/server.rb
corps: Yeah, that makes sense.  Part of my goal of having an e2e integration is so that I can pass this back to you and we can start actually shaping the API the way you'd like, with a (somewhat) backing skeleton in place.  Let me make some more progress and then we can work together and shape the api further.

vulcan/lib/server/controllers/workflows_controller.rb
corps: Yessir.  I intend to change this section here soon anyways.

vulcan/lib/server/models/workflow.rb
corps: Don't worry, this isn't yet the part that connects to the api, this is more a functional part that helps me back the api.

I will keep all the information you need preserved when I return the actual json.

corps: Yes, I will preserve the filenames in all the outputs, I assure you.  This model is backing to other functionality, i haven't gotten to the part where this ties directly into the api yet.

vulcan/lib/server.rb
corps: So, just to be super clear, I imagine that you will use, abstractly, these apis:

1.  workflow route to get, the steps / workflow (minus any states, just definitions of them).
2.  status / input routes to add new user inputs & get the current state of work (which is synchronous in this version, so it basically will run everything and give you the status of those completed jobs)
3.  data endpoints, referenced in the status values, of where to stream each step's outputs.

coleshaw: Okay, cool that makes sense. At some point once we move beyond just UMAP, I think we'll also need a "here are the workflows you can access" API?

corps: Absolutely.  Basically, there should be a crud interface to listing, creating, editing, and fetching workflows.  @graft also wants it to be possible to adhoc edit workflows, as well, which is doable with a backing crud.  But that is beyond umap.

corps:
@graft @coleshaw This is done, as far as the backend is concerned, complete with tests.

Let's get this reviewed address concerns and get it merged so that we can start integrating the frontend more deeply, adding the actual umap scripts, and all the other little tasks we can split up on top of this branch.

bin/regen-completions
```diff
@@ -32,4 +32,5 @@ gen_for "magma_app"  "./bin/magma"
 gen_for "polyphemus_app"  "./bin/polyphemus"
 gen_for "polyphemus_app"  "./bin/etl_tools"
 gen_for "etna_app"  "./bin/etna"
+gen_for "vulcan_app"  "./bin/vulcan"
```
corps, line 35: Adding vulcan commands to the completions regen script

docker/archimedes-base/Dockerfile
```diff
@@ -0,0 +1,60 @@
+FROM python:3.8-slim
```
corps, line 1: This is basically all the python and binary requirements for archimedes.  I decided to fork this into a separate base image, rather than pile it into the etna-base image, simply because these requirements have less overlap and machines that eventually are running archimedes scripts likely won't be the same machines that are web servers (say, C4).

docker/etna-base/entrypoints/build.sh
```diff
@@ -80,4 +80,4 @@ fi
 echo 'for file in /app/*.completion; do source $file || true; done' >> /root/.bashrc
 echo 'export PATH="/app/bin:$PATH"' >> /root/.bashrc
 # Allow other users to use the root bash setup
-chmod -R 777 /root/
+chmod 744 -R /root/
```
corps, line 83: I realized that I only needed this slightly lower permission grade so this change is pretty minor.

etna/lib/etna/command.rb
```diff
@@ -39,6 +39,10 @@ def boolean_flags
       def string_flags
         @string_flags ||= []
       end
+
+      def multi_flags
+        @multi_flags ||= []
```
corps, line 44: I added the ability to include flags that are repeatable in etna.

For instance

`etna my bommand --env A=1 --env B=2` would result in env: ["A=1", "B=2"]

I was planning on using this for a command line demo of the backend, but didn't end up completing it.

etna/lib/etna/directed_graph.rb
```diff
@@ -18,6 +18,27 @@ def add_connection(parent, child)
     parents[parent] = parent_parents
   end
 
+  def paths_from(root)
```
corps, line 21: This determines, in an acyclical directed graph, (but with potential convergence and divergence), one of each of the shortest paths to every node that isn't a strict subset of some other longer shortest path.  Which is how we can present a converging and diverging directed graph in a linear stable way.

etna/lib/etna/route.rb
```diff
@@ -176,13 +177,21 @@ def update_params(request)
       )
     end
 
+    def separator_free_match
```
corps, line 180: I noticed an issue with the way we handle named args; when I want to serve workflows or data files from a named argument, it considers `.` a separator, which may make sense in some cases, but in our data and cwl endpoints, it does not.  So, I added an option to enable matching named args at the extension level, too.

etna/packages/etna-js/utils/fetch.js
```diff
@@ -80,3 +80,26 @@ export const form_post = (path, params, performCheckStatus = true) => {
     body: form
   }).then(performCheckStatus ? checkStatus : (v) => v);
 };
+
+export const handleFetchError = (e) => {
```
corps, line 84: This is just part @coleshaw branch, a bunch of them are.

make-base/etna-ruby.mk
```diff
@@ -32,4 +33,4 @@ release:: Dockerfile
 
 release-test:: docker-ready
 	docker-compose up -d $(app_db_name)
-	docker run --rm -e $(app_name_capitalized)_ENV=test -e APP_NAME=$(app_name) -e RELEASE_TEST=1 -e CI_SECRET=$${CI_SECRET} -e IS_CI=$${IS_CI} --network monoetna_default $(fullTag) /entrypoints/development.sh bundle exec rspec
+	docker run --rm ${EXTRA_DOCKER_ARGS} -e $(app_name_capitalized)_ENV=test -e APP_NAME=$(app_name) -e RELEASE_TEST=1 -e CI_SECRET=$${CI_SECRET} -e IS_CI=$${IS_CI} --network monoetna_default $(fullTag) /entrypoints/development.sh bundle exec rspec
```
corps, line 36: Enables the injection of the docker.sock volume mount for vulcan release-tests

vulcan/docker-compose.yml
```diff
@@ -10,12 +10,13 @@ x-base-etna-service:
...
     - vulcan-tmp-dir:/app/tmp:rw
     - vulcan-public-dir:/app/public:rw
-    - vulcan-data-dir:/app/data:rw
     - vulcan-opt-dir:/usr/opt:rw
     - ../etna:/etna:delegated
+    - /var/run/docker.sock:/var/run/docker.sock:ro
   environment: &base-etna-env
     'VULCAN_ENV': 'development'
     APP_NAME: 'vulcan'
+    HOST_DIR_MAP: "/app=${PWD}/../vulcan"
```
corps, line 19: Data directories are mounted from the host, so we need the mapping from the host directory to the container directory to enable sending the correct docker commands.

vulcan/lib/client/jsx/spec/fixtures/sample_cwl.yaml
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env cwl-runner
```
corps, line 1: Simple adding cwl used to test against.

etna/lib/etna/route.rb
graft: Err, missed this the other day but just noticed it - is this necessary? Can't you use the `*` to catch these names? E.g. in vulcan:  `get 'api/:project_name/data/:cell_hash/*data_filename', action: 'data#fetch', as: :data_view, auth: { user: { can_view?: :project_name } }`. I guess it couldn't handle two globs like `*cell_hash/*data_filename` but otherwise it ought to work.

PR #208, opened 2021-01-27
Consolidate js components

coleshaw:
This PR does several main things, in preparation of standing up the separate Vulcan app to serve the UMAP UI:

* Moves some of the useful Timur UI components into `etna-js`. Namely the input components, Magma / Metis / Location reducers / selectors / api. Also moves the specs for those components. This may be useful to @graft as you rebuild the Janus UI, specifically the inputs?
* Creates a Vulcan docker container and makes it visible in the browser. Make sure to add the right entry in your `/etc/hosts` file for local development.
* Fixes some small Timur styling issues.
* Enforces an `https` redirect for Janus. This should fix the issue with Polyphemus logging in.

The Vulcan app itself is very much a WIP and doesn't reflect the latest approach we've been talking about, with CWL parsing. But that's next on my list and then plan to get @dtm2451 involved with writing some of the UI components.


etna/packages/etna-js/utils/fetch.js
```diff
@@ -1,3 +1,6 @@
+import downloadjs from 'downloadjs';
+import * as Cookies from './cookies';
+
 export const checkStatus = (response) => {
   let content = isJSON(response) ? response.json() : response.text();
```
coleshaw, line 5: `checkStatus` used to be semi-duplicated. There was a version in Timur that returned the response object, and this version in `etna-js` that returned a Promise. I've consolidated to the Promise version and updated the consumers accordingly.

janus/lib/server/controllers/authorization_controller.rb
```diff
@@ -121,7 +121,7 @@ def login_form
 
   def respond_with_cookie(token, refer)
     # Set redirect.
-    @response.redirect(refer, 302)
+    @response.redirect(refer.gsub("http://", "https://"), 302)
```
coleshaw, line 124: This makes all Janus redirects use `https` after logging in, regardless of what the query param in the URL provides. Should fix the issue with Polyphemus on production and also local dev (for non-HTTPS everywhere users).

corps: :+1:

vulcan/build/create-js-routes
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env sh
+
+set -e
+
+# The webpack build process depends on gem setup in order to build the routes, but it cannot run the bundle install
+# itself without conflicting with the ruby server, so it merely checks and waits before invoking.
+if [ -n "$RUN_NPM_INSTALL" ]; then
```
corps, line 7: Is... this a pattern we want to repeat?  I'm honestly not sure this is wise.  I feel that compiling routes into the bundle makes building more complicated, tests more complicated (those route functions have to get mocked), and still don't really guarantee much about the integration (the exchange, headers, response payload schemas aren't captured anyways). But it's not a blocker.

corps (approved):

vulcan/build/create-js-routes
corps: Normally I am definitely in favor of code generation, I guess my concern is really that this code generation doesn't give a lot (just routes, nothing about endpoint schemas), and it comes with acost.  If it could generate _more_ useful things, I'd be more keen.  But, in retrospect it's probably just fine.

coleshaw: What would you recommend as an alternative? Would you just bake in the paths in the methods that call the API?

corps: Yeah, I would.  If we wanted, the paths could be considdated in a global or group of functions (similar to how it is, but just baked into the code).

In reality, if the api path changes, it's highly likely other aspects of the api change too, so the code is likely in need of udpate.  The main advantage is that tests themselves can be written without worrying too much about mocks, and you're not forced to codegen just to run js tests.

graft:
This looks great! Looks like vulcan is still pretty stubby - an initial version of vulcan might take the document components out of Timur (i.e. manifests, plots) and leave the browse/search components behind.

coleshaw:
I didn't want to remove those components from Timur yet, since I wasn't sure how to handle plots -- some of the Timur views reference plots, so somehow they would have to be similarly reference-able from Vulcan?

graft:
Yeah we can leave them duplicated in Timur for the moment. Eventually what I am imagining is you could attach plots to Timur views using a view editor (either loading from a list populated from Vulcan or from a saved/copied blob of JSON from Vulcan), so the editing/production could happen in Vulcan and later be attached into Timur without needing an editor there.

PR #209, opened 2021-01-28
Vulcan progress

coleshaw:
This PR is really just to get the sync-subtrees action working. It's currently broken because the Vulcan UI files weren't cleaned up in #208  and didn't compile. They compile in this PR so sync-subtrees should work again after this is merged in.

This PR also adds in the Vulcan icon from @graft . Most of Vulcan is still pretty non-functional though and much a WIP.


vulcan/lib/client/jsx/components/vulcan_nav.jsx
```diff
@@ -33,21 +32,18 @@ const Halo = ({radius}) => (
...
-  >
+  <div id='vulcan-logo' className='etna-vulcan-logo'>
     <div className='image' />
     <Halo radius={25} />
   </div>
 ));
 
 const getTabs = () => ({
-  workflows: Routes.workflow_path(CONFIG.project_name),
+  workflow: `/${CONFIG.project_name}/workflow`,
```
graft, line 42: I would much favor moving these to a constant (e.g. just a static version of the compiled routes file) rather than having them embedded all over in individual functions, which is pretty hard to track.

corps (approved):

vulcan/lib/client/jsx/components/vulcan_nav.jsx
coleshaw: Sounds good, I can change that. But right now I'm guessing there won't be many Vulcan routes, since I expect the UI to make calls directly to Archimedes...

coleshaw: Will include this change in a subsequent PRs, so we can keep building.

corps:
Merging this to keep the build pipeline working.  @coleshaw can follow up on any unresolved feedback.

PR #210, opened 2021-01-28
Fix shibd issues caused from servername -> implied proto, speed up sh…

corps:
…ibd startup by using pre-verified metadata

Comments explain the shibd issue around https.  :shrug:   It works in staging.  Proxies being proxies I suppose.  Feel free to ask questions for clarification here, I realize this docker deployment stuff is still new.

The other fix is to pre-bake and pre-verify the incommon metadata to improve shibd startup performance.  There is a chef side change to `shibboleth2.xml` (on my docker branch) that uses this pre-baked file rather than the exhausting performance of its internal signature validation algorithm (which pegs the cpu for minutes at a time, woah).

### DIFF
--- a/docker/etna-apache/bin/prepare_shib.sh
+++ b/docker/etna-apache/bin/prepare_shib.sh
@@ -1,0 +1,17 @@
+#!/usr/bin/env bash
+
+set -e
+
+VERIFIED_SHIB_DIR=/opt/shibboleth/
+mkdir -p $VERIFIED_SHIB_DIR
+
+cd $VERIFIED_SHIB_DIR
+curl -o InCommon-metadata.xml https://md.incommon.org/InCommon/InCommon-metadata.xml
+curl -o inc-md-cert.pem https://md.incommon.org/certs/inc-md-cert.pem
+
+# Verify the metadata's signing
+xmlsec1 --verify --pubkey-cert-pem inc-md-cert.pem --id-attr:ID "urn:oasis:names:tc:SAML:2.0:metadata:EntitiesDescriptor" InCommon-metadata.xml
+
+
+
+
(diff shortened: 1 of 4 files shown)


docker/etna-apache/bin/prepare_shib.sh
```diff
@@ -0,0 +1,17 @@
...
+
+VERIFIED_SHIB_DIR=/opt/shibboleth/
+mkdir -p $VERIFIED_SHIB_DIR
+
+cd $VERIFIED_SHIB_DIR
+curl -o InCommon-metadata.xml https://md.incommon.org/InCommon/InCommon-metadata.xml
+curl -o inc-md-cert.pem https://md.incommon.org/certs/inc-md-cert.pem
+
+# Verify the metadata's signing
+xmlsec1 --verify --pubkey-cert-pem inc-md-cert.pem --id-attr:ID "urn:oasis:names:tc:SAML:2.0:metadata:EntitiesDescriptor" InCommon-metadata.xml
```
corps, line 13: Xmlsec verifies the metadata signatures before hand.  It seems that this is the step that, for whatever reason, `shibd` takes like forever to accomplish, but `xmlsec` takes like 3 seconds.  The result is that the metadata can be referenced "as is" in the `shibboleth2.xml` file.

graft:
When would shibd restart in this new regime? Only Janus really talks to shibd at all, so it should be a relatively rare event in any case. The use of xmlsec maybe obviates the painful bootup costs so it might not be so bad to be restarting it all the time.

corps:
That's an excellent question!  My concern is that in this new regime, shibd restarts are currently tied to the etna-apache docker image, so they may run anytime we change that.  In a slightly more ideal world, shibd would be extract into separate container which is communicated with via tcp.  That way, shibd would basically only restart if we changed that docker image.

That may be something I need to address before going to production.

In any case, it seemed valuable for whomever has to play with this again that they not have to 10 minutes every time they make a change to shibd configuration, and it's still a work in progress to improve.

PR #212, opened 2021-02-02
Add script to notify on deployment, bundle git shas into image.

corps:
This improves the base notifications provided by watchtower with custom notifications to slack.  It also bundles the git sha used in CI to build production images so that it can report a diff view of what changes between deploys.


corps:
This is not a high priority change, and it was part of some musing.  Focusing back on vulcan backend.

coleshaw (approved):

PR #213, opened 2021-02-03
Vulcan hooks to some API

coleshaw:
This PR has Vulcan in a state where some API hooks to an imaginary Archimedes API exist and can fetch data, so hopefully we can start building UI components on top. Will probably need tweaking as we move forward with the UI and actual API.


corps:
Awesome work dude!  Gonna spend more time reading through it later this afternoon.

vulcan/lib/server.rb
```diff
@@ -17,6 +17,7 @@ class Server < Etna::Server
 
     get 'api/workflows', action: 'workflows#fetch', as: :workflows_view, auth: { user: { active?: true } }
```
coleshaw, line 18: Note, I basically expect to throw all these server-side routes out, after we hook into Archimedes.

vulcan/lib/server/controllers/data_controller.rb
```diff
@@ -11,9 +11,17 @@ def fetch
 
     raise Etna::NotFound, "No data for value #{@params[:data]}." unless ["steps", "pools"].include?(@params[:data])
 
+    case @params[:data]
```
coleshaw, line 14: Note, I basically expect to throw all this controller out, after we hook into Archimedes.

vulcan/lib/server/controllers/workflows_controller.rb
```diff
@@ -11,5 +11,17 @@ def fetch
         symbolize_names: true)
     })
   end
+
```
coleshaw, line 14: Note, I basically expect to throw this controller out, after we hook into Archimedes.

vulcan/lib/server/data/status1.json
```diff
@@ -0,0 +1,20 @@
+{
```
coleshaw, line 1: This may not be the most accurate format, just what I'm theorizing right now. I'm assuming the server will return status with steps as a Hash, like CWL format.

vulcan/lib/server/views/client.html.erb
```diff
@@ -7,7 +7,19 @@
...
     <link rel='stylesheet' media='screen' href='/css/vulcan.bundle.css'/>
     <link rel="stylesheet" href="https://use.fontawesome.com/releases/v5.8.1/css/all.css" integrity="sha384-50oBUHEmvpQ+1lW4y57PTFmhCaXp0ML5d60M1M7uH2+nqUivzIebhndOJK28anvf" crossorigin="anonymous">
-    <script src='/js/routes.js'></script>
+    <script type='text/javascript'>
+      // All these Routes are mocked for development only.
+      const ROUTES = Object.assign(
+        window.ROUTES || {},
+        {
+          workflow: () => (`/${CONFIG.project_name}/workflow`),
+          fetch_workflows: () => ('/api/workflows'),
```
coleshaw, line 16: These will get updated to match the actual API in the future.

vulcan/lib/client/jsx/components/workflow/browser.jsx
```diff
@@ -30,24 +32,52 @@ const errorDiv = (
 
 export default function Browser() {
   const invoke = useActionInvoker();
-  const browserState = useReduxState(browserStateOf());
+  const [workflows, setWorkflows] = useState(null);
```
coleshaw, line 35: Should probably use the context hook here instead...will refactor as we get into UI components.

corps:
OK!  Just wrapped up my archimedes runner PR, so I'll start looking at this closely and thinking about the integration of the backend based on these ideas.

vulcan/lib/server.rb
graft: My current understanding is Vulcan will hold onto workflow documents and orchestrate execution while Archimedes just receives and executes cells + stores data, so some of this might remain.

coleshaw: That'd be fine, though probably need some clarification since I thought the workflow CWL docs would be held elsewhere. Perhaps that's the "Vulcan - Archimedes Bridge" card in the Kanban board? Just want to make sure we know which component has responsibility for what.

vulcan/lib/client/jsx/components/workflow/manager.jsx
```diff
@@ -0,0 +1,86 @@
+// Framework libraries.
```
coleshaw, line 1: @dtm2451 , I think this is an example component you can build off of and take a look at. You'll want to clean up the rendered stuff and break that into multiple child components, but this should work and give you an idea of how data flows and how to manipulate it. We can talk through it in detail when you have time and have fiddled with it.

vulcan/lib/server.rb
corps: Hey @coleshaw 

My goal in terms of responsibility is for `ruby` to hold most of the logic around workflow definitions, session stage, storage management, and scheduling.  archimedes is intended to be `stupid as balls`.  It has one job -- run some python code in a sandbox with with some input / output specifications.  So actually, vulcan will take some responsibilities, including storage of CWL documents.  We (the backend team) will be building that out.  I need to proliferate more cards on this point.

vulcan/lib/client/jsx/spec/fixtures/sample_cwl.yaml
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env cwl-runner
```
corps, line 1: We probably don't actually need the bash line, I don't think we'll want to have this shebang in our files in general.

corps: but eh nevermind don't mind me

vulcan/lib/server.rb
coleshaw: Then the Vulcan APIs should definitely be re-worked from their current implementation.

vulcan/lib/client/jsx/api/archimedes_api.js
```diff
@@ -0,0 +1,69 @@
+import {
+  checkStatus,
+  handleFetchSuccess,
+  handleFetchError,
+  headers
+} from 'etna-js/utils/fetch';
+
+import CwlSerializer from '../serializers/cwl';
+
+const archimedesPath = (endpoint) => `${CONFIG.archimedes_host}${endpoint}`;
```
corps, line 10: I think we'll want to remove most of this -- I'm thinking that vulcan communicates directly to vulcan, and that archimedes is not a standalone api.  The main motivating factors:

1.  Keeping as much of our core application logic in ruby, and not split between languages.
2.  Archimedes can be just about scheduling and sandboxing, without the added overhead of serialization and transports (removing a whole leg of work).
3.  Vulcan frontend can be self contained, only interacting with one core api (vulcan).

coleshaw: My original concern about that was streaming large amounts of data from Archimedes -> Vulcan -> UI (instead of just Archimedes -> UI), but that was when I thought Archimedes would be providing the interface to the data store. If all that will be coming through Vulcan instead, I'll change all the UI routes to talk directly to Vulcan.

vulcan/lib/server/data/status1.json
corps: That's totally fine.  We'll work with this if it's easier.  I can work the backend to fit any shape you want, as long as the correct information is captured by _some_ means.

vulcan/lib/client/jsx/api/archimedes_api.js
corps: Yessir!  Vulcan can stream.  In fact, _it might be possible for apache to stream it_.  For atleast this version, the storage will actually just be files on the file system locally.  We could point apache to that directory and serve them using X-Send just like metis.

corps: In fact, even if the storage backend changes, I imagine that it would make sense to always atleast buffer files somewhere locally, even if it's just in chunks, so that we could enable apache to do the streaming and not ruby.  It's possible to configure apache to handle streams from, say, a file system fifo or socket that the backend arranges on behalf of a more complex storage.  The more linux filesystem magic we can work so that we don't have streaming in our scripted languages the better!

vulcan/lib/server/data/status1.json
coleshaw: Yeah, I'm not sure right now if this is easier, it's just that the UI needs to re-map the original steps from the CWL from a Hash into an Array. But for subsequent steps, I think the options are either use a Hash (like CWL), or just provide the step `name` in each Hash in the Array. Agree basically we just have to send the step name data somehow...easier on the UI if we keep things in an Array of Steps so there's less manipulation.

corps: Then that's totally fine.  _you tell me_ what the UI would prefer and I can make it happen.  I'm ok with complexity in the data transformation on the ruby side because frankly, _testing complexity in ruby is easier than javascript_.

vulcan/lib/client/jsx/api/archimedes_api.js
graft: Right now it is fine for Vulcan to run cells via some process instead of a service - but at some point there will be a scheduler involved, and Vulcan will be submitting cells to this scheduler, with Archimedes executing via the scheduler job in some worker. How does all of this relate? It seems logical to put the scheduler into Polyphemus since we will want to schedule other things besides Vulcan running cells; the Archimedes worker could run anywhere, but must in the end deposit data in a place where the user can query it - this might mean keeping everything (except the scheduler) Vulcan-proximal as you suggest, so it's all easily streamable. I guess this would look like:
Now:  Vulcan calls sandboxed Archimedes instance to run cell, which writes results to local disk for Vulcan to stream
Then: Vulcan submits cell to scheduler, which runs it via a Vulcan worker, which writes results to local disk for Vulcan to stream

corps: There's tons of different iterations of what is possible.  For sure, our scheduler may have other work than vulcan jobs.  So basically, by having vulcan just make the invocation directly now, we don't bake any assumptions about how our job infrastructure will work in the end, and we have complete flexibilty about how to proceed, for sure.

coleshaw:
Merging this preliminary Vulcan work in, summarizing the changes here since a lot has happened since the PR opened:

1) Building out UI framework for Vulcan workflow UI. Still a WIP and will be continued in subsequent PRs. A lot of the Vulcan work has changed already, but in a different branch.
2) Moved plot components from Timur to etna-js, for when we need custom plot components in Vulcan. Includes consignment components.

PR #214, opened 2021-02-05
Moves plot components from Timur to etna-js

coleshaw:
Not sure how much of the plotting UI components we'll re-use, but at the very minimum I think it's these. I've left anything to do with consignment and manifest in Timur for now, since I think for Vulcan most of those configuration items will come via the workflow and not be user-constructed.

I also replaced the color scheme with the color-blindness-friendly set of colors from dittoSeq (@dtm2451 ).

Proposing to merge this into the current Vulcan branch instead of master.

I did not have a great way to test this, but I was able to make a new plot in Timur with simple, linear data, and it rendered....so I'd call that success?


graft:
The canvas and axis components along with the various series are probably useful. The main problem with them is they rely on the matrix and vector json types and could probably be more generic. The config and plotScript stuff will probably not survive in the end...

coleshaw:
Hm, okay. I was planning to also move the consignment stuff over and have the Python workflows reformat their output data into consignment-JSON. Right now it looks like export options from scanpy is a set of TSV values, so the other option is  to just read raw TSVs in, but I figured that all the UI components currently rely on consignment format, so initially we can go with that as an interface.

etna/packages/etna-js/dispatchers/work-dispatcher.jsx
```diff
@@ -1,4 +1,4 @@
-import {createWorker, terminateWorker} from '../upload/workers/index';
+import {createWorker, terminateWorker} from '../web-workers/index';
```
corps, line 1: :+1:

etna/packages/etna-js/plots/components/category_plot/category_plot.jsx
```diff
@@ -0,0 +1,130 @@
...
+const SeriesComponent = ({
+  series,
+  index,
+  count,
+  gutter,
+  gap,
+  xScale,
+  ...props
+}) => {
+  let Component = COMPONENTS[series.series_type];
```
corps, line 53: super minor stye nit, but we could use an eager return to remove the need for the `Component &&` and any cascading nil checks.:

```
let Component = 
if (Component == null) return null;
// rest of the code can assume it is non null
```

corps (approved):

etna/packages/etna-js/plots/components/category_plot/category_plot.jsx
coleshaw: Sure, I can take a look at that!

PR #217, opened 2021-02-09
Gvaihir/timur view editor

Gvaihir:
View Editor feature. Allows to edit views in the form of JSON. 
NOTE: deleting an existing view deletes view form the database. New view is not created if submitted with a bad JSON string


timur/lib/client/jsx/components/document/document_view.jsx
```diff
@@ -19,7 +19,7 @@ export default class DocumentView extends React.Component {
   }
 
   render(){
-    let { document, onUpdate, editing, children } = this.props;
+    let { document, documentName, onUpdate, editing, children } = this.props;
```
coleshaw, line 22: nit: overwriting the global `document` here and in the related components seems slightly risky? Should this be named something else?

graft: This is my bad from way back. It probably doesn't break anything except the local scope, we can refactor it out later (these components may not survive vulcan).

timur/docker-compose.yml
```diff
@@ -46,7 +45,7 @@ x-base-etna-services: &base-etna-services
       <<: *base-etna-env
       RUN_NPM_INSTALL: "true"
       SKIP_RUBY_SETUP: "true"
-    command: npm run webpack
+    command: npm run webpack-dev
```
graft, line 49: should probably be reverted? But I also have a weird kludge to convince monoetna to build my files in dev mode (futzing with package.json) - is there a canonical way this should be done @corps?

timur/lib/client/jsx/components/plotter/plotter.jsx
```diff
@@ -85,6 +85,10 @@ const Chart = connect(
   </div>
 );
 
+const accessFilter = (access, documents)=>{
+  return documents.filter(m => m.access == access).sort((a,b) => a > b);
+};
+
```
graft, line 91: I think I left this in here by accident, I don't think it is used here.

timur/lib/client/jsx/components/view_editor/view_editor.jsx
```diff
@@ -0,0 +1,146 @@
...
+  return useSelector(selector, shallowEqual);
+};
+
+
+// Main component for viewing/editing views.
+
+const ViewEditor = ({view_id}) => {
+  let [editing, setEditing] = useState(useSelector(state => state.editing));
+  const dispatch = useDispatch();
+  let [view, setActualView] = useState(null);
```
graft, line 25: elsewhere we have `useActionInvoker` from `etna-js/hooks/useActionInvoker` - should we prefer useDispatch to this?

timur/lib/client/jsx/components/timur_nav.jsx
```diff
@@ -39,6 +39,7 @@ const getTabs = () => ({
   map: Routes.map_path(CONFIG.project_name),
   manifests: Routes.manifests_path(CONFIG.project_name),
   plots: Routes.plots_path(CONFIG.project_name),
+  'view editor': Routes.views_path(CONFIG.project_name),
```
graft, line 42: I would favor calling the tab `views` for consistency. Also since only admins can see it can we somehow suppress the link for non-admins?

coleshaw:
super-nit: in FF, the bottom of the view editor component has no border ... so it blends into the background and looks confusing. It might help to add a bottom border, so that users know the box is scrollable and there's more content?

Screenshot from 2021-02-10 07-51-54

coleshaw (approved):

timur/docker-compose.yml
corps: Nah that's fine.  This is the correct place to setup webpack in dev mode.  We don't use docker-compose in production.  This is all dev setup.

In the past I think the reason I didn't use webpack-dev is that I saw massive spikes in build times due to the symbols and maps being built, but perhaps that depends on the app and webpack version (we have inconsistent versions).

timur/lib/client/jsx/components/document/document_view.jsx
corps: It's an argument for a low scope linter, too, I think.  I'm not a fan of many linters, but it's possible to have a linter that can also _auto correct_ and focus on just the big issues (like shadowing an important global) and get some big value out of it.

timur/lib/client/jsx/components/view_editor/view_editor.jsx
corps: They are similar.  If you checkout the code for `useActionInvoker` it is extremely similar to useDispatch except for adding on piece of functionality:

```
export function useActionInvoker() {
  const { store } = useContext(ReactReduxContext);
  return useCallback((action) => {
    if (typeof action === 'function') {
      return action(store.dispatch);
    }

    return store.dispatch(action);
  }, [store]);
}
```

Not only does it dispatch actions, it also can dispatch the results of functions that wrap taking the dispatcher (which we have scattered in many places), where as I think the default behavior of useDispatch assumes functions are themselves thunks (a reasonable assumption).  I think it'll take some time before we can consolidate our frontend patterns.

timur/lib/client/jsx/components/view_editor/view_editor.jsx
```diff
@@ -0,0 +1,146 @@
...
+    let savedView = { ...view, document: JSON.parse(view.document) };
+    if (view_id == 'new')
+      saveNewView(savedView)(dispatch);
+    else
+      saveView(savedView)(dispatch);
+    if (editing) toggleEdit();
+  };
+
+  const revertView = () => {
+    let {model_name} = view.model_name;
```
corps, line 109: Hmm, this looks like a bug.  Did you mean `let {model_name} = view` ?

timur/lib/client/jsx/components/plotter/plotter.jsx
Gvaihir: Yeh, I forgot to remove it here

timur/lib/client/jsx/components/view_editor/view_editor.jsx
Gvaihir: I tried to use hooks over classes in React and instead of `connect` in Redux.

Gvaihir:
Anything else I should do to get this merged?

coleshaw:
> Anything else I should do to get this merged?

I think @corps pointed out a possible bug in variable declaration / destructuring -- did you check that? Otherwise I think it looks okay to merge.

Gvaihir:
replied to that

PR #218, opened 2021-02-10
Addresses a bug in the bin/seed_janus and parallel code of the README

dtm2451:
Adjusts janus Permit calls to avoid a potential bug, though doesn't address the potential bug.

Not sure I'm able to diagnose ruby code yet, but
For the janus Permit command, the `privileged` input appears meant to be optional, however, when not given, I get this error
```
        2: from ./bin/janus:12:in `<main>'
        1: from /app/vendor/bundle/2.5.7/ruby/2.5.0/bundler/gems/monoetna-725e91bdc5df/lib/etna/application.rb:108:in `run_command'
/app/lib/commands.rb:99:in `execute': undefined method `to_sym' for {}:Hash (NoMethodError)
Did you mean?  to_s
               to_set
```
with traceback to https://github.com/mountetna/monoetna/blob/69c1cee3c6783d75e176c4e53dc8136015fc3c36/janus/lib/commands.rb#L99

This PR simply adds a value of `true` to skirt the issue.


graft:
I would rather set this to `false` as I think it is safer to test features as a non-privileged user generally so that if some bug reveals privileged data improperly you would tend to see it rather than not.

As for the underlying issue, looks like this is overwriting the first non-required command argument (that is, the first argument with a default) with an empty hash:
https://github.com/mountetna/monoetna/blob/69c1cee3c6783d75e176c4e53dc8136015fc3c36/etna/lib/etna/application.rb#L108

There are a few commands that still use default args, and the code is ambiguous about using them - the careful parsing of required parameters suggests they ought to work, but mostly we are now using keyword arguments (which would be passed in as --options), which seems a better and less confusing paradigm than the possibly-missing positional argument.

In this case I think we could remediate by changing `privileged=:false` to `privileged: :false` and probably deleting the `usage` as the auto-generated usage is likely better.

dtm2451:
ahhh, that all makes sense.  This isn't super high priority since all our dev setups should be good now...  

Would that be a quick fix?  And maybe do you or someone else who would be faster than me want to (& have time to) implement it ? I'm quite rough on ruby still, and currently more focused on learning JS(X) for our calc & viz goal! (Perhaps such a true fix should be a separate PR because there's no need for my changes if the underlying bug goes away.)

Alternatively, I'll change the `true`s to `false`s in a sec and we could just shove in this quick fix for now.  And I can reference this PR in a new issue so we don't forget.

dtm2451:
@graft @corps Feel free to answer the above plan by just closing (true fix soon in a separate PR) or merging (this fix now, true fix later with an issue reminder) this PR!

dtm2451:
I imagine this is nolonger needed!

PR #220, opened 2021-02-11
Vulcan plot refactor

coleshaw:
This PR is my initial attempt to mimic the work done in the old PlotConfig / PlotScript classes, except from the standpoint of CWL steps. Because there isn't an explicit user-definition of variables for each series as in the Timur plotting model, we infer or calculate the series information from the consignment itself and some user-defined series data in the CWL. Similarly, layout is calculated from the rendered components.

I'm not sure if these models are too tightly coupled to the UMAP example we're working through...

### DIFF
--- a/vulcan/lib/client/jsx/models/xy_plot.js
+++ b/vulcan/lib/client/jsx/models/xy_plot.js
@@ -1,14 +1,16 @@
+import XYPlot from 'etna-js/plots/components/xy_plot/xy_plot';
 import Plot from './plot';
 import Domain from './domain';
 import Series from './series';
 
 export default class XYPlotModel extends Plot {
-  constructor(step, consignment, dimensions) {
-    super(step, consignment, dimensions);
+  constructor(step, consignment, parentWidth) {
+    super(step, consignment, parentWidth);
     this.type = 'xy';
 
     this.plotObj = {
       plot_type: this.type,
+      component: XYPlot,
       configuration: {
         layout: {},
         variables: {
@@ -26,7 +28,8 @@ export default class XYPlotModel extends Plot {
         configuration: {
           config: {}
         }
-      }
+      },
+      parentWidth: this.parentWidth
     };
     this.dataObj = {
       xdomain: new Domain([0]).asArray,
@@ -138,8 +141,8 @@ export default class XYPlotModel extends Plot {
     // Calculate the plot layout given the dimensions
     //   of a parent container.
     this.plotObj.configuration.layout = {
-      height: Math.round((3 * this.dimensions.width) / 4),
-      width: this.dimensions.width,
+      height: Math.round((3 * this.parentWidth) / 4),
+      width: this.parentWidth,
       margin: {
         bottom: 100,
         left: 100,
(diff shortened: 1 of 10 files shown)


vulcan/lib/client/jsx/models/xy_plot.js
```diff
@@ -0,0 +1,155 @@
...
+    super(step, consignment, parentWidth);
+    this.type = 'xy';
+
+    this.plotObj = {
+      plot_type: this.type,
+      component: XYPlot,
+      configuration: {
+        layout: {},
+        variables: {
+          x_label: 'x', // ??
```
coleshaw, line 17: Does it make sense to label the axes with the column names from the first series?

graft: You might make a plot like (pseudocode) `plot(df = df, x = 'Column 1', y = 'Column 2')`, or you might make it like `plot(x=df['Column1'], y=df['Column2'])` - I think the latter will be more common and will need labels to be passed in.

dtm2451: Still trying to catch up on our code base T_T, but @graft's first option looks very much like how ggplot works. I would be a bigger fan of that first option which has obvious labels + directs to where to grab the data, all in one. I think it would benefit us to have column names generated automatically whenever possible. Seems to me that it would be easy to shove any raw data vectors (option 2) into a df in a step beforehand (-> enables option 1)?

coleshaw: > You might make a plot like (pseudocode) `plot(df = df, x = 'Column 1', y = 'Column 2')`, or you might make it like `plot(x=df['Column1'], y=df['Column2'])` - I think the latter will be more common and will need labels to be passed in.

In the second case, couldn't you make the labels equal to `Column1` and `Column2` so you can extract them from the DF itself and not need to be passed in separately?

For the first option, how do you programmatically know what `Column 1` and `Column 2` are called for a plot, without being told explicitly / separately, or standardizing on first column / second column, in which case you extract them from the DF? Or do you still have to be told the labels explicitly?

dtm2451: I think we're building this so for option 1 we build the plot function know to expect  x/y to be column names of data, but also the labels directly.

Your question, Cole, about option two sounds like turning it into option 1 to me. df would need to be given somehow?

For reference, actual ggplot2:
```
### These would be basically the same
# Option 1: aes() part tells it find these columns in the data,
# to use them as as plot/legend titles,
# and to use an appropriate scale (default scales are built in for ggplot)
# for mapping the raw data to plot values.
ggplot(data = df, mapping = aes(x='column1', y='column2', color='column3')) + ... 

# Option 2: Given outside of aes(), ggplot authors decided they couldn't know as much for building defaulting.
# There would be no legend for this.
# And values & colors are expected to be explicit (so literally '=green, blue, orange'
# but NOT '=1,2,3' as no scale will be used!
ggplot(x=df['column1'], y=df['column2'], color=df['column3']) +
    xlab('column1') + ylab('column2')

coleshaw:
Going to merge this in, since might be useful for custom plotting in the future.

PR #222, opened 2021-02-16
some hotfixes for failed view handling and magma dependencies

coleshaw:
This PR fixes some minor things from my cleaning up / migrating of components to `etna-js`:

1) Some bad request paths weren't behaving properly in the change to `checkStatus`, so I had to update them to expect the body contents instead of the error response.
2) Taking out the carrierwave dependencies broke Magma, because there is assumed ActiveSupport syntax in some of the code (namely `Hash.except`). So putting those back in for now, since I'm not sure how deeply that syntax is used in the code base.

I'd like to get these fixes in before trying to re-deploy Timur, since they are known issues. Then I'd like to redeploy Timur to see if the fixes to Timur browser.jsx fix the Chrome issues folks have been seeing.


graft:
Could we simply maintain the `active-support` dependency in place and remove `carrierwave`? It seems likely impossible to remove all the pernicious ActiveSupport code so maybe it is best to just acknowledge reality and write it in the Gemfile.

PR #223, opened 2021-02-16
update package ref

coleshaw:
Wondering if this will fix the broken push / sync subtrees action, since it seems to be getting stuck on the metis image build?


magma/lib/magma.rb
```diff
@@ -1,4 +1,5 @@
 require 'sequel'
+require 'active_support'
```
coleshaw, line 2: @graft , importing ActiveSupport here but removing carrierwave from the gem, per your feedback on #222.

corps (approved):

PR #225, opened 2021-02-16
Janus app

graft:
Finally eking out this Janus PR. The main bits:

1. Create a React app for Janus along with component versions of all of its UI. I have attempted to avoid using the Redux store as much as possible. Unfortunately, the RootView component in etna-js uses the store to parse user tokens, and user-token parsing mostly happens in Redux selectors, so this remains an entanglement (to be addressed elsewhere). Since Janus is so simple I left out /api and /actions folders, perhaps incorrectly, and put all of the fetch logic into the component itself, where it is surprisingly terse.

2. Moved the router logic into etna-js. There is a surprising amount of discord amongst the various React applications at the "application/ui" level (even disagreeing about how to mount the base component to the DOM); little of this is settled here.

3. Finally removed 'first_name' and 'last_name' from the user models. I think I caught all of the places this is used across the various applications, but stragglers may lurk.

Also I tweaked the icon - looks much nicer at scale with less eye-strain from trying to peer at it.


janus/Dockerfile
```diff
@@ -9,4 +9,4 @@ ARG APP_NAME
 ARG RUN_NPM_INSTALL
 ARG SKIP_RUBY_SETUP=1
 RUN /entrypoints/build.sh
-RUN $(npm bin)/sass lib/client/scss/janus.scss:lib/client/css/janus.css
+RUN npm run build
```
corps, line 12: :+1:

janus/lib/client/jsx/janus.jsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import { Provider } from 'react-redux';
+import Cookies from 'js-cookie';
+
+import createStore from './store';
+import JanusUI from './janus-ui';
+
+function Janus() {
+  this.store = createStore();
```
corps, line 10: Hmm, should this be a const or useState instead of a `this.store` reference?  The `this` here will become the window, so we're putting the store in a global.  It works, but it may be safer to keep it more local.

janus/lib/client/jsx/project-view.jsx
```diff
@@ -0,0 +1,191 @@
...
+const NEWUSER_INPUTS = {
+  select: [ 'role' ],
+  text: [ 'name', 'email' ]
+};
+
+const updateNewUser = function(e) {
+  let select = e.target;
+  let form = select.closest('.item');
+
+  if (inputsChanged(form, NEWUSER_INPUTS)) showForm(form);
```
corps, line 17: Might be missing something -- where does `inputsChanged` come from?

janus/lib/client/jsx/project-view.jsx
```diff
@@ -0,0 +1,191 @@
...
+  else hideForm(form);
+}
+
+const cancelNewUser = function(e) {
+  let select = e.target;
+  let form = select.closest('.item');
+
+  resetInputs(form, NEWUSER_INPUTS);
```
corps, line 25: Also curious about this function, I don't see it defined or imported anywhere.

janus/lib/client/jsx/project-view.jsx
```diff
@@ -0,0 +1,191 @@
...
+      <div className='cell'>Privileged</div>
+      <div className='cell submit'></div>
+    </div>
+    {
+      permissions.length > 10 && <div className='item'>
+        <input className='filter' type='text' placeholder='Filter rows' name='filter' onChange={ e => setFilter(e.target.value) }/>
+      </div>
+    }
+    {
+      permissions.sort((a,b) => (a.role+a.user_email).localeCompare(b.role+b.user_email) ).filter(
```
corps, line 169: might make sense to to the sort / localeCompare up closer to the deconstruction rather than down here.  Will make the view code a little cleaner.

janus/lib/client/jsx/project-view.jsx
```diff
@@ -0,0 +1,191 @@
...
+        p => !filter || [ 'name', 'user_email', 'role', 'affiliation' ].some( column => column in p && p[column].match(new RegExp(filter, 'i')))
```
corps, line 170: Yeah, in general it might be nice to upll some of this heavy data filtering into either a function or a local variable than inside this JSX block.

janus/lib/commands.rb
```diff
@@ -15,12 +15,14 @@ def setup(config)
   end
 
   class AddUser < Etna::Command
-    usage '<email> <first_name> <last_name> [<password>]'
-    def execute(email, first_name, last_name, password=nil)
-      user = User.find_or_create(email: email)
+    usage '<email> <name> [<password>]'
```
corps, line 18: We will also want to update the `seed_janus` script to ensure it only passes in a quoted full name as opposed to separate first and last name arguments.

corps (commented):
Overall a very big step forward, addresses a lot of outstanding janus issues, nice work!

PR #228, opened 2021-02-17
Finalize magmaR for release

dtm2451:
Updates documentation. Simplifies and comments some code. Creates an added `retireveProjects()` function.


etna/packages/magmaR/R/retrieve-special.R
```diff
@@ -82,16 +88,15 @@ retrieveIds <- function(
...
     
-    .retrieve(
-        projectName, modelName, recordNames = rec,
-        attributeNames = "all", format = "tsv",
-        names.only = TRUE, token = token, ...)
+    template <- retrieveTemplate(projectName, token, ...)
+    
+    atts <- names(template$models[[modelName]]$template$attributes)
+    
+    atts[!atts %in% c("created_at", "updated_at")]
```
dtm2451, line 101: I'm not sure if it makes sense to hide (as this line does) versus show these???
These attributes are generally marked 'hidden' and thus don't get /retrieved when attributeNames = "all", nor are they displayed in timur/map, so hiding them makes sense to me.

graft: Can't test right now (🚿) but it looks like they are only marked `hidden` so they should show in a retrieve; the decision to show things called hidden is left to clients. If hidden is really censoring retrieve i might favor relaxing its stringency for these (which are super useful if a bit dodgy being phi-ish).

dtm2451: The output of the deleted code (a retrieve with attributeNames = "all", then gave the rownames) was the same.  magma/retrieve does not give these attributes, at least not for `format = tsv`.

dtm2451: Will probably revert this specific set of code changes anyway cuz looking at the fixtures reminded me that templates can actually be gigantic (because of matrix 'options' especially!) compared to the output of a `retrieve(..., recordNames = IDs[1], attributeNames = "all")`

graft: Yeah the matrix/template thing is dumb and unsustainable, we need a better way to distribute that information that is less cumbersome, I think it really slows down a lot of data retrieval that could be much faster to have these huge templates always being handed down.

corps: @graft I've been thinking one easy fix would to make an optional argument to _not_ return the template in payloads.  That would keep the system completely backwards compatible, but allow new clients to set this argument and avoid the slow painful realizations of the template on every load.

dtm2451: pretty sure @graft and I have discussed that as a possibility at some point too!  As long as the default is `true` / `give the template`, shouldn't break any current code!

dtm2451:
Main thing left here is just:
- [ ] vignette finalization.

Then for CRAN submission:
- [ ] Build the "source" tar.gz tarball of the package
- [ ] Submit this tarball and a short description of the package + justification of why we have .MAGMAR_* environmental variables to CRAN

dtm2451:
Actually ending here probably because the `Rebuild the .MAGMAR_TOKEN / .MAGMAR_URL components into a single object that the user sets` task deserves its own PR which will involve its own major vignette & doc re-works.

PR #230, opened 2021-02-17
Gvaihir/magma py client

Gvaihir:
Python client for magma `magby`. Please first merge the `view editor` branch.
Documentation is not ready - only docstrings in the code


Gvaihir:
Any comments here before I merge?

etna/packages/etna-js/utils/fetch.js
```diff
@@ -57,6 +57,17 @@ export const headers = (...types) => {
   return _headers;
```
coleshaw, line 57: Is this file change really part of this PR?

graft:
Hmm I was waiting for the view editor branch to be merged before commenting so I didn't have to wade through so many commits :)

etna/packages/etna-js/utils/fetch.js
Gvaihir: no-no, sorry I accidentally rebased one of view editor commits on it. I tried to recover but seems easier if I merge view editor to master and them merge master into this branch

coleshaw:
Just two other comments...is this based off of your view-editor branch? Seems like there shouldn't be JS file changes in this PR ... the file changes would look cleaner if you merged in view-editor first and then updated this PR.

Also, it would be nice if you could tie `pytest` into the GitHub actions under `.github/workflows/branches.yml` so they run on commit, just to make sure there are no reversions. I may have missed it if you already did that.

Gvaihir:
Yes, merging in master after VE is in - trying to do it. 
I used `unittest` not `pytest` if that's ok. I'll update workfolws thanks @coleshaw

coleshaw:
> Yes, merging in master after VE is in - trying to do it.
> I used `unittest` not `pytest` if that's ok. I'll update workfolws thanks @coleshaw

Thanks! Yes, whatever test framework you used is great.

dtm2451:
Was also waiting for the simplifying merge, but will review today

etna/packages/magby/tests/test_Magby.py
```diff
@@ -0,0 +1,95 @@
...
+    def test_token(self):
+        self.assertEqual(self.magby.token, token)
+        self.magby.token = 'changed_token'
+        self.assertNotEqual(self.magby.token, token)
+        self.assertEqual(self.magby.token, 'changed_token')
+
+    def test__constructPayload(self):
+        payload = self.magby._constructPayload(projectName='ipi',
+                                               modelName='sample',
+                                               recordNames=["IPIADR001.T1"],
```
dtm2451, line 36: I'm not sure if the seeded development ipi data is real or fake, but for magmaR, we were weary of fixtures creating a route for exposing patient data to the public.  And thus, the example project was born.

I haven't looked too closely at what the ipi seeding code does, but perhaps we could seed with the example project instead and then switch your tests over to that project?

etna/packages/magby/magby/Magby.py
```diff
@@ -0,0 +1,199 @@
...
+                 modelName: Union[List, str],
+                 recordNames: Union[List, str]='all',
+                 attributeNames: Union[List, str]='all',
+                 dataType: str='meta',
+                 filter='',
+                 **kwargs) -> Union[pd.DataFrame, Dict]:
+        '''
+        Method to retrieve data from Magma
+        :param projectName: str. Project name in Magma
+        :param modelName: List[str] or str. Name(s) of model(s) in Magma
```
dtm2451, line 47: Curious what use cases you are imagining that involve multiple models. I don't think I have any in magmaR except for the `retrieveMetadata()` endpoint... Is this just in prep for a potential future endpoint like that?

dtm2451:
Looks good to my fairly python-naive eye aside from the potential use of real data in test fixtures.

etna/packages/magby/tests/test_Magby.py
Gvaihir: Fair, but fixtures expose some random ID's those are not PHI, and no additional data to trace to the individual. I can change, but would prefer to populate my dev environment rather than testing agains prod

etna/packages/magby/magby/Magby.py
Gvaihir: yeh, why not

etna/packages/magby/tests/test_Magby.py
dtm2451: hmm, the matrix fixture is even all `null`s so agree there's maybe not much data here at all... but still a potential route for future tests to expose things.

I'll bring up potential of seeding dev envs with the example project in standup tomorrow.

Gvaihir: Sure, but keep in mind - there are no "future tests" which we don't control. All the data is in fixtures, so users outside the project won't be able to extend testing beyond those fixtures

dtm2451: Right. I was on your side of this before, Anton. Any thoughts @graft?

Gvaihir: I'll test agains the example project. Thanks

Gvaihir: updated fixtures agains Example project in the prod

PR #231, opened 2021-02-18
Archimedes dittoseq

coleshaw:
This PR installs dittoSeq into the Archimedes base image. Be forewarned, it can take awhile to install, so I included in the image instead of using the `build/` hooks.

Tested with:

```
$ make -C archimedes bash
-> R
> library('dittoSeq')
Loading required package: ggplot2
```


dtm2451:
Hmm, when I run `make -C archimedes bash` for the first time, it seems to run through install of the packages.

I'm not sure if the docker image was saved properly after the install?  The code I'd given should skip installation of the `remotes` package if it's around, but that got installed for me too.

coleshaw:
Did you run that command once or multiple? The way to verify would be that any subsequent `make -C archimedes bash` should not re-install remotes or dittoSeq. @dtm2451

dtm2451:
Confirming: 2nd build of the up-to-date branch did not invoke R package reinstalls on my Mac.

coleshaw:
I'm going to merge this, so @dtm2451 if you decide to use `dittoSeq` it'll be available in the workflows.

PR #233, opened 2021-02-18
Add scanpy to Archimedes and scripts to Vulcan for UMAP Workflow

dtm2451:
Not at all ready, just adding the PR now for some visibility and the handy `update branch` button


coleshaw:
I think you'll also want to add an archimedes wrapper method around the dittoseq functions you'll need.

dtm2451:
Hmmm.  that makes sense.  But... my question would be how. We didn't go through the R version of the python __init__.py that I made here, and I'm not seeing anything that looks to me like an R linter.  Not sure if @corps has built that out?

My plan for this PR was to focus on testable calc-side code, which is all in python.  But I could expand it to the R-side Viz bit too if that system is ready now, or will be by the time this might be merged!

coleshaw:
Ah, I was thinking you would use [RPy](http://heather.cs.ucdavis.edu/~matloff/rpy2.html) or just open a [subprocess in Python](https://docs.python.org/3/library/subprocess.html?highlight=popen#subprocess.Popen) to make calls to R, and wrap that into an Archimedes function call just like the `scanpy` stuff. Otherwise, if you were going to call `dittoSeq` directly in the Vulcan scripts, then I've got it installed in the wrong place :smile: 

However, since the Vulcan workflow scripts are also Python, you'd have to use the same libraries to call into R regardless. I guess this comes back to the question of "what belongs in Archimedes" versus "what belongs in Vulcan"? Not sure if @graft  or @corps have opinions in this case ... perhaps depends on how re-usable you think the `dittoSeq` method will be for other workflows, or if it's too UMAP-specific, then it might not make sense to put it into Archimedes. If that's the case and it should be in the Vulcan scripts, just let me know and I'll change which docker image it gets installed into in #231 .

graft:
To be honest I think anything involving using dittoSeq should be out of scope this season - there are serious transport issues involved in getting any amount of data in and out of R, and there is no way that the overhead of solving those issues is worth it to make use of dittoSeq for formatting a scatter plot, which is probably ~ 2 lines of code in python plotly. My vote would be tabling the R integration work until later when we have more bandwidth.

coleshaw:
Oh, also ignore what I said above about moving `dittoSeq` to another container -- I forgot that even the Vulcan scripts will be run in Archimedes containers :facepalm: Still trying to learn how stuff all plugs together...

dtm2451:
@graft, dittoSeq does more than plot generation.  It also handles the grabbing of needed data & turning that all into a plot-lib ready data format, then plot generation.  But the R code would need to be a bit more than a one-liner (each of these would be a separate function call: 1- convert from anndata to SCE, 2- Call dittoSeq to org data and output plotly plot, 3- output that plotly as json or html). I'm not sure if this can work through RPy?

If not I'm with @graft in thinking dittoSeq usage may need to be a question for a future quarter. And for now, we could rely on scanpy's -> matplotlib plot functionality and just target static plot images?  As Cole started here https://github.com/mountetna/monoetna/blob/cs/vulcan-plotly/vulcan/lib/server/workflows/scripts/convert_scanpy_umap_to_plotly.py

coleshaw:
Well, sounds like our goal needs to be coloring by various options in the UI -- how do we achieve that? Is `dittoSeq` the best way (and we can figure out how to get the R scripts running), or matplotlib / plotly in Python?

One question I had from standup was -- how much back and forth between the browser and the server is required for coloring by the various options? Is all the data sent to the UI, and coloring is taken care of in the browser alone, or does info need to be sent back to cells?

dtm2451:
In the cases mentioned above, where we rely on readymade plotting functions of scanpy (or dittoSeq), it would all go back to a cells and that cell would output an image.

dtm2451:
Alternatively, we could alternatively hold the data as a consignment, or another format, in the browser and render plots all via the browser, but as you noted sometime recently, these data get rather large.  Because that size owes to the data directly underlying the color options (expression of ~20k genes, or a handful of metadata, per all cells), I'm not sure it can be avoided without some creative hack(?). But I can't plan toward such a hack myself due to my limited JS tool-base.

coleshaw:
I actually think having it as cells makes it a bit easier, since the UI is dumber and simpler...so coloring done by the server (scanpy / dittoSeq), and not in the browser. And the server step just re-calculates the image based on a new input.

dtm2451:
This route may be slow due to needing to reopen the large file each time in a new cell, but because of output cacheing (done or not I still think that's in scope for the coming weeks?), should be just a 1 time long wait per color-choice.  Then if a user cycles back to a color that's been displayed before, the plot would come more immediately.

graft:
What I was imagining is something like:

1) UMAP calculations result in a cells x (U1, U2) matrix. This only needs computing once and can remain cached.

2) A metadata matrix of cells x (genes, metadata) is calculated. This also only needs computing once.

3) A user input cell selects a column out of the metadata matrix. This may be re-computed as needed. The output is a single vector of length(cells).

4) An output scatter plot production is the final cell, feeding in 1.U1 => x, 1.U2 => y, 3.vector => color. A plotlyjs is the output.

5) The UI downloads the output from (4) since it is an output production. Does this include the actual data? Depends on what plotly allows us to get away with, but ideally it just lets us pass references to external data somehow, and we can separately download the references (x: 1.U1, y: 2.U2, color: 3.vector); once we've done this once for U1 and U2 there is no reason to do it again. However, if plotly requires us to ship the actual data in the json, we'll just have to send the contents of the vectors in the output production. Still, this 3x(numcells) matrix doesn't seem to burdensome to recalculate, cache and download every time the user iterates.

coleshaw:
Thanks for the step by step layout. From my limited experimentation with plotly, I think it sends the data together in the JSON. However, step 4 could just pull the data from 1’s cached result and  combine it with 3?

Another small detail we will have to work out is passing the users token into the Magma query cells...

vulcan/lib/server/workflows/scripts/magma_retrieve_paths.py
```diff
@@ -1,47 +1,47 @@
...
+    if rec.find('POOL') != -1:
         ids_ret = magma.retrieve(
-            projectName: project_name,
-            modelName: seq_pool_model_name,
-            recordNames: id,
-            attributeNames: seq_model_name,
-            dataType: str='meta')
+            projectName= project_name,
+            modelName= seq_pool_model_name,
+            recordNames= '[' + rec + ']',
```
Gvaihir, line 23: This is fixed in the last update, so you won't need to encapsulate a single string into a list, it will recognize data type internally.

Gvaihir: actually not even sure how line 23 would work - you should pass a string or a list not a string with `'[]'`

dtm2451: Oh this section doesn't work yet... I was planning to ask if we could work on it together via zoom at some point.

But 👍 your first comment addresses something that I was also planning to report as a bug!

vulcan/lib/server/workflows/scripts/magma_retrieve_paths.py
```diff
@@ -0,0 +1,47 @@
+from archimedes.functions.dataflow import output_tsv, input_var
+from archimedes.functions.magby import Magby
+
+input_records = input_var('record_ids').split(", ")
+token = input_var('token')
```
coleshaw, line 5: I think this should be an environment variable in the future, too? Like `project_name` and `magma host`

vulcan/lib/server/workflows/scripts/merge_anndata_from_raw_h5.py
```diff
@@ -0,0 +1,30 @@
+from archimedes.functions.dataflow import output_path, input_path
+from archimedes.functions.metis import metis_file_path
+from archimedes import scanpy
+
+magma_output = input_tsv('magma_output')
```
coleshaw, line 5: Do you need to import `input_tsv` from `archimedes.functions.dataflow`?

dtm2451: yup.  And likely need to add `.functions.scanpy` in line 3. This script is untested (only `magma_retrieve_paths.py` has been) and has some pseudo-code lines remaining.

vulcan/lib/server/workflows/scripts/magma_retrieve_paths.py
dtm2451: Agreed.  I moved this around after, but have that noted in the comments in lines 11&12...

How to implement environment variables seems a discussion for the larger group, perhaps in standup?

vulcan/lib/server/workflows/scripts/magma_retrieve_paths.py
dtm2451: Got it working!  Fixed in the latest commit.

dtm2451:
One of the next steps needed will be accessing data from metis via urls.  Some notes on a potential method for that from a quick pairing up with @graft earlier today:
```
# mount with httpfs2.  And have urls that are retrieved via magma mounted.
# They'd be read-only.
# Simple method for any client to be able to just use.
```

dtm2451:
Gonna merge this in today likely so others can keep going while I have to focus elsewhere for a bit!

coleshaw:
I've got some fixes that I'll put into a new branch.

PR #236, opened 2021-02-19
Hotfix timur manifests

coleshaw:
This PR fixes the manifest page on Timur, because the routes changed but the components didn't follow...


graft:
Hmm, looks like some kind of merge-related regression, these were definitely updated in the view editor branch by me. See b0d28b07c5c6a00a59807156d832732cc24902cd

PR #240, opened 2021-02-22
Zc/archimedes local executor qol improvements

corps:
This PR includes fixes to getting the local archimedes script executor (for running and testing individual scripts), as well as the introduction of a workflow executor (running cwls with scripts included in them) from the CLI.

This also demonstrates my idea of using run names to signal to a client UI how to behave.  I'll annotate more details.

NOTE don't pull this branch yet, I'm fixing a couple of other things first.  Let me merge and I'll explain how to use these tools tomorrow morning.


archimedes/archimedes/runner.py
```diff
@@ -184,6 +185,7 @@ class LocalIsolator(Isolator[LocalProcess]):
 
     def is_running(self, t: LocalProcess) -> bool:
         p, io, f = t
+        p.poll()
```
corps, line 188: This fixes the main issue with the local executor, wherein essentially the python side does not update the returncode from the kernel process table without explicit polling here.  Tests were passing due to the short test timeout kicking in, which happened to cause the python process to update the process table, and pass, which is kind of hilarious.

archimedes/archimedes/runner.py
```diff
@@ -213,14 +215,22 @@ def start(self, request: RunRequest, exec_script_path: str) -> LocalProcess:
         os.mkdir(outputs_dir)
 
         for input_file in request.input_files:
-            os.symlink(input_file.host_path, os.path.join(inputs_dir, input_file.logical_name))
-            print(input_file.host_path, os.path.join(inputs_dir, input_file.logical_name))
+            in_path = os.path.abspath(input_file.host_path)
+            isolated_path = os.path.join(inputs_dir, input_file.logical_name)
+            print("Linking input: ", in_path, isolated_path, file=sys.stderr)
```
corps, line 220: Just improving debug output when running with the local executor.

archimedes/docker-compose.yml
```diff
@@ -17,6 +17,7 @@ services:
     - /var/run/docker.sock:/var/run/docker.sock:ro
     - archimedes-pycharm-opt:/opt/.pycharm_helpers
     - archimedes-exec:/archimedes-exec
+    - ../vulcan/lib/server/workflows/scripts:/app/vulcan_scripts
```
corps, line 20: To make it easier to develop scripts in archimedes for vulcan, I mount the vulcan scripts directory as /app/vulcan_scripts.

vulcan/lib/commands.rb
```diff
@@ -18,6 +19,174 @@ def setup(config)
...
+    include Etna::CommandExecutor
+
+    class Run
+      def execute(project_name, workflow_name)
+        @project_name = project_name
+        @workflow_name = workflow_name
+
+        while true
+          orchestration.run_until_done!(storage)
+          find_all_output_ui_queries.each do |query_name, source, path|
```
corps, line 32: So, this was just one novel interpretation of how to handle UI displaying outputs.  Essentially,

1.  primary outputs are mainly used for composing workflows together.  primary outputs of one workflow can be used as step outputs in another workflow in CWL.  Unfortunately that means that primary outputs don't get their own run specification and cannot explicitly define a ui query.  In this case, I assign primary outputs a default ui_query based on the type, although that does bring up the original point; does the frontend need to map types to outputs?  Maybe, but I'd like to avoid that.  Maybe we don't show primary outputs without an explicit cell anyways, thus allowing a workflow to both be a data pipeline as well as an interactive workflow -- the interactive inputs and outputs only get triggered, essentially, on a client that specifically handles them.  :shrug:  There are several options, but this one is simple.

2.  explicit steps with a run of `ui-queries/XXX` that themselves have no outputs, are considered UI components that display their inputs.  I encode that here as simply using `send` to dynamically dispatch output based on the ui-queries name.

vulcan/lib/commands.rb
```diff
@@ -18,6 +19,174 @@ def setup(config)
...
+            if self.respond_to?(query_name.to_sym)
+              self.send(query_name.to_sym, source, path)
+            else
+              show_data(source, path)
+            end
+          end
+
+          while (input = find_next_primary_input)
```
corps, line 40: Primary inputs, like primary outputs, are definitely the special case of having types but no run information encoded.  Perhaps we could provide type information and use that as the determinate of gathering inputs, but I still have skepticism if that is wise.  In any case, for primary inputs, we definitely only have the type to go off of.

vulcan/lib/commands.rb
```diff
@@ -18,6 +19,174 @@ def setup(config)
...
+            val = collect_primary_input(input.id, input.type)
+            session.define_user_input([:primary_inputs, input.id], val)
+          end
+
+          while (ui_query = find_next_input_ui_query)
```
corps, line 45: And these are actual ui-query steps that can use the run to encode a specific 'component'.  Like before, I use the query name to dynamically dispatch handling how to query data from the json, but otherwise fall back to a generic "give me a json input" workflow.

corps:
My main conclusions:

1.  We probably do want type information for the frontend to fallback to when decide how to display or input data.  Currently, *we do not have this for steps* because normally that would be defined in the backing cwl (but for us, the cwls are 'fake').  In the future, we will materialize the "fake" cwls based on the script and infer types, but that does not exist yet.
2.  Still, I think there are many cases where we want to do more than just "a default output based on type".  Remember that cwl's typing system is really... naive.  It doesn't include nominal typing or parameterized typing (so you can't have a type that is an array of gene names, you could only have an array type) or any way to extensively "signal" the different types of frontend inputs you likely will want.  I think the best approach is to first attempt to use the ui-queries run to select a specialized component if available, and fall back to a default based on type that is completely ignorant of any specialization when all else fails.  This last prt is still only possible once #1 is solved in terms of inferring types and providing them for non primary inputs and non primary outputs.

archimedes/docker-compose.yml
dtm2451: I may be putting my scripts in the wrong place... `vulcan/spec/fixtures/scripts`. Wasn't clear to me where they should go, but the one you reference here organizes within workflows. I picked the one I did because I thought eventually we foresee some scripts being used for more than just one workflow, and by-workflow organization wouldn't make sense then.  @coleshaw ?

corps: @dtm2451 let's discuss this tomorrow.

corps:
I will demo this tomorrow @9am and explain in detail.

archimedes/docker-compose.yml
coleshaw: Yeah, let's talk about it during Zach's run-through, I think that'll help. Also, @dtm2451 , just as an FYI, the `spec/` directory in our projects houses the Ruby tests, so any scripts or fixtures you find there are used for testing.

No worries where you've put the scripts for now, we can always move things around!

PR #244, opened 2021-02-24
Provide an MD5 as Content header in download controller

corps:


coleshaw:
Is the idea that you would start the download, but would then cancel it if you just wanted the md5sum, as in @Gvaihir 's use case?

coleshaw (approved):

corps:
> Is the idea that you would start the download, but would then cancel it if you just wanted the md5sum, as in @Gvaihir 's use case?

Yes.  Download itself is managed by apache so this doesn't overly bother our ruby processes.  Http/tcp is buffered anyways, so large files wouldn't see much consumption as long as the body isn't actually read from on the client.  An ideal use case would be an explicit HEAD request type and handling, but that would involve a lot further work.  I suspect it's reasonable to not worry though as long as apache is the backing system for files.

graft:
The HEAD request could probably be handled by Etna::Route, which ought
to wrap all requests in Etna::Route#call and could swap HEAD for GET
to match the route, then dump the body after the request returns?

Gvaihir (approved):
We approve thine changesth.

Gvaihir:
Please merge soon if all is good. Need this functionality

PR #247, opened 2021-02-25
Vulcan error handling

coleshaw:
This PR updates Vulcan orchestration to capture any errors thrown by individual cells, instead of allowing them to bubble up to the controller. In turn, errors are reported in the `status` Hash that is returned by the controller, and therefore a `status` and `session` are always returned to the UI. Steps with errors are then rendered in the main feed and include the Archimedes-provided stacktrace.


vulcan/lib/orchestration.rb
```diff
@@ -18,8 +20,13 @@ def run_until_done!(storage)
       [].tap do |ran|
         i = 0
         until (runnables = next_runnable_build_targets(storage)).empty? || (i += 1) > MAX_RUNNABLE
-          run!(storage: storage, build_target: runnables.first)
-          ran << runnables.first
+          begin
+            run!(storage: storage, build_target: runnables.first)
```
corps, line 24: Hmm... I gotta say that I don't like this approach at all.  This approach would have the runner keep executing in the face of errors, which likely will just result in it trying the same cell over and over until i > MAX_RUNNABLE.  Furthermore, I feel like catching errors here is the wrong place to handle.  How does a caller to `run_until_done!` know if something went wrong?  Placing it in an errors hash seems wrong as well.  Who clears the errors hash?  How do you know if this run was safe but the previous wasn't?  I don't like this approach at all I think.

corps: If the idea is that you want to attempt to run _other_ nodes before raising an exception, I'd recommend iterating the runnables value (it is an array itself), trying to run each of those, collecting errors, and then immediately throwing possibly a new error type that collects inner errors by step.

corps (changes requested):
I understand the desire to report back 'error' state rather than just returning 500s right now, but I don't think I like this approach.  Better would be

1.  nothing stateful in orchestrator
2.  still throw an exception -- but that exception can contain additional state (such as moving the build_target_has_error? logic into that)
3.  Not changing the next_runnable_build_targets logic.  Instead, just iterate the runnables and attempt steps that can be run, then immediately fail once no more progress can be made.

corps:
Basically -- I'm ok with any solution where the state of errors is local to the call, not the class.

vulcan/lib/orchestration.rb
coleshaw: Thanks for the suggestions! Okay, I'll rework this. The idea was that users would be able to see exactly which step(s) failed and why, but with the previous behavior, that was impossible -- the 500 didn't report back the step name or any related context, just a stacktrace without an identifiable relation to a step (i.e. `script.py` doesn't correlate to a specific step). It also seemed undesirable from a debugging standpoint if multiple steps would have failed (i.e. parallel steps), but having to iterate through them one-at-a-time.

Also, I thought the overall philosophy we were following was for the UI to be thin / simple, so thus the `status` response would encapsulate all the information about each step? But if it's preferable to have the controller and orchestrator throw exceptions / 500's instead of updating the status elements, and leave the UI to figure out how that information ties together with the "good" status, I'll just re-think how to present it.

corps: absolutely.  I agree with the direction of this pr, only the implementation am I concerned about it.  I think we can get the same exact effect, but keep the state local, and return it back to the controller caller, who can unwrap it.

coleshaw: @corps , could you take another look and see if the updated approach works better?

vulcan/lib/orchestration.rb
```diff
@@ -16,11 +16,21 @@ def initialize(workflow, session)
...
       [].tap do |ran|
+        errors = RunErrors.new
         i = 0
-        until (runnables = next_runnable_build_targets(storage)).empty? || (i += 1) > MAX_RUNNABLE
-          run!(storage: storage, build_target: runnables.first)
-          ran << runnables.first
+        num_cells_successful = 1
+        until (runnables = next_runnable_build_targets(storage)).empty? || (i += 1) > MAX_RUNNABLE || num_cells_successful == 0
+          begin
+            num_cells_successful = 0
```
corps, line 24: on other suggestion that might help:

```
next_runnable = runnables.find { |r| !errors.include?(r) }
break if next_runnable.nil?
```

corps: This would allow the loop to natural keep running as long as there is some non errored runnable in a simple way I think.

vulcan/lib/server/controllers/sessions_controller.rb
```diff
@@ -16,35 +16,47 @@ def submit
       raise Etna::NotFound.new("Workflow by the name #{session.workflow_name} could not be found.")
     end
 
-    orchestration.run_until_done!(storage)
+    begin
+      orchestration.run_until_done!(storage)
+    rescue Vulcan::Orchestration::RunErrors => e
+      run_errors = e
```
corps, line 22: Nice.  Yeah, admittingly this does feel odd in some sense.  One style would argue that you could always return (rather than throw an exception) a run errors, and an empty one when it happens there are no errors.  That would atleast prevent the nil vs non nil case.  But one could also argue that throwing an exception ensures that code that is written without explicit error handling is forced to see the failure.  I said use exceptions before so this is criticism of my own thinking lol.  Anyways, I don't care too much either way, just a style note.

corps (approved):
Yeah, I appreciate, this is much better.  I made one suggestion that might? improve the loop, but it's your call if you want to address or when.  I approve this either way.

PR #248, opened 2021-02-25
Query md5s

graft:
A bit heftier than @corps's slim one-liner :sweat_smile:, here is an attempt to provide an ::md5 querying interface via Magma. E.g., from the specs, you may post to `/query` : `{ project_name: 'labors', query: [ 'monster', '::all', 'stats', '::md5' ] }` and get back a list of MD5s from Metis, with Magma doing the work for you.

My main interest here however was making it easier for applications to hmac-sign requests to each other, which is something I believe I will need for finishing long-lived token verification (in progress), and seems a useful tool in general. I think I made it easier to implement hmac-signing on the client-application-side, but it remains excruciating to test.


etna/lib/etna/client.rb
```diff
@@ -17,6 +17,29 @@ def initialize(host, token, routes_available: true, ignore_ssl: false)
 
     attr_reader :routes
 
+    def signed_route_path(route, params)
```
corps, line 20: Veeery minor, but would it be possible to make the expiry a variable with default here?
I can imagine us wanting to generate signed route paths and pass them off to a service or hook where the turn around could be a bit longer.  We could still restrict the expiry to a max, say, 5 minutes?  It's also not important to address in this PR, just a thought.

magma/lib/magma/attributes/file_copier.rb
```diff
@@ -15,49 +15,12 @@ def bulk_copy_files
...
-
-      # At some point, when Metis supports changing project names,
-      # this parameter should be the old file project name (metis_file_location_parts[2]))
-      # and the new project name in the HMAC headers should
-      # be project_name
-
-      path = client.route_path(
-        bulk_copy_route,
-        project_name: project_name
+      client.file_bulk_copy(
```
corps, line 18: nice.

coleshaw:
Hm, so I'm getting an exception when I try to run this against patient / flojo_file locally:

```
magma_app_1      | ERROR:2021-02-25T20:29:17+00:00 ae02bi Caught unspecified error
magma_app_1      | ERROR:2021-02-25T20:29:17+00:00 ae02bi undefined method `[]' for nil:NilClass
magma_app_1      | ERROR:2021-02-25T20:29:17+00:00 ae02bi /app/lib/magma/query/predicate/file.rb:36:in `block (2 levels) in <class:FilePredicate>'
magma_app_1      | ERROR:2021-02-25T20:29:17+00:00 ae02bi /app/lib/magma/query/verb.rb:83:in `instance_exec'
magma_app_1      | ERROR:2021-02-25T20:29:17+00:00 ae02bi /app/lib/magma/query/verb.rb:83:in `get_extract'
magma_app_1      | ERROR:2021-02-25T20:29:17+00:00 ae02bi /app/lib/magma/query/verb.rb:12:in `do'
magma_app_1      | ERROR:2021-02-25T20:29:17+00:00 ae02bi /app/lib/magma/query/predicate/column.rb:20:in `extract'
magma_app_1      | ERROR:2021-02-25T20:29:17+00:00 ae02bi /app/lib/magma/query/predicate.rb:92:in `child_extract'
magma_app_1      | ERROR:2021-02-25T20:29:17+00:00 ae02bi /app/lib/magma/query/predicate.rb:87:in `extract'
```

Perhaps something to do with empty entries for that attribute, for some records? I'll do some more poking around...

etna/lib/etna/application.rb
```diff
@@ -91,6 +91,10 @@ def environment
     (ENV["#{self.class.name.upcase}_ENV"] || :development).to_sym
   end
 
+  def id
+    self.class.name.snake_case.split(/::/).last
```
corps, line 95: Hmm.  There is also the `APP_NAME` environment variable that I set in all application environments.  Would it make sense to prefer that value if set?

corps:
Btw I am very excited about hmac signing.  I'm setting up some monitoring tooling with the docker rollout that could use signing as the main authentication mechanism.

etna/lib/etna/client.rb
graft: Hmm, since the approval happens right away (the message will be signed just before the request is sent) it actually should not matter how long the request is, and even a 10s expiration time is probably too much.

etna/lib/etna/application.rb
graft: > md5s can also be stub values and nil in metis, as they are calculated asynchronously by a background job. Does this PR test against those cases?

I didn't test against those cases, but the method is very lazy and will give up and say "null" if metis doesn't know the file exists (common in dev). If metis hasn't hashed the file yet you get back a `temp-` hash, which seems like the right thing to do.

graft: Err, to be clear I didn't test in specs, I tested it against my dev instance.

coleshaw:
@graft , did you test this against `File` entries with `::blank` as the filename? I think that's what is killing the MD5 part when I test locally.

coleshaw:
> @graft , did you test this against `File` entries with `::blank` as the filename? I think that's what is killing the MD5 part when I test locally.

Hm, maybe not...more digging I guess.

coleshaw (approved):
Latest version works for me!

graft:
Hmm, i didn't consider ::blank, but it should fall out because metis will
ignore it.

PR #249, opened 2021-02-25
Vulcan now returns a linear, serialized path as the first result

corps:
As per @coleshaw request, vulcan now returns, as the first path in workflows and sessions output, a single linear path that traverses all nodes in one ordering that ensures all dependencies occur before and convergent graph points.  Worth noting, in the future when we have async processes, there is no guarantee that downloads are available in a strictly linear fashion -- divergent paths will execute in a non deterministic order and thus this 'flattened' graph cannot plan for the true ordering.

However, it is sufficient, for the sake of gathering  input from the user in a linear fashion, to find the longest path where all previous steps are complete, to 'estimate' the progress of divergent parallel paths.

### DIFF
--- a/docker/etna-apache/bin/httpd_with_shibd.sh
+++ b/docker/etna-apache/bin/httpd_with_shibd.sh
@@ -3,5 +3,5 @@
 set -e
 
 /etc/init.d/shibd start
-httpd -DFOREGROUND
+exec httpd -DFOREGROUND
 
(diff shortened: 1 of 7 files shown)


docker/etna-apache/bin/httpd_with_shibd.sh
```diff
@@ -3,5 +3,5 @@
 set -e
 
 /etc/init.d/shibd start
-httpd -DFOREGROUND
+exec httpd -DFOREGROUND
```
corps, line 6: Incidental change I need to fix the janus shutdown issue with watchtower.

coleshaw (approved):
Thanks! Looks great! Better than me just flattening in the UI...

PR #250, opened 2021-02-26
Vulcan plotly output component

coleshaw:
This PR adds a `ui-outputs` option for UI "sinks" -- output components that have no output and just inputs, like plots. There is also a corresponding UI component when the CWL step's run command is defined as `ui-outputs/plotly.cwl`.

Adding this explicit outputs directory (instead of re-using `ui-queries`) helps make it simpler to differentiate behavior in the status response and in the UI code.

This depends on #247 and should be merged afterwards.


archimedes/archimedes/functions/dataflow/__init__.py
```diff
@@ -1 +1,2 @@
-from .path import input_path, output_path
\ No newline at end of file
+from .path import input_path, output_path
+import json
```
coleshaw, line 2: Importing `json` to mock out data for the plots, we can remove this later if it isn't really needed for the workflows.

vulcan/lib/orchestration.rb
```diff
@@ -220,16 +232,18 @@ def build_target_for(step_name, cache = {})
         elsif (step = workflow.find_step(step_name))
           if step.ui_query_name
             script = {}
+          elsif step.ui_output_name
+            script = {}
```
coleshaw, line 236: @corps , I had to add cases for the `ui-outputs` directory for output scripts in `Orchestration`. Not sure if these were the right places to account for the new case, but it seems to work correctly?

vulcan/lib/server/controllers/sessions_controller.rb
```diff
@@ -41,23 +42,44 @@ def orchestration_status(orchestration, run_errors)
...
         [
             output_name,
             storage.data_url(project_name: sf.project_name, cell_hash: sf.cell_hash, data_filename: sf.data_filename),
         ]
       end.to_h : nil
     }
   end
+
+  def step_status(bt, ui_output)
+    # UI Sinks (plot, download data, etc), will not have outputs
```
coleshaw, line 69: Note that because UI output steps have no actual outputs, status has to be calculated differently.

vulcan/lib/client/jsx/api/vulcan.js
```diff
@@ -119,6 +120,13 @@ export const submit = (context) => {
     })
     .then((extractions) => {
       extractions.forEach((datum, index) => {
+        // We don't currently have accurate Content-Type headers
```
corps, line 123: :+1: 

I will get to work on content-type headers on the backend once I return to working fulltime on vulcan proper.  Thank you for being flexible.

vulcan/lib/client/jsx/components/workflow/steps/step.jsx
```diff
@@ -6,7 +6,7 @@ import StepName from './step_name';
 export default function Step({step, index, onClick}) {
   const {status, pathIndex} = useContext(VulcanContext);
 
-  if (!status || !pathIndex) return null;
+  if (!status || null === pathIndex) return null;
```
corps, line 9: minor style nit, but what about `pathIndex == null` ?  Double equals null is safer because it catches both undefined and null.

vulcan/lib/orchestration.rb
corps: Yup, this works for me.  It ponders me if atleast on the backend we should have another method `ui_behavior?` which simply returns the boolean of wether ui_output_name or ui_query_name are set to make these conditionals simpler.

vulcan/lib/server/controllers/sessions_controller.rb
corps: That is a good point.  Yeah, I'm ok with this.

corps (approved):
:+1: Seems like the right direction.

PR #252, opened 2021-02-26
suppress templates if asked

graft:
Cf. Zach's remark in slack, the ability to not return a template if you so desire.


coleshaw (approved):

dtm2451 (approved):

dtm2451:
Just double checking my understanding... I'll test it out & add explicitly into magmaR later.

It works via a new `hide_templates` input which defaults to `nil`? What happens if it is given explicitly as `false`?

graft:
> It works via a new `hide_templates` input which defaults to `nil`? What happens if it is given explicitly as `false`?

`@hide_templates = !!@params[:hide_templates]` converts the input to `true` or `false`, so it should be equivalent to leaving it out.

dtm2451:
I'll ping again once I have a chance to test!

dtm2451:
Just kidding lol... I can't test in production/staging until it's deployed, and I can't actually use magmaR with my dev setup until after I complete part of my magmaR PR.  Deploy whenever as far as I'm concerned!

PR #254, opened 2021-03-01
Give the upload webworker it's own fetch utils

coleshaw:
To prevent issues with the upload webworker relying on unnecessary global variables like `document` and `CONFIG` that are used in the shared `etna-js/utils/fetch` methods, we split out the required upload methods into their own `etna-js/upload/utils/upload_fetch` module that don't use those globals.

This works for me locally in Metis and Timur.


graft:
It might make sense to go the other way - to have a base utility that doesn't use globals, which the uploader can inherit, and another utility that does use globals, which everyone else can inherit and which inherits from the base utility - this might avoid some code duplication.

graft:
This no longer seems to require CONFIG.baseUrl, dunno what happened to that integration test. Maybe i dreamed it?

coleshaw:
Hm, when I do a `grep` on `monoetna`, I don't see `baseUrl` in any of our code ... was it something on one of your branches? Or do you recall specifically which test was involved?

graft:
https://github.com/mountetna/monoetna/blob/d975fee21232840e05071eaa4123b436c5550e49/janus/jest.config.js#L5

coleshaw:
Oh, `baseURL` not `baseUrl`? Ha, interesting...adding that line to the fetch wouldn't have made a test pass anyways, the casing was wrong...

coleshaw:
Oh, I see, it fails on `npm test` because `node_fetch` (not browser `fetch`) expects an absolute URL, not just the path. It passes CI because `make -C janus release` doesn't currently run node tests, it just runs Ruby tests. You have to add `include ../make-base/node.mk` to the Janus  Makefile in order to run the new UI tests as part of CI.

coleshaw:
I'll post a PR for this soon.

PR #258, opened 2021-03-02
Etna gem json attributes

coleshaw:
This PR is a small quality of life change, mostly for my workflow when updating Magma records with a combination of attributes that include File / Image / FileCollection attributes as well as non-JSON attributes in a single CSV. To avoid having to escape non-JSON attribute values (which required string values be enclosed by triple double-quotes, `"""new value"""`), this change will detect the three attribute types mentioned above and automatically parse those cell values as JSON, when you don't provide the `--json-values` flag. This change shouldn't affect the previous behavior, in case anyone had built scripts that did the extra escaping.

This also changes the workflow to always call the Magma update route with a JSON payload instead of a form. Since we don't upload large binary blobs directly to Magma anymore, I think this is an okay change.


corps:
Hahaha, yeah.  I love CSV. 

`"""string"""` -> the thing that quotes string data is used also for boundaries to quote spaces which is also the thing that typically quotes data which itself is a quote which is escaped with, a quote.

corps (approved):

corps:
Apologies for this conflict... soon dude, all of this will go away.

PR #260, opened 2021-03-05
Vulcan UI tweaks

coleshaw:
This PR adds the following small changes:

1. Several Archimedes imports for plotly, pandas, and numpy. These are just to support the random scatterplot example and can / should be removed once we move to real workflows.
2. A UI select-widget with typeahead (use `ui-queries/select-autocomplete.cwl` to trigger this in your CWL). This acts like a select box with a text input at the top, so as you type the options are narrowed down to match what you type. This should be useful for the color-by options, especially genes, where there may be a lot of options.
3. Stream results data back (see comments in code). This is to work around default Firefox timeouts that break longer-running workflows (> 10 seconds).
4.  Removes the modal in favor of a rotating clock for pending steps.
5. Adds in the "throb" animation for the Vulcan icon when Archimedes is running.
6. Adds a Link to download raw data as an output step (use `ui-outputs/link.cwl` to trigger this in your CWL). This is also the default behavior if an unknown output widget is specified.
7. Auto-runs the workflow when the page loads, to eagerly run as many steps as possible.
8. Removes the unused google fonts for faster page loading.


archimedes/archimedes/functions/matrix/__init__.py
```diff
@@ -1,4 +1,5 @@
 from pandas import Series
+import pandas as pd
```
coleshaw, line 2: These Archimedes imports should all be removed once the demo UMAP workflow is replaced.

edge-apache/development-edge-httpd.conf
```diff
@@ -66,6 +66,8 @@ LoadModule ssl_module modules/mod_ssl.so
 
 
 <VirtualHost *:443>
+    TimeOut 300
```
coleshaw, line 69: The default Apache 2.4 timeout is 60 seconds, which can kill long-running workflows. We set it to 300 seconds here, but we may need to tweak that depending on how long the actual UMAP stuff runs. Note that there is also the same setting that should be adjusted for the `app_fe` containers.

vulcan/lib/server/controllers/sessions_controller.rb
```diff
@@ -11,6 +11,65 @@ def session
   end
 
   def submit
+    if can_hijack?
```
coleshaw, line 14: This probably needs a bit of scrutiny. As noted in the comments, apparently Firefox has very strict default timeouts for active connections (without streaming, a long workflow ran fine with curl, Postman, and Chrome for me, so it is a browser-specific issue). To work around Firefox automatically aborting the POST submit request, we do two things:

1. Immediately return headers stating a 200 JSON response.
2. Stream a null byte back every 5 seconds. This keeps the connection alive and Firefox happy, and also does not impact JSON-parsing the response on the client-side.

graft: Sounds like https://en.m.wikipedia.org/wiki/Comet_(programming), should be plenty of libs to support this.

coleshaw: When we move to an async model, that should also remove the problem...in addition to just a general "comet"-type philosophy, I think another of our issues is that it can take more than 10 seconds to send back data since running the cells can take awhile. Since we're not sending individual cell results back as they finish, it looks like a hung connection. We still need to stream "filler, keep-alive data" back, whether the protocol is SSE or WebSocket or anything similar.

There are Rails / ActiveSupport libraries that support streaming SSE back like this, this implementation is just a bit closer to the metal and bare Rack interface since client-side we're expecting simple JSON at the moment.

PR #264, opened 2021-03-11
UMAP integration

coleshaw:
This PR presents a basic integration of @dtm2451  and @graft 's UMAP scripts with the UI. Woo hoo! Some issues that still need to be addressed:

* I was trying to optimize the front-end and broke the LINK output. I need to fix this.
* @corps has pointed out some better, cleaner refactorings for the Vulcan - Archimedes with the DockerIsolator that would make this a lot cleaner to use in development:
  * Bake Magby into the `archimedes-base` image instead of mounting it
  * Update our dev configs to refer to `magma_app` instead of `https://magma.development.local` for Magma and Metis hosts in Vulcan's `config.yml`. However, this requires Magma's `config.yml` to also update the storage host to `metis_app` instead of `metis.development.local`, which then breaks the HMAC URL calculation. This step would clean up a lot of the SSL verify issues with Magby and curl in the UMAP scripts.

Also issues from standup last week:

* Splash screen that focuses on workflow instead of project.
* Option to pick different workflows.

Putting this PR out so that we can divide up further work on a functional UMAP workflow.


vulcan/docker-compose.yml
```diff
@@ -1,21 +1,21 @@
-
 version: '3.4'
 
 x-base-etna-service:
   &base-etna-service # Base configuration used by most etna services.
   image: etna-base-dev
   volumes: &base-etna-service-volumes
     - .:/app:delegated
+    - /var/run/docker.sock:/var/run/docker.sock:ro
```
corps, line 8: Yeah, the loss of this was definitely a merge bug, my apologies.

vulcan/lib/orchestration.rb
```diff
@@ -49,7 +49,10 @@ def load_json_inputs!(storage)
     # bad inputs.  Validating here would start to conflate potential storage or logic errors with simple user input
     # errors.
     def load_json_payload!(storage, reference)
-      if (payload = reference[:json_payload])
+      if (reference.key?(:json_payload))
```
corps, line 52: I'm actually a little curious what `false files` are?

vulcan/lib/orchestration.rb
```diff
@@ -69,12 +86,20 @@ def command(storage:, input_files:, output_files:)
...
+          "/etna/packages/magby/magby/:/app/magby/:ro",
+          "-v",
           "#{Vulcan.instance.config(:archimedes_exec_volume)}:/archimedes-exec",
           Vulcan.instance.config(:archimedes_run_image),
           "poetry",
           "run",
           "archimedes-run",
-          "--isolator=docker",
+          "--isolator=docker"
+      ] + (is_dev? ? dev_options : []) + [
```
corps, line 97: this goes away once we take the approach mentioned before with baing magby into the image.

vulcan/lib/orchestration.rb
coleshaw: There are default boolean values that can be set to `false`. When that's the case, the old logic would evaluate `payload` as `false` and not generate a file with that value written to it...

vulcan/lib/orchestration.rb
coleshaw: Yeah, that would be great to get rid of the `--local-packages` flag. I think most of the `dev_options` stuff would go away once we fix the Magma HMAC stuff, too. Perhaps only adding it to the `monoetna_edge_net` network would be required at that point?

vulcan/lib/orchestration.rb
coleshaw: Which would then mean the orchestrator would never consider those steps as ready to run, because they were missing input files :smile:

dtm2451: Is this regarding `regress_<cell-feature>` inputs?  If so, the files would be a string recording either "True" or "False" as UI-inputs are transferred from vulcan to archimedes as files.

coleshaw: Is, that's how the issue showed up. Hm, so you might have to update the script to do some Python manipulation of the string value instead of just reading it in an casting it to a bool, because I'm pretty sure it will be recorded as `false` and `true` since Ruby uses lowercase...

coleshaw: In which case, casting it to `bool` in Python might cast both values as `True`

etna/packages/magby/magby/Magby.py
```diff
@@ -7,14 +7,16 @@
 _session = Session()
 
 class Magby(object):
-    def __init__(self, url: str, token: str):
+    def __init__(self, url: str, token: str, verify: bool = True):
```
coleshaw, line 10: @Gvaihir unfortunately I had to edit the Magby constructor slightly to get the UMAP Vulcan work functional in dev. This should go away once we clean up the docker networking issues, but that will require a bit of work on the Magma app.

vulcan/lib/orchestration.rb
dtm2451: It did, and so I added an interpreter function that ignores casing and calls yes/true/y/1 -> True, else False. (We can update/trim that.)

PR #266, opened 2021-03-11
Bake magby into archimedes image.

corps:
Simplification by building magby into the archimedes image directly.  Future other libraries can also be included in a similar fashion.  This ensures no struggling with extra mounts.

### DIFF
--- a/archimedes/docker-compose.yml
+++ b/archimedes/docker-compose.yml
@@ -26,7 +26,7 @@ services:
       - archimedes-pycharm-opt:/opt/.pycharm_helpers
       - archimedes-exec:/archimedes-exec
       - ../vulcan/lib/server/workflows/scripts:/app/vulcan_scripts
-      - ../etna/packages/magby/magby:/app/magby/
+      - ../etna/packages/magby/:/magby/
     environment:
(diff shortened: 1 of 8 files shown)


archimedes/docker-compose.yml
```diff
@@ -26,7 +26,7 @@ services:
       - archimedes-pycharm-opt:/opt/.pycharm_helpers
       - archimedes-exec:/archimedes-exec
       - ../vulcan/lib/server/workflows/scripts:/app/vulcan_scripts
-      - ../etna/packages/magby/magby:/app/magby/
+      - ../etna/packages/magby/:/magby/
```
dtm2451, line 29: Should the second `/magby/magby:` here be retained?

perhaps no because the package is being "properly built and installed" now rather than imported as a folder?  Does the `archimedes/archimedes/functions/magby/__init__.py` need update for this?

corps: Noppe, because this is now properly built and installed, it should use the top level directory.
The import works as is, `from magby import Magby` will look at the python path, which will now include the dependency's path from poetry's installation.

PR #267, opened 2021-03-12
trying to fix rbuf_fill error

coleshaw:
When using the etna gem against REDCap, we occassionally run into this exception:

```
2.7.2/lib/ruby/2.7.0/net/protocol.rb:225:in `rbuf_fill’: end of file reached (EOFError)
```

Unsure if it's related to [this post](https://stackoverflow.com/a/9227933), so trying to re-explicitly set the `use_ssl` flag to `true` and see if it re-occurs.

I'm unsure if the issue is also just a long connection with REDCap itself causes this, i.e. the form is huge or there are many forms, so the connection from the gem to polyphemus somehow dies? I did see this error as well when we were using net-http-persistent, but that shouldn'thave been purged everywhere...

### DIFF
--- a/etna/etna.gemspec
+++ b/etna/etna.gemspec
@@ -1,6 +1,6 @@
 Gem::Specification.new do |spec|
   spec.name              = 'etna'
-  spec.version           = '0.1.30'
+  spec.version           = '0.1.31'
   spec.summary           = 'Base classes for Mount Etna applications'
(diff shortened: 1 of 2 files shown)


coleshaw:
Bah, I'm thinking we're actually running into timing issues, perhaps should just refactor the query code for fewer requests.

PR #268, opened 2021-03-12
send redcap updates individually instead of bulk

coleshaw:
This PR updates the polyphemus REDCap loading job to send individual update requests to Magma instead of a single bulk request. Seems more stable, but certainly slower to load. We could probably also chunk it into some range, like 10 records at a time, if performance becomes an issue.


corps (approved):

coleshaw:
Sorry @bsamad , I'll work on performance optimization later, this is super slow for a lot of records but I want to deploy the comorbidity updates, too, so merging this in.

PR #269, opened 2021-03-12
Upgrade puma, do not query all attributes from polyphemus job.

corps:
This should, at the very least, make the issue go away for now.  Figuring out why certain queries cause the ruby process to spin out is important, but it will take some more instrumentation to narrow down.  For now, I just want the existing parts to work better.


coleshaw (approved):
:shipit:

PR #270, opened 2021-03-15
Update umap plotly

graft:
A WIP, this "should work", but does not. I.e., I think I successfully aped the example Cole included, but there is no attempt to render.

The main issue I have been having, however, is with rebuilding the archimedes image, which seems to be required whenever I do much involving archimedes (like update any python imports). However, the build takes a loooong time (hours), so long that I have never had the patience to watch it finish. This, naturally, makes development glacial.

The build looks like it gets through most of step 7/10:
```
Step 7/10 : RUN /entrypoints/build.sh                                                                                                                                       
 ---> Running in c0fcde5ffd54                                                                                                                                               
+ shopt -s globstar                                                                                                                                                         
+ '[' -z '' ']'                                                                                                                                                             
+ poetry install                                                                                                                                                            
Installing dependencies from lock file                                                                                                                                      
                                                                                                                                                                            
No dependencies to install or update                                 
                                                                
Installing the current project: archimedes (0.1.0)       
+ '[' -e /app/build ']'                                              
+ '[' -e config.yml ']'                                                     
+ '[' -e config.yml.template ']'                             
+ poetry completions bash                                                                                                                                                   
+ touch /root/.bashrc                                                                                                                                                       
+ echo 'for file in /app/*.completion; do source $file || true; done'
+ echo 'export PATH="/app/bin:$PATH"'
+ chmod -R 744 /root/
```

But then it hangs for hours, for reasons that are not clear. CPU usage is nonexistent during this time. At some point, it remembers life again and completes, but I'm not sure what is going on here.


coleshaw:
Hm, not sure about the archimedes image build....

Very cool throbber, I feel like I should duck and hide when I see it. A couple of issues:

* Some of the input styling seems off on FF. Perhaps we should re-set the widths. I can imagine if you can't get the container up, there's no way to check that :stuck_out_tongue_closed_eyes: 
* That step does take a little longer for me, but not hours. Have you tried manually removing the `etnaagent/archimedes` image and letting it rebuild? `docker image rm etnaagent/archimedes`. Perhaps there is some intermediate container that has cruft or something.
* The plot is not getting rendered. I bet this is because of how I changed the data fetching, to minimize grabbing non-UI-related output. I'll try to dig into this so I can give more helpful guidance.


Screenshot from 2021-03-15 08-56-34

coleshaw:
Pushed an update with spec to this branch to fix the plot output. The UMAP looks nice, though a bit blobby to my untrained eye.

archimedes/archimedes/functions/plotting/__init__.py
```diff
@@ -1,2 +1,4 @@
 import plotly.express as px
 import plotly.io as pio
```
dtm2451, line 2: Wonder if it would it be easier for someone to jump into scripting for Vulcan if we leave these called plotly.express and plotly.io here?

Means the `import plotly.express as px` would be in the downstream scripts, but I think the immediate recognizability of "Oh, `px` is plotly.express" saves a new coder from needing to either dive into our backend system from the get-go or else ask us what those are.

coleshaw: I think that's from my hacking around to work on the UI earlier. I definitely think it's a good idea to rename these to whatever makes more sense.

archimedes/archimedes/functions/plotting/colors.py
```diff
@@ -0,0 +1,42 @@
...
+  '#A3A3A3',
+  '#8A5F00',
+  '#1674A9',
+  '#005F45',
+  '#AA9F0D',
+  '#00446B',
+  '#803800',
+  '#8D3666',
+  '#3D3D3D'
+]
```
dtm2451, line 42: dittoSeqq::dittoColors(reps=100) actually repeats this set 100 times by default in order to accommodate complexity of more than 40 groups.  (total length = 4000 is probably overkill there, but the thought was towards accommodating coloring by TCR/BCR clones!)

Not saying we need to implement the same way, but we definitely need our code to handle 40+ groupings somehow.

archimedes/archimedes/functions/plotting/__init__.py
graft: Hmm, alternatively: do we want to actually expose all of plotly.express functionality to the workflow writer, or do we want to give them the `scatter` function with a defined interface, which may or may not use plotly.express under the hood?

archimedes/archimedes/functions/plotting/colors.py
graft: In python this is just `colors * 100` :)

corps: You can do better by using iterators instead of materialized lists!   `itertools.repeat` for instance.  In this way, even if we need 600000000 different colors, we never exhaust the source. :P

https://www.geeksforgeeks.org/python-itertools-repeat/
:nerd_face:

archimedes/archimedes/functions/plotting/__init__.py
dtm2451: God question!  I think both? I expect sometimes a workflow might use packages with their own visualization functions (like scanpy has, though we're ignoring them) and it's be nice to be able to at least let people start with those (perhaps staying that way forever, or perhaps someone *cough, me* would be in charge of re-implementing on our viz-tooling). In such cases, it's probably good to have the raw tooling available for any necessary tweaks.

graft:
Cutting this off here to avoid rebasing more :) if it looks okay.

vulcan/lib/client/jsx/utils/__tests__/workflow.spec.js
```diff
@@ -608,6 +608,7 @@ describe('Workflow Utils', () => {
       expect(result).toEqual(true);
     });
 
+<<<<<<< HEAD
```
coleshaw, line 611: Looks like there is still an unresolved merge conflict here? @graft

PR #271, opened 2021-03-15
allow you to inject a prod token in dev

coleshaw:
As @corps suggested, it may be easier to inject a production token into dev, especially as it will only get more onerous to duplicate a project structure and data locally. This PR adds an optional `archimedes_token` value to `config.yml`, where you can put a production token. If you also edit the `magma -> host` value to production, the Archimedes cells can then talk directly to production Magma.

I've left the convoluted local stuff in for now ... not sure if there is a use case where folks might want to run cells against local data for testing (either Metis files, or Magma records) because it will involve manipulation they don't want in production? If that seems not useful, I can take out the previous local dev setup for Magby / Archimedes and clean that up.


vulcan/lib/server/controllers/sessions_controller.rb
```diff
@@ -154,5 +154,17 @@ def send_headers(stream)
...
   end
+
+  def token
+    # For development, we can inject a production token
+    #   via config.yml, to talk directly to production services.
+    #   This should reduce duplication of data.
+    # Note: You'll need to configire the Magma host to
+    #   in config.yml :development to also point to
+    #   production
+    return @user.token if :production == Vulcan.instance.environment
```
graft: I think it is dangerous to have the production behavior running behind a guard clause; better to check for the exceptional environment (:development) and isolate that, so the fallback behavior is production.

graft:
Seems useful, cloning data into the dev environment remains somewhat ponderous. Should you also change the magma host entry in config.yml at the same time in order for this to work?

vulcan/lib/server/controllers/sessions_controller.rb
coleshaw: Yes, I already made that change in the updated version of this.

coleshaw:
> Seems useful, cloning data into the dev environment remains somewhat ponderous. Should you also change the magma host entry in config.yml at the same time in order for this to work?


You mean in the template? Yes, the user needs to also change their config.yml entry for Magma host.

dtm2451:
I haven't modified any of the config.yml files in the past, but this is something that would need to be done directly, by replacing the `your-production-token-to-use-in-development` with an actual token, correct?

coleshaw:
> I haven't modified any of the config.yml files in the past, but this is something that would need to be done directly, by replacing the `your-production-token-to-use-in-development` with an actual token, correct?


Yes, because the tokens expire and are user-specific, you would have to make that edit.

dtm2451 (approved):

vulcan/lib/server/controllers/sessions_controller.rb
graft: Derp

PR #272, opened 2021-03-15
Fixing build_image caching issue

corps:
Super minor fix for the docker development.

`build_image is a script that builds our base images before docker-compose up is run.  It checks the source code file modification dates against a 'marker' file (empty file with the last touch of when a build was complete) in order to reduce unnecessary builds.  However, if a user runs `make clean` or removes the image, the marker file still exists even when the image clearly needs to be rebuilt.  This change ensures that if an image does not exist, it will be rebuilt regardless of the marker.


Gvaihir (approved):
We approve thine request-eth

PR #281, opened 2021-03-16
Vulcan dashboard

coleshaw:
This PR adds several UI updates from standup a couple weeks ago:

* Stores the workflow sessions in local storage. This means if you navigate back to a given workflow or reload the page, your inputs are remembered and the session key persists, so Vulcan will return you the same data and outputs.
* Revamps the root view with a dashboard showing workflows instead of projects. It's unclear to me what this will look like long-term, so I didn't refactor out projects everywhere....but visually, there are no projects in the URL or on the page. However, users should only be able to see the workflows where they can view all the tagged projects. The idea is that the user could also filter or search for workflows in the dashboard via text-search, author, or tags via a non-existent Search component.
* I've stubbed out metadata for each workflow, including authors, tags, and allowed projects. My thought was that the workflow authors will have to specify which projects will work for a given workflow, since there are different models and data requirement, so cells may not be pan-Data Library compatible. Unclear how this metadata will be authored or stored in the long-term, so JSON files for now.
* Vignettes can be shown for each workflow. They need to be edited as markdown in the metadata JSON file.

This PR does not yet do the following. Will add them in subsequent PRs.

* Groups the primary inputs via a new "group" label in the CWL.
* Polling for "live" status update.
* Some sort of drag-and-drop multiselect input for large groups? Needs to be discussed with @dtm2451 .


vulcan/lib/client/jsx/components/dashboard/card.jsx
```diff
@@ -0,0 +1,52 @@
+import React from 'react';
+
+import Icon from 'etna-js/components/icon';
+
+import {workflowName} from '../../utils/workflow';
+import ImageMemo from './image_memo';
+
+// To get webpack to pick up the files.
+require('../../../img/umap-thumbnail.png');
```
coleshaw, line 9: Note, the images are basically stubbed in right now. Once we figure out how we want to do actual workflow metadata and editing, I would imagine the images become served up by Apache or some other means.

graft (approved):

PR #282, opened 2021-03-17
Vulcan group inputs + help

coleshaw:
This PR adds some additional functionality to the UI:

* If a `group` attribute is provided for primary inputs in the CWL, the UI will present the inputs as grouped panes.
* If a `doc` attribute is provided for primary inputs in the CWL, the UI will show a help icon with the given text, next to the input box. @graft not sure if this conflicts with your current style changes, feel free to adjust how this behaves.
* Refactors inputs and outputs to distinct components. @graft may conflict a bit with where you put your Plotly component.
* Fixes the vignette page and only shows the link if a workflow has that defined.

@dtm2451 , when you have time, could you take a look at the inputs' `group`, `doc`, and workflow `vignette` values in the CWL and the `metadata.json` file, and populate those with text that makes sense?


corps:
This is fine for now, but I'll be honest -- I think we'll probably eventually throw this idea away.  We need to remember that any user using an external CWL tool will not handle grouping, which means that grouping is likely to arbitrarily "get lost" when pushed through another tool and only a user who combines intimate knowledge of hand writing yaml and our system's unique behaviors is going to know how to recover that.

What about grouping by naming convention?  For instance, if a primary

corps:
oh crap cat hit button

corps:
What if instead, primary inputs can use naming conventions like 'groupa-myinput', and the ui looks for a dash prefix as a grouping name?

The advantages:
1.  Is 100% compatible with all existing CWL tools, which of course, allow users to edit primary input names however they wish.
2.  Is intuitive, both visually (just looking at a series of inputs with similar prefixes immediately is clear that they are grouped somehow) and in terms of teaching it (is grouping down by the attribute group, group_name, section, section_name, grouping, ???, oh wait it's just the name which I already have to know).

But, in any case, moving ahead is most important.

dtm2451:
:+1: with Zach on all points, including being down for moving forward with this implementation for now.  gonna try testing this out...

coleshaw:
:+1: It shouldnt' be much effort to change it. The thing will be it has to be a valid YAML control character, so things like `-`, `:` are probably out (?)...but we can settle on something.

dtm2451 (approved):
Looks lovely!

PR #284, opened 2021-03-17
Async Scheduler

corps:
No ready for review, but checking in against losing local work.


etna/docker-compose.yml
```diff
@@ -22,6 +22,7 @@ services:
       - ../docker/etna-base/entrypoints:/entrypoints:ro
       - /etc/passwd:/etc/passwd:ro
       - /etc/group:/etc/group:ro
+      - $HOME/etna.yml:/root/etna.yml
```
corps, line 25: Not, entirely convinced this is wise.  If you don't have etna.yml when you start this up, it creates a directory lol.

etna/lib/etna/command.rb
```diff
@@ -203,6 +203,7 @@ def subcommands
       @subcommands ||= self.class.constants.sort.reduce({}) do |acc, n|
         acc.tap do
           c = self.class.const_get(n)
+          next unless c.respond_to?(:instance_methods)
```
corps, line 206: Unrelated bug I found in the etna gem, due to using a const that wasn't another command.

etna/docker-compose.yml
graft: Why would you want to mount etna.yml anyway?

corps: Eh, for testing the etna command stuff more conveniently.

graft: oh ah - yeah i have been just running gem build/install for testing, but it is painful.

PR #285, opened 2021-03-18
Gather tube recs from multiselect filtering

dtm2451:
Certainly not a finished product, but I think it could be after the multiselect piece is updated by @coleshaw & after some review and/or pairing with @graft and @coleshaw.


dtm2451:
Think it's ready now!  I can't test via vulcan, only via `make -C archmedes bash` > `archimedes-run ...`, cuz of something I haven't known how to fix on my system... so let me know if it's not working!

Cole - could we show docs for the multiselects?

coleshaw:
> Cole - could we show docs for the multiselects?

Hm, perhaps I can do that as part of the "put multiselect steps into a single drawer" work. So I would say, add them into the step docs, and I'll figure out how to make them appear in the UI.

coleshaw:
For the error, did you refresh your token in your `config.yml`? Did you remember to update the key to `:archimedes_token`?

coleshaw:
@dtm2451  I pushed a commit to make this work with Vulcan ... but when I run it, I don't get any records back for "Confirm record names". Should I? I picked an experiment and a tissue (only 1 option for each -- is that right?)...

dtm2451:
Well, there weren't actually any values for the tissue (biospecimen:biospecimen_type) before.  So the return there was wonk and selecting any of them actually *should have* led to zero options =).  Try again?  I just filled in that attribute for the two current samples.

Re token: that's been weird.  I'll mention in standup but my production token didn't seem to change at all last week in janus.  I have a fresh one today, so I'll try again soon.

dtm2451:
Should be up-to-date with the current modeling!

I will need to remove the biospecimen_group IDs from the options given for tissues.  Their inclusion is not intended.  For any testing in the meantime, please ignore the "XCRS1-MM170-SCPYMT3PT1" tissue option!

dtm2451:
Likely gonna merge this in later today after I link in some more of the data and can test it more! =)
👍 / 👎 @coleshaw?

coleshaw:
I haven’t had a chance to test, but if it does what you need go ahead and merge. I will fix my PR tomorrow on top of yours. Should get these two in before demo meeting tomorrow.

PR #291, opened 2021-03-23
Vulcan UI inputs groups

coleshaw:
This PR addresses some UI feedback for Vulcan:

* Increases heading size for output cells.
* Adds text to the Vignette and Run buttons. Increases the button size.
* Shows the steps-list by default and automatically hides it only when there are output cells.
* Larger icons for the Plotly tools.
* Default Plotly download set to PNG ... this seems up for further development to perhaps add a custom PDF button.
* UI component for `nested-select-autocomplete` input, specifically for color-by selections.
* Groups UI Input steps into a drawer, if the step names follow the convention of `Group_Name__step_name`. Passes through the step `doc` help strings for each step to the corresponding individual input.
* Dashboard card images are regular resolution at full opacity.
* Changes the steps-list drawer icon to `<`.


dtm2451:
Hmm The color-by option shows up perhaps too early?  I have to fill it in before I can have the record selection run.  (That's as far as I get before the SSL error I get from magby-calls cuts things off.)

coleshaw:
You and @grafy can move it to wherever makes sense. I think it shows up early now because it basically has no requisite steps.

coleshaw:
Oh, I also think the "color by" step doesn't do anything right now anyways, because right now the options aren't even real options :laughing: Still my simple ones, waiting for @graft to populate it and link it to the real "umap" step.

PR #293, opened 2021-03-23
add file delete method to etna metis client

coleshaw:
This is a small PR that adds in a `delete_file` method to the etna gem's Metis client. Useful for scripting clean-up when bulk upload goes awry.


corps:
:+1: 

Would, eventually, be nice to have a recursive delete (with a big confirmation warning behind administrator privs) but this helps already.

corps (approved):

PR #300, opened 2021-03-25
Failed coloring

graft:
WIP to update coloring


archimedes/archimedes/functions/magma/magma.py
```diff
@@ -0,0 +1,17 @@
...
+        verify=(app_env == 'production')
+    )
+
+def question(magma, question, strip_identifiers=True):
+    query_result = magma.query(project_name, queryTerms=question)
+
+    if not 'answer' in query_result:
+        raise 'No answer to magma query'
+
+    return [ v[1] for v in query_result['answer'] ] if strip_identifiers else query_result['answer']
```
dtm2451, line 17: This is certainly much improved compared to the incomplete version I'd started making. 👍

coleshaw (approved):
I think this looks good ... at least the earlier revision worked for me. Hopefully you can make it more robust? I will fix the two UI things (temporary showing of color-by step and nested-autoselect typing).

coleshaw:
Oh, and I guess just need to update the snapshots for vulcan tests.

dtm2451:
I see that the back-end of the experiment/tissue option grab & selection parsing changed, and so did how it works...

Can I ask a use case question then:
1. How does a user ask to get "Blood" tissue from all experiments?  Do they need to click on all of the experiments?

dtm2451:
We should update
https://github.com/mountetna/monoetna/blob/334bc17b08a70bed1bdb303a944e8de5d8511cd3/vulcan/lib/server/workflows/umap.cwl#L93 and 
https://github.com/mountetna/monoetna/blob/334bc17b08a70bed1bdb303a944e8de5d8511cd3/vulcan/lib/server/workflows/umap.cwl#L100 based on the new usage.  Unchanged, so probably still reference the 'No Selection' option which will no longer exist.

graft:
Hmm, I removed "No selection" perhaps overzealously - I was confused by how it works in the multiselect, since you could select a bunch of experiments and "No selection" as well, so it seemed redundant. But we need to allow a 'blank' or 'all' option somehow. I think the component itself ought to allow you to select an "empty" state, which should behave differently than selecting a bunch of experiments, rather than trying to inject this in as a python option (which results in some weird behavior).

dtm2451:
Ah, so 'No Select' was the "all"-like option.
Old behavior:
I had the behavior set to go with 'No Selection' if selected whether other items were actually selected or not.  And if all the multi-selects had 'No Selection', the return for the checkboxes was all_tubes.  These mechanics were/are explained in the in-line docs.

Re:
> rather than trying to inject this in as a python option (which results in some weird behavior)
What was this weird behavior?  'No Selection' never went into a query how I'd had the parse script's logic set up 🤷. Either no query was done for that set, or 'No Selection' wasn't there.

coleshaw:
So if you control it via the UI component, how would you differentiate between not-selected and empty?

coleshaw:
Perhaps an alternative way to show this in the UI is to use the checkboxes component (like verify record names), instead of multi select? That way all experiments can be selected at once, or individually. Downside is that takes up more screen space and the current component may not behave exactly right.

graft:
> What was this weird behavior?

The weird behavior is that the user can create a list with both "Pymt3" and "no selection" in it. The fact that you can remediate this later in python doesn't make it less weird.

The solution @coleshaw is to have the component itself present "no selection" (that is, it isn't in the cwl input) as an additional option and disallow other options once it has been chosen.

coleshaw:
So we just have to agree on the word "No selection"....at some point, all this hidden documentation will need to be made obvious to other workflow authors, or just let them author the UI directly somehow.

coleshaw:
Just a semantic question, but does “All” make more sense than “None”? In a software conditional, None and (something) == None, usually?

dtm2451:
Agree it could make more sense to have the multi-select component itself add the 'No Selection'-esque option.

I think "All" would make more sense than "None" here.  In terms of the records being ultimately selected, the meaning is more 'all' than 'none'.

Maybe the display could be **"All"**, and the output when that's selected could be literally all-of-the-input-options?  I think that would work directly with Saurabh's code as written, removes any need to allow users to move forward without selecting anything, and works in the intuitively predicted way for a script-writer!

dtm2451:
Side Note: I want to add a 3rd multi-select option for "Experiments" "Tissues" **+ "Cell Fraction"**.  I'll branch off of here for now to do it.

coleshaw:
I think having a default “All” in the UI component makes sense, and it would just send back all the options (like a ui-queries/multiselect-string-all.cwl component). I’ll see if I can tweak that into the component Monday morning.

graft:
"All" seems the right choice in this case. In the future we will have to figure out all the input cases we might encounter, started to attempt that in #306.

dtm2451:
> We should update
> https://github.com/mountetna/monoetna/blob/334bc17b08a70bed1bdb303a944e8de5d8511cd3/vulcan/lib/server/workflows/umap.cwl#L93
> 
> and
> https://github.com/mountetna/monoetna/blob/334bc17b08a70bed1bdb303a944e8de5d8511cd3/vulcan/lib/server/workflows/umap.cwl#L100
> 
> based on the new usage. Unchanged, so probably still reference the 'No Selection' option which will no longer exist.

Just noting that I took care of this within my PR/single-commit made yesterday

graft:
Okay, other than expecting the "all" behavior in the multiselect component from @coleshaw, I think there are no longer any outstanding issues to resolve before demo (once the component behaves correctly the workflow should work fine as is). I merged in @dtm2451's fractions branch and also added the inputs Bushra was inquiring about, in addition to resolving outstanding plotly drawing issues.

I also investigated enabling multi-threading for scanpy, which apparently can be done across the suite via `scanpy.settings.n_jobs = NUM_JOBS`. However my attempt to set this resulted in a bizarre timeout error with no error result produced, so further experimentation is required here, i.e. we should wait til after Tuesday.

vulcan/lib/server/workflows/scripts/calc_umap.py
```diff
@@ -1,10 +1,26 @@
...
+sc.tl.umap(scdata, min_dist = min_dist, spread = spread, maxiter = num_iters)
+
+# color options
+color_options = {
+    'Cluster': None,
+    'Experiment': None,
+    'Tissue': None,
+    'Pool': None,
+    'Biospecimen Group': None,
+    'Gene': dict([ [ gene_id, None ] for gene_id in scdata.var_names ])
```
graft, line 19: This uses the list of reduced genes in the final UMAP for coloring - since the original matrix is stripped down this is the best we can do here, but some important genes might not be visualized this way (e.g. cd3e didn't show up in my test) - does this seem reasonable @dtm2451?

dtm2451: Oh right this is my fault.  There is definitely a way to have the non-reduced data stick around.  It's always seemed a bit clunky to me, but I can fix that in the scripts after this gets merged!

PR #304, opened 2021-03-26
Speed up ci

coleshaw:
So this PR puts our app tests into a matrix and separates out the Python ones into a standalone test set...was an attempt to speed up CI and thus our overall dev cycles. Does halve the time, about, which seems like a win?


etna/Makefile
```diff
@@ -13,6 +13,7 @@ bash::
 	@ docker-compose run -e SKIP_RUBY_SETUP=1 --rm etna_app bash
 
 release-test::
+	docker network create monoetna_default
```
coleshaw, line 16: I had to add this step to the `etna` Makefile to create the `monoetna_default` network manually, since it kept failing because the `monoetna_default` network didn't exist. I suspect this is because it's created during some app_fe or edge-apache container, and since etna doesn't have that, the network wasn't being created.

corps: :+1:

corps (approved):
Love it

corps:
Note however, this will exhaust our github actions concurrency faster.  We should consider asking gabi for an upgraded plan.

coleshaw:
That's a good point. I assume we're on the free plan? In which case, 20 concurrent jobs would be...ha, like 2 commits being tested at any given time. Which is not a lot given how much activity we have. I think sometimes we've got like 6-7 things happening at once?
 
 
-----Original Message-----

corps:
True.  But, ultimately, I think this is the right step.  Let's follow up next week with Gabi.  The paid plan and the faster builds are a huge team win and we can demonstrate it pretty clearly.

coleshaw:
Seems like we're on a GH Academic plan, [which is the equivalent of Team](https://docs.github.com/en/education/explore-the-benefits-of-teaching-and-learning-with-github-education/apply-for-an-educator-or-researcher-discount). So that should be 60 concurrent jobs, which may be enough? Especially since the Python job is pretty fast, so that should unblock any additional jobs.

PR #308, opened 2021-03-29
Vulcan multiselect all

coleshaw:
This PR does several things:

* Sorts the options in a multiselect, in alphabetical order ignoring case.
* Adds a new UI component that lets you select "All" of a multiselect, or clears your selection. This has been set to use for the three  Select Records inputs.
* Fixes a bug around determining dependent steps for output cells with multiple, named outputs.


coleshaw:
Have tested this locally, going to merge it in so everything is ready for standup.

PR #310, opened 2021-03-29
Timur view editor fixes

coleshaw:
This PR fixes a couple of small issues I had when using the view editor in Timur:

* Creating a new view correctly adds it to the menu. Previous behavior was to throw an error and present a blank page.
* Deleting a view blanks out the main pane (previously it kept the just-deleted view).
* Updates the CSS so that the view document takes up most of the vertical space ... a bit nicer when editing long views.
* Only displays the view for Admin users for a given project. For simplicity's sake, this now appears to the right of "Help", but if anyone feels strongly about it appearing to the left of "Help", I can move it back.

### DIFF
--- a/timur/lib/client/jsx/actions/view_actions.js
+++ b/timur/lib/client/jsx/actions/view_actions.js
@@ -60,7 +60,7 @@ export const deleteView = (view, success) => (dispatch) =>
     .then(
       () => {
         dispatch(removeView(view.id));
-        tryCallback(success, view);
+        tryCallback(success);
       }
     )
     .catch(showErrors(dispatch));
--- a/timur/lib/client/jsx/actions/view_actions.js
+++ b/timur/lib/client/jsx/actions/view_actions.js
@@ -3,7 +3,7 @@ import {fetchAllViews, getView, updateView, createView, destroyView} from '../ap
 import {Exchange} from 'etna-js/actions/exchange_actions';
 import {showErrors, tryCallback} from 'etna-js/actions/action_helpers';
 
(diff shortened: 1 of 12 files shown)


graft:
Thanks for these - one of the things I noted when editing it is how dangerous it felt - if I deleted something there was no recovery. Might be nice to also check the sanctity of the JSON somewhat before updating it.

coleshaw:
There is a JSON check for general conformance to the spec, but the results only show up in the dev console. Do you mean surface that, or provide more of a "view-JSON validation" check?

graft:
> There is a JSON check for general conformance to the spec, but the results only show up in the dev console.

Hmm, i can't see where that runs. I am imagining just, when you hit save, if your json is bad you get a visible warning and the action is canceled.

coleshaw:
> Hmm, i can't see where that runs. I am imagining just, when you hit save, if your json is bad you get a visible warning and the action is canceled.

Yeah, I couldn't see it today :cry: Maybe I imagined it, when I was editing views for COMET clinical data. 

In any case, I've added some in-line linting, so errors will appear directly on the page as such:

Screenshot from 2021-03-30 09-57-43

Let me see if I can figure out how to disable save when there are errors.

coleshaw:
Okay, so this should hide the "save" button if there are any explicit errors in the linting, so users don't accidentally save malformed JSON. Warnings (like trailing commas, or `'` instead of `"`) are ignored and still let the user save (I'm assuming warnings are okay for the server-side of things).

PR #311, opened 2021-03-29
Graft/workflow bugs

graft:
A few last-minute bugfixes, fixing the color options query to magma, moving the controls in the multiselect component, and labeling the color map correctly in the plot.


dtm2451 (approved):
Not in linux to be able to test directly and see the jsx updates, but the color-by and use_weights code looks right to me. 👍

coleshaw:
Hm, there seems to be weird behavior with the multiselect -- not sure if it's worth fixing now, or how easy that would be. But basically if you start adding an item to the list (so a dropdown appears), and then click "all", it only adds items up to the next-to-last one. And keeps the dropdown.

Makes me wonder if the "all" should first call "reset" or at least clear all the values, before attempting to add everything?

Screenshot from 2021-03-29 18-24-43

PR #312, opened 2021-03-30
add "Tube" color-by option

dtm2451:
A very small, but very important addition!

Thinking we use Lymphoid fraction samples and then we color by 'Tube', then 'Cluster', then 'Cd3*', 'Ms4a1', and 'Cd14'


dtm2451:
I'm going to merge this in tonight, but tagging you @graft so you know!

PR #315, opened 2021-04-02
Fix rename attribute on etna gem

coleshaw:
This PR fixes some minor things:

* The `etna` gem was throwing an exception when trying to renaming an attribute using a CSV. This fixes that feature and bumps the gem version.
* Adds a script to update Gemfile.lock to match the new `etna` gem version, inside all the apps. When you bump the `etna` gem version, even though the apps are mounting the source code and using a path to install the gem instead of from RubyGems, the version in the Gemfile.lock still has to match the version mounted at `/etna`. This script provides a handy way to update the lockfiles.
* Updates the autocompletion regeneration script with Timur.
* Patches the Attribute model so that the gem can update an attribute's `desc` value via CSV.


graft (approved):

graft:
Would the usage be - after you have made a change bumping the version of etna-gem, you run `bin/update-gem-locks` and `make restart`?

coleshaw:
Yup, otherwise next time you restart the app containers, they will fail on bundle install because of the version mismatch.
 
 
-----Original Message-----

graft:
Could we then add a call to this to /entrypoints/build.sh so this runs before `bundle install`? Or I guess a split-by-app version of this...

coleshaw:
> Could we then add a call to this to /entrypoints/build.sh so this runs before `bundle install`? Or I guess a split-by-app version of this...

I think I understand what you're getting at, I'll see if this can be baked in without an extra command.

coleshaw:
So just a note, this does already happen in `build.sh`, so I guess a specific command isn't strictly necessary. Just have to remember to run `make restart` and commit the new lockfiles.

corps:
I've been tempted to add a file watcher process in our docker images that rebuilds things like this, but it could be super finnicky too.  Hard to know what the ideal behavior really is.

PR #316, opened 2021-04-02
Task tokens

graft:
This adds task tokens to Janus, which are project-specific, editor-only tokens that last for a week (or whatever is configured). The token has a 'task' flag; tokens with this flag are always checked with Janus. Also a UI for copying them on the Janus Settings page.

Some weaknesses:
- The token is not stateful, Janus just checks that the task token is well formed and the permissions are current. Thus, we cannot invalidate individual tokens, we can still only change user permissions or cycle the Janus secret.
- Checking for the presence of the `task` flag seems brittle - if the flag name changes, these tokens might just be treated as a regular (albeit very long-lived) token.


coleshaw:
Might be nice to also add a command to the etna gem to generate a task token? That way DS folks could create their own without having to figure out the REST API or use the UI?

Is it possible to create your own task token, with your SSH key if you've got that set up?

graft:
Ok, updated with a token-generation workflow (`etna administrate token generate`) that lets you generate both task and regular tokens with a nonce or a regular token.

coleshaw (approved):

coleshaw:
Do you plan to bump the `etna.gemspec` version so it gets published as a gem for folks to use?

PR #323, opened 2021-04-06
Change IPI Population Loader ETL to Hash-based scanner

coleshaw:
Sigh, okay I think I've fixed the time comparison between when a file is updated versus when the Magma record is updated. It feels hacky, but I'm not sure there is any other way to make the comparison work because the Metis file will always have an `updated_at` that is earlier than the Magma record's `updated_at` (since the file is created / copied before the Magma record is updated). So I've included a 5-second fuzzy window of time check -- if the file was updated up to 5 seconds before the record itself, I call that a "new" WSP file and load population data from it. Hopefully will not result in too much redundant processing.

Will let this percolate. Running it off my computer for now to help in the debugging, will migrate it to production tomorrow.


corps:
Hmm.  What about implementing the etl so that instead of doing time comparisons, the cursor just stores the last md5 for each relevant file for each record?  
That way, allow the etl to tell you that the magma record has updated, then just compare the md5 to see if it is the same or different than the last update, and keep the cursor that way.  Timing isn't as relevant, just the hash.

But, this works and I'm ok with it as a stop gap.

coleshaw:
> Hmm. What about implementing the etl so that instead of doing time comparisons, the cursor just stores the last md5 for each relevant file for each record?
> That way, allow the etl to tell you that the magma record has updated, then just compare the md5 to see if it is the same or different than the last update, and keep the cursor that way. Timing isn't as relevant, just the hash.
> 
> But, this works and I'm ok with it as a stop gap.

Oh thanks, that's a great idea! Let me try swapping in a different cursor implementation.

graft:
> Hmm. What about implementing the etl so that instead of doing time comparisons, the cursor just stores the last md5 for each relevant file for each record?

I agree this is a good idea and very flexible. The other way I imagined polyphemus working (which requires no memory) is looking for the presence of valid output, i.e. check updated_at for your population records to see if they are after the wsp updated_at.

magma/lib/magma/query/md5_set.rb
```diff
@@ -4,6 +4,6 @@ class Md5Set < MetisMetadata
   def [](file_path)
     super
 
-    @requested_files[file_path][:file_hash]
+    @requested_files[file_path] ? @requested_files[file_path][:file_hash] : nil
```
coleshaw, line 7: This wasn't when you queried for multiple file-related attributes, i.e.

```
{"project_name": "ipi", "query":["patient", ["::or", ["::has", "flojo_file_raw"]], "::all", ["updated_at", ["flojo_file_raw", "::md5"], ["flojo_file_processed", "::md5"]]]}
```

Because one of the file paths might be `nil` for any given record.

magma/lib/magma/server/query.rb
```diff
@@ -14,7 +14,10 @@ def action
         show_disconnected: @params[:show_disconnected],
         restrict: !@user.can_see_restricted?(@project_name),
         user: @user,
-        timeout: Magma.instance.config(:query_timeout)
+        timeout: Magma.instance.config(:query_timeout),
+        page: @params[:page],
```
coleshaw, line 18: This should work as expected, at least for the top-level model being queried. `Magma::Question` already takes these arguments and uses them for the `/retrieve` endpoint, so this just passes them through via the `/query` endpoint.

polyphemus/lib/etl_scanner.rb
```diff
@@ -0,0 +1,43 @@
+# Abstract base class for an etl
```
coleshaw, line 1: Extraced these out of `TimeScanBasedEtlScanner` to use across ETLs.

polyphemus/lib/magma_record_file_etl.rb
```diff
@@ -0,0 +1,119 @@
...
+        query_request.page_size = @limit * i
+
+        begin
+          self.magma_client.query(query_request).answer.map do |result|
+            { result.first => result.last.map.with_index { |value, i|
+              [queried_attribute_names[i], value]
+            }.to_h }
+          end
+        rescue Etna::Error => e
+          {} if e.message =~ /Page.*not found/
```
coleshaw, line 36: This is the error thrown when there are no query results...feels like there should be a better way, but I don't think we can rely on status code.

coleshaw:
I've made a hash-based scanner, and it does seem a lot cleaner. @graft you should be able to use it for triggering a raw WSP -> processed WSP ETL, too. Thoughts or feedback on this approach?

magma/lib/magma/server/query.rb
graft: Does it need page_size?

polyphemus/lib/magma_record_file_etl.rb
graft: Seems like a bug... Shouldn't it return a 200 with an empty answer?

coleshaw: Hm, let me investigate. I thought the Timur UI throws a 422 on the Search page when there are no results, so I assumed that was the same behavior for `/query`

coleshaw: Oh, so when you page a query or retrieve request, and you include both `page` and `page_size` options, `Magma::Question` will throw a ["Page X not found" exception with a 422](https://github.com/mountetna/monoetna/blob/master/magma/lib/magma/query/question.rb#L307) when the specified doesn't exist. So this is the correct behavior within an ETL context.

PR #324, opened 2021-04-07
allow multicolumn format for panes

graft:
Just a quick one based on a random thought the other day, allowing Timur panes to be organized into several columns if so desired. Now that we have a view editor, you can just stick "group": "whatever" into your pane definition to put it into a new column for that group.


coleshaw (approved):
Pretty neat! Works for me.

PR #328, opened 2021-04-08
Vulcan small UI fixes

coleshaw:
This PR fixes some of the small UI issues that @dtm2451 identified. Definitely some really neat updates to the UI architecture, @corps ! I like the live updates.

* Workflow link in the nav button points to the right page now.
* Vignette is a modal instead of it's own page.
* Clocks animate in the step list when a step is running.
* Can select color-by gene.

Open issues:

* Halo animation, I think @graft is looking at.
* I can reproduce the issue with Metis error log for the `import to scanpy` step. I'll try digging into this more. I'm going to remove the SSL cert stuff since it seems like none of us are using that. The full error trace is getting cut off, so we can't tell what is really happening.
* For the UMAP not appearing at the final step, I feel like there is some weird timing issue. I'll have to dig into it more. The UI stops polling because there are no "running" steps, but the backend is actually running those last two steps. I suspect there happens to be a small time window where the status isn't updated in time and  the UI thinks nothing is running.


vulcan/lib/client/jsx/components/vulcan_nav.jsx
```diff
@@ -145,11 +158,18 @@ const ModeBar = ({mode, user}) => (
...
-const VulcanNav = ({mode, user}) => (
-  <Nav user={user} logo={Logo} app='vulcan'>
-    {mode !== 'home' && <ModeBar mode={mode} user={user} />}
-  </Nav>
-);
+const VulcanNav = ({mode, user}) => {
+  let {state} = useContext(VulcanContext);
+  let {workflow} = state;
+
+  if (!workflow) return null;
```
graft, line 165: Hmm, is this right? The nav bar won't appear if the workflow has not loaded?

PR #331, opened 2021-04-09
fix throbber

graft:
Makes the throbber run when any step status is "running".


vulcan/lib/client/jsx/components/vulcan_nav.jsx
```diff
@@ -116,8 +117,12 @@ const Halo = ({radius}) => {
 };
 
 function Logo() {
-  let {state} = useContext(VulcanContext);
-  const {calculating} = state;
+  const {state} = useContext(VulcanContext);
+  const {workflow, status} = state;
+
+  let calculating = workflow ? workflow.steps[0].some(
+    step => statusStringOfStepOrGroupedStep(step, workflow, status) == 'running'
```
coleshaw, line 124: Just curious why you used `statusStringOfStepOrGroupedStep` instead of just `statusOfStep`? If you're pulling steps directly from the workflow, I don't think any will be Grouped Steps...?

corps: It still works either way.  The function behaves the same.  I think using this function is fine.

corps:
Incidentally, to make the selectors easier to discover, I plan to put some namespaces in place that allow us to autocomplete and search for selectors by subject domain (steps / data / etc).  I think right now it's a grab back and hard to memorize the purpose of all these pieces, but ican be improved.

vulcan/lib/client/jsx/components/vulcan_nav.jsx
```diff
@@ -116,8 +117,12 @@ const Halo = ({radius}) => {
...
+  ) : null;
```
corps, line 125: The error is this line -- step status should not be null, either undefined or StepStatus.  I'd make this undefined to fit the types.

vulcan/lib/client/jsx/selectors/workflow_selectors.ts
```diff
@@ -24,7 +24,7 @@ export function stepOfStatus(stepStatus: StepStatus | string, workflow: Workflow
 
 export function statusOfStep(step: WorkflowStep | string, status: VulcanState['status']): StepStatus | undefined {
   const stepName = (typeof step === 'string' ? step : step.name);
-  return status[0].find(s => s.name === stepName);
+  return status[0] ? status[0].find(s => s.name === stepName) : null;
```
corps, line 27: Same.  Return undefined here to match the type.

vulcan/lib/client/jsx/components/vulcan_nav.jsx
graft: It seems statusString... does some additional work: `return statusOfStep(step, status)?.status || STATUS.PENDING; ` - I tried using statusOfStep at first but duplicating this logic in the throbber seems like a bad idea.

coleshaw: Ah, okay! :+1:

coleshaw (approved):
Looks good!

PR #333, opened 2021-04-09
provide clearer error message on Archimedes OOM

coleshaw:
This PR is a small fix to provide better feedback to the user when an Archimedes container fails due to a 137 error (out of memory).

Not sure there is anything else we can do, since it seems like it's on the cell author to refactor their code?


graft (approved):

corps (approved):
Nice!

Issue #329, opened 2021-04-09
issues & ideas raised after recent Vulcan back-end update


dtm2451:
Recording these here...

Issues:
- [x] need to manually refresh the page to see the umap show up [likely check this box after testing]
- [x] Hammer nolonger electrified during running.
- [x] Gene input not work for coloring -- 2nd selection box does not show up.
- [ ] Needing to hit run right away may be confusing. (Previously, the first cell (a magma query to populate selection options) was proactively run.) 
- [x] Clicking `Workflow` button from the Vignette page does not navigate properly back to the workflow page because the project_name was not filled in to the url: https://vulcan.development.local/null/workflow.
- [x] magma query parameterization problem at `import into scanpy` step when there are many records
- [x] Clocks did not move continuously while steps were running. The running animation seemed short-lived and only applied to the first step after "Run" was hit.

Enhancements:
- [ ] Zoom/window button, with the plotly buttons, that opens a new window with just the UMAP widget.
- [x] Changes to the color-by output could automatically trigger running for live-ish plot updates... they do now for previously loaded options
- [x] Text box with live typing-based filtering in the experiment/tissue/fraction multi-selects. (Togglable on/off at UI-engineer level could be nice?)
- [x] Perhaps the workflow vignette could be an .md iframe

coleshaw:
I think "Needing to hit run right away" may change anyways as we add in support for more projects ... I think we'll have to create a widget before "Run" but after you click on a card, that says which project you want to run the workflow for.

dtm2451:
Gonna close this as the remaining "issue" is not an actual issue and the remaining enhancement is not a high priority item.

Issue #327, opened 2021-04-08
Change multiselect component to use select_autocomplete instead of just a dropdown?


coleshaw:

dtm2451:
I'm certainly in favor of this, and had it among #329

Issue #326, opened 2021-04-08
Sort group inputs for stable appearance?


coleshaw:
When selecting / deselecting grouped inputs (Select Records step), the inputs seem to jump around in the drawer based off of completion status? Unclear, but it is a big awkward to see them changing order in the list. Perhaps we should sort them alphabetically.

coleshaw:
Note that this only seems to happen on a previously submitted (COMPLETE) step, and the "blank" option is always moved to the bottom of the group.

dtm2451:
:+1: I definitely have seen this too.

Issue #325, opened 2021-04-08
Unchecking records in "confirm record names" step resets on polling status return


coleshaw:
If you go through the workflow and deselect some records at the "confirm record names" step, after you click "Run", the polling status result seems to reset to all as checked, disregarding your user selected ones?

PR #335, opened 2021-04-12
Vulcan minor UI updates

coleshaw:
This PR implements sime minor UI changes:

* #329 / #327, changed the multiselect component to use a dropdown with autocomplete.
* #326, inputs in a group or drawer are sorted alphabetically by label, so they shouldn't re-order if you change their values.
* #325, fixes the checkbox inputs to not reset to all-checked after you Run.

### DIFF
--- a/etna/packages/etna-js/components/inputs/dropdown_autocomplete_wrapper.jsx
+++ b/etna/packages/etna-js/components/inputs/dropdown_autocomplete_wrapper.jsx
@@ -1,4 +1,4 @@
-// Wraps the Dropdown component so it will work with the
+// Wraps the DropdownAutocomplete component so it will work with the
 // list_input to act as a multi-select.
 import React from 'react';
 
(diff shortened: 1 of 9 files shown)


etna/packages/etna-js/components/inputs/dropdown_autocomplete_wrapper.jsx
```diff
@@ -0,0 +1,25 @@
...
+
+import DropdownAutocomplete from './dropdown_autocomplete';
+
+export default function DropdownAutocompleteInput({
+  onChange,
+  list,
+  onBlur,
+  ...otherProps
+}) {
+  function onChangeWrapper(value) {
```
corps, line 13: Minor nit -- you can use `const onChangeWrapper = useCallback((value) => { onChange(value); onBlur(); })` to slightly improve performance and make the onChangeWrapper reference stable.

corps (approved):

PR #337, opened 2021-04-13
use non-rails methods for finding user

coleshaw:
This PR fixes instantiating a user from a token in Janus, which is used by the task token API. I was getting this exception:

```
Caught unspecified error
docker[26165]: ERROR:2021-04-13T13:53:37+00:00 9hwba4 undefined method `symbolize_keys' for #<Hash:0x0000555581ab1220>
docker[26165]: ERROR:2021-04-13T13:53:37+00:00 9hwba4 /app/lib/models/user.rb:36:in `from_token'
docker[26165]: ERROR:2021-04-13T13:53:37+00:00 9hwba4 /app/lib/server/controllers/authorization_controller.rb:75:in `generate'
docker[26165]: ERROR:2021-04-13T13:53:37+00:00 9hwba4 /etna/lib/etna/controller.rb:25:in `response'
```

Because `symbolize_keys` and `except` are Rails / ActiveSupport methods. The tests pass because `factory_bot` includes `active_support` as a dependency, but in the deployed version of the app, task token generation fails (locally and in production).

The alternative approach to this PR, if you prefer, is that we just include `active_support` as a production dependency in the Gemfile and app.

~~Note that I'm working on the Vulcan select-a-project work, and that requires this feature to be working in production, first.~~

### DIFF
--- a/janus/Gemfile
+++ b/janus/Gemfile
@@ -18,5 +18,6 @@ group :test do
   gem 'factory_bot'
   gem 'simplecov'
   gem 'pry'
+  gem 'pry-byebug'
   gem 'database_cleaner', '1.8.5'
 end
--- a/janus/Gemfile
+++ b/janus/Gemfile
@@ -10,6 +10,7 @@ gem 'etna', path: '/etna'
 gem 'jwt'
 gem 'puma', '>=5.0.2'
 gem 'concurrent-ruby'
+gem 'activesupport', '>= 4.2.6'
 
 group :test do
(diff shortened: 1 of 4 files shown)


corps (approved):
Super unfortunate that the test environment here doesn't reflect production... how many other places are we going to run into these quiet issues?  Hmm.

I guess, just adding activesupport wouldn't be a bad idea, atleast then we know our production and test environments are closer.  But this works as well.

PR #340, opened 2021-04-14
Vulcan check scripts

coleshaw:
This PR runs all the Vulcan python scripts through the `archimedes-checker`, to make sure there are no syntax errors and they conform to the DSL. It doesn't catch all errors, but should help prevent some. For example, undefined variables would be flagged by a linter / IDE but are not flagged by the checker. Missing colons and unsafe imports are reported by the checker.


corps (approved):
Nice!  This is a really good idea.

PR #341, opened 2021-04-15
Magma table debugging

coleshaw:
@graft @corps , this PR probably is more a conversation starter around the Magma patient table issue, so it will be long. Happy to pair if it makes more sense, but will try to include all relevant information about replicating the problem, what I think the root cause is, and a patch for the `/retrieve` endpoint to make it not hang / crash the server. I do not think this solves the root problem -- I think a "real" solution would require re-architecting the table models, which may be out of scope.

### Replicating the issue locally

There is a Polyphemus command in this branch to generate some fake data in your dev MVIR environment for the `patient` model plus the three clinical tables. It inserts 200 patients with 50 rows of clinical data per patient in each table (similar to production, with a bit more clinical rows than the real data).

1. Copy the `mvir1` project into your dev environment, using the `etna` gem.
2. `$ make -C polyphemus bash`
3. `-> ./bin/polyphemus generate_fake_mvir_patient_tables`

If you try to run a retrieve on a single table, it works fine:

```
{"model_name": "patient", "record_names": "all", "attribute_names": ["comorbidity"],"project_name":"mvir1"}
```

But trying to run a retrieve on all three, it will hang:

```
{"model_name": "patient", "record_names": "all", "attribute_names": ["comorbidity", "symptom", "admission_lab"],"project_name":"mvir1"}
```

You can monitor top and see that postgres is sucking up memory until the container dies.

###  Root cause -- replicating in psql

I think the root cause is all the joins that happen when you request tables, along with the fact that each table is many-to-one with patient. You can actually go into `psql` and duplicate the memory issue above. Using `EXPLAIN`, you can also see that the join queries explode in complexity.

```
magma_development=#  EXPLAIN SELECT "patients"."name" AS "patients_name", "comorbidities"."id" AS "comorbidities_id", "admission_labs"."id" AS "admission_labs_id", "symptoms"."id" AS "symptoms_id" FROM "mvir1"."patients" AS "patients" LEFT OUTER JOIN "mvir1"."comorbidities" AS "comorbidities" ON ("patients"."id" = "comorbidities"."patient_id") LEFT OUTER JOIN "mvir1"."admission_labs" AS "admission_labs" ON ("patients"."id" = "admission_labs"."patient_id") LEFT OUTER JOIN "mvir1"."symptoms" AS "symptoms" ON ("patients"."id" = "symptoms"."patient_id") WHERE ("patients"."project_id" IS NOT NULL) ORDER BY "patients"."name";
                                                         QUERY PLAN                                                          
-----------------------------------------------------------------------------------------------------------------------------
 Nested Loop Left Join  (cost=1.00..3297066.24 rows=26530200 width=25)
   ->  Nested Loop Left Join  (cost=0.72..78328.74 rows=520200 width=25)
         ->  Nested Loop Left Join  (cost=0.43..1548.24 rows=10200 width=21)
               ->  Index Scan using patients_name_key on patients  (cost=0.14..18.74 rows=200 width=17)
                     Filter: (project_id IS NOT NULL)
               ->  Index Scan using mvir1_comorbidities_patient_id_index on comorbidities  (cost=0.29..7.14 rows=51 width=8)
                     Index Cond: (patients.id = patient_id)
         ->  Index Scan using mvir1_admission_labs_patient_id_index on admission_labs  (cost=0.29..7.02 rows=51 width=8)
               Index Cond: (patients.id = patient_id)
   ->  Index Scan using mvir1_symptoms_patient_id_index on symptoms  (cost=0.29..5.68 rows=51 width=8)
         Index Cond: (patients.id = patient_id)
(11 rows)
```

You can see that the cost is pretty high ... because as we join more tables, the complexity grows exponentially. If we reduce to just two tables, you can see that each patient id + comorbidity id is duplicated for each admission_lab id ... 

```
magma_development=# SELECT "patients"."name" AS "patients_name", "comorbidities"."id" AS "comorbidities_id", "admission_labs"."id" AS "admission_labs_id" FROM "mvir1"."patients" AS "patients" LEFT OUTER JOIN "mvir1"."comorbidities" AS "comorbidities" ON ("patients"."id" = "comorbidities"."patient_id") LEFT OUTER JOIN "mvir1"."admission_labs" AS "admission_labs" ON ("patients"."id" = "admission_labs"."patient_id") WHERE ("patients"."project_id" IS NOT NULL) ORDER BY "patients"."name";
 patients_name | comorbidities_id | admission_labs_id 
---------------+------------------+-------------------
 MVIR1-TEST1   |           193631 |            165528
 MVIR1-TEST1   |           193630 |            165528
 MVIR1-TEST1   |           193629 |            165528
 MVIR1-TEST1   |           193628 |            165528
 MVIR1-TEST1   |           193627 |            165528
 MVIR1-TEST1   |           193626 |            165528
 MVIR1-TEST1   |           193625 |            165528
 MVIR1-TEST1   |           193624 |            165528
 MVIR1-TEST1   |           193623 |            165528
 MVIR1-TEST1   |           193622 |            165528
 MVIR1-TEST1   |           193621 |            165528
 MVIR1-TEST1   |           193620 |            165528
 MVIR1-TEST1   |           193619 |            165528
 MVIR1-TEST1   |           193618 |            165528
 MVIR1-TEST1   |           193617 |            165528
 MVIR1-TEST1   |           193616 |            165528
 MVIR1-TEST1   |           193615 |            165528
 MVIR1-TEST1   |           193614 |            165528
 MVIR1-TEST1   |           193613 |            165528
 MVIR1-TEST1   |           193612 |            165528
 MVIR1-TEST1   |           193611 |            165528
 MVIR1-TEST1   |           193610 |            165528
 MVIR1-TEST1   |           193609 |            165528
 MVIR1-TEST1   |           193608 |            165528
 MVIR1-TEST1   |           193607 |            165528
 MVIR1-TEST1   |           193606 |            165528
 MVIR1-TEST1   |           193605 |            165528
 MVIR1-TEST1   |           193604 |            165528
 MVIR1-TEST1   |           193603 |            165528
 MVIR1-TEST1   |           193602 |            165528
 MVIR1-TEST1   |           193601 |            165528
 MVIR1-TEST1   |           193600 |            165528
```

Taking out one table also reduces the query complexity by quite a bit:

```
magma_development=# EXPLAIN SELECT "patients"."name" AS "patients_name", "comorbidities"."id" AS "comorbidities_id", "admission_labs"."id" AS "admission_labs_id" FROM "mvir1"."patients" AS "patients" LEFT OUTER JOIN "mvir1"."comorbidities" AS "comorbidities" ON ("patients"."id" = "comorbidities"."patient_id") LEFT OUTER JOIN "mvir1"."admission_labs" AS "admission_labs" ON ("patients"."id" = "admission_labs"."patient_id") WHERE ("patients"."project_id" IS NOT NULL) ORDER BY "patients"."name";
                                      QUERY PLAN                                      
--------------------------------------------------------------------------------------
 Sort  (cost=66716.22..68016.72 rows=520200 width=21)
   Sort Key: patients.name
   ->  Hash Left Join  (cost=437.00..6655.09 rows=520200 width=21)
         Hash Cond: (patients.id = admission_labs.patient_id)
         ->  Hash Right Join  (cost=6.50..346.84 rows=10200 width=21)
               Hash Cond: (comorbidities.patient_id = patients.id)
               ->  Seq Scan on comorbidities  (cost=0.00..313.00 rows=10200 width=8)
               ->  Hash  (cost=4.00..4.00 rows=200 width=17)
                     ->  Seq Scan on patients  (cost=0.00..4.00 rows=200 width=17)
                           Filter: (project_id IS NOT NULL)
         ->  Hash  (cost=303.00..303.00 rows=10200 width=8)
               ->  Seq Scan on admission_labs  (cost=0.00..303.00 rows=10200 width=8)
(12 rows)
```

### Proposed fix to /retrieve

Because [`@collapse_table` was used in the Retrieve Controller when constructing the Retrieval object](https://github.com/mountetna/monoetna/blob/master/magma/lib/magma/server/retrieve.rb#L150), the joins happened automatically if the flag was not passed in the retrieve params. I think we can patch `/retrieve` to 1) never do the joins, and 2) to add in table data from the extra table queries themselves, thus working around the joins (not as elegant, but perhaps scales better). Locally I can now run the 3-table query and get data back, without crashing Magma.

### Long-term fix?

I'm not actually sure what a good long-term fix would be, because it seems like the problem is inherent to the database design and what we're trying to query out. We may run into this issue if we build a query UI or allow users to do joins directly ... but that could happen with any database design, and we may have to teach users how to work around it, or put in limits?


magma/lib/magma/model.rb
```diff
@@ -26,6 +26,9 @@ def load_attributes(attributes = [])
             alias_method "#{attribute.attribute_name}=", "#{attribute.column_name}="
           end
         end
+
+        # Table models should always have an identifier attribute
+        self.attributes[self.identity.name] = self.identity unless attributes.include?(self.identity)
```
coleshaw, line 31: This is required because for some reason, table models weren't getting the `id` loaded initially, so first queries always returned no `id` field. Subsequent queries worked, so that was getting added in somewhere in the loop. We should just enforce that table models always load with an identifier attribute.

magma/lib/magma/payload.rb
```diff
@@ -16,6 +16,31 @@ def add_records model, records
       @models[model].add_records records
     end
 
+    def add_table_records parent_model, records, table_model
```
coleshaw, line 19: Basically doing the joins in Ruby, not SQL...

corps:
The thing that surprises me is that in production when I replicate the issue, the memory explosion does not happen in the psql process, but in the inner ruby process of puma.  Like, puma is pumped with anywhere from 2-20gb, while the psql process remains stable.  I haven't had super strong focus the past couple of days due to local distractions, but I was hoping to throw a memory profiler under it in production to see the explanation.

If this is just a query problem, why does the ruby process in production gain gigs of memory weight that are then retained?

corps:
Like in general, it is fine if a system accepts requests that do not complete because they are simply too heavy, so long as the system can retain stability (the request is ended, memory reclaimed, resources freed), but that is not what I see.  After these requests, if the process hasn't already collapsed, it still retains a high memory / swap profile that hurts overall performance until it finally keels over.

One fix should be -- if a request is too big, and it fails due to timeout, all resources should be freed and the system return to normal performance.

coleshaw:
Hm, I'll keep poking around. You're right, in production the same SQL statement returns fine and doesn't consume all the memory.

graft:
Okay, let me try to state what is happening:

You have 200 rows. You join one table to it with LEFT OUTER, now you have 10200 rows (x51 for some reason). Each subsequent table grows x51, because each join is compounded and not reduced.

The culprit is here: https://github.com/mountetna/monoetna/blob/f220f78ae73f724f9bcb4d51488de67ea0f09ae9/magma/lib/magma/retrieval.rb#L132-L133

This adds a join across another table in an attempt to deliver an array of ids as the value for the requested attribute for each record.

A couple of solutions suggest themselves:
1. The immediate, easy thing to do would be to restrict the behavior of `/retrieve` to prevent this happening - that is, set `@collapse_tables=true` for `record_names: 'all'`. I'm not sure any endpoints currently require this behavior anyway, and the data can be put together simply by querying each of the join tables individually.

2. Deliver the ids through another mechanism in the `/retrieve` endpoint. I.e., rather than using the query/Question interface to get this data via a join, the retrieval fetches it separately (probably through a separate flat query) and collates it with the query. This could probably be written into a verb for the table and collection attributes, there are several verbs that do this sort of collating already.

3. Fix the SQL. The combinatorial problem happens because we are LEFT OUTER joining across all of these rows - but we don't actually care about the join information, it is just a nuisance (i.e., not suited to our purpose).

In general this is a problem with the way Magma does queries, it mostly joins things into a big table and does a lot of reducing in Ruby. In many cases this is quite bad and makes large parts of query space inaccessible because Magma runs into combinatorial problems. Instead of doing all of this reducing in Ruby, Magma should just use better SQL, e.g. in this case it would require reducing the join information to an array of ids in a single row. This would hopefully produce more efficient queries without combinatorial explosion problems, and it would reduce the size of memory footprints for queries in Ruby since information would come back from postgres more compactly. Mostly this hasn't been done because, frankly, my SQL is quite bad.

coleshaw:
Yeah, I think #1 is similar to the patch I provided in this PR. And we should probably implement one of the solutions you outlined, since as we get more data, this may prove to be another bottleneck.

But as @corps pointed out, none of this probably addresses the current problem on production, because the query there actually completes from the SQL-side (the number of joined rows-per-table are lower). We're still working on pinpointing the issue...

magma/lib/magma/model.rb
graft: Hmm, the logic around this bit of weirdness is very warped, tread carefully here.

corps:
Right, we have two issues.  One is a performance issue with joining, which we could fix.
The other is the production issue that explodes memory in the ruby process, that memory sticks around (is a leak), and eventually kills the ruby server (while the psql process has no such issue).

We need to address the second issue first, and we can rethink the first.  While the two are related, I think the second is actually more pressing.  It's ok and expected to have performance issues, but it should not be causing exponential memory leaks and killing the server.

graft:
> Yeah, I think #1 is similar to the patch I provided in this PR. And we should probably implement one of the solutions you outlined, since as we get more data, this may prove to be another bottleneck.
> 
> But as @corps pointed out, none of this probably addresses the current problem on production, because the query there actually completes from the SQL-side (the number of joined rows-per-table are lower). We're still working on pinpointing the issue...

As I said I think the combinatorial-explosion issue creates two humps, one is the large table in SQL, the other is when this is unfurled in Ruby. I think the former is easier to survive, because postgres is made to do hefty queries and eventually struggles through it, whereas the Ruby side is not made to do much of anything and thus falls apart more easily when trying to complete its half.

corps:
> > Yeah, I think #1 is similar to the patch I provided in this PR. And we should probably implement one of the solutions you outlined, since as we get more data, this may prove to be another bottleneck.
> > But as @corps pointed out, none of this probably addresses the current problem on production, because the query there actually completes from the SQL-side (the number of joined rows-per-table are lower). We're still working on pinpointing the issue...
> 
> As I said I think the combinatorial-explosion issue creates two humps, one is the large table in SQL, the other is when this is unfurled in Ruby. I think the former is easier to survive, because postgres is made to do hefty queries and eventually struggles through it, whereas the Ruby side is not made to do much of anything and thus falls apart more easily when trying to complete its half.

No, I don't actually buy that explanation.  For one, running this query in production there isn't any hump -- it actually executes fast and just fine due to the sheer power of the machine.  In other words, the first half of your theory isn't demonstrible in production at all. 

Secondly, we can materialize the full query in magma console (ruby side) just fine without any issue.  The payload is less than 10mb.  And that memory is recovered by the GC as expected.  But something between the Retrieval class and the actual controller returning json explodes this to 2-20gb per run.  This is not explained by the size of the query, even if exponential.  Further, it doesn't explain why the memory leaks.  Ruby may be garbage collected, but it's not "that bad" that it would leave gigs of memory for indefinite periods of time.

PR #344, opened 2021-04-20
Etna gem bug descriptions

coleshaw:
This PR fixes my changes to the gem in #315, specifically regarding attribute descriptions. The previous PR was flagging all descriptions as "new" because of the discrepency between `description` and `desc` in the different systems we're using. This PR patches that and forces the comparison to look at `desc` only, to detect if the attribute has been edited.


graft:
Can we do it the other way? I have wanted to deprecate "desc" for years.

coleshaw:
Just get rid of `desc` and consolidate on `description`? Let me look into how far that's propagated.

corps:
@coleshaw Let me know if you need help recording a new VCR tape for those failing tests.

coleshaw:
Thanks @corps , I'll let you know. A bit slower because of half-days this week, but I'll try re-recording later this morning.

PR #346, opened 2021-04-22
Magma search improvements

coleshaw:
This PR implements some small search improvements:

- Adds support for `<`, `>`, `<=`, and `>=` for string predicates. They will only return values where the value is inherently a "numeric string" and can be cast to a numeric value, i.e. works for `"20"` but not for `"platypus"`. Works for both `/query` and `/retrieve`. This should let folks use the Search page to filter table columns that are numeric but typed as strings.
- Updates the `/retrieve` `filter` param to also accept a JSON array of filters. This means that we can support spaces in any filter to that endpoint -- will require changes to the front-end (future PR).

### DIFF
--- a/magma/lib/magma/query/predicate.rb
+++ b/magma/lib/magma/query/predicate.rb
@@ -279,6 +279,24 @@ EOT
       )
     end
 
+    def double_cast_constraint column_name, operator, value
+      Magma::Constraint.new(
+        alias_name,
+        Sequel.lit(
+          "CAST(? as DOUBLE PRECISION) #{operator.sub(/::/,'')} ?",
+          Sequel.qualify(alias_name, column_name),
+          value
+        )
+      )
+    end
+
+    def is_numeric_constraint column_name      
+      Magma::Constraint.new(
+        alias_name,
+        Sequel.qualify(alias_name, column_name) => Regexp.new(/\d+/)
+      )
+    end
+
     def invalid_argument! argument
       raise QuestionError, "Expected an argument to #{self.class.name}" if argument.nil?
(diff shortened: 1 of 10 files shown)


magma/lib/magma/query/predicate.rb
```diff
@@ -279,6 +279,24 @@ def basic_constraint column_name, value
...
+          Sequel.qualify(alias_name, column_name),
+          value
+        )
+      )
+    end
+
+    def is_numeric_constraint column_name      
+      Magma::Constraint.new(
+        alias_name,
+        Sequel.qualify(alias_name, column_name) => Regexp.new(/\d+/)
```
graft, line 296: Do you want to allow floats or scientific notation here as well?

graft (approved):

graft:
I might worry that comparison operators on strings might be assumed to behave as alphabetic comparators, which might produce some confusing results. Should we overload this behavior so it behaves as might be expected with "platypus"?

coleshaw:
Hm, when I tested this on “float” strings it seemed to work (will add some specs), but didn’t test scientific notation. Are there fields which use that? Seems difficult to enforce in a string input because of variability... I would probably punt unless we have a clear use case.

coleshaw:
Unfortunately I’m not sure that is possible, because of the casting. Basically you would have to know beforehand if the values are strings or “numeric” with the given implementation. 

If we really wanted it to work with strings and numbers, I think we would have to implement some natural sorting procedure and sort, then limit results outside of the current Constraint architecture. 

I assume that most non-CS folks don’t think about comparing strings with > or < operators...?

PR #348, opened 2021-04-23
make tsv filename more specific

coleshaw:
This is a small PR to make the Magma TSV download filename include more data, so it's a bit clearer to the user what they downloaded. Filename now includes project name, model name, and a date / timestamp.


graft (approved):
Nice one!

PR #350, opened 2021-04-23
Etna gem load tables

coleshaw:
This PR fixes a performance issue when loading tables via the gem. When many records are already populated, trying to do an initial fetch of the templates would time out / fail, because of the large number of existing records. By not fetching all records (which aren't needed anyways for table loading), the process does not time out and runs more smoothly.

### DIFF
--- a/etna/lib/commands.rb
+++ b/etna/lib/commands.rb
@@ -339,7 +339,7 @@ class EtnaApp
             request = Etna::Clients::Magma::RetrievalRequest.new(project_name: project_name)
             request.model_name = model_name
             request.attribute_names = 'all'
-            request.record_names = 'all'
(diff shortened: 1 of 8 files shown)


etna/lib/commands.rb
```diff
@@ -339,7 +339,7 @@ def execute(project_name, model_name, file_path, execute: false)
             request = Etna::Clients::Magma::RetrievalRequest.new(project_name: project_name)
             request.model_name = model_name
             request.attribute_names = 'all'
-            request.record_names = 'all'
+            request.record_names = []
```
coleshaw, line 342: Basically `all` here was not necessary and caused performance issues for models with many records.

PR #356, opened 2021-04-27
Prometheus in dev, vulcan uses task token

corps:
Two changes I'm playing with

1.  Run a prometheus instance in development.  Using this as a staging ground for testing the configuration we might use in production.  This is just a basic setup, will start to spend more time on this post DNS resolution.

2.  Vulcan exchanges token for task token before passing to scheduler

3.  User must select their project before loading into a workflow.


etna/lib/etna/clients/base_client.rb
```diff
@@ -29,7 +29,8 @@ def token_expired?
       def token_will_expire?(offset=3000)
         # offset in seconds
         # Will the user's token expire in the given amount of time?
-        epoch_seconds = JSON.parse(Base64.urlsafe_decode64(token.split('.')[1]))["exp"]
+        payload = JSON.parse(Base64.urlsafe_decode64(token.split('.')[1]))
```
corps, line 32: This non functional change just made it easier for me to debug by examing the payload during a test.

graft:
I like the read_only though you lose the restricted bit in the process, which is maybe a good idea from vulcan but should maybe be its own option to generate?

Rather than requiring the vulcan client to keep track of the task token, why not simply grab a new task token every time just before running a cell using the user's login token? This is pretty cheap and avoids confusing token swapping...

corps:
> I like the read_only though you lose the restricted bit in the process, which is maybe a good idea from vulcan but should maybe be its own option to generate?

Unclear, what do you mean by this?

> Rather than requiring the vulcan client to keep track of the task token, why not simply grab a new task token every time just before running a cell using the user's login token? This is pretty cheap and avoids confusing token swapping...

So this is fairly problematic for a few reasons.

1.  It causes an additional failure point between every cell call.  It's a small performance and failure quirk that could be quite annoying over time.
2.  When we do change the scheduler, we want to keep the logic required to run cells minimal.  If we require that each cell run also require exchanging a token, it means that any change to airflow or otherwise requires increasing additional logic and state.  It is true we will need some logic anyways (handling the storage process and setting up the environment), anyway we can reduce that logic pays off.
3.  If we're persisting a user's token and exchanging that for a task token each time a cell runs, the original user token must be persisted and must not expire for the entire duration.  This gets tricky, requiring us to ALSO refresh the user's token AND exchange it for a task token over and over.  Making a one time exchange into a task token kills two birds with one stone -- we get a new, stronger expiry that is bound arond the start of the task (and not just the start of the user's login session) and we also achieve the ultimate restriction we intend.

Overall I think the small amount of effort on the frontend (minimal) beats the impact of not doing so on the backend by a far margin.  And furthermore, I believe having an explicit entry point to a session as being a good hook to hang other work that is likely to follow.  I believe ultimately, some amount of initialization logic will become necessary for clients entering into a workflow.

If the concern is about the token object being stored on the client side and having to be returned to the server, to be honest we can easily move the boundaries of where the session / token is stored or handled as we evolve (for instance, server side stored and referenced via a cookie or other token, for instance).  There is, I think, some simplicitly overall gained in the implementation to have this token exchange on the client side, and no unwarranted even if how pipelines use the task token now.

corps:
One other alternative -- the task token is injected into the session object directly, which the client deals with opaquely (returning it to the server without inspection) anyways.  In this case, the session controller simply validates first that the user's login session is valid for the project, that the task token is also for the given project, and then uses the inner task token of the session. 

This has a few benefits:

1.  The server can, optionally, refresh or exchange a task token if necessary without client awareness (opaque state)
2.  No explicit action or changing of behavior by client in terms of which token to send up

The only complexity introduced is that the server must validate both the user's permission view and the task token's view.  This isn't without precedent given how oauth works with scoped sub tokens given from a major login token.

corps:
I guess the last alternative is simply that the token generation occur on submit, but be passed around in scheduler state.  This might work.

graft:
I like the idea of storing the task token in the session, but how about - the only undetermined component of the token at this point is the user's identity, which in a read-only scenario is going to produce, at best, some log lines in magma or metis logs (this also becomes confusing when we have people running each other's workflows and contributing to a common cache). Instead we could have a single calculation token, issued to a generic "vulcan@ucsf.edu" user, which can be issued per-project, has viewer-only permissions, and lasts indefinitely. This needs to be computed once by vulcan (or whenever it becomes invalid) and recorded somewhere per-project; it could be re-used across project sessions. The user's token gives them the right to instruct Vulcan to query magma and metis on their behalf using Vulcan's calculation token for that project.

graft:
> I guess the last alternative is simply that the token generation occur on submit, but be passed around in scheduler state. This might work.

Err, wrote all that before seeing this - this seems like the simplest and best!

coleshaw:
> Now I know what you're think -- what about data loading jobs we may want to run?

Perhaps related (or may be what you meant by data loading jobs...!), but that also seems like it will be useful once we get around to letting users persist data, i.e. I want this data to not be garbage-collected in a week, but persisted somewhere long-term (i.e. Magma, Metis).

coleshaw:
So one of Gabi's requests (for reproducibility) is that we add the ability for users to download the parameters they used to generate the AnnData object / plot. For the time being, we don't have to worry about uploading or auto-ingesting any sort of downloaded file, but the assumption would be that the user manually re-enters parameters based on said document.

Thinking about a more advanced version of that, does that have any implications for how the task token is handled? Does the task token need to be included in such a downloaded file, or re-injectable somehow? If someone creates a session and then downloads a parameters file, can they later re-use the same session key with a different (newly- or auto- generated) task token?

Perhaps that is less important once we have server-side saving of state, but I think people will want to track locally how they generated each of their plots, especially if they use them in an article.

corps:
@coleshaw @graft Ready for review again, this time moved the token handling into the submit endpoint.

Added the modal to select a project, and ensure that project is valid for a user.
Now, we show any workflow for which the user has ANY matching project, and filter the resulting json output based on that, too.

corps:
In the mean time, I'm continuing to document prometheus and make more config changes there.

corps:
> So one of Gabi's requests (for reproducibility) is that we add the ability for users to download the parameters they used to generate the AnnData object / plot. For the time being, we don't have to worry about uploading or auto-ingesting any sort of downloaded file, and I think the assumption would be that the user manually re-enters parameters based on said document.
> 
> Thinking about a more advanced version of that, does that have any implications for how the task token is handled? Does the task token need to be included in such a downloaded file, or re-injectable somehow? If someone creates a session and then downloads a parameters file, can they later re-use the same session key with a different (newly- or auto- generated) task token?
> 
> Perhaps that is less important once we have server-side saving of state, but I think people will want to track locally how they generated each of their plots, especially if they use them in an article.

The task token is a better form of reproducibility, since atleast it should be restricting behavior more deterministically (for any given cell hash, it can only hav fetched or interacted with data in the exact same project scoping as the cell hash itself).

For the greater reproducability question, my thoughts are that we shouldn't make too strong of guarantees.  Better that we focus on integrity and _the ability to reproduce_ without necessarily guaranteeing that we will, ourselves, maintain historical data to do so.

For instance, the user could download the set of md5s associated with all metis downloads resulted from magma queries.  It is up to the user if they want to then download and store those md5 binaries (for a paper's purpose).  

For "re-injecting" data into a historical workflow, it's simple -- @graft once we have a binary level cache (one that hashes purely on the value, not the parameterization), we can allow users to provide external sources of data (metis files they keep locally) and have those "injected" at the same point that these binary cached values would occur.  In other words, if you have any cached cell's output (a cell that downloaded raw resources) stored somewhere locally, you can provide them again and rerun the workflow in that manner.  This puts some of the long term storage and reprocability requirements on end users, which is actually maybe not a bad thing -- they could use a private metis bucket if they wanted, or store locally, or whatever based on their precise requirements without us having to necessarily create an encompassing solution in the short term.

So basically, if we allow a user to download the raw cached binaries, a manifest of the binary cache, and then "rerun" work based on those two, we have a reprocability solution without ourselves having to take full ownership of that immediately.  (long term, we could have better policies about when data goes into long or short term storage, and recovering that for a user, but doing that is a big project I think initially, and allowing externalization has many benefits as well).

corps:
Notably though, we still don't have a good answer for published data and access to that.  Once we're talking about reprocability, it needs to be clear that the data required to reproduce must also somehow be escalated to public access.  We don't really have a solution right now for isolating out data for a paper into a publicly accessible bucket so that the historical form of that data is avaiaible for other researchers, do we?

coleshaw:
So for at least the download of parameters, it's certainly not at the level of published papers. I think that's more like...I used the workflow, downloaded an h5ad file + a PNG....now I come back a week later, how do I know what parameters and records to generate those files? What if I want to re-create the plot and fiddle with other params?

corps:
> So for at least the download of parameters, it's certainly not at the level of published papers. I think that's more like...I used the workflow, downloaded an h5ad file + a PNG....now I come back a week later, how do I know what parameters and records to generate those files? What if I want to re-create the plot and fiddle with other params?

Right.  That is also I think @graft 's use case, the "forking" and "sharing" of a document and its parameters.
The tricky party really comes down to -- when you say you want to know which records generated those files, do you mean semantically what query at X time generated those files? (for which currently there can be a difference, and you want the most updated view, because you really want to perform a similar analysis with the latest dataset).  Or do you mean, no literally, what are the raw bytes that resulted in this PNG and h5ad?

janus/lib/models/token.rb
```diff
@@ -62,19 +62,23 @@ def self.valid_task_token?(token, janus_user)
...
-        raise Token::Error, "Cannot write invalid permission on task token!"
-      end
+      # ensure read only
+      payload[:perm] = "v:#{project_name}" if read_only
 
       # degrade admin permissions
       payload[:perm] = payload[:perm].tr('Aa', 'Ee') if payload[:perm] =~ /^[Aa]/
 
+      # Ensure the resulting permission is valid.
+      unless payload[:perm] =~ /^[AaEeVv]:#{Project::PROJECT_NAME_MATCH.source}$/
```
graft, line 78: Nit, the permissions are degraded in line 75 but since this is now running later it incorrectly checks for Aa anyway.

coleshaw:
This looks neat! Just a small nit: would it be useful to show the selected project somewhere in the workflow UI itself? Just wondering if someone has access to multiple projects at some point, they may forget which set of data they are working with when they are adjusting the parameters. I see it in the URL, but that may escape many people's notice.

coleshaw (approved):

PR #359, opened 2021-04-29
Graft/update redcap

graft:
Here is what I think is a working inverse-loader that loads data into existing magma records from redcap. The main purpose of this kind of loader is for when it is impossible to construct a magma identifier from redcap because data is not tracked there, e.g. assay identifiers which do not appear in a redcap clinical database.

The gist: if a model is defined as `invert: true` instead of iterating across redcap records to build new records, the loader will iterate across existing magma records, pulled by query to magma. The magma identifier is used to construct the redcap name corresponding to the magma record, and the update to the magma record is constructed based on the data from the matching redcap record. The patch + casting process is the same, so date shifting happens the same way.

In addition to this inverse-loading change I also attempted to refactor the redcap loader a bunch - the loading is now much more efficient, with fewer queries to redcap and a little better code organization. I also pared down the config format (scripts are no longer subdivided by form, each iteration is defined on the model) to remove some complexity that was no longer required. I don't anticipate many more major config format revisions of this sort going forward, though who can say?

I have made corresponding changes to the project-specific redcap loaders in a separate PR.


corps:
Nice!  Digging in.

polyphemus/lib/etls/redcap/lib/model.rb
```diff
@@ -1,38 +1,60 @@
...
-      project.logger.write("Attempting to load model #{name}.\n")
+    def events?
+      @config[:each] == "event"
+    end
+
+    def invert?
+      @config[:invert]
+    end
+
+    def magma_records
```
corps, line 34: minor nit, should this be called `existing_records` to differentiate more clearly from `records` ?

corps (approved):

polyphemus/lib/etls/redcap/lib/script.rb
```diff
@@ -1,37 +1,104 @@
...
+      return records
+    end
+
+    def inverse_load
+      update = {}
+
+      redcap_records = records
+
+      @model.magma_records.each do |record_name, magma_record|
+        next unless @attributes.values.all?{|v| v.valid?(magma_record) }
```
corps, line 44: I am a bit confused about on thing however-- the magma_records line seems to query only identifiers, will these magma_record entries support this attributes matching algorithm for other attributes somehow?

We definitely shouldn't query "all" attributes, but we may need, another approach as well?

coleshaw (approved):

PR #363, opened 2021-05-03
Magma retrieve matrices

coleshaw:
This PR adds the following features for MatrixAttributes in the Timur Search page (plus support in the Magma /retrieve API):

- Can add matrix `In` filters, which correspond to `::slice` actions. You can specify a subset of the validation options in the Search filter, and only those results will be returned. They will be visible in a modal "Matrix Row" that you click in the table cell.
- A new toggle option to unmelt matrix attributes in the TSV. This will render a matrix into columns in a TSV when selected, instead of just outputing `#<Magma::MatrixPredicate...`. Works in conjunction with the filter above.


magma/lib/magma/attributes/matrix.rb
```diff
@@ -31,6 +33,18 @@ def revision_to_payload(record_name, new_value, loader)
       [ name, new_value ]
     end
 
+    def expand(matrix_value)
+      # Provide an embedded data frame inside of the TSV
+      #   that includes the selected columns names + values.
+
+      # matrix_value.to_json is actually a JSON string.
+      JSON.parse(matrix_value.to_json)
```
corps, line 41: There, might be `as_json` which would return the json compatible structure itself pre stringifying.  It's not standard but many apis do this.  If not, this works.

magma/lib/magma/tsv_writer.rb
```diff
@@ -17,5 +30,40 @@ def write_tsv
...
+        yield array_to_tsv_row(row)
+      end
+    end
+
+    private
+
+    def tsv_row_to_array(row)
+      # Need to set `limit` for split()
+      #   to "space-pad" the resulting array
+      row = row.gsub("\n", "")
```
corps, line 54: Hmm, is this safe?  Any reason we arent using the `CSV` module with `sep: "\t"`?  My concern is that this kind of hand parsing won't work anytime there is an escaped string or other weird edge cases.

magma/lib/magma/attributes/matrix.rb
coleshaw: Unfortunately seems like not included.

magma/lib/magma/tsv_writer.rb
coleshaw: Thanks, good suggestion! I'll investigate.

PR #367, opened 2021-05-05
Generalize the UMAP workflow to ready it for other use with more projects 

dtm2451:
(WIP)
Adds utilization of a `project_data` definition which tells the model#attribute of the target raw data, model#attributes of record-selection and color-by options, and the paths from the raw data model to all other utilized models.

Implementation:
- Adds a new step to the CWL corresponding to a script where the `project_data` object is stored.
- `project_data` is coded as a python `dict`, which gets output (& later read into later steps) as a JSON hash.
- Adds project parsing functions to archimedes which:
  1. Interpret "<model>#<attribute>"
  2. Convert a given "<model>#<attribute>" into the necessary query path.
- Implements these new methods within all magma interaction code of the workflow


vulcan/lib/server/workflows/scripts/parse_record_selections.py
```diff
@@ -1,34 +1,42 @@
...
-seq_model_name = 'sc_seq'
-seq_pool_model_name = 'sc_seq_pool'
 magma = connect()
 
-experiments = input_json('experiments')
-tissues = input_json('tissues')
-fractions = input_json('fractions')
+select1 = input_json('select1')
+select2 = input_json('select2')
+select3 = input_json('select3')
```
coleshaw, line 13: Does this mean projects will only be limited to 3 record selection methods? Is that normal -- not sure if you think some might have more?

vulcan/lib/server/workflows/scripts/retrieve_selection_options.py
```diff
@@ -1,42 +1,35 @@
...
+q_start = [
+    seq_target['model'],
+    [ '::has', seq_target['attribute']],
+    '::all'
+]
 
 magma = connect()
 
-experiments = question(
+select1 = question(
```
coleshaw, line 18: Do these need to check if `select#` is defined for the given project, if it has fewer than 3?

dtm2451:
If we go with the "less coding now option" currently here:
- we would need to enforce a set number of select-by features.
- those feature could be described in the main workflow documentation (I implemented that, and point users to that via the individual (?) buttons)
- Whenever a DS provides the necessary info for their project, Dan would manually add the feature descriptions to the main doc.

coleshaw:
Sounds good. I've got a basic widget working that at least renders correctly, but it's messy and has bugs around reset and all, so I'm a bit stuck for now. Going to refocus on some other priorities early next week and hope to get back to the widget later in the week.

PR #368, opened 2021-05-06
Zc/janus token validate urls

corps:
The janus client, which is used in the etna/auth.rb, did not use the correct url.

How we didn't notice before?  The second commit demonstrates how our logic before would ensure that in dev, we never triggered the task token.  I thought that by removing the archimedes token I was hitting the create task token, but I never was, thus my manual testing did not hit this case.

As for automated testing, I have tests that verify the creation of task tokens work and that those tokens are correct and valid and that they are passed along correctly -- I merely didn't write an integration test deep enough that verified that the etna/auth.rb actually correctly handled these tokens in magma itself.  And because of the above logic, I didn't notice any issues when I attempted to test locally.

Womp womp.  TLDR; changes to auth can be tricky.


corps:
Also extra fun -- the reason why this slows down vulcan is because janus would actually _redirect_ the request to validate the token due to the incorrect url, which double slowed down and confused all etna/auth.rb endpoints for a bit.  Anyways, this should resolve all the things.

coleshaw:
For some reason "add my review" doesn't let me actually add a review...but looks good to me. :ship:

graft:
Hmm, egg on my face for missing this.

Your e2e test seems like it would not catch issues since vcr is recording the endpoint - so if we break the client it would work, but if we do what i did to create this bug (move the endpoint) the test would still pass until someone rerecords.

The client ought to have enough info to know the route is invalid, even in specs, since Etna::Client retrieves route defs when it boots; could we validate hard coded routes this way before phoning? Not sure that would help the test..

graft:
Could you also clarify what the failure mode was? One of the things I am worried about the way this check is written is that if Etna::Auth fails to recognize a token as a task token it might approve of it (as a normal token) until the token expires, even if Janus would invalidate it.

corps:
> Could you also clarify what the failure mode was? One of the things I am worried about the way this check is written is that if Etna::Auth fails to recognize a token as a task token it might approve of it (as a normal token) until the token expires, even if Janus would invalidate it.

Nope, that's not what we were seeing.  Definitely, if it fails to validate as a token due to the client call failing, it returns into a 302 (authentication redirect) as expected in that code.

corps:
> Hmm, egg on my face for missing this.
> 
> Your e2e test seems like it would not catch issues since vcr is recording the endpoint - so if we break the client it would work, but if we do what i did to create this bug (move the endpoint) the test would still pass until someone rerecords.
> 
> The client ought to have enough info to know the route is invalid, even in specs, since Etna::Client retrieves route defs when it boots; could we validate hard coded routes this way before phoning? Not sure that would help the test..

Absolutely.  VCR tests are a weird compromise on e2e tests because they don't catch changes to the server, only changes to the client, but they are fast to write and reliable (most integration testing is hard to make reliable or fast).  That said, in this case, the main benefit was -- to make it easier to rerecord and retest the logic in the future.  that's one of the main VCR benefits, that if someone does make a big refactoring and wants to quickly verify the sanity of their new logic , they have a starting place to reverify the server the changes and lock in a behavior that the client would be tested against.

Testing the route information is possible, but a bit onerous.  You have to intercept the url as it was generated from the client right before the call somehow.  One idea I think I could try to do that would help all VCR tests is changing the matcher logic to attempt this verification, although it will be hard to verify any more than one application's routes at a time (routing is handled at the application singleton level, so for instance janus specs cannot verify magma routes easily).  I'll try adding a minimal change here for that, but in general, there is always a tension because verifying full integration is valuable but also a lot harder to do right than seems right (no good having super complex, slow, unreliable testing).

Issue #306, opened 2021-03-27
list of inputs


graft:
Here is an attempt to lay out the requirements for all possible input variants we might want in a workflow.

Integers
- single int
- range of ints (min and max)

Floats
- single float
- range of floats

Strings
- single string
  - from a list
  - from a long list
  - from a nested list
  - free text
  - regexp constraint
- multiple strings
  - from a short list
  - from a medium list
  - from a long list
  - from a nested list
  - with replacement
  - without replacement
  - up to n
  - 1 to n

Booleans
- single boolean
- multiple booleans

Files
- single file
- list of files

Tables
- tsv file

dtm2451:
Yes!

For tsv, part of me is like this may be a hard requirement for some users to make. But at the same time, a use case might be a user running differential expression in our system and then wanting to make either a Volcano plot or UMAPs / heatmaps / box- or violin-plots / DotPlots based on genes (column X) that meet "Y-column" fold change cutoff and "Z-column" p-value cutoff.  For that, WE would output the tsv to them, then they would take that tsv to the visualization workflow they want!

coleshaw:
I think some of these are pushing the bounds of what CWL currently includes in the spec, so we may have to figure out our strategy for that. May affect adoption of any existing authoring tool.

graft:
Going forward i think we should not try to build any of these, we should pick a component lib and use that (e.g. material-ui), there are many of them and it will save a ton of work that none of us are currently that proficient at.

coleshaw:
I'm pretty sure that even with something like Material UI, we'll still want to build custom wrappers or components on top of the base ones. There are some pretty specific and custom behaviors outlined above (like selecting multiple strings, or the color-by picker from UMAP) that don't look like they are available "out of the box". But certainly we can achieve a more consistent look & feel and take advantage of all the layout components using a component library.

PR #370, opened 2021-05-07
Allow flexibility in the number of model#attribute features to use for UMAP workflow record selection

dtm2451:
Adjusts:
- the output of `retrieve_selection_options.py` to be a JSON hash, which can have any number of keys, instead of being exactly 3 separate lists. output here => input of the widget in which users make selections
- the expected input of `parse_record_selections.py` to be a similarly structured JSON hash.
- the CWL accordingly for these and the selection widget in between them

These changes would be in anticipation of a `multi-multiselect` widget in which:
- CWL `label` encodes a main title & `doc` encodes a single (?) hover-button
- input and output could both look like this for HuMu (output might often be just a subset of these values for each hash element, but the keys will always match between in and out!) :
```
{
  "Experiment": [
    "NMR2",
    "PyMT3",
    "Gnoto 1 - 4T1",
    "B16F10 ABx vs HFD",
    "10X GRN NINA",
    "LLC ABx vs HFD",
    "MC38 ABx vs HFD",
    "B78mChOVA with Kelly",
    "RENCA ABx vs HFD",
    "4T1 ABx vs HFD",
    "CT26 ABx vs HFD",
    "PDAC 6",
    "PyMT1",
    "PyMT2"
  ],
  "Tissue": [
    "Spleen",
    "Tumor",
    "Blood",
    "dLN"
  ],
  "Fraction": [
    "lymphoid",
    "myeloid",
    "tumor associated macrophages 1",
    "tumor associated macrophages 2"
  ]
}
```

That's to go along with the HuMu selection_options project-data which looks like:
```
{
  'Experiment': 'experiment#alias',
  'Tissue': 'biospecimen_group#biospecimen_type',
  'Fraction': 'sc_seq#cell_fraction'
}
```

### DIFF
--- a/vulcan/lib/server/workflows/scripts/parse_record_selections.py
+++ b/vulcan/lib/server/workflows/scripts/parse_record_selections.py
@@ -4,28 +4,19 @@ from archimedes.functions.list import unique, flatten
 from archimedes.functions.environment import project_name
 
 pdat = input_json("project_data")[project_name]
-selection_options = list(pdat['selection_options'].values())
+selection_atts = pdat['selection_options']
+
+selected = input_json('selected_options')
 
 magma = connect()
 
-select1 = input_json('select1')
-select2 = input_json('select2')
-select3 = input_json('select3')
-
 # Experiment and Tissue (AND logic)
 filters = []
-if len(select1) > 0:
-    filters.append(
-        buildTargetPath( selection_options[0], pdat ) + ['::in', select1]
-    )
-if len(select2) > 0:
-    filters.append(
-        buildTargetPath( selection_options[1], pdat ) + ['::in', select2]
-    )
-if len(select3) > 0:
-    filters.append(
-        buildTargetPath( selection_options[2], pdat ) + ['::in', select3]
-    )
+for set in list(selected.keys()):
+    if len(selected[set]) > 0:
+        filters.append(
+            buildTargetPath( selection_atts[set], pdat ) + ['::in', selected[set]]
+        )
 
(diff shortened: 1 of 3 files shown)


dtm2451:
If we go with the "some widget coding now" option here:
- Projects could use any number of model#attribute pairs for allowing record specification
- Those features could be described in place (by pulling from the input's keys)
- Whenever a DS provides the necessary info for their project, Dan would only need to 1) add that data to where all project_data is stored.

coleshaw:
@graft @dtm2451 is this idea similar to the "multiple strings, from a nested list" [specified in the list of inputs](https://github.com/mountetna/monoetna/issues/306)? Perhaps we should build the UI widget regardless, if it's useful in other contexts.

graft:
>CWL label encodes a main title & doc encodes a single (?) hover-button
>a "sub" label for each multi-select would come from the input's keys

I don't see any reason to use CWL features to configure the ui widget if we are passing in a JSON anyway, the configuration should just be in the JSON; the less we have to overload CWL features the better.

dtm2451:
> > CWL label encodes a main title & doc encodes a single (?) hover-button
> > a "sub" label for each multi-select would come from the input's keys
> 
> I don't see any reason to use CWL features to configure the ui widget if we are passing in a JSON anyway, the configuration should just be in the JSON; the less we have to overload CWL features the better.

That's the idea here!  Aside from the normal 'label' and 'doc' bits that all the CWL steps have, everything is in the JSON

PR #372, opened 2021-05-10
REDCap timout

coleshaw:
For deployed versions, I think we actually need the changes as noted in the Chef PR. I've deployed those changes to staging, but not production yet.

In any case, this PR includes two changes that I think will help with the REDCap timeouts. I was seeing `edge-apache` indicate that the timeout was exceeded from the FE app containers, so increasing the timeouts in its `httpd.conf` file (plus in Chef for deployments). I also increased the Magma timeouts because the table update took a bit over one minute, so was being killed sometimes in my local dev testing for large updates.

I feel like REDCap caches API responses, because when I've been testing this morning, typically the first request for a large table fails, and subsequent ones succeed in fetching data. I haven't seen that again after tweaking the timeouts in edge-apache and the etna gem...will try to wait awhile before trying again (seems to at least be more than an hour before it expires...will try again later this afternoon).

### DIFF
--- a/edge-apache/development-edge-httpd.conf
+++ b/edge-apache/development-edge-httpd.conf
@@ -43,6 +43,8 @@ LoadModule ssl_module modules/mod_ssl.so
 </VirtualHost>
 
 <VirtualHost *:443>
+    TimeOut 300
+    ProxyTimeout 300
     ProxyPass / http://magma_app_fe/ disablereuse=on
     ProxyPassReverse / http://magma_app_fe/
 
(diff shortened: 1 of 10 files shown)


corps:
Seems reasonable.  We will make these streamable background jobs proper in the future, but that requires atleast two pieces:

1. Generalizable jobs (almost there)
2. Backend for streaming like redis (ideally this would happen only after we have consolidated polyphemus onto the swarm cluster to make the redis deployment easier)

corps (approved):

PR #373, opened 2021-05-10
Vulcan flexible UI multiselect

coleshaw:
This PR is a WIP for a flexible UI widget that lets projects create multiple select dropdowns. Note that this includes updates from `master`, so the diff is a bit messy and you'll probably want to update your branch to get a clean diff. 

I had to modify the CWL, scripts, and Vulcan server-side code a teeny bit, but the button should render and work if you test it @dtm2451 . The main issues now are behavioral around clicking "reset" (all three lists will reset instead of just the one you click) and loading the session's saved value. Plus cleaning up the code, which is still super messy.


dtm2451:
Did some testing with this and it looks great!

In addition to the bugs you mention, however, I did notice another -- that the records confirmation step didn't seem to run.  Seems to me that the cause is an issue in the output of selections from this widget. The then got swallowed, without error, in the `parse_record_selections.py` script that follows.  I've just added an error for when there are no valid records at the end of that script to prevent such swallowing.

I also fast-forwarded the db/umap-democratize-flexible branch.  There are some merge conflicts now, but I've scanned through them all and I can easily resolve them if you'd like!

dtm2451:
saw the commit called 'clean up' so figured I'd test again =p.  Seems to work great!

coleshaw:
Thanks, thought I was there, but it's very, very close. One small bug left ... I've got to rework the logic a little.

etna/lib/etna/directed_graph.rb
```diff
@@ -52,7 +52,12 @@ def as_normalized_hash(root, include_root = true)
             result[grandparent] << child_node if result.include?(grandparent)
           end
 
-          result[child_node] = []
+          # Depending on the graph shape, diamonds could lead to
+          #   resetting of previously calculated dependencies.
+          # Here we avoid resetting existing entries in `result`
+          #   and instead concatenate them if they already exist.
+          result[child_node] = [] unless result.include?(child_node)
+          result[n].concat(result[child_node]) if result.include?(child_node) && result.include?(n)
```
coleshaw, line 60: Updated the path dependencies algorithm to account for the new shape of the UMAP CWL, otherwise it was getting stuck on "Confrm records" because it was not flagged as a depencendies of the parse_record_selections step.

vulcan/lib/client/jsx/contexts/session_sync.tsx
```diff
@@ -72,7 +73,7 @@ export function useSessionSync(
 
     if (!post &&
         hasNoRunningSteps(state.current.status) &&
-        lastCompletedPollingRequest === state.current.session.inputs &&
+        _.isEqual(lastCompletedPollingRequest, state.current.session.inputs) &&
```
coleshaw, line 76: UI was constantly polling once we introduced a Hash-like object as an input value because this equality was always `false`, so had to convert this to do a deep equality check.

vulcan/lib/orchestration.rb
```diff
@@ -78,7 +78,9 @@ def load_json_payload!(storage, reference)
         unless ::File.exists?(ms.build_output&.data_path(storage))
           storage.with_run_cell_build(run_cell: storage.run_cell_for(ms)) do |build_dir, build_files|
             path = build_files.first.data_path(storage)
-            ::File.write(path, payload)
+            # If the payload is a Hash, we need to save it as a standard JSON
+            #   string, not a Ruby Hash with symbolized keys.
+            ::File.write(path, payload.is_a?(Hash) ? payload.to_json : payload)
```
coleshaw, line 83: JSON stringify if the input value is a Hash, so we don't get Ruby symbolized keys in the output file.

vulcan/lib/server/workflows/scripts/parse_record_selections.py
```diff
@@ -6,17 +6,18 @@
...
+selected = input_json('selected_options')["a"]
 
 magma = connect()
 
 # Create filters for all the 'select-bys' that the value for this attribute
 # must be among the options selected in the previous step.
 filters = []
-for set in list(selection_atts.keys()):
-    if len(selected[set]) > 0:
+for selected_model in list(selected.keys()):
```
coleshaw, line 17: `set` is a keyword in Python, so I think it's safer to use a different variable name here, to avoid issues.

vulcan/lib/server/workflows/scripts/parse_record_selections.py
```diff
@@ -6,17 +6,18 @@
...
+selected = input_json('selected_options')["a"]
 
 magma = connect()
 
 # Create filters for all the 'select-bys' that the value for this attribute
 # must be among the options selected in the previous step.
 filters = []
-for set in list(selection_atts.keys()):
-    if len(selected[set]) > 0:
+for selected_model in list(selection_atts.keys()):
```
coleshaw, line 17: `set` is a keyword in Python, so I think it's safer to use a different variable name here.

coleshaw:
@dtm2451 , I think this is ready to go! Would appreciate if you could test again and try to break it :smile:

vulcan/lib/server/workflows/scripts/parse_record_selections.py
dtm2451: ahhh right!  That's a base data-type isn't it!  Oops.  Can we change this to `target` instead of `selected_model`?

`selected_model` might be confusing. The direct target is the 'label' of the `<label>: <model>#<attribute>` definition, and it's foreseeable that a project definition might actually include two separate attributes of the same model in the future.

dtm2451:
So the only 'bug' I notice is that sometimes if I'm clicking to remove an item from the `Record Selection`s, it flashes away and then comes back.  BUT not a big problem because it's highly noticeable and the items always have gone away for me after the next click :shrug:.

dtm2451:
Another thought, but this would probably be a different PR if we do decide to act on it...

The error report that pops up when selections yield nothing:

```
Python error while executing script: Traceback (most recent call last):
  File "/archimedes-exec/tmpa0hhkcmi/script.py", line 36, in <module>
    raise RuntimeError('No records with data meet the selected criteria.')
RuntimeError: No records with data meet the selected criteria.
```
I wonder if some novice users won't read far enough to get to the real issue here.  I think it's possible to define a custom error type in python. Perhaps we could define for which whenever Vulcan notices it in the traceback, it presents the error text first (I made 'No records with data meet the selected criteria.') and then gives the traceback afterwards?

dtm2451 (approved):

dtm2451:
Oh! Actually can you rebase to `db/umap-democratize`?  I merged in the `*-flexible` branch, so the open PR -> master is just with the 'base' one.  The two branches are matched up exactly, so should be no conflicts.

coleshaw:
> So the only 'bug' I notice is that sometimes if I'm clicking to remove an item from the `Record Selection`s, it flashes away and then comes back. BUT not a big problem because it's highly noticeable and the items always have gone away for me after the next click shrug.

Hm, is there any consistent way to reprouce this? I guess we'll have to notice a pattern, otherwise super difficult to debug / fix...

Renamed to `target` and changed the merge target of the PR.

coleshaw:
> I wonder if some novice users won't read far enough to get to the real issue here. I think it's possible to define a custom error type in python. Perhaps we could define one for which whenever Vulcan notices it in the traceback, it presents the error text first (I made 'No records with data meet the selected criteria.') and then gives the traceback afterwards?

There's probably some larger idea here, but different types of Vulcan errors where this might be useful. I agree, probably a different PR, but I think we should also think about generalizing that idea to other errors. Perhaps there are some Archimedes errors (out-of-memory?) or similar that would also be useful to present differently.

Also, this would probably happen in the scheduler / orchestrator on the server rather than in the UI. The UI is pretty agnostic about the text that comes through, and that kind of re-ordered text might also be useful in other non-browser contexts.

dtm2451:
> Hm, is there any consistent way to reprouce this? I guess we'll have to notice a pattern, otherwise super difficult to debug / fix...

yea... :laughing: I really don't notice much of a pattern!  I think it only happens once there are selections for all the sets, but it's rare enough that I'm not sure.

dtm2451:
Oh!  But I did just find one consistent bug:

If I select "all" for all the sets, then the step registers with a check mark and minimizes. The next step doesn't start calculating, but I do need to reopen the `Record Selection` cell.

coleshaw:
> If I select "all" for all the sets, then the step registers with a check mark and minimizes. The next step doesn't start calculating, but I do need to reopen the `Record Selection` cell.

I'm not clear what the expected behavior is here. Does it work if you click "Run" or are you expecting the record generation step to automatically happen after you click All? I think the latter expectation is inconsistent with the rest of the workflow...? Like usually you need to click Run I think to move on...

dtm2451:
Right, neither of those are my expectation.  I don't plan to hit run right after (though I have tested with hitting run having all-"all"s and it works fine.). What I do expect is to be able to then remove selections before I hit run.  I just don't expect the window to minimize.

I think the cell-ui minimizes early because it's status becomes set, incorrectly, at least momentarily, as "completed", even though I never hit `run`.
closes_early

coleshaw:
Is that on an existing session where you've previously hit run and gotten past this step? And now you're changing your previous selection? This is on the latest version of this branch -- you've pulled, restarted Vulcan, and webpack has recompiled / you've done a hard refresh?

I can't quite reproduce that behavior on my machine, using FireFox or Chrome. What browser are you using? What are the server-side logs or Dev Tools actions that happen when you click through -- it's weird that even though the status changes to completed, no further steps are pending / running?

It should never set as "completed" unless the value has been sent to the server, because the server determines what is completed or pending. I do see a similar interaction happen on my computer, if you go past that step, go back to change the values, and you pick a set of values that you've picked previously -- because of the cache, the records will immediately pop up without you clicking Run. But that doesn't quite look like what you're describing.

coleshaw:
I'm thinking what you're seeing is some version of this:

> I do see a similar interaction happen on my computer, if you go past that step, go back to change the values, and you pick a set of values that you've picked previously -- because of the cache, the records will immediately pop up without you clicking Run. But that doesn't quite look like what you're describing.

? In that you've picked all / all / all before, and because that value matches, it considers the step as "completed" even though you want to edit it. I think there's a tough tradeoff there to make -- I don't think it's a bug, rather which behavior we want. Currently the behavior is to try and help you "optimize" by showing you cached results automatically. Perhaps that is too distracting / trying-to-be-smart though, and we want the user to always have to click "Run" even if they've done that combination of input values before. :shrug:

dtm2451:
You are right!  I don't see this in a clean session (`make down`, cleared storage in the dev tools window for `Dev Vulcan` tab, checked out this branch and `git pull`, `make -C vulcan bash` then `exit` that after completion, `make down`, `make up`, then back in my `Dev Vulcan` tab, `ctrl`+`shift`+`R`).

It was doing this because there was a cached output of all-"all"s that I had previously `Run`.

I see similar behavior for any sets of selections that are cached.  Like below.
closes_for_cached_answers

coleshaw:
Yeah, so I think that's a "feature" :smile: If we decide that's a feature that we don't like, I'd propose to alter the behavior in a separate PR.

dtm2451:
:+1: I do wonder if this behavior might get confusing when we start sharing sessions?  Because a new person/collaborator won't know at all what has been previously run. But also so many things might change before then that Idk if it's worth worrying about!

Def would be a separate PR... my approval stands lol.

Also, to answer your prev Q, I'm in Chrome.

coleshaw:
Kind of interesting that in the first screencast, you don't see the "Interpret Record Selection" step turn into a green checkmark to indicate that you've moved on, after the "Record Selection" step turns completed and minimizes, but that does happen in the second screencast (which is what I see, too, and what I would expect given the cache-feature).

dtm2451:
I mayyyy have switched branches in between the original run and my screen grab, since that session was open for a while.  If only the `parse_record_selections.py` script changed, the server seeing the step as needing to be re-run is expected, no?

coleshaw:
:thinking: Hm...yeah, not sure what the behavior would be in that case. Certainly possible!

dtm2451:
:shrug: I think I must have actually `Run` my all-"all"s while a different version of the parsing script was current, cuz just swapping branches back and forth is not enough.

Anyway, I'll just take this is as reason to ensure I test for bugs with fresh sessions, or at least sessions where I know exactly what I've done!

Great work here, Cole :tada:

PR #379, opened 2021-05-13
Add per-cluster differential expression to UMAP

dtm2451:
Adds Differential Gene Expression by cluster vs every other cluster to the UMAP workflow & also sets the top10 genes to be shown & will also add a download button for the DGE data

lots of it are already working!
image

Need to add the download side next.
And perhaps also removal of ribosomal and mitochondrial genes in the selection of top-marks?


dtm2451:
I've worked in all the bits I'd planned!  That includes making the `ignore_prefixes` bit work across both humans and mice, as talked about in today's meeting, by making the underlying code case-insensitive.

vulcan/lib/server/workflows/umap.cwl
```diff
@@ -195,10 +200,10 @@ steps:
     run: scripts/calc_leiden.cwl
     label: 'Calculate Leiden clustering'
     in:
-      nn_anndata.h5ad: neighbors/nn_anndata.h5ad
+      nn_anndata.h5ad: calc_umap/umap_anndata.h5ad
```
dtm2451, line 203: I've rewired the calc_leiden step to take in the output from calc_umap in order for us to have an object containing all pieces at the end.

This will be important for the now-planned, separate, sc Differential Expression workflow as that workflow will take in the downloaded .h5ad, and so will need 'leiden' stored within the object, but will also aim to provide a grid of umap plots as an output option.

coleshaw (approved):

vulcan/lib/server/workflows/umap.cwl
```diff
@@ -223,7 +229,20 @@ steps:
   downloadRawData:
     run: ui-outputs/link.cwl
     in:
-      a: calc_umap/umap_anndata.h5ad
+      a: calc_leiden/leiden_anndata.h5ad
```
dtm2451, line 232: Connected with my last comment, I've re-wired the download button to output the leiden_anndata output.

vulcan/lib/server/workflows/umap.cwl
```diff
@@ -88,6 +88,11 @@ inputs:
     label: 'Number of iterations?'
     default: 0
     doc: 'The number of iterations for optimization - by default (0) either 200 for small datasets or 500 for large ones.'
+  6_Cluster_Differential_Expression__ignore_prefixes:
+    type: string
+    default: 'MT-,RPL,RPS'
+    label: 'Gene prefixes to ignore'
+    doc: 'A set of strings, separated by commas, for which gene symbols starting with these strings should NOT be shown in the umap overlay. Not case-sensitive, so the same strings will work for both human and mouse ribo & mito genes. Note: this does not affect the full differential expression table that can be downloaded.'
```
dtm2451, line 95: @coleshaw I noticed that this input does still get shown right from the beginning even though I did not add this primary input as a "mock" input to the first cell.  Perhaps that bug is fixed and we can remove that long list of unnecessary inputs from the first cell?

dtm2451:
Need to add method options to the ui before merge

coleshaw:
For the drop-down, if you are providing the options for that via JSON, you should be able to use `ui-queries/select-autocomplete.cwl` as the step's `run` option, for it to show up in the UI.

dtm2451:
Hmmm is it possible to set a default for `ui-queries/select-autocomplete.cwl`?  I'm alternatively thinking that we could have the input be a textbox where I state that "allowed options are 'logreg’, ‘t-test’, ‘wilcoxon’, ‘t-test_overestim_var'" in the (?) `doc`

coleshaw:
Hmm.....how would you specify that? In the JSON? I don't think the CWL has a good spot for that...hm.

coleshaw:
@dtm2451 , do you think this format would work for your cell that provides the options? If so, can you make an example so I can build around it?

```
{"default": "the-string-you-want-as-default", "options": ["list-of-options","plus","the-string-you-want-as-default"]}
```

dtm2451:
I suppose it could be via the json. Either the first value, or a second slot, perhaps...?  But I think we'll want a distinction between drop-downs with a default, that are therefore skip-able, versus when sometimes we will not want to fill in anything by default and will want to force the user to make a choice.

We could potentially set that up by making two separate `select-autocomplete*`s.  One with, one without defaulting.
Alternatively, we could set up for the JSON to be a nested list where options are given first and a default, optional, is given second. When only the first slot is given, the dropdown needs filling in... but how to document that.....

I'm really leaning towards the string / textbox approach!

coleshaw:
Yes, I was going to make a new component, like `select-autocomplete-default` and leave the current one with the existing behavior. I think we can just wrap it and provide a default if given in the JSON.

I'm not sure what you mean by the string / textbox approach...?

dtm2451:
For "string / textbox" I meant making the `dge_method` input be a primary input with `type: string` and `default: 'wilcoxon'`

coleshaw:
Ah, okay, and it's just up to the user to know what the valid values are, at this point? :+1: 

So I won't build any new components, if that's the way you're going to move forward.

dtm2451:
Yup, and also `doc: 'A string indicating what scanpy method option to use for calculating differential expression. Options are: "logreg", "t-test", "wilcoxon", "t-test_overestim_var". See documentation for "scanpy.tl.rank_genes_groups" for further details.` so users can know the options!

Additional perk of this method is that it still gets grouped with the other DGE-related input!  (The dropdown would have been its own separate thing, right?)

coleshaw:
Yup, a dropdown would have been separate because we don't have good support for CWL input options right now. Docstring looks good!

dtm2451:
> @dtm2451 , do you think this format would work for your cell that provides the options? If so, can you make an example so I can build around it?
> 
> ```
> {"default": "the-string-you-want-as-default", "options": ["list-of-options","plus","the-string-you-want-as-default"]}
> ```

If we do need this later, :+1:

PR #380, opened 2021-05-14
add clearer error messaging if timeout reached

coleshaw:
Was testing Vulcan production, and it seemed to error out with a 137 at various steps, depending on how many samples I had selected, though I'm not really convinced it was an out-of-memory (was monitoring on the server itself). I'm wondering if some steps just take so long that the container gets killed, and the error messaging to the user should be better.


dtm2451 (approved):
Seems to work!  I see this when I try to run with probably ~250k cells with otherwise all default parameters ("all" experiments, "lymphoid" and "myeloid" fractions, and "Blood" tissue.)

image

coleshaw:
Using this branch?

Yeah, there might be other resource limitation errors going on that aren't timeout or memory related, that we aren't catching and labelling correctly, too. I ran into that issue on the Regress step on production, too, but I don't think it's memory related (was < 5% on the server) nor time related (appeared after ~3 minutes). So not sure what else to track down, but we might be running into something else when regressing on a large number of cells. On the server, all the CPU cores are pegged for that step, so maybe something there.

dtm2451:
:facepalm: :facepalm: :facepalm: :facepalm: :facepalm: that was on Production & clearly also not even the new error message option that you added!

PR #381, opened 2021-05-14
remove output feed scroll

coleshaw:
Removing the code that scrolls the output feed down to the bottom, since it interferes with fetching data higher up in the pane. It's not really needed and while seeing the plot come into focus is nice, let's just remove it for now.


dtm2451 (approved):
Tested and this does allow me to scroll where I want *and* stay there.  :+1: 

I mean don't love that this throws away code you'd put together, but I suppose we could always revert this PR, and make further edits from there, if we do come up with workable implementation edits in the future!

Issue #382, opened 2021-05-14
UMAP workflow: Changing a "Regress by"-step parameter causes all steps to be re-run.


dtm2451:
The expected behavior would be to only need to re-run from the `regress_and_pca`/"Regress params and run PCA" step onwards.

I imagine this is because regression inputs, and many others, are attributed to the first step of the CWL (`projectData`/"Fetch project settings").  Cole tells me that this was implemented to get around some issue in the path-finding algorithm.  It seems to me that this original issue may be fixed, as putting `in: []` for this step still does seem to allows all inputs to render properly in the ui!
![image](https://user-images.githubusercontent.com/38870347/118317523-061f7980-b4ad-11eb-9dd5-8eed0901262a.png)

BUT doing so causes a different issue related to that empty input definition. Upon clicking `Run` (when the cwl has `in: []` for its first step) I get this error:
![image](https://user-images.githubusercontent.com/38870347/118317876-79c18680-b4ad-11eb-8cc3-e89b5e19e19d.png)
with logs:
```
vulcan_app_1      | Traceback (most recent call last):
vulcan_app_1      |   File "<string>", line 1, in <module>
vulcan_app_1      |   File "/app/archimedes/runner.py", line 332, in main
vulcan_app_1      |     input_files=[make_storage_file(s) for s in args.inputs],
vulcan_app_1      | TypeError: 'NoneType' object is not iterable
vulcan_app_1      | WARN:2021-05-14T19:03:19+00:00 archimedes-run failure: 
vulcan_app_1      | ERROR:2021-05-14T19:03:19+00:00 Error Failure to run archimedes-run command!  Check logs.
vulcan_app_1      | INFO:2021-05-14T19:03:19+00:00 Removing ecf11bde1d25b18766545f95ef8c9952219e9432 from running
```

Path forward:
- Perhaps we can set up the runner to skip adding inputs when given `[]`, and then we can remove them from the initial step?
- Or else perhaps there is a better way to set up the CWL for that step?  But if I remove the `in:` option for the step altogether, vulcan doesn't render at all:
![image](https://user-images.githubusercontent.com/38870347/118318426-3ca9c400-b4ae-11eb-861d-091d30368b82.png)

PR #383, opened 2021-05-14
archimedes runner handles steps with no inputs or outputs gracefully

coleshaw:
This PR allows the Archimedes runner to handle commands when no inputs or outputs are specified. This can happen if a CWL step in Vulcan does not have any listed. Related to #382, but does not actually fix the root cause of that issue. There still seems to be an implied "when a single primary input changes, all steps dependent on any primary input re-runs" instead of "when a single primary input changes, only steps dependent on that specific primary input re-run". I think this is because primary inputs are treated collectively when we construct the directed graph + dependencies -- not sure if we can break those up.

### DIFF
--- a/archimedes/archimedes/runner.py
+++ b/archimedes/archimedes/runner.py
@@ -329,8 +329,8 @@ def main():
 
     request: RunRequest = RunRequest(
         isolator=args.isolator,
-        input_files=[make_storage_file(s) for s in args.inputs],
-        output_files=[make_storage_file(s) for s in args.outputs],
+        input_files=[make_storage_file(s) for s in args.inputs] if args.inputs else [],
+        output_files=[make_storage_file(s) for s in args.outputs] if args.outputs else [],
         environment=args.env or [],
         script=(args.file and open(args.file, 'r').read()) or sys.stdin.read(),
         image=args.image
--- a/archimedes/archimedes/runner.py
+++ b/archimedes/archimedes/runner.py
@@ -321,16 +321,16 @@ def main():
     parser.add_argument('--file', help="The script file to run")
     parser.add_argument('--isolator', default='local', choices=['docker', 'local'])
     parser.add_argument('--image', default='etnaagent/archimedes:latest')
-    parser.add_argument('--input', dest='inputs', action='append', help="input files of the form name:/path/on/host")
-    parser.add_argument('--output', dest='outputs', action='append', help="output files of the form name:/path/on/host")
+    parser.add_argument('--input', dest='inputs', default=[], action='append', help="input files of the form name:/path/on/host")
+    parser.add_argument('--output', dest='outputs', default=[], action='append', help="output files of the form name:/path/on/host")
     parser.add_argument('-e', '--env', dest='env', action='append', help="environment variables of the form ABC=abc")
     
     args = parser.parse_args()
 
     request: RunRequest = RunRequest(
         isolator=args.isolator,
-        input_files=[make_storage_file(s) for s in args.inputs] if args.inputs else [],
-        output_files=[make_storage_file(s) for s in args.outputs] if args.outputs else [],
+        input_files=[make_storage_file(s) for s in args.inputs],
+        output_files=[make_storage_file(s) for s in args.outputs],
         environment=args.env or [],
         script=(args.file and open(args.file, 'r').read()) or sys.stdin.read(),
         image=args.image
--- a/vulcan/lib/server/workflows/umap.cwl
+++ b/vulcan/lib/server/workflows/umap.cwl
@@ -98,19 +98,7 @@ steps:
   projectData:
     run: scripts/umap-projects.cwl
     label: 'Fetch project settings'
-    in:
-      a: 1_Cell_Filtering__min_nCounts
-      b: 1_Cell_Filtering__max_nCounts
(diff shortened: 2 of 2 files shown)


archimedes/archimedes/runner.py
```diff
@@ -329,8 +329,8 @@ def main():
 
     request: RunRequest = RunRequest(
         isolator=args.isolator,
-        input_files=[make_storage_file(s) for s in args.inputs],
-        output_files=[make_storage_file(s) for s in args.outputs],
+        input_files=[make_storage_file(s) for s in args.inputs] if args.inputs else [],
```
corps, line 332: This works but is not robust -- you are forced to provide thi default anywhere args.input or arg.outputs is referenced.
It should be possible to set a default value to the arg parser to be an empty list when no options are set on the command line.

corps (changes requested):

corps:
> There still seems to be an implied "when a single primary input changes, all steps dependent on any primary input re-runs" instead of "when a single primary input changes, only steps dependent on that specific primary input re-run". I think this is because primary inputs are treated collectively when we construct the directed graph + dependencies -- not sure if we can break those up.

Yes and this is intentional, we should definitely _not_ change this behavior.  Instead, we can introduce the binary cache which would allow input level caching based on contents.

There is 1 level cache right now, which is more about _output layer_ caching.  That is, given a step's dependency output layers and script, we know that the resultant _output layer_ is the same.  This tells us immediately, given any primary input specification, which steps are already cached or not, including the last step or middle step, efficiently.  Note that with output layer caching one does not need to read files or even have access to those files to determine whether a step is done.

That said, there is a level 2 cache we should introduce, the binary cache, which addreses _input layer caching_.    _Input layer caching_ is, given a steps' dependency input _raw binaries_, we can calculate their hash to find a potential existing equivalent output layer.  In this case, finding the raw binary form of the input json and determining that it and all other input hashes have an existing built output layer is possible _at build time_.  This process is slower than output layer caching, and has some caveats, but will often at the very least greatly increase rebuild time.  It also requires access to raw files or metadata to calculate, which may not always be accessible (when we break the system up).  So it is left as a best effort optimization, not a core system assumption.

Primary inputs are a special case because there often isn't big outputs or complex logic involved in the transformation from the json values in the cwl.  That said, I feel that addressing this via binary caching in our first pass better since we know we need the binary cache for other purposes, and it doesn't involve inventing more specialized cases for the primary inputs step.

corps (approved):

dtm2451:
> > There still seems to be an implied "when a single primary input changes, all steps dependent on any primary input re-runs" instead of "when a single primary input changes, only steps dependent on that specific primary input re-run". I think this is because primary inputs are treated collectively when we construct the directed graph + dependencies -- not sure if we can break those up.
> 
> Yes and this is intentional, we should definitely _not_ change this behavior. Instead, we can introduce the binary cache which would allow input level caching based on contents.
> 
> There is 1 level cache right now, which is more about _output layer_ caching. That is, given a step's dependency output layers and script, we know that the resultant _output layer_ is the same. This tells us immediately, given any primary input specification, which steps are already cached or not, including the last step or middle step, efficiently. Note that with output layer caching one does not need to read files or even have access to those files to determine whether a step is done.
> 
> That said, there is a level 2 cache we should introduce, the binary cache, which addreses _input layer caching_. _Input layer caching_ is, given a steps' dependency input _raw binaries_, we can calculate their hash to find a potential existing equivalent output layer. In this case, finding the raw binary form of the input json and determining that it and all other input hashes have an existing built output layer is possible _at build time_. This process is slower than output layer caching, and has some caveats, but will often at the very least greatly increase rebuild time. It also requires access to raw files or metadata to calculate, which may not always be accessible (when we break the system up). So it is left as a best effort optimization, not a core system assumption.
> 
> Primary inputs are a special case because there often isn't big outputs or complex logic involved in the transformation from the json values in the cwl. That said, I feel that addressing this via binary caching in our first pass better since we know we need the binary cache for other purposes, and it doesn't involve inventing more specialized cases for the primary inputs step.

I may have a caveat to that:
In #379, I added a primary input, but did not add it as an input of cell#1.  If I change that primary input, nothing upstream gets re-run, just the Differential Expression step where that input is used.  I know for a fact I didn't follow (yet) everything that you just said Zach, but I'm not sure if the primary input interpreter follows the intended logic?

corps:
Right.  That is the limitation of output layer caching.  _It doesn't matter that the input layer of the step is the same_ it only matters that the primary input's output layer is not the same.  That is our primary cache level 1.  

We can improve the situation by adding a level 2 cache that is best effort.

I'm not in favor of treating the primary input layer different than other steps, and I'm not in favor of discarding the concept of output layers.

corps:
Also I want to be clear -- if things rebuilding is a big problem, we should not crutch too hard on the caching layer's logic.  The cache should be at best a way to increase the performance of iterative exploration.  At some point, I highly believe we must export built data "as canonical" (say saving it to metis) and make other steps reconsume it.  This basically means that expensive steps are not rebuilt on a whim but are deliberately versioned and handled in a special way.

There are other reasons why this approach will be necessary -- in addition to various caching layer logic, we will need to introduce reaping logic that enforces space constraints and include the possibility that _at any given time_ a previously built, cached data payload could disappear.  Anything that _needs_ longevity needs a deliberate mechanism for storage and reconsumption.

PR #384, opened 2021-05-18
Update magmaR for CRAN resubmission, v1.01

dtm2451:
magmaR is back in CRAN.

The CRAN team requested that all code utilizing [`Suggests:`](https://github.com/mountetna/monoetna/blob/master/etna/packages/magmaR/DESCRIPTION) packages be made conditional upon package availability, even if it was only in the vignette.

I'd missed their deadline for addressing that issue, :facepalm:, but it's corrected, resubmitted, re-accepted, and fully back https://CRAN.R-project.org/package=magmaR.

Additionally, I'd noted that some of the functions failed when tests were run in earlier versions of R, so I have added an `R (>=4.0)` version requirement to ensure they don't pester about that next.  Might be worth addressing the code to remove that requirement, but I'd rather focus elsewhere for now.

### DIFF
--- a/etna/packages/magmaR/DESCRIPTION
+++ b/etna/packages/magmaR/DESCRIPTION
@@ -1,12 +1,13 @@
 Package: magmaR
 Type: Package
 Title: R-Client for 'Magma' of the 'UCSF Mount Etna Data Library'
-Version: 1.0.0
+Version: 1.0.1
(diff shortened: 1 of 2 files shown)


dtm2451:
The "make code conditional, [clarified: even in the vignette]" ask is still odd to me cuz really the other 6 `Suggests:` packages are all used in the vignette / elsewhere in build materials, but their infrastructure can't be made conditional.  Anyway, it's done and I'm gonna merge this in.

PR #386, opened 2021-05-19
prep UMAP scripts & cwl for manual cluster annotation (WIP)

dtm2451:
1. Initializes a dictionary of cluster ids (keys) + new names (values; initialized as the ids) within the `calc_leiden` cell.
2. Adds a cell referring to a `multiinput-string` ui-query where users might update those if inclined to do so. Idea: https://www.figma.com/file/m0meOQSzYdZNm0c9YG39fH/Annotation?node-id=0%3A1
3. Adds this 'Manual Annotations' `color_by` option.
4. Moves `color_by` option collection from `calc_umap` to `calc_leiden` script in order to hopefully synchronize the materialization of the new ui-query with the color_by option selection.  Then, a single `Run` click says yes to both!


vulcan/lib/server/workflows/scripts/calc_leiden.py
```diff
@@ -4,10 +4,36 @@
...
+clusts = scdata.obs['leiden'].tolist()
+
+output_json(clusts, 'leiden.json')
+output_json(
+    dict(
+        [
+            str(clust),
+            str(clust)
+        ] for clust in range(max(clusts)+1)
+    )
```
coleshaw, line 23: I think the tests are failing because of a missing `,` here?

vulcan/lib/server/workflows/scripts/plot_umap.py
```diff
@@ -41,6 +41,9 @@ def get(ids, value):
         ] for clust in list(sets.keys()) ])
     hover_name = 'top10 markers'
     hover_text = [texts[str(val)] for val in leiden]
+elif color_by == 'Manual Annotations'
```
coleshaw, line 44: I think tests are failing because of a missing `:` at the end of this line.

vulcan/lib/server/workflows/scripts/calc_leiden.py
dtm2451: :+1: thanks!  maybe I over-talked the completeness of these bits... zero testing so far both here and the #385 scripts, so typos and incorrectly assumed python logic are erm *likely*.

coleshaw:
> Biggest issue now is that it is incredibly slow -- to the point of being non-responsive and missing keyboard inputs unless they are typed quite slowly. I'm not sure why?

There's probably a recursive re-render introduced because of the component self-updating state -- this is what we're trying to avoid with the new validation context. Can you try merging in my branch `cs/vulcan-validation-context` to see if the behavior improves? You'll have to add a validator to the new component in `user_input.tsx` -- probably the `AllInnerValuesNotEmptyValidator` one. You also might have to update your component to avoid the re-rendering (you can see the updated MultipleMultiselect as an example of how to simplify the component). Ping me if you have any questions or want to pair on it.

I'm just working on specs for my new branch, so the validation stuff should work. Hope to put up a PR for that today so it gets merged in soon.

coleshaw:
> Biggest issue now is that it is incredibly slow -- to the point of being non-responsive and missing keyboard inputs unless they are typed quite slowly. I'm not sure why?

Oh, I think for this you might want to look at the `debounce` pattern: https://github.com/mountetna/monoetna/blob/master/etna/packages/etna-js/components/inputs/slow_text_input.jsx#L22

You don't want React to try updating state on every key stroke, it's too slow. So adding a debounce lets React only update state after some amount of time, to capture "chunks" of typing.

dtm2451:
Hmm, pairing is definitely a good idea!  The way this component works is VERY different right now from all the other input components you made and that should probably be fixed?  But I'll see if I can work in debouncing in the meantime for practice and cuz I it does seem the fix that's needed.

corps:
> Hmm, pairing is definitely a good idea! The way this component works is VERY different right now from all the other input components you made and that should probably be fixed? But I'll see if I can work in debouncing in the meantime for practice and cuz I it does seem the fix that's needed.

For sure!  As a rule of thumb,

For Text Inputs,
When the OnChange has a wide scale impact (changing inputs in the top level state, trigger a lot of refresh)
Debounce it or buffer it (for instance, with onBlur)

When the OnChange has a local impact (triggering local validation)
Let it floowww, let it floowwww


Welcome to data flow management, it never gets easy :)

dtm2451:
Seems to be working!

And without the need for debouncing I think... it's responsive && ...

vulcan/lib/client/jsx/components/workflow/user_interactions/inputs/multiple_string.tsx
```diff
@@ -0,0 +1,57 @@
...
+    // Otherwise set to mirror input.value
+    if (null == input.value) {
+      onChange(name, options);
+    }
+  }, [options, input.value, name, onChange]);
+
+  const updateLabel = (newValue: string, key: string, prevLabels = input.value) => {
+    prevLabels[key] = newValue;
+    onChange(name, prevLabels);
+    console.log(input.value);
```
dtm2451, line 36: This console.log only seems to be called once or twice per keystroke, so seems there's little recursion introduced!

The use of input.value rather than direct state seems to have been enough =)

dtm2451: (I do plan to remove this before merge.)

vulcan/lib/client/jsx/components/workflow/user_interactions/inputs/multiple_string.tsx
```diff
@@ -0,0 +1,57 @@
+import React, {useState, useMemo, useEffect, useCallback, useRef} from 'react';
+import * as _ from 'lodash';
+
+import {InputBackendComponent} from './input_types';
+import TextInput from 'etna-js/components/inputs/text_input';
```
dtm2451, line 5: I don't understand why this line is causing the vulcan tests to fail.  AFICT, this line is the same format as all the other etna-js component imports?

coleshaw: Oh, you'll need to edit `etna.d.ts` and add a type definition for that component, if it hasn't been used before

coleshaw: Basically giving typescript a hint for the parameter types and return format. There should be examples in there you can follow.

vulcan/lib/client/jsx/components/workflow/user_interactions/inputs/multiple_string.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+function equalKeys(hash1: {}, hash2: {}): boolean {
```
coleshaw, line 7: super nit: I think you can remove this function now, since it's cleaned up?

Also, if you want this to validate the annotation inputs as empty string is not allowed, you'll want to add that comparison to `selectors/workflow_selectors.ts/inputValueNonEmpty`, here: https://github.com/mountetna/monoetna/blob/master/vulcan/lib/client/jsx/selectors/workflow_selectors.ts#L112. Currently if you leave it as empty string, it would pass that along to the server -- not sure if you want that behavior or not?

dtm2451: good nit!

For the validation, I'm not sure how widespread the `inputValueNonEmpty` use is, but I think we will want to allow selections of `''` in the future for other things.  Maybe we could pair briefly on the ideal way for me to set up only sometimes allowing `''`?

dtm2451:
I suppose I don't actually want to hold this PR up until after we update automated running vs additional `Run` button workings.  The annotation functionality is usable, just slightly annoying.

Once the validation Q is settled, I'm going to merge.

dtm2451:
Validation is set up and unit-tests (for both the new validator and the component) are set up and working =).  Planning on merging once tests run!

PR #389, opened 2021-05-21
Janus flags editor

coleshaw:
This PR adds a "flags" management page to Janus, for superusers only. This should help us add feature flags to user tokens for use with Vulcan, not only for access but I can imagine for things like resource level control (i.e. user A can consume more CPU than user B).

The page should render with a list of all users, their projects, and their current feature flags. You should be able to filter based on text regex (across name + email), a set of projects, or a set of current flags. You can add or remove a flag from a set of users, at the same time. For example, I can imagine filtering for all users of HuMu, and then selecting the specific users we want and adding flags for them.

Apologize for the styling, it is basically out-of-the-box Material UI. Does not quite match our current styling, and some Janus global style rules make it "less pretty", but we can work on that.


janus/lib/client/jsx/etna.d.ts
```diff
@@ -0,0 +1,153 @@
+// TODO: In the future, we'd want to convert etna to typescript, or atleast type its .d.ts files there, rather
```
coleshaw, line 1: Basically a slightly modified version of the one from Vulcan. Would definitely reduce copy / paste if we add in a standard version to `etna-js` itself...

janus/lib/client/jsx/flags/flags-context.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+export type FlagsContextData = typeof defaultContext;
+export const FlagsContext = createContext(defaultContext);
+export type FlagsContext = typeof FlagsContext;
+export type ProviderProps = {
+  params?: {};
+  children: any;
+};
+
+export const FlagsProvider = (
```
coleshaw, line 26: @corps , when we reviewed the Vulcan UI changes you made, you mentioned not using Redux with a Context. Is this approach more like what you  meant? This way we avoid using `dispatch` and the async nature of it, just keep the `state` via `useState` in the Context?

corps: Yeah, i like this approach alot.  You can even choose to expose 'action' functions.  So instead of putting setState or setProjects into the context, you could `useCallback` and create little functions that modify the state like an action, and pass those to the context only if you preferred.  Many options for organizing.  This works great.

janus/lib/client/jsx/flags/flags-context.tsx
```diff
@@ -0,0 +1,52 @@
...
+  props: ProviderProps & Partial<FlagsContextData>
+) => {
+  const [state, setState] = useState(defaultFlagsState);
+  const [projects, setProjects] = useState([] as Project[]);
+  const [users, setUsers] = useState([] as UserFlagsInterface[]);
+
+  useEffect(() => {
+    setState(props.state || defaultFlagsState);
```
corps, line 34: Any reason not to just have `useState(props.state  || defaultFlagState)`?

janus/lib/client/jsx/flags/flags-flags-cell.tsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+
+import Chip from '@material-ui/core/Chip';
```
corps, line 3: Cool.  Can't wait for us to get this into etna-js too :o)

corps (approved):

janus/lib/client/jsx/etna.d.ts
corps: Yeah, I want to work on getting etna-js to typescript sometime soon.

graft:
I am excited about the introduction of material-ui, I'm confident we can wrestle it into some other shapes over time.

A few UI things I noted, maybe immaterial given it is superuser only and they are only nuisances:
- When I add a flag like 'test one', it tells me: "Flags should be an array of words", similarly with 'test-one'. Perhaps a format hint of some kind in the error would be helpful.
- Selecting a user brings up the "add flag" modal; dismissing this modal seems to require deselecting all users, which is not obvious.

PR #394, opened 2021-05-23
Save & load params

graft:
This is a mostly-working WIP - basically, you can save and load your session in the JS client/local storage to and from a json file on disk. This works okay (especially if you reload the page).

However, there is some confusion remaining about UI state to be resolved which can result in persistence of old session data upon loading. I've traced this down to an issue in `stepInputDataRaw` but need some help sorting this out.

### DIFF
--- a/vulcan/lib/asynchronous_scheduler.rb
+++ b/vulcan/lib/asynchronous_scheduler.rb
@@ -113,7 +113,7 @@ class Vulcan
 
         started = []
 
-        runnables.each do |build_target|
+        runnables.each do |step_name,build_target|
           if schedule!(orchestration: orchestration, build_target: build_target, token: token, storage: storage)
             started << build_target.cell_hash
           end
--- a/vulcan/lib/client/jsx/components/workflow/session/primary_inputs.tsx
+++ b/vulcan/lib/client/jsx/components/workflow/session/primary_inputs.tsx
@@ -30,7 +30,7 @@ export default function PrimaryInputs() {
               workflow.inputs[name].default
           ].find(a => a != null)
     }))
-  }, [session]);
+  }, [workflow, session]);
 
   let groupedInputs = primaryInputs.reduce((result, input) => {
     let groupName = inputGroupName(input.name);
--- a/vulcan/lib/client/jsx/components/workflow/session/primary_inputs.tsx
+++ b/vulcan/lib/client/jsx/components/workflow/session/primary_inputs.tsx
@@ -10,6 +10,7 @@ import {InputSpecification} from "../user_interactions/inputs/input_types";
 export default function PrimaryInputs() {
   const {state, dispatch} = useContext(VulcanContext);
   const {workflow, session} = state;
+
   if (!workflow) return null;
   if (!workflow.inputs) return null;
 
@@ -29,7 +30,7 @@ export default function PrimaryInputs() {
               workflow.inputs[name].default
           ].find(a => a != null)
     }))
-  }, []);
+  }, [session]);
 
   let groupedInputs = primaryInputs.reduce((result, input) => {
     let groupName = inputGroupName(input.name);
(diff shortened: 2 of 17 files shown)


vulcan/lib/asynchronous_scheduler.rb
```diff
@@ -113,7 +113,7 @@ def schedule_more!(token:, storage:, orchestration:)
 
         started = []
 
-        runnables.each do |build_target|
+        runnables.each do |step_name,build_target|
```
coleshaw, line 116: Just curious, did you mean to use this variable in the block?

vulcan/lib/client/jsx/reducers/vulcan_reducer.ts
```diff
@@ -82,7 +82,7 @@ export default function VulcanReducer(state: VulcanState, action: VulcanAction):
       return {
         ...state,
         session: action.session,
-        inputs: {...state.inputs, ...action.session.inputs},
```
coleshaw, line 85: Are you sure you want to remove the `...state.inputs` bit? I think the original intent was to preserve inputs that you may have entered that have not been saved in the local storage session, yet. I'll play around with it in a bit to observe behavior, but seems like this may change how that works.

coleshaw:
Are we trying to save some SHA values from the server, too, not just the values stored in Local Storage in the browser?

coleshaw:
nit: might it be useful to also save the project name with the workflow name, in the downloaded JSON? Also not sure if a timestamp would be helpful as well.

vulcan/lib/client/jsx/reducers/vulcan_reducer.ts
coleshaw: Can you clarify what you mean by UI confusion issues when you load settings, and the exact behavior you saw? I was able to reproduce something, but putting this back to include `...state.inputs` seemed to fix what I was.

Here are the steps I took that led to UI confusion:

1) Ran the UMAP workflow until the "Color by" option appeared.
2) Loaded up a JSON file with only the default primary inputs, with one value changed for # of UMIs.
3) UI reset back to the three "Record Selection" inputs. Run button was active, and the UI widget showed the three previous values that I had selected.
4) Clicking RUN didn't work. I think because even though the UI widget had values, those values had not been translated into state (overwritten by my JSON upload).
5) Had to reset and then re-select things in the widget to continue on.

Is that what you mean by UI confusion?

However, adding the above back into the reducer allowed me to:


1) Ran the UMAP workflow until the "Color by" option appeared.
2) Loaded up a JSON file with only the default primary inputs, with one value changed for # of UMIs.
3) UI reset back to the three "Record Selection" inputs. Run button was active, and the UI widget showed the three previous values that I had selected.
4) State automatically continued back to the "confirm record selection" step, saw that the same 4 values were present and selected, and continued running the workflow beyond that (to filtering, PCA, etc.), with no additional user interactions.

Not sure if the latter behavior is what you expect, when you upload a new JSON file?

vulcan/lib/asynchronous_scheduler.rb
corps: It's an adaptation for a change we made else where, this is correct.  It used to iterate on just the target, now it is a tuple.

vulcan/lib/client/jsx/components/workflow/session/primary_inputs.tsx
```diff
@@ -29,7 +30,7 @@ export default function PrimaryInputs() {
               workflow.inputs[name].default
           ].find(a => a != null)
     }))
-  }, []);
+  }, [session]);
```
corps, line 33: Interesting find.  This memo should technically also include `workflow`.  It should read `[workflow, session]` since both of those values are used from the closure of the inner useMemo function.

PR #396, opened 2021-05-24
Vulcan gate via flags

coleshaw:
Not sure how we'd like the Vulcan "feature flag" behavior to work on user denial -- this PR is for a "hard deny" of any users who do not have a `vulcan` feature flag attached in Janus. What I mean is that upon loading the main vulcan page, the user will receive a 403 when the client tries to load all the workflows, and the user sees a message like "You are not authorized for this action" on top of an empty page. They basically have nothing to look at or interact with. This is purely server-side authorization checks on the routes but blocks all access to the Vulcan app.

Another alternative, if we want a more "soft deny" is to load the workflows, and a user can view all the workflows they have access to (currently the cards on the dashboard). But when they try to click into a workflow and run it, they would be denied and see a message at that point. Not sure if this would be more frustrating, since it would initially seem like you have access, but then, not really... To present this to the user in a nice way, I think this approach might also lead to proliferation of the `vulcan` flag in the UI code and the server code, which seems messy and I'd like to avoid.

Thoughts?


etna/lib/etna/route.rb
```diff
@@ -148,7 +148,10 @@ def user_authorized?(request)
       params = request.env['rack.request.params']
 
       @auth[:user].all? do |constraint, param_name|
-        user.respond_to?(constraint) && user.send(constraint, params[param_name])
+        user.respond_to?(constraint) && (
+          params.has_key?(param_name) ?
+            user.send(constraint, params[param_name]) :
+            user.send(constraint, param_name))
```
graft, line 154: Maybe safer to do this on `param_name.is_a?(Symbol)` rather than params.has_key?, since that is subject to the user input.

PR #397, opened 2021-05-26
Vulcan UI fixes, linter, scheduler fix

corps:
This gets eslinter with the hook checker in and fixes many many hook dependency arrays.
It also:

1.  Puts node_modules into a volume, hopefully reducing impact to developers when permissions go weird.
2. Adds a WorkflowContext that can be used to get a non-nullable workflow value out (because that context is only mounted after workflows have loaded).
3. Fixes scheduler issue so that it will never report complete for a step until it has alteast attempted to start dependent steps downstream.

Still testing locally.


docker/archimedes-base/entrypoints/build.sh
```diff
@@ -24,7 +24,6 @@ if ! [ -e config.yml ] && [ -e config.yml.template ]; then
   cp config.yml.template config.yml
 fi
 
-poetry completions bash > /app/poetry.completion
```
corps, line 27: This was what caused the excessively archimedes rebuilding, sadly.  This ran on every build, which updated the source file, which required a new rebuild everytime.

vulcan/lib/asynchronous_scheduler.rb
```diff
@@ -45,18 +45,23 @@ def running?(build_target)
     def status(storage:, build_target:, step: nil)
       hash = build_target.cell_hash
 
-      requires_inputs = step&.ui_output_name && !build_target.is_buildable?(storage)
```
corps, line 48: This is the bug fix that ensures that we never report "complete" for a step just because it is built -- we ensure it also removed from the `@running` hash, which only ever occurs after we have attempted to schedule more dependent tasks.  Check the `ensure` block below.

docker/archimedes-base/entrypoints/build.sh
coleshaw: Does that mean no poetry autocompletions in the Archimedes container now, or that is still available if you do `make -C archimedes bash`?

vulcan/lib/client/jsx/components/workflow/session/input_group.tsx
```diff
@@ -10,18 +10,19 @@ import {
   sortInputsByLabel
 } from '../../../selectors/workflow_selectors';
 import {InputSpecification} from "../user_interactions/inputs/input_types";
+import {WorkflowContext} from "../../../contexts/workflow_context";
 
 export default function InputGroup({inputs, onChange}: {inputs: InputSpecification[], onChange: (name: string, val: any) => void}) {
   const {state} = useContext(VulcanContext);
-  let {session, workflow, status, data} = state;
+  const workflow = useContext(WorkflowContext);
```
corps, line 17: the new workflow context allows us to fetch non null workflow.  It will not mount any child component of workflow manager until workflow is non null.  this removes a lot of annoying checks.

coleshaw:
Should we have ESLint also run in CI, or that'll be up to us for now to enable in our IDEs?

docker/archimedes-base/entrypoints/build.sh
corps: No, it still works.  This actually exists in the source directory anyways thanks to the mount point in docker-compose dropping it.  So it gets checked in.  what it does mean is that if we want updated autocompletions, we just have to rerun it inside the docker-compose bash shell and commit it.

corps:
> Should we have ESLint also run in CI, or that'll be up to us for now to enable in our IDEs?

Good call, can add to CI as well.  If we keep the linter very conservative (only serious issues, not every stylistic nit pick) I'm ok keeping it in CI to catch real issues.

I'm definitely against using one of the big presets as I feel it would result in a lot of broken CI builds due to silly things (whitespace) and slow us down.

corps:
@graft @coleshaw Take another look?  In particular, worth examing my refactor of https://github.com/mountetna/monoetna/pull/397/files#diff-6917f93a03f512ce6894166b9075776d41af63451a951badba28e518dd9347b5

1.  For query components that are not using the "named argument" part of the CWL input (as in, it really just takes in a single argument and it doesn't matter the name), the pattern we can use is to flatten the data dictionary down into a single dictionary with labels, merging them as we go.  "What about label conflicts?"  Theoritically possible, in reality I don't think this will happen as the main use case is to consume just one, anonymous argument, but CWL models everything as 1 or more named arguments.  I've thought about introducing proper type classes (for instance, the `Monoid` would really simplify this) in the future to make this kind of argument flattening natural and "for free", but not this time.

2. "Source of truth" of state is tricky, so let's clarify how this should work:
   a.  `state.inputs` is the "source of truth" for what the valid input states are in the browser, and includes a null value for either invalid inputs or inputs that have no default value yet.  Components may want to buffer partial, incomplete selections without pushing to state.inputs via onChange.  In that case, to keep those in sync, _make sure to keep referential integrity_ between the input state and your buffered local state.  Check for a valid change to sync back from state.inputs by comparing the references.    In a future PR, this will get simpler when we extract the notion of "invalid inputs" out so that most components do not need to buffer, and can use the state.inputs for both valid and invalid values.

3.  Use more `useCallback` to direct imperative behavior, rather than a web of `useEffect`.  `useEffect` is very useful _in small numbers_ but large webs of them are worse than just an explicit callback with logic.  Rule of thumb, again: `useEffect` is for *I always want this behavior when X and Y change*.  `useCallback` is for *I sometimes want this behavior, and I want control of exactly when that sometimes is*.

coleshaw:
> For query components that are not using the "named argument" part of the CWL input (as in, it really just takes in a single argument and it doesn't matter the name), the pattern we can use is to flatten the data dictionary down into a single dictionary with labels, merging them as we go. 

I think this means that we also need to slightly update the `parse_record_selection.py` cell to not use the sub-input  name, to not break the current workflow?

corps:
> single

I actually already did this, and it was pretty simple.

vulcan/lib/server/workflows/scripts/parse_record_selections.py
```diff
@@ -6,8 +6,7 @@
 pdat = input_json("project_data")[project_name]
 selection_atts = pdat['selection_options']
 
-# because a is the input name from the previous CWL input step
-selected = input_json('selected_options')["a"]
+selected = input_json('selected_options')
```
corps, line 9: @dtm2451 @coleshaw @graft This is the change in the workflow to support the flattened data approach.

coleshaw:
I've noticed that this branch constantly polls, whereas the previous behavior was that it would only poll on Run. Is that to be expected?

corps:
> I've noticed that this branch constantly polls for status, whereas the previous behavior was that it would only poll on Run. Is that to be expected?

Looking into it!  I'm definitely still slightly adjusting things here, I'm manually testing things as I go.
It should ONLY poll after run, so there is a slight but.  gonna write a test, too.

corps:
@coleshaw @graft I created a helper hook that "solves" the general state problem we have (components want local, buffered values, but also should update from the source of truth `state.inputs`): https://github.com/mountetna/monoetna/pull/397/commits/a883e72e05758a554181319b886c763a1306cb61

Basically, components can alter and control when their local buffered input state reaches `input.state` via `onChange`, but only when the `input.state` has a unique new value, that should propogate down to the local buffered value.

With one consistent hook implementation for this behavior, other components become simpler.

PR #398, opened 2021-05-26
(WIP) Vulcan UI widget for single dropdown with checkboxes

coleshaw:
This PR adds the UI widget for a single dropdown + checkboxes, for the differential gene expression workflow.

While this does work, @corps  has indicated that it's not a performant solution and we may want to re-think how the front end handles data loading and syncing in general, to avoid recursion issues (which would clean up a lot of the complex widgets). So leaving this PR WIP for now, until we sort that out.


corps:
Taking a look at this, I'll have some thoughts to share soon.

coleshaw:
Going to implement this component on a different branch.

PR #399, opened 2021-05-26
Relax the temporary id regex for censor

corps:
I _think_ this could be improved, but it got me thinking --

the censor hook is actually called _before_ upsert.  If it was called _after_ upsert, we could actually still just do a check (either exchange the temp ids out with real ones, or even just a ::first check for tables), but right now it happens in this order which means new records aren't yet visible.

Which makes me think -- does this prevent creating new records to a child relationship of a restricted model right now?  My experiments showed no, we don't.  

I think the long term fix is that we hold an open transaction, do the upsert, _then_ run the censor query within that transaction, validate, and then commit.  But that is a big refactor.  We should add this to part of our data ingestion work.

### DIFF
--- a/etna/etna.gemspec
+++ b/etna/etna.gemspec
@@ -1,6 +1,6 @@
 Gem::Specification.new do |spec|
   spec.name              = 'etna'
-  spec.version           = '0.1.36'
+  spec.version           = '0.1.37'
   spec.summary           = 'Base classes for Mount Etna applications'
   spec.description       = 'See summary'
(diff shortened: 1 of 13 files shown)


graft:
Since in order to attach a table the revision must include the `parent` attribute, for records with a temp id, we could simply change record_names to be the list of parent identifiers (gleaned from the revisions) and the model name to be the parent model. I think this would take care of searching the restriction correctly and prevent updates to tables for restricted records.

graft (approved):

Issue #402, opened 2021-05-28
Material UI theming


graft:
This is still a WIP, but looks mostly as intended.

I added a base theme for material-ui in Janus and made use of it throughout the application. Fortunately Janus only has a few view components so this was not so bad; elsewhere it will take time to propagate. The conversion to material-ui components is pretty easy in general.

Mostly what the theme does is set some base component props and styles to make things look more etna-ish (mostly the amount of rounding and disabling elevation). I also used disableRipple which seems to suppress many irritating animations.

I have not yet, but will before closing this, move the `createMuiTheme` stuff into etna-js, so that other etna apps can call `etnaTheme(primary, secondary)` to generate a consistent theme (although I won't attempt to extend this beyond Janus for the moment).

A few issues I had:
- I couldn't hang on to the old button styling in Janus, because I am not sure how to inject css that depends on the "primary" theme variable. Buttons will consequently look different.
- For some reason ScopedCssBaseline was breaking text boxes and had to be axed, perhaps it is incompatible with ThemeProvider?
- The alignment remains a little weird between the flags controls and the flags table. Unfortunately this seems hard to control, because of grid vs. table differences.

PR #403, opened 2021-05-28
Graft/use themes

coleshaw:
I think this is the PR attached to #402?


coleshaw:
I think it looks good, @graft . Once you get the theme into `etna-js`, I'll use it on the Query mockup I'm working on, too, so it'll look consistent.

PR #404, opened 2021-06-02
Addressing test, fixing issue with polling

corps:
Ok, I spent most of the day trying to get deeper testing of this, but I'm realizing there's still a few issues with the way our testing in react + promises works that is frustrating.

Mostly, react testing requires us to wrap a lot of promise calls in `act` calls, which... gets tricky.  We can't just use a hook in the global Promise constructor because frustratingly `jest` itself needs the normal Promise constructor behavior for other purposes, too.  

What we need is a more disciplined dependency injection like approach that would allow us to inject asynchronous wrappers into the contexts so that all our fetches and promises and jazz would magically get the `act` wrapper, but not the code unrelated to our code.

I could also do with reducing the amount of debug output in tests right now, so working on that bit by bit.

For now, this atleast fixes the polling behavior you saw.

### DIFF
--- a/etna/packages/etna-js/spec/helpers.js
+++ b/etna/packages/etna-js/spec/helpers.js
@@ -5,6 +5,12 @@ import {appSubscription} from "../utils/subscription";
 
 export const mockStore = (state, middleware = []) => configureMockStore([thunk, ...middleware])(state);
 
+function nockDebug(...args) {
+  if (process.env['NOCK_DEBUG'] === '1') {
+    console.log(...args);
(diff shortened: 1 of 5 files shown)


coleshaw:
Looks good for polling, but the UI widgets seem to behave oddly. I'll try merging this into my branch and see if the validation context fixes the UI portion.

coleshaw (approved):
Thanks! I think this will help with the UI interactions. I'll use it in my branch and see if I can make more progress.

PR #405, opened 2021-06-03
Vulcan validation context

coleshaw:
This PR changes how the UI input widgets validate the user selection, by introducing a validation context. This helps clean up a lot of the internal logic in the complicated UI widgets.

I added the single-dropdown-multicheckbox that @Gvaihir and @dtm2451 wanted to use for differential gene expression, so that should be available once you start working on that workflow.

Also fixes some minor bugs, renames `input.default` to `input.value` for clarity, and adds some specs.


graft:
This seems to fix the outstanding UI issues allowing me to run to the end of the workflow for the first time in a while, which is fantastic.

The behavior of the "Run" button remains a little odd. For example, if I get to the "Color By" step and have not selected a value, I can hit "Run", but then am told that my selection is incorrect, and I can't "run" after all. This is not wrong per se, but there is no hint that this will happen, and especially if an input drawer is collapsed it may not be obvious that you are not ready to run yet. It might be nice if instead the component (and the containers if collapsed) hinted its own validation status, and maybe even if the "run" button had another state to indicate that it can't run until inputs are resolved.

There is also some oddness around loading params and cache discrepancies, but I think this will involve a lot more thought to resolve. The validators depend on input.data, which results in some odd behavior if there are conflicts between the cached state and params. E.g., if a file in params drops out of the cached state, the UI will not show the non-existent file, but hitting "Run" will result in a complaint about an invalid input that you are unable to resolve.

Regarding the validation functions themselves, my main concern is that the validation function is somewhat opaque and may miss edge cases. I feel like eventually we ought to use something like json-schema here, which might give us the hope of versioning the component behavior somewhat. Since these json outputs are an essential component of the session params, this all has implications for how saved documents are versioned eventually, and our ability to faithfully reload those documents correctly. The brittleness of this arrangement across time is a real plum.

etna/packages/etna-js/components/inputs/dropdown_autocomplete.jsx
```diff
@@ -67,17 +68,25 @@ export default function DropdownAutocomplete({
 
   useEffect(() => {
     onChange = debounce(onChange, waitTime || 500);
-    setSelectedValue(defaultValue ? defaultValue : '');
+    setSelectedValue(defaultValue ? defaultValue : null);
```
corps, line 71: Do we feel comfortable with this change in general?  The other DropdownAutocomplete uses do not depend on this?

vulcan/lib/client/jsx/components/workflow/session/__tests__/session_manager.spec.tsx
```diff
@@ -0,0 +1,184 @@
...
+import {
+  setDownloadedData,
+  setStatus,
+  setWorkflow,
+  setWorkflows,
+  addValidationErrors
+} from '../../../../actions/vulcan';
+
+describe('SessionManager', () => {
+  const workflow = createWorkflowFixture({
```
corps, line 24: Cool!  I intend to actually improve these fixture functions and give them a nicer "builder" like interface with some sane defaults and the ability to do things like autocompute the `dependencies_of_outputs` inner piece.

vulcan/lib/client/jsx/components/workflow/session/session_manager.tsx
```diff
@@ -82,15 +85,35 @@ export default function SessionManager() {
     [state.inputs, workflow]
   );
 
+  const hasValidationErrors = useMemo(
```
corps, line 88: In cases where the computation itself is fairly cheap, it's actually totally ok to just `const hasValidationErrors = Object.keys(state.validationErrors).length > 0`.  This saves complexity in the code as well.  useMemo is definitely most useful when it is important to keep the instance stable or to save against hard computations, but simple predicates like this are also totes fine as a regular old variable :P

vulcan/lib/client/jsx/contexts/session_sync.tsx
```diff
@@ -47,7 +49,7 @@ export function useSessionSync(
     return () => cancellable.cancel();
   }, [lastPollingRequest, dispatch, pollStatus, postInputs, state]);
 
-  const statusIsFresh = lastCompletedPollingRequest === state.current.session.inputs;
+  const statusIsFresh = _.isEqual(lastCompletedPollingRequest, state.current.session.inputs);
```
corps, line 52: aaaaangh, this change worries me alot  and could potentially really bite us from a performance point of view.  Is there a known case where we _need_ the deep comparison? Lately, I've been thinking to change how the session sync works to make this go away either way.

coleshaw: Yes, because if you don't do the deep comparison, it will poll constantly.

etna/packages/etna-js/components/inputs/dropdown_autocomplete.jsx
coleshaw: I don't think they do, but not 100% sure. Can try to verify. I only changed it because of some warning / recommendation from running tests, I think end behavior is the same even if we leave it as `''`.

coleshaw:
> Regarding the validation functions themselves, my main concern is that the validation function is somewhat opaque and may miss edge cases. I feel like eventually we ought to use something like json-schema here, which might give us the hope of versioning the component behavior somewhat. Since these json outputs are an essential component of the session params, this all has implications for how saved documents are versioned eventually, and our ability to faithfully reload those documents correctly. The brittleness of this arrangement across time is a real plum.

Can you clarify how we might use json-schema in lieu of input.data? I thought json-schema would be like the TypeScript type hinting and only tell you the format of the JSON object to expect (i.e. string, array, boolean, etc.). Not sure how that helps validate an input's value?

coleshaw:
> The behavior of the "Run" button remains a little odd. For example, if I get to the "Color By" step and have not selected a value, I can hit "Run", but then am told that my selection is incorrect, and I can't "run" after all. This is not wrong per se, but there is no hint that this will happen, and especially if an input drawer is collapsed it may not be obvious that you are not ready to run yet. It might be nice if instead the component (and the containers if collapsed) hinted its own validation status, and maybe even if the "run" button had another state to indicate that it can't run until inputs are resolved.

Would this be more like "constant validation" where the drawer is left open, and perhaps highlighted somehow, until the input is valid? Like a login form where your password is wrong, you have immediate feedback?

graft:
> > Regarding the validation functions themselves, my main concern is that the validation function is somewhat opaque and may miss edge cases. I feel like eventually we ought to use something like json-schema here, which might give us the hope of versioning the component behavior somewhat. Since these json outputs are an essential component of the session params, this all has implications for how saved documents are versioned eventually, and our ability to faithfully reload those documents correctly. The brittleness of this arrangement across time is a real plum.
> 
> Can you clarify how we might use json-schema in lieu of input.data? I thought json-schema would be like the TypeScript type hinting and only tell you the format of the JSON object to expect (i.e. string, array, boolean, etc.). Not sure how that helps validate an input's value?

The validators seem to be doing three kinds of work:
1. Check that the input.value format is correct
2. Check that the input.value format is "filled in"
3. Check that the input.value entries agree with input.data

This latter is I think in a different category than the others and should maybe occur separately (and perhaps not even in JS), mostly because it raises so many concerns about magma cache state that I am not sure we really understand yet. In most instances (3) should not really be relevant, since when running forward through a workflow the input.value will be composed directly from input.data anyway; it only really becomes an issue when we start dealing with strange save/load and cache invalidation use cases.

If we don't do (3) here the result may be an invalid magma query down the road, resulting in an error. However, there's no guarantee that we won't produce an invalid magma query anyway through some combination of totally valid inputs (e.g. if I ask for samples in two non-overlapping categories), so I'm not sure that the work of (3) is saving us from something in the end.

So leaving out (3), (1) and (2) are both some kind of format-checking, which seems pretty amenable to json-schema.

corps:
> > > Regarding the validation functions themselves, my main concern is that the validation function is somewhat opaque and may miss edge cases. I feel like eventually we ought to use something like json-schema here, which might give us the hope of versioning the component behavior somewhat. Since these json outputs are an essential component of the session params, this all has implications for how saved documents are versioned eventually, and our ability to faithfully reload those documents correctly. The brittleness of this arrangement across time is a real plum.
> > 
> > 
> > Can you clarify how we might use json-schema in lieu of input.data? I thought json-schema would be like the TypeScript type hinting and only tell you the format of the JSON object to expect (i.e. string, array, boolean, etc.). Not sure how that helps validate an input's value?
> 
> The validators seem to be doing three kinds of work:
> 
>     1. Check that the input.value format is correct
> 
>     2. Check that the input.value format is "filled in"
> 
>     3. Check that the input.value entries agree with input.data
> 
> 
> This latter is I think in a different category than the others and should maybe occur separately (and perhaps not even in JS), mostly because it raises so many concerns about magma cache state that I am not sure we really understand yet. In most instances (3) should not really be relevant, since when running forward through a workflow the input.value will be composed directly from input.data anyway; it only really becomes an issue when we start dealing with strange save/load and cache invalidation use cases.
> 
> If we don't do (3) here the result may be an invalid magma query down the road, resulting in an error. However, there's no guarantee that we won't produce an invalid magma query anyway through some combination of totally valid inputs (e.g. if I ask for samples in two non-overlapping categories), so I'm not sure that the work of (3) is saving us from something in the end.
> 
> So leaving out (3), (1) and (2) are both some kind of format-checking, which seems pretty amenable to json-schema.

I am in favor of adding various forms of validation and feedback, naturally.  But I have a couple of specific thoughts here:

1.  I have reservations for `json-schema` specifically.  Let's not jump into that as being the solution quickly.
2.  If we get CWL <-> python types to check, then really we're only concerned about the ui-query interface remaining stable.  And _type checking is not strong enough to guarantee that_.  As @graft points out, plenty of ui-query inputs can be invalid for reasons that have nothing to do with the format of the data.  I think what we're really talking about is 1.  bugs  2.  stability of the interface over time  and 3.  reproducibility.

This is definitely a step in the right direction for handling the biggest concern: _did the user make a complete and valid selection_ as far as we are concerned?  Type checking is pretty meaningless to the user, and only at best a diagnostic for us when things go wrong.  Imagine the user's reaction to us telling them their selection in a UI "doesn't match the expected type validation".  "So you mean there's a bug?  Or like did I click something wrong?  do I refresh?"

corps:
> Since these json outputs are an essential component of the session params, this all has implications for how saved documents are versioned eventually, and our ability to faithfully reload those documents correctly. 

This is really the crux of the issue.  This has most to do with reproducibility and how we want to handle it.  I have many thoughts on this topic, but I'm going to be honest -- none of them involve json-schema or type checking.

coleshaw:
> Check that the input.value entries agree with input.data

I agree that this item could / should be handled by the cells, under typical web application scenarios, since it's more like server-side validation, and is different than the other two. However, because that feedback can happen after many seconds in this case (at some future step that happens in some unknown amount of time), I feel like if we can provide more immediate user-feedback, that helps the overall experience.

etna/packages/etna-js/components/inputs/dropdown_autocomplete.jsx
coleshaw: Just tried this against the other Vulcan UI components, and the change seems to be okay / have no impact.

coleshaw:
> It might be nice if instead the component (and the containers if collapsed) hinted its own validation status, and maybe even if the "run" button had another state to indicate that it can't run until inputs are resolved.

How do you imagine this would look? I'll make a super-annoying initial version, but it'll be ugly...

etna/packages/etna-js/components/inputs/dropdown_autocomplete.jsx
coleshaw: reverted this to be safe, since the original didn't seem to result in negative behavior.

vulcan/lib/client/jsx/reducers/vulcan_reducer.ts
```diff
@@ -81,8 +82,8 @@ export default function VulcanReducer(state: VulcanState, action: VulcanAction):
 
       return {
         ...state,
-        session: action.session,
-        inputs: {...action.session.inputs},
+        session: {...state.session, ...action.session},
```
coleshaw, line 85: One additional change to how `setSession` works, because it seems like parallel steps would overwrite user inputs if they were made during the "waiting" stage. Best example of this is color-by and differential gene expression clustering. The color-by UI widget appears before the differential gene expression cells finish running. If a user selected a color-by option while waiting, after the diffexp cells finished, their "status" update would wipe out the selected color-by option. This meant that clicking the Run button would do nothing, and the user would have to re-select a color-by option. Kind of confusing with no good way to provide feedback.

This change preserves any user inputs made during in the interim.

coleshaw:
@graft , do you want to take a look at my attempt at warning a user regarding validation errors / styling the Run button differently?

@corps , additional thoughts on deep equality comparison versus the constant polling tradeoff?

PR #408, opened 2021-06-07
Vulcan plot and toggle updates

coleshaw:
This PR addresses three items from Tristan's UMAP feedback:

1) Adds angle bracket icons to the UI step input drawers, to indicate that they can be opened and closed.
2) Removes all the plotly menu bar icons except for downloading a snapshot.
3) Adds a Hover effect to the plot legend, where hovering over a legend highlights that trace in the plot.


dtm2451:
1. I like the carrots / angle brackets and think it's the right icon for the job!
2. Wonderful.  Feels a bit sparse now lol, but that's better than giving useless/confusing extra buttons.  Maybe we'll add back the zoom or some other in the future, but I think this is the right call for nnow.
3. WONDERFUL.  I love this so much.  Color uncertainties basically don't matter, within the interactive session, when we have this =).

Not sure if this is related, but the color-by drawer seems to disappear for a bit after I click Run.

coleshaw:
Thanks for the feedback. I'll take a look at the color-by, I think that may be normal behavior due to status changing and updating cycles, but I am also seeing some regression around "cancelling" dependent UI steps, so I'll be tweaking on this branch. Maybe that will also fix this issue.

I'll also include the useMemo part that @corps requested in this PR.

coleshaw:
Just a note, I think removing the auto-status-POST on changing an input will have downstream consequences that affect "clearing" of dependent UI steps, so that may be a bigger chunk of work than initially estimated. (@dtm2451 , based on the conversation at standup today)

corps:
> Just a note, I think removing the auto-status-POST on changing an input will have downstream consequences that affect "clearing" of dependent UI steps, so that may be a bigger chunk of work than initially estimated. (@dtm2451 , based on the conversation at standup today)

True.  That raises the broader question -- what is the intended behavior?  If I change an input, but I'm not expecting that change to _yet_ change the interface or start new work, would I also expect to clear other inputs automatically or no?  

I think the rule we might be looking for here is one of least surprise: don't, cancel or change anything until a user commits to a change on certain inputs.  Keep the working input buffered out enough that it doesn't depend on or replace the working state until a user is ready.

coleshaw:
So it seems like the UI component flicker happens with every other input widget, too. I'm a bit reluctant to try and address it, if we're going to move to an explicit status polling action as proposed at standup. I think that (plus changing the behavior around auto-submitting) would probably fix or alter this flickering behavior anyways.

coleshaw:
Oh @corps , I think the deep equality comparison is required due to the new multiple-multiselect-string-all input type rather than the validations. I think it's because the `input.value` is now a Hash and requires deep equality comparison, whereas before none of the input types had a Hash value. :shrug: The validations don't change any of the logic around moving state from `state.inputs` to `session.inputs`...

coleshaw:
> True. That raises the broader question -- what is the intended behavior? If I change an input, but I'm not expecting that change to yet change the interface or start new work, would I also expect to clear other inputs automatically or no?

> I think the rule we might be looking for here is one of least surprise: don't, cancel or change anything until a user commits to a change on certain inputs. Keep the working input buffered out enough that it doesn't depend on or replace the working state until a user is ready.

I think there will be some edge cases to be aware of -- at the very least, I think we internally will need to clear dependent inputs, even if not reflected in the UI. I think there are at least two, potentially-conflicting (?) scenarios:

1) When doing cluster annotations, the user wants to see both the UMAP and the UI input for the annotations, without the UMAP cell being removed from the UI until the Run button is clicked.
2) If editing previous inputs, we need to clear dependencies of the output from session.inputs, otherwise we risk generating a UMAP with unintended records. For example, if I've already made Record Selections and am currently on the "Confirm record names" step, if I go back to change the Record Selection entries, we need to clear "Confirm record names" from session.inputs. Not clearing the latter could lead an internal conflict between the two sets of inputs, because the downstream python cells don't check that they are internally consistent -- they just assume that the record names belong to the set from record selection. 

Perhaps this is "clear downstream inputs" but don't clear any outputs from the UI?

corps:
> > True. That raises the broader question -- what is the intended behavior? If I change an input, but I'm not expecting that change to yet change the interface or start new work, would I also expect to clear other inputs automatically or no?
> 
> > I think the rule we might be looking for here is one of least surprise: don't, cancel or change anything until a user commits to a change on certain inputs. Keep the working input buffered out enough that it doesn't depend on or replace the working state until a user is ready.
> 
> I think there will be some edge cases to be aware of -- at the very least, I think we internally will need to clear dependent inputs, even if not reflected in the UI. I think there are at least two, potentially-conflicting (?) scenarios:
> 
>     1. When doing cluster annotations, the user wants to see both the UMAP and the UI input for the annotations, without the UMAP cell being removed from the UI until the Run button is clicked.
> 
>     2. If editing previous inputs, we need to clear dependencies of the output from session.inputs, otherwise we risk generating a UMAP with unintended records. For example, if I've already made Record Selections and am currently on the "Confirm record names" step, if I go back to change the Record Selection entries, we need to clear "Confirm record names" from session.inputs. Not clearing the latter could lead an internal conflict between the two sets of inputs, because the downstream python cells don't check that they are internally consistent -- they just assume that the record names belong to the set from record selection.

Now hold on -- we're getting confused here.  I believe that yes, editing session.inputs (committing inputs) should cler downstream dependents.  BUT what I'm suggesting is that some components do _no_ push their changes to session.input directly, but buffer them until the user hits "confirm", a local button.  Thuse, downstream dependents only clear when the confirm button is clicked, and incidentally, the state and server and python cells only ever see consistent inputs.

coleshaw:
Ah, I see, new, local control widgets. Thanks for clarifying, I think I understand better, I was thinking there would be a blanket change where all UI widgets would behave the same with regards to pinging for status automatically (or not-automatically).

dtm2451:
> I think the rule we might be looking for here is one of least surprise: don't, cancel or change anything until a user commits to a change on certain inputs. Keep the working input buffered out enough that it doesn't depend on or replace the working state until a user is ready.

This sounds right to me.

It opens two other questions though: 1) Unfortunately it'd mean that inputs might be stale (often would, but not always), yet still presented, so I thinkkk we would want to add a warning about that.  But if we do, could we reliably have the warning show up specifically in/over/next to inputs that are likely to be stale? And 2) What do we do with previous downstream input sets? We do not have validator-assurance of whether inputs will work for future states, and that seems cumbersome to set up(?), so do we just discard those and make the user make fresh selections?  The think Yes for 2.

corps:
> > I think the rule we might be looking for here is one of least surprise: don't, cancel or change anything until a user commits to a change on certain inputs. Keep the working input buffered out enough that it doesn't depend on or replace the working state until a user is ready.
> 
> This sounds right to me.
> 
> It opens two other questions though: 1) Unfortunately it'd mean that inputs might be stale (often would, but not always), yet still presented, so I thinkkk we would want to add a warning about that. But if we do, could we reliably have the warning show up specifically in/over/next to inputs that are likely to be stale? And 2) What do we do with previous downstream input sets? We do not have validator-assurance of whether inputs will work for future states, and that seems cumbersome to set up(?), so do we just discard those and make the user make fresh selections? The think Yes for 2.

1.  Stale in which sense, specifically?  In general, by the way, I do intend to lock down inputs and prevent any changes while the run is executing.  We should require a user to cancel their run in order to make changes to inputs.

2.  We don't do anything to downstream input sets until a user hits the save button.  Until a user hits the save button the UI is identical to having never done anything.

dtm2451:
> > > I think the rule we might be looking for here is one of least surprise: don't, cancel or change anything until a user commits to a change on certain inputs. Keep the working input buffered out enough that it doesn't depend on or replace the working state until a user is ready.
> > 
> > 
> > This sounds right to me.
> > It opens two other questions though: 1) Unfortunately it'd mean that inputs might be stale (often would, but not always), yet still presented, so I thinkkk we would want to add a warning about that. But if we do, could we reliably have the warning show up specifically in/over/next to inputs that are likely to be stale? And 2) What do we do with previous downstream input sets? We do not have validator-assurance of whether inputs will work for future states, and that seems cumbersome to set up(?), so do we just discard those and make the user make fresh selections? The think Yes for 2.
> 
> 1. Stale in which sense, specifically?  In general, by the way, I do intend to lock down inputs and prevent any changes while the run is executing.  We should require a user to cancel their run in order to make changes to inputs.
> 2. We don't do anything to downstream input sets until a user hits the save button.  Until a user hits the save button the UI is identical to having never done anything.

1. Stale in the sense that changing an experiment in the Record Selection input would update the Confirm Records options shown.  A user might still make changes in the Confirm Records component after making a change in the Record Selection box.  This is a fairly obvious case, but I could still see a user thinking that their non-checks might be saved, so I'm wondering if we want to warn them over inputs that are downstream and likely to not matter.  :+1: to having users cancel their run before making edits.
2.  Right, my question is separate from that. It's about what happens when the user hits run and the session state catches up to their downstream input.  If color-by, those options actually will not change.  If Manual annotations, it's likely that there will be a different number of clusters so the old state will be incompatible with the current h5ad single-cell data.  I'm saying that our input.value validation accounts for if an input seems valid, but not if it actually would be valid for the current data, so I think we need to discard all downstream selections.  Really, 🙃 that's our current behavior upon upstream change, so I guess this one is actually more a kudos on that algorithm choice + commentary that drives my wanting some staleness hinting, a.k.a. drives 1.

vulcan/lib/client/jsx/components/workflow/user_interactions/inputs/multiselect_string.tsx
```diff
@@ -40,6 +40,7 @@ const MultiselectStringInput: InputBackendComponent = ({input, onChange, onClear
       }}
       onAll={onAll}
       onClear={onClear}
+      maxItems={25}
```
dtm2451, line 43: Future question now (and that's fine!) but I wonder what happens if there are 26 items?  That could happen easily if a certain project has reason to directly expose the biospecimen or subject id levels here.

coleshaw:
Yeah, certainly an issue. At that point you would have to type into the box so the list is filtered down. Basically I would say it’s a trade off between performance (you can’t render the entire list of Gene IDs) versus convenience of seeing more items. If 25 seems too small, we can adjust, it was arbitrary on my part. The default is 10, so you may notice that you don’t actually see all experiment aliases on production now anyways. 

Also there is probably some general UX question of teaching people to use the search box, it’s not a select where all options render.

PR #409, opened 2021-06-07
reset button

graft:
A WIP because the button works - sometimes? Seems similar to issues with save/load, which may not always suffice to overwrite the UI state for reasons not yet clear to me. Usually some combination of hitting reset, reloading, and repeating will suffice to reset the state entirely, but the gestures required remain arcane.


coleshaw:
Is it related to any active running `status` polling?

coleshaw:
Hm, it seems to work pretty well for me, testing at various stages of the workflow (with some intermediate steps showing the UI widgets in various steps of open / close before removing them entirely) -- do you observe any patterns that lead to the UI state not overwriting correctly?

dtm2451:
Seems to work for me as well.  On quick inspection, I needed to refresh in order to have the 'previous' param boxes automatically open, but the step progress reset even before then.

PR #410, opened 2021-06-07
Fixing some noisy errors from rollbar

corps:
I upgraded the rollbar account so that we see items flow through that system again.
I took a jab at a couple of common errors our users are using that could have improved responses.

We could probably take a look at rollbar from time to time to see common errors users are hitting and see if we can improve the error messaging around them.

### DIFF
--- a/etna/lib/etna/controller.rb
+++ b/etna/lib/etna/controller.rb
@@ -19,25 +19,30 @@ module Etna
       @logger.warn(request_msg(line))
     end
 
+    def handle_error(e)
+      case e
+      when Etna::Error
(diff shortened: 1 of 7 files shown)


etna/lib/etna/controller.rb
```diff
@@ -19,25 +19,30 @@ def log(line)
       @logger.warn(request_msg(line))
     end
 
+    def handle_error(e)
+      case e
```
corps, line 23: Extracting this allows subclasses to customize error handling a bit better.

coleshaw (approved):
Nice, this will help a lot of users!

PR #413, opened 2021-06-11
Order color mappings of the UMAP workflow + other tweaks

dtm2451:
Adds an `order()` function to `archimedes/functions/list` for this purpose.

Also:
- **pretty impactful for the fututre!** Separates data grab and plot-tweak planning code from direct scatter plot generation code, && turns the latter into its own separate function in archimedes! (`scatter_plotly` of `archimedes.functions.plotting`)
- adds a `do_call()` function to `archimedes/functions/utils` to simplify conditional coding by allowing common inputs to be set in a dict, conditional inputs to be added, then the function called on those inputs later.
- Sets point opacity to full & make legend keys larger.
- adds 'Fraction' as a color-by option for HuMu.

How do we feel about this ordering?
ordering

### DIFF
--- a/archimedes/archimedes/functions/list/manipulations.py
+++ b/archimedes/archimedes/functions/list/manipulations.py
@@ -1,10 +1,11 @@
 from collections.abc import Iterable
+from typing import List, Union
 import re 
 
-def unique(ls):
+def unique(ls: list):
     return list(dict.fromkeys(ls))
 
-def flatten(ls):
+def flatten(ls: list):
     def _flatten(l):
         for el in l:
             if isinstance(el, Iterable) and not isinstance(el, (str, bytes)):
@@ -13,14 +14,14 @@ def flatten(ls):
                 yield el
     return list(_flatten(ls))
 
-def order_num(ls: list, return_indexes: bool = False):
+def order_num(ls: List[Union[int, float]], return_indexes: bool = False):
     if return_indexes:
         ret = 0
     else:
         ret = 1
     return list(i[ret] for i in sorted(enumerate(ls), key = lambda x:x[1]))
 
-def order_str(ls: list, return_indexes: bool = False): 
+def order_str(ls: List[str], return_indexes: bool = False): 
     """ Sort a string iterable in the way that humans expect.""" 
     convert = lambda text: int(text) if text.isdigit() else text 
     alphanum_key = lambda key: [ convert(c) for c in re.split('([0-9]+)', key[1].lower()) ]
--- a/archimedes/archimedes/functions/list/manipulations.py
+++ b/archimedes/archimedes/functions/list/manipulations.py
@@ -13,8 +13,34 @@ def flatten(ls):
                 yield el
     return list(_flatten(ls))
 
-def order(ls): 
-    """ Sort the given iterable in the way that humans expect.""" 
+def order_num(ls: list, return_indexes: bool = False):
+    if return_indexes:
+        ret = 0
+    else:
+        ret = 1
+    return list(i[ret] for i in sorted(enumerate(ls), key = lambda x:x[1]))
+
+def order_str(ls: list, return_indexes: bool = False): 
+    """ Sort a string iterable in the way that humans expect.""" 
     convert = lambda text: int(text) if text.isdigit() else text 
-    alphanum_key = lambda key: [ convert(c) for c in re.split('([0-9]+)', key.lower()) ] 
-    return sorted(ls, key = alphanum_key)
+    alphanum_key = lambda key: [ convert(c) for c in re.split('([0-9]+)', key[1].lower()) ]
+    if return_indexes:
+        ret = 0
+    else:
+        ret = 1
+    return list(i[ret] for i in sorted(enumerate(ls), key = alphanum_key))
+
+def order(ls: list, return_indexes: bool = False):
+    """
+    Output: ls reordered in "increasing" order
+    types: str, bool, 'number' (else case):
+    """
+    if all(map(lambda x: isinstance(x, str), ls)):
+        return order_str(ls, return_indexes)
+    elif all(map(lambda x: isinstance(x, bool), ls)):
+        num_ls = list({True: 0, False: 1}[x] for x in ls)
+        order_inds = order_num(num_ls, True)
+        if return_indexes: return order_inds
+        else: return ls[order_inds]
+    else:
+        return order_num(ls, return_indexes)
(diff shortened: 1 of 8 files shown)


vulcan/lib/server/workflows/scripts/plot_umap.py
```diff
@@ -55,29 +55,32 @@ def get(ids, value):
...
 
 ##### OUTPUT
+plot_args = {
+    'data_frame': scdata.obsm['X_umap'],
+    'x':0, 'y':1,
+    'color_discrete_sequence': colors,
+    'color': color
+    }
+
+# if 'color is dicrete'
```
dtm2451, line 65: What's a good way to check for discrete vs continuous in python?

graft: Do you mean numeric vs. categorical, i.e. number vs. string?

dtm2451: Yea, I suppose those are the only options so perhaps a silly question, but I wanted to do it safely without knowing all options.

dtm2451: There's also booleans and perhaps other types allowed as a record attribute

archimedes/archimedes/functions/plotting/scatter_plot.py
```diff
@@ -0,0 +1,66 @@
...
+    px_args['data_frame'] = data_frame
+    px_args['x'] = x_by
+    px_args['y'] = y_by
+    px_args['color'] = color_by
+    px_args['color_discrete_sequence'] = color_panel
+
+    # FUTURE! Make this work for bool too.  And maybe other types
+    if not all(map(lambda x: isinstance(x, str), data_frame[color_by])):
+        color_order_legend = 'unordered'
+    if (color_order_legend == 'increasing'):
```
dtm2451, line 45: Question posed above is addressed here.  Not sure if there might be a better way, but the ordering code that follows only works for strings so I have it ensuring the data is string-type

dtm2451:
Ended up having some time to wrap this up today!

Here's what I'm thinking a dedicated scatter plot Vulcan workflow could look like, and the "Color Options" widget inside is what I would like to make for general scatter plot mods. I have a couple ideas for how that widget itself could be made viable for either full functionality versus restricted functionality within the UMAP workflow, but all of that would be a future PR! https://www.figma.com/file/XTf27hua9EohijktVm4q6A/Scatter_plotly-Workflow?node-id=0%3A1

dtm2451:
Shoulda asked this in standup perhaps... I’m pretty confident about the scatter_plotly function, but far less confident that the `order` function [here](https://github.com/mountetna/monoetna/pull/413/files#diff-9644c5a235ba3a05f3c6c38e30ed420b4efb54edc5af7e66c5a5732887fd0596) and `do_call` function [here](https://github.com/mountetna/monoetna/pull/413/files#diff-7578d09b14b35c7277c51b0a123261e12cc251c4a9f01727d278e93710ccc0e0) are quality python code.

vulcan/lib/server/workflows/scripts/plot_umap.py
```diff
@@ -55,29 +55,32 @@ def get(ids, value):
...
-        )
-else:
-    fig = px.scatter(
-        scdata.obsm['X_umap'], x=0, y=1,
-        color_discrete_sequence=colors,
-        color=color)
+    plot_args['data_frame'] = dat
+    plot_args['hover_data'] = [hover_name]
+
+fig = do_call(px.scatter, plot_args)
```
corps: rather than relying on this, do_call function, why not inline just with `px.scatter(**plot_args)` ?
The do_call function seems superfulous.

archimedes/archimedes/functions/list/manipulations.py
```diff
@@ -1,16 +1,20 @@
...
         for el in l:
             if isinstance(el, Iterable) and not isinstance(el, (str, bytes)):
                 yield from _flatten(el)
             else:
                 yield el
-    return list(_flatten(ls))
\ No newline at end of file
+    return list(_flatten(ls))
+
+def order(ls):
```
corps: I'd document that ls here is intended ot be a list of strings -- any other list will bail on the convert line here.

archimedes/archimedes/functions/list/manipulations.py
```diff
@@ -1,16 +1,20 @@
...
                 yield el
-    return list(_flatten(ls))
\ No newline at end of file
+    return list(_flatten(ls))
+
+def order(ls): 
+    """ Sort the given iterable in the way that humans expect.""" 
+    convert = lambda text: int(text) if text.isdigit() else text 
+    alphanum_key = lambda key: [ convert(c) for c in re.split('([0-9]+)', key.lower()) ] 
+    return sorted(ls, key = alphanum_key)
```
corps: Yeah this works fine.  I see the logic and it makes sense.

archimedes/archimedes/functions/list/manipulations.py
```diff
@@ -1,16 +1,20 @@
...
 def flatten(ls):
     def _flatten(l):
         for el in l:
             if isinstance(el, Iterable) and not isinstance(el, (str, bytes)):
                 yield from _flatten(el)
             else:
                 yield el
-    return list(_flatten(ls))
\ No newline at end of file
+    return list(_flatten(ls))
```
corps, line 14: Yup, this works great.  Unfortunately there isn't a natural built in to do this, I believe.  I'd nit that using recursion here is not great for python in general and is not necessary (you can achieve this with a list and a loop), but honestly I can't imagine the nesting in our data to be deep enough for that to actually matter.

vulcan/lib/server/workflows/scripts/plot_umap.py
dtm2451: No reason other than that I'm a python noob.  I'll make it that instead.

archimedes/archimedes/functions/list/manipulations.py
dtm2451: looks like you commented somehow on the outdated version, but point taken and I'll put that where it's needed now.

archimedes/archimedes/functions/list/manipulations.py
dtm2451: Hmmm I think I agree that our data is unlikely to yield deep enough recursion for efficiency here to be the major concern.  My main goal was to have the output be assured to be a completely flat list with no sub-lists, which seems easier with recursion.

PR #415, opened 2021-06-14
Additional notification for production metis deploys

coleshaw:
This should add an `@channel` to the bioinformatics-ping channel when Metis is redeployed in production (@corps how would I test that behavior locally?). Pretty much a bandaid for when the metis_client retry doesn't work. I've also added another exception catch that I saw when testing, to the retry code (not sure if that is what triggers the production issues).

Testing locally this seems to work okay, though I'm not sure I'm reproducing the production issue. Restarting and stopping edge-apache and the front-end container don't seem to affect metis_client at all (even without these changes), so perhaps there is something more complex going on.

Thoughts? Will this be too noisy?

### DIFF
--- a/docker/deploy.sh
+++ b/docker/deploy.sh
@@ -8,6 +8,10 @@ function notifySlackOfVersion() {
     if ! [ -e "/var/run/deployment/notified/${APP_NAME}/$(cat /built-from-sha)" ]; then
       touch "/var/run/deployment/notified/${APP_NAME}/$(cat /built-from-sha)"
       post-to-slack "${APP_NAME} in ${ENV}" "bioinformatics-ping" "Deployed https://github.com/mountetna/monoetna/commit/$(cat /built-from-sha)"
+
+      if [[ $APP_NAME == "metis" && $ENV == "production" ]]; then
+        post-to-slack "${APP_NAME} in ${ENV}" "bioinformatics-ping" "<!channel> ${APP_NAME} restarted in ${ENV} -- please check your upload / download jobs!"
+      fi
     fi
   fi
 }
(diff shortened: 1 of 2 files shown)


corps:
Feels noisy.  :shrug: but I'm not strictly against it. 
1.  I can prevent the restarts of edge-apache.  This would address the downloads case better.
2. Obviously I don't want to prevent restarts of metis when metis is deployed, which still causes uploads to fail for a short period.
3. No matter what we do, random network blips will happen and all tooling should be prepared for retry where possible.  

We should feel this out, and make sure we have healthy expectations.  I'm not really often going to be looking for permission to deploy -- I'm going to deploy when it makes sense, which is often, and especially when we need to get work deployed.  Tooling can get to a point of reliable retrying, we'll need to hunt down root causes.

corps (approved):

graft:
I generally agree with the spirit of Zach's rant that this is just a fact of life and clients should be robust to the server being down as much as possible. Some other other things we might try, if we care to expend the effort:
1) Isolate uploads/downloads to metis in a separate service that isn't rebooted as often.
2) Reduce the need to deploy individual apps (i.e. don't deploy metis on every production deploy).
3) Do a more graceful blue/green type of deploy to ease downtime.

corps:
I've been thinking about #2 and I think it should be doable.

corps:
#3 is doable as well, and some refactoring I'm making right now should bring us closer.  The trick is, however, that no amount of blue / green is 100% gap covering, it requires proper behavior of puma and apache at the very least, and furthermore generally will be constrained to within a certain time window.  Long, or even just network slow requests will need to be killed.

But yeah.  I mean, I'm hoping with the monitoring up soon, we'll have some numbers to understand our SLA better for making the right decisions to improving it, too.

PR #416, opened 2021-06-15
Timur query UI

coleshaw:
So this is a large PR and includes a very rough / MVP Query page for Timur. Most certainly not feature-complete and I'm sure has tons of bugs. Note that my goal was not to replicate 100% of the query API functionality, but hopefully hit the most commonly asked questions, as reflected in the feedback from my lab meeting. 

My hope is to deploy this to production and let @graft and @bsamad play around with it a bit to give more feedback. It's easier to test when there is real data to hit. You should be able to build queries and fetch them, with results rendered into a table in the browser. @bsamad , I'd be happy to do a Zoom with you and walk through the UI, once it's deployed. It's kind of confusing and busy, even though it is all clicky.

Note that the query page is behind a feature flag, and you'll need to add the `timurquery` flag to your user to use it.

Main missing features are ability to download your final data frame, rename the column headings, and load previous queries.

Also introduces TypeScript and Material UI to Timur.


timur/lib/client/jsx/components/query/ant_switch.tsx
```diff
@@ -0,0 +1,41 @@
+// From https://material-ui.com/components/switches/#CustomizedSwitches.tsx
+import {withStyles, Theme, createStyles} from '@material-ui/core/styles';
+import Switch from '@material-ui/core/Switch';
+
+const AntSwitch = withStyles((theme: Theme) =>
```
corps, line 5: Cool.  This is definitely another place where typescript will help -- going to the definition of withStyles and following the definitions around clarifies this api.

timur/lib/client/jsx/timur_ui.jsx
```diff
@@ -21,6 +21,7 @@ import Search from './components/search/search';
 import {selectUser} from 'etna-js/selectors/user-selector';
 import {ModalDialogContainer} from 'etna-js/components/ModalDialogContainer';
 import {Notifications} from 'etna-js/components/Notifications';
+import QueryPage from './components/query/query_page';
```
graft, line 24: Could you try using `createEtnaTheme` in here like we do in Janus?

PR #417, opened 2021-06-17
More metric improvements

corps:
1.  Measure all etna routes -- capture response times, controllers, and unique users.  Unique users are tracked as tags, hashing their email against a production secret to ensure anonymity.
2.  Measure rollbar error rates
3. Measure command runtimes and status.
4. -- WIP measure the number of rows for each project + model pairing with a polyphemus job.  To be complete, still working on this.

### DIFF
--- a/etna/lib/etna/route.rb
+++ b/etna/lib/etna/route.rb
@@ -1,3 +1,5 @@
+require 'digest'
+
 module Etna
   class Route
     attr_reader :name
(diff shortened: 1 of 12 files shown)


etna/lib/etna/route.rb
```diff
@@ -58,6 +60,43 @@ def parts
...
+        process_call(app, request)
+      ensure
+        try_yabeda(request) do |tags|
+          dur = Process.clock_gettime(Process::CLOCK_MONOTONIC) - start
+          Yabeda.etna.response_time.measure(tags, dur)
+        end
+      end
+    end
+
+    def hash_user_email(email)
```
corps, line 80: @graft @coleshaw 

This is how I intend to label and track unique visitors.  The hashing function includes 3 parts:  a production specific secret value, and the controller + action labels.

This ensures that
1.  Hashes are consistent _within a given day_, so that any measurement whose window is less than 1 day (almost certainly all our windows will be in that window) are still consistent and correct (although the windows which intersect with 12am PST will be inaccurate).
2. Without knowing the production secret, no one can predict what any given email's hash will be day to day, action to action.  *If you know that a particular person is using a specific action often you can infer their hash!*  However, that "knowledge" of which hash a person is tied to does not transfer to any knowledge about what other controller actions they are using (different hashes) and does not transfer to multiple days (different hash).  At worst, a person's hash can only be inferred by knowledge of their behavior within any particular day and controller/action usage.

Let me know what you think.

corps: Note that this means that we cannot count unique users to _any_ part of our system.  We can track the unique users to _each_ part of the system.

graft: The daily granularity of the hash means we can't see if there are power users of the system (e.g., Arjun makes 90% of uploads in the last month) because we can't follow user activity across multiple days. This seems fine here but we might want to setup a longer-window hash for certain routes (mostly the metis upload/download routes).

magma/lib/magma.rb
```diff
@@ -134,6 +134,19 @@ def find_descendents(klass)
     end
   end
 
+  def setup_yabeda
+    Yabeda.configure do
+      group :magma do
+        gauge :data_rows do
+          comment "The number of data rows by project and model for magma."
```
coleshaw, line 141: We should keep in mind (when communicating to users) that this might be a rough number because of disconnected or otherwise "bad quality" records. So it's not a "QC passed" number of records.

PR #419, opened 2021-06-18
Skeletal Polyphemus UI

graft:
I'm adding in a skeletal UI for polyphemus with the basic etna client-side bells in place. I don't have anything to put in here yet, but I am inspired to this move because of the continual experience of having to edit the redcap loader configuration for various projects, a problem which will only grow in scope as time goes on. There are a number of additional polyphemus-related pieces of work (e.g. various ETLs) that are at the moment relatively inaccessible and opaque to most users - I hope to collect these items in this interface, which will increase their visibility and thus hopefully their cohesiveness in the collective imagination, and eventually make the work of managing them available to non-engineers. This seems especially timely if we intend to focus heavily on ingestion work in the near future.


corps (approved):
Looks good.

PR #420, opened 2021-06-21
Metis: set folder and file authors on rename

coleshaw:
This PR fixes the Metis behavior for renaming files and folders, to set the `author` field correctly. This works in the UI and via `metis_client`. When folders are renamed, only the renamed folder has its `author` field set to the new value -- any children folders or files keep their previous `author`.


metis/lib/models/file.rb
```diff
@@ -216,11 +216,15 @@ def unprotect!
...
-    def rename!(new_folder, new_file_name)
-      update(
+    def rename!(new_folder, new_file_name, user=nil)
+      new_params = {
         file_name: new_file_name,
         folder_id: new_folder ? new_folder.id : nil
-      )
+      }
+
+      new_params[:author] = Metis::File.author(user) if user
```
graft, line 225: if !user, author is empty?

coleshaw: If `!user`, `author` gets set to the default of `metis|Metis` (which is the current behavior you see on rename).

graft:
I'll comment in general that the Metis "author" field was hastily constructed and poorly thought out, as evidenced by the kludgy `|` field separator in the string. I think my main idea was that this is a field we should only approximately tie to a particular user (that is, there is no foreign key to a User). This avoids the pain of maintaining user models in each of the apps and keeping them in sync, and has the side effect of keeping permissions at the per-project role and not per-user access level. There is obviously also the documentary value of this field in terms of tracking who is responsible for what file, but I don't think the usage is extremely clear here (for example, would we ever want to search this by email, etc.). Clarifying this might suggest some obvious reformations to this process.

coleshaw:
The desire to show some sort of documentation makes sense. For some future need where we may need find the email of a user, would it make sense to expose a Janus API for that? I agree we don't want to duplicate user models / users, but we should able to query and use the "source of truth" to establish other information we might need.

PR #421, opened 2021-06-21
Production deploy

coleshaw:
This deploy would basically get the Timur query stuff out into production -- everyone on the engineering team already has the flag set up for it.


corps (approved):

dtm2451 (approved):
👍 as long as Tristan also has the vulcan flag.

coleshaw:
> +1 as long as Tristan also has the vulcan flag.

Yup, looks like he still has it!

PR #422, opened 2021-06-21
Metis client more exceptions

coleshaw:
This PR adds a couple more Net HTTP type exceptions to retry with during metis_client uploads. The latest error seen by Arjun was `Errno::EPIPE`, and the other exceptions I added are the ones that `Net::HTTP` will retry on as well (might as well try to catch them all).

Also adds some retry logic around downloading files (though, I think Apache handles not closing download streams already ... but the new logic might help if there is a forced Apache restart or network hiccup, to not interrupt downloading a set of files).


metis/bin/metis_client
```diff
@@ -264,7 +284,7 @@ class MetisClient < Client
         return retry_blob(upload_path, upload, blob, retries)
       end
       raise e
-    rescue Errno::ECONNREFUSED, Errno::ECONNRESET, Net::OpenTimeout  => e
+    rescue *net_exceptions  => e
```
corps, line 287: :+1:

corps (approved):

PR #426, opened 2021-06-22
More compact query UI

graft:
This is an attempt to wrestle the Query UI @coleshaw is building into a little better layout. The main things I am learning to watch out for with Material UI:

1) Grid components often get nested too deeply unnecessarily, if a component is wrapped in a `Grid` and its top level component is also a Grid.
2) Grid spacing seems to set negative margins, which has strange effects that are hard to control. Often it seems best to use `spacing={0}` and set spacing by hand.
3) Card layouts don't seem to compact easily.

The only actual functionality I have changed here is the "select model" dropdown, which now does not include attribute selection. You can still do this through "select" (e.g., choose patient in 'From', then add 'patient -> ice_time' in Select).

I also changed the base button style yet again, as I found that colored border to be quite ugly. Hopefully this one sticks.


timur/lib/client/jsx/components/query/query_select_pane.tsx
```diff
@@ -48,61 +44,47 @@ const QuerySelectPane = () => {
...
-            </IconButton>
-          </Tooltip>
-        ) : null}
-      </CardActions>
-    </Card>
+        let choiceSet = [
+          ...new Set(state.graph.allPaths(state.rootModel).flat())
+        ]
+          .filter(
+            (m) => !intersectionModels.includes(m) && m !== state.rootModel
```
coleshaw, line 55: This filters out the root model, so you can't actually select attributes from the root model in the Select pane anymore. I think you'll need to remove this filter, to get the desired behavior? For example, if I select `patient` in the From pane, I can no longer select `patient` in any Select model selector.

coleshaw:
I like the styling changes, and the switch to attribute-selection only in Select makes sense!

coleshaw:
One other issue I've found -- if you select a model that's only linked to project (i.e. document for IPI), it doesn't also appear underneath the Select pane so you can't get any attributes for it. But I think it should appear. I suspect this is because the QueryGraph ignores all paths through project right now, so there are no paths generated...

PR #427, opened 2021-06-22
add some indices to metis foreign keys

graft:
Ibid, hopefully this will clear up the issue with listing the root folder on any metis project. I'll be glued to my Metis dashboard.


corps:
Yeah, the cool part is that we can validate theories and fixes and watch the trends!  Dude metrics are game changers for this sort of thing.

metis/db/migrations/022_add_indices.rb
```diff
@@ -0,0 +1,20 @@
+Sequel.migration do
+    change do
+      alter_table(:files) do
+        add_index :bucket_id
+        add_index :folder_id
+        add_index :data_block_id
+        add_index :project_name
```
corps, line 7: Hold on -- is this one worth it though?  I feel that project_name is dense and not sparse.  I believe there will be a performance drop from indexing this on every write with very little gain on read performance.  Same applies to all project_name entries.

graft: Yeah I'm probably being overzealous indexing this, I'm not sure we actually even query files by project_name alone at any point.

PR #432, opened 2021-06-24
Timur query download tsv

coleshaw:
This PR updates the Timur query page a little bit:

* Adds functionality to download the TSV data frame.
* Changes usage of `::first` in filters to `::any` (@graft , I think the end behavior of the query is still correct, but probably need you to verify...at least the queries are runnable!). ~~-- still futzing with this ... found some edge cases I need to test further.~~
* Updates human-readable wording of `::has` and `::lacks` to `Is Present` and `Is Missing`.
* Refactors some of the table calculations into a custom effect hook and adds testing around it.
* Fixes an issue around `::count` returning non-unique counts from Magma.
* Allow you to add slices to any `collection` in the Columns pane, not just `table` and `matrix`


graft:
Tis a bit weird that we map out the identifier as a column automatically AND return it as a rowName, so both show up in every TSV. Looks like the empty query `[ 'patient', '::all', [] ]` does not work to give just identifiers and no columns, but maybe it should?

coleshaw:
> Tis a bit weird that we map out the identifier as a column automatically AND return it as a rowName, so both show up in every TSV. Looks like the empty query `[ 'patient', '::all', [] ]` does not work to give just identifiers and no columns, but maybe it should?

So that's basically just how the TSV is constructed, I can change that behavior.

That query does not work against Magma -- you get a `New error: ArgumentError: No columns were requested!` -- which I think makes sense from the server perspective?

coleshaw:
I think I've solved all the issues I was having with `::any`, and the query builder formats stuff correctly. I'd like to merge this in as-is, so at least the TSV download functionality is added.

I think next steps would be to do some server-side improvements to the query code (move Ruby processes into SQL if possible, add in `::every`), and then make corresponding UI-side changes to reflect the usage of `::every`. And another big thing would be saving previous queries and enabling queries-as-query-params in the URLs.

graft (approved):

PR #433, opened 2021-06-30
(WIP) Create a widget that can be used for creating/tweaking scatterplots in vulcan

dtm2451:
The goal here is to create a `ui-inputs/scatter-plotly` widget to go with the `scatter_plotly` function recently added to archimedes.  In doing so, I'll also set up a basic scatter plot workflow for testing, but the data for this workflow will be a set  bit of data.  (Allowing users to pick their data is out of scope for this PR!)

Goals of the widget:
- Provide ui for adjusting various scatter_plotly settings - `x_by`, `y_by`, `color_by`, titles, `size`, `color_order`, `order_when_continuous_color`.
  - [x] Basic logic to parse the type of component based on a key/name
  - Component types:
    - [x] string input
    - [x] numeric slider
    - [x] drop-downs
      - [x] with either standard options being the keys/column names of the data frame OR given as a `data_options` -- for certain inputs)
      - [x] Must work as nested_select_atuocomplete to allow hidden-at-first gene options!!
    - [x] checkbox
- [x] 'Additional Options' button that toggles visibility of all but `x_by`, `y_by`, and `color_by` inputs
- [x] Utilize a 'hide' input option to allow a workflow creator to set up certain inputs to be hidden (such  as x and y for umap) when those inputs might be fixed for the given workflow

Then:
- [ ] Make the umap workflow work with this widget


dtm2451:
Plenty to go, but the most basic bits are working now =)
scatter_widget

vulcan/lib/client/jsx/components/workflow/user_interactions/inputs/scatter_plotly.tsx
```diff
@@ -0,0 +1,152 @@
...
+}) => {
+  const {data, name} = input;
+
+  const options: {[k: string]: string} = useMemo(() => {
+    if (data == null) return {};
+    
+    let given = Object.values(data['data_options']);
+    let cols = Object.keys(data['data_frame'])
+
+    if (given == null && cols != null) {
```
corps, line 39: you could probably simplify this logic to `cols || given` which produces the logic `if cols is not null, return that, otherwise return given`.

vulcan/lib/client/jsx/components/workflow/user_interactions/inputs/scatter_plotly.tsx
```diff
@@ -0,0 +1,152 @@
...
+  };
+  
+  const component_use = (key: string, value: any, extra_inputs: any) => {
+    const comps: {[k: string]: fxn} = {
+      'plot_title': string_input,
+      'legend_title': string_input,
+      'xlab': string_input,
+      'ylab': string_input}
+    
+    const comp_use: fxn = comps[key]
```
corps, line 87: In this case you can just use the `Function` typing, which is similar.

vulcan/lib/client/jsx/api_types.ts
```diff
@@ -20,7 +20,8 @@ export const TYPE = {
   NESTED_SELECT_AUTOCOMPLETE: 'nested-select-autocomplete',
   MULTIPLE_MULTISELECT_STRING_ALL: 'multiple-multiselect-string-all',
   MULTIPLE_STRING: 'multiple-string',
-  SINGLE_DROPDOWN_MULTICHECKBOX: 'single-dropdown-multicheckbox'
+  SINGLE_DROPDOWN_MULTICHECKBOX: 'single-dropdown-multicheckbox',
+  SCATTER_PLOTLY: 'scatter-plotly'
```
corps, line 24: Just, to clarify, by the way.  @dtm2451 @coleshaw Is this component a ui-query, or a ui-output?

That is to say, is the intention of this component that it show the result of a workflow, or is the intention of the component to pass results onto to further steps in a workflow?

I suspect that we need to transition this from being an input type to being an output type.

vulcan/lib/client/jsx/components/workflow/user_interactions/inputs/input_types.ts
```diff
@@ -15,6 +15,30 @@ export type InputType = string;
 // This type represents the unflattened raw data format coming in from the cwl.
 export type DataEnvelope<Inner> = { [k: string]: Inner };
 
+export function get_vals(de: DataEnvelope<any>): any[] {
```
corps, line 18: There's a function that already does this, `Object.values(de)`.  I'd use that and replace this function.

vulcan/lib/client/jsx/components/workflow/user_interactions/inputs/input_types.ts
```diff
@@ -15,6 +15,30 @@ export type InputType = string;
...
+  // could add any[] typing to the return if that's not automatic?
+  let vals = new Array();
+  for (var key in de) {
+    vals.push(de[key]);
+  }
+  return vals
+}
+
+export function get_keys(de: DataEnvelope<any>): string[] {
```
corps, line 27: Same, called `Object.keys(de)`.  I'd use that function and replace this.

vulcan/lib/client/jsx/components/workflow/user_interactions/inputs/input_types.ts
```diff
@@ -15,6 +15,30 @@ export type InputType = string;
...
+  // could add string typing to the return if that's not automatic?
+  let keys = new Array();
+  for (var key in de) {
+    keys.push(key);
+  }
+  return keys
+}
+
+export function nulled_vals(de: DataEnvelope<any>): DataEnvelope<null> {
+  let de_ = de;
```
corps, line 37: As written, `de` will still be mutated, since you are still preserving the original reference.  i worry that will introduce mutation state bugs.  Let's change this line to `let de_ = {...de};` -- this will make a shallow copy (which is fine for referential integrity, we're not mutating any inner values themselves.

vulcan/lib/client/jsx/components/workflow/user_interactions/inputs/scatter_plotly.tsx
```diff
@@ -0,0 +1,232 @@
...
+  'size': 5,
+  'plot_title': 'make',
+  'legend_title': 'make',
+  'xlab': 'make',
+  'ylab': 'make',
+  'color_order': 'increasing',
+  'order_when_continuous_color': false
+};
+
+const remove_hidden = (values: DataEnvelope<any>, hide: string[] | null | undefined) => {
```
corps, line 46: Same as before, I highly recommend doing a shallow copy.  `values = {...values}`.  This ensures the deletes do not break referential integrity.

vulcan/lib/client/jsx/api_types.ts
coleshaw: I think the component itself looks like an input -- perhaps the names could be changed to make it more clear? Like `plotly-scatter-config`? It does kind of sound like an output widget with the current name.

vulcan/lib/client/jsx/components/workflow/user_interactions/inputs/scatter_plotly.tsx
```diff
@@ -0,0 +1,232 @@
...
+      'y_by': ['Y-Axis Data', options],
+      'color_by': ['Color Points By', options],
+      'color_order': ['Point render order', null, ['increasing', 'decreasing', 'unordered']],
+      'order_when_continuous_color': ['Follow selected render ordering when color is continuous?'],
+      'size': ['Point Size', 0.1, 50]
+    }
+  }, [options]);
+
+  // Component set constructor
+  const component_use = (key: string, value: any, extra_inputs: any) => {
```
corps, line 174: Very interesting.  The structure of this component is as a nested set of other configurable inputs.  This, works for now, but I imagine is fragile and prone to breakage / difficulty to maintain.  Later, when we have the behavior in a more stable place, let's imagine how we can deconstruct some of the roles of this component into composable separate pieces.

vulcan/lib/client/jsx/components/workflow/user_interactions/inputs/validators/all_output_values_not_empty_validator.tsx
```diff
@@ -0,0 +1,16 @@
...
+import {DataEnvelope, ValidationInputSpecification} from '../input_types';
+import AllInnerValuesNotEmptyValidator, { AllInnerValuesNotEmptyValidatorStrong } from './all_inner_values_not_empty_validator';
+import NotEmptyValidator from './not_empty_validator';
+
+const AllOutputValuesNotEmptyValidator = (input: ValidationInputSpecification<DataEnvelope<any>, DataEnvelope<any>>): string[] => {
+  // When this is used, we don't care if keys in input.data are missing from input.value
+  if (!inputValueNonEmpty(input.value)) {
+    return ['Value cannot be empty.'];
+  } else {
+    input.data = input.value as DataEnvelope<any>
```
corps, line 11: This is fairly dangerous.  You are mutating the passed input reference, which will effect not just your local variable _but also magically all references to the passed input from other code_.  I believe what you want is this:

```
AllOutputValuesNotEmptyValidator({...input, data: input.value})
```

This syntax essentially creates a shallow copy of input, such that the data key is replaced with input.value (but not mutated).
```

File: vulcan/lib/client/jsx/contexts/vulcan_context.tsx [hunk 10]
```diff
@@ -50,16 +50,21 @@ export const VulcanProvider = (props: ProviderProps & Partial<VulcanContextData>
...
+    const [state, setState] = useState(stateRef.current);
+
+    const dispatch = useCallback((action: VulcanAction) => {
       if (props.logActions) {
         console.log(action.type, action);
       }
 
-      const result = VulcanReducer(state, action);
-
+      console.log('processing', action.type, stateRef.current.data)
```
corps, line 60: I'll want to remove this pre merge.

vulcan/lib/client/jsx/api_types.ts
dtm2451: Hmmm, interesting.  I didn't think of there being any distinction like that.

It could go either way perhaps?  But right now, it's written around being a ui-input.  the output of the step is passed to a python script which does some extra bits before calling the scatter_plot function that's in archimedes.

corps: If the output is passed on, it's a ui-query.
The reason why you would want a ui-output and not a ui-query is if you intentionally did not want to _promise the interface of the output_.  For instance, a component which may change its internal data format quite frequently as features are added will be hard to maintain as a ui-query, since any other workflow that depends on that component will instantly break when value type changes.

As a rule, once we use a ui-query in a workflow, _we may never again change its value type_, since we must remain compatible with all existing workflows.

If we think that a component is oft going to change its value, we're talking about a ui-output that makes no promises to connect other steps in other workflows.

dtm2451: Hmmm.  I'd thought about the issue of the component <-> script connectivity a bit before, but certainly not with the full scale of this maintainability issue in mind.  I think we do want this to be a ui-output then.  

That raises 2 issues:
1. The plotting code is currently in python, though we could probably migrate it to React-python.
2. Color-data retrieval currently happens after the step where I was going to plug this into UMAP.  I think we want to keep that retrieval behind a ui-input because querying all options of data may get expensive for large data with many options & possibly also buggy if there's missing data.

For 2, we can leave the color-by selection method as-was before this PR, then have plot adjustments be a separate (often optional) ui bit.

For 1... I imagine it would be better to plan on making future visualizations directly in React than to create a back-end path where a ui-output can require a paired python script be run after inputs are commited.

vulcan/lib/client/jsx/components/workflow/user_interactions/inputs/input_types.ts
dtm2451: Tried this, but it wasn't giving the right format?  Maybe inexperience and just how the developer tools console renders, but I thought It's output was still keyed, rather than an Array...

Will do / try putting this and the others below back as requested.

vulcan/lib/client/jsx/components/workflow/user_interactions/inputs/scatter_plotly.tsx
dtm2451: Yea, I hate it myself, and that's definitely the plan lol.

dtm2451:
These in place transformations that I'd thought I wasn't committing!... I'll need to learn myself away from R's strong concern for not doing that unexpectedly to users.

vulcan/lib/client/jsx/api_types.ts
corps: Or, alternatively, perhaps we can build a `stable bridge` between react <-> plotly?

So instead of a bunch of on off scripts + one off configurations, there is a single plotly script that accepts a full plotly configuration?  And therefore, all python side plotly interactions involve just one script and one type of input (but the ui-queries can be of various kinds, each simply always outputing the bridge config type?)

coleshaw: Might that basically be the same as the Plotly.JS JSON `layout` (and possibly `config`) payloads that get sent from the python to the `Plot.tsx` output component? So any "plot config" UI inputs would write to the Plotly format, regardless of what kind of plot they are configuring? Hopefully that made sense...

So the interface would be the same [as outlined in this Plotly.js documentation](https://plotly.com/javascript/reference/layout/).

vulcan/lib/client/jsx/components/workflow/user_interactions/inputs/scatter_plotly.tsx
```diff
@@ -0,0 +1,230 @@
...
+              min={min}
+              max={max}
+              valueLabelDisplay="auto"
+            />
+          </div>
+      )
+    }
+
+  const extra_inputs: DataEnvelope<any[]> = useMemo(() => {
+    return {
```
coleshaw, line 158: Are some of these values "default" values for the inputs? It seems inconsistent to include these in the UI component, when most other default values come from the server...will that be confusing to someone making a new workflow?

vulcan/lib/client/jsx/api_types.ts
corps: Yeah, that's more or less what I'm thinking.  Especially if the python side isn't really "doing" much beyond that.

dtm2451: `stable bridge` sounds like what I was already planning for the archimedes plotting folder... A single function per plot type that parses our plot setups into how those actually need to provide those to plotly.

dtm2451: It's already the case that the script following this widget in UMAP just adds some set-within-the-workflow inputs (behind the `hide` option of this widget) to the payload of inputs for the `import scatter_plot from archimedes.functions.plotting` call.  Then it makes the call.  Then it `output_json`s that return ("to" the ui-output step that renders the plot).  So, the code change on the workflow side would be a reorganization rather than an algorithmic restructure :+1:

vulcan/lib/client/jsx/components/workflow/user_interactions/inputs/scatter_plotly.tsx
dtm2451: Text labels and default values yes, but for the bits behind the `advanced options` button.  Initial values, but I don't think a workflow editor should ever need to concern themself with them... none are workflow context dependent, so easier to create them within the tsx than to require the workflow author to make a `defaults` input.

This is part of the ugliness of way I have the "The structure of this component is as a nested set of other configurable inputs." set up currently.

corps: I get that, and I think it's a reasonable approach.  Certainly, _assuming reasonable complexity in the component to make workflow creation easier_ is a worthy goal.  I think we can reduce the complexity of this design together, let's pair early next week on it.

dtm2451:
Getting close!  I've pushed the reorganization we did together @corps, plus a little bit more.  I also did break out more of the Viz type-specific bits so now there is the overall component, `VisualizationUI`, which `scatter_plotly` and future `bar_plotly`/`y_plotly`/... can call on!

PR #435, opened 2021-07-04
Entity iteration in redcap

graft:
This provides a more general framework for iterating across items in redcap, allowing us to group individual redcap values into a magma record based on a multi-level iteration key. I am allowing four types of entity right now: records, events, redcap repeat instances, and fields. For records and events these are pretty straightforward (there is a single field for each in the redcap record); the redcap repeat instances produce two fields in the redcap record, `redcap_instrument_name` (a string) and `redcap_repeat_instance` (a counter). Fields are a little weirder, mostly because a repeating field (like a checkbox) produces multiple values, which might be combined with a single value (e.g., ten checked checkbox items + 1 date field yields 10 records containing 1 checkbox value each + the same date repeated 10 times).

In addition to iteration we also can restrict iterated entities by regexp (e.g., if i only want to load records from a certain event name).

The main changes to the config format include a more complex `each` specification for models which may also be used in scripts, + nesting script attributes back into an `attributes` key to accommodate the new `each`. A separate PR for the redcap projects repo includes these updates for mvir1 and ipi.

Finally, a significant flaw I noted along the way is that identifier construction currently can only be done based on the redcap iterables (i.e., `record`, `redcap_event_name`, `redcap_repeat_instrument`, etc.). If we wish to include other fields in the identifier construction, currently it cannot be done. It's possible we could fix identifiers like this in the `patch` method for each model, but this might be difficult, and especially so when you're trying to construct parent identifiers for linking, where the ancillary information to construct the parent identifier might not be easily accessible. This will almost certainly bite us in the ass later.


polyphemus/spec/fixtures/etls/redcap/test2.rb
```diff
@@ -0,0 +1,94 @@
...
+        scripts: [
+          {
+            attributes: {
+              type: "car_class"
+            }
+          }
+        ]
+      },
+      year: {
+        each: [ "record", "event", "repeat" ],
```
coleshaw, line 59: Can you explain a bit this new notion in the project config? I see three versions here, not clear what the expected outcome would be or the syntax differences?

polyphemus/spec/fixtures/redcap_test2_metadata.json
```diff
@@ -0,0 +1,122 @@
...
+    "matrix_ranking": "",
+    "field_annotation": ""
+  },
+  {
+    "field_name":"feature",
+    "form_name":"cars",
+    "section_header":"",
+    "field_type":"checkbox",
+    "field_label":"Features",
+    "select_choices_or_calculations":"0, Window | 1, Door | 2, Spoiler | 3, Radio | 4, Engine | 5, Carbuerator | 6, Wheels",
```
coleshaw, line 88: Is this what the checkbox output looks like, from REDCap?

coleshaw (approved):

polyphemus/spec/fixtures/redcap_test2_metadata.json
graft: This is not the output, but the metadata for what the checkbox values are. The output will then be like `{ "field_name": "checkbox_field", "value": 0 }, { "field_name": "checkbox_field": "value": 1 }`, i.e. several records/values per field_name.

corps:
Reading through now!

polyphemus/lib/etls/redcap/lib/client.rb
```diff
@@ -54,52 +54,30 @@ def initialize eavs, flat_record
...
       end
 
-      def make_record(record_eavs, use_flat=true)
+      def make_record(record_eavs)
         record = record_eavs.reduce({}) do |rec,eav|
-          if add_eav?(eav)
-            rec[eav[:field_name].to_sym] = best_value(
-              eav, use_flat
-            )
+          if !empty?(eav[:value])
```
corps, line 72: A little curious of this logic.  We check if the eav value is not empty, but then we only use it if the inner @flat_record's entry is also empty.  If @flat_record[eav[:field_name]] is not empty, but eav[:value] is, do we intend to ignore it?

corps (approved):
Looks good, just had one logic check question.

polyphemus/spec/fixtures/etls/redcap/test2.rb
graft: Apologies - there are two main introductions in this syntax:

1) The `each` Array defines the items in the redcap record that will be used to group a set of redcap records into magma records. Each term in the each Array corresponds to an iterable property of the redcap record - this doesn't exactly map cleanly to a single field in the record, so there is some abstraction here (e.g., "repeat" encompasses several fields). When trying to build Magma records, the loader will extract each of the values in the `each` Array from every redcap record and group according to the corresponding key of strings. These values are also passed in to the model's `identifier` method to construct the Magma identifier from the various redcap values.

2. The entities can be filtered according to the identifier values in order to restrict or exclude certain records. If the `each` term is a single-value hash instead of a string, the hash value is used to filter the key (e.g. `each: [ "record", { "entity" => /Entity [123]/ } ]`).

polyphemus/lib/etls/redcap/lib/client.rb
graft: I think it would not be possible for @flat_record[ eav[:field_name] ] to be present while eav[:value] is absent, although I am a little uncertain. This whole thing is a kludge to account for the fact that the redcap API is broken and does not translate ontology values for `eav` mode (e.g., if you are using the OMIM ontology, it will return the OMIM numeric code in `eav` mode and the OMIM term string in `flat` mode). Therefore, we need to get the record in eav format and in flat format to have the required information. The intent here is to conservatively paste the translated value in the flat record over the untranslated eav value. However, whenever one exists, the other probably ought to as well.

corps: Got it.  Hmm, yeah overall the redcap api seems confusing, and worse, not open source.  Who, picks this kind of software? 🤷

PR #436, opened 2021-07-06
Magma query verb refactors

coleshaw:
This PR adds in `::every` and (slightly modified) `::any` support for Magma queries, such that you can do the following. Really the `::any` behavior is exactly the same, though the generated SQL is slightly different:

1) Get a boolean `true` or `false` depending if every or any record matches your condition, i.e. `query = ["prize", ["worth", "::>", 4], "::every"]` would return true if every "prize" in the system has value > 4. Correspondingly, `::any` would return `true` if any prize has worth > 4.

2) Filter records based off of an `::every` or `::any` condition. i.e. to get all Labors where every prize has worth > 4, you can do `query = ["labor", ["prize", ["worth", "::>", 4], "::every"], "::all", ":: identifier"]`. If you wanted all Labors where any prize has worth > 4, `query = ["labor", ["prize", ["worth", "::>", 4], "::any"], "::all", "::identifier"]`. Can also use these in combination with any set of `::and` or `::or` filters, to get fairly complex queries.

3) Adds in various specs around the above functionality. 

This PR also fixes a small bug in the Timur Query UI when switching models in the Select / Column pane.

The abstraction still feels a little funny, so welcome any feedback.


magma/lib/magma/query/predicate/subquery_filter.rb
```diff
@@ -0,0 +1,71 @@
...
+  class SubqueryFilter < Magma::SubqueryPredicateBase
+    def create_subqueries(query_args)
+      parent_model = predicate.model
+      join_table_alias = predicate.alias_name
+      args = query_args
+
+      loop do
+        verb, subquery_model_name_args, subquery_args = predicate.class.match_verbs(args, predicate, true)
+
+        raise Exception, "This does not appear to be a valid subquery filter, #{args}." if subquery_model_name_args.first.is_a?(Array)
```
graft, line 13: You might raise a QuestionError here to have this better caught.

graft:
This is another one of those PRs I've been dreaming about for years, so thanks for turning fantasy into reality!

The specs look great in general, it is nice to have more coverage around some of the complex intersections of these verbs. Looks like there is an old test there which tests `::any` as a boolean returned value, might be nice to have one for `::every` also. Similarly with `::count` and `::every`, which had some weird intersection with `::any`.

Haven't quite finished reading through all of the subquery code yet. One thing I might wish is that the conditional logic referencing specific verbs (the operator specification) might bubble up into the verb's block instead of being buried in the code.

It seems like `::first`, which happens currently in Ruby, might also be amenable to using this subquery code?

coleshaw:
>  Looks like there is an old test there which tests ::any as a boolean returned value, might be nice to have one for ::every also. Similarly with ::count and ::every, which had some weird intersection with ::any.

Buried in all the specs, there should be [one for `::every` as a boolean](https://github.com/mountetna/monoetna/blob/cs/magma-query-verb-refactors/magma/spec/query_spec.rb#L255). I'll add one for `::count` and `::every` -- good point, missed replicating that one.

> One thing I might wish is that the conditional logic referencing specific verbs (the operator specification) might bubble up into the verb's block instead of being buried in the code.

That's a good point. I'll see how I can refactor that into the verb itself.

`::first` might use similar hooks, will have to think about how to add it in, since that seems like a SQL `LIMIT` instead of a subquery?

coleshaw:
So just poked around with `::first` for a bit, I think I'll punt on that to prioritize data ingestion work because it at least works with the current implementation. It does seem like you could use subqueries with `LIMIT` instead, however because `::first` has child predicates, it doesn't seem as straightforward...

PR #440, opened 2021-07-19
Polyphemus etl sftp file metis

coleshaw:
This PR takes several steps towards an ingestion pipeline:

* Updates `aspera` client to `4.1.0`, in order to get an updated `net-ssh` dependency. This shouldn't break anything (since it looks like a minor update), but I guess we'll see... :crossed_fingers: 
* Adds an new table to Polyphemus for `ingest_files`. This tracks basic filename / host information and has a flag for `should_ingest`.
* Adds a REST API on top of this new table to list files / subdirectories given a directory and host, as well as enqueue files / subdirectories for ingestion, given a directory + host.
* Adds an ETL, scanner, and cursor to scrape the CAT (using `rsync`) and mirror available files in the new `ingest_files` table.
* Adds an ETL, scanner, and cursor to copy ingest files marked as `should_ingest` from the source to Metis (currently hardcoded to a `triage` project, `waiting_room` bucket). In the future, I can imagine parameterizing the project / bucket in the ETL, depending on how we want the triage workflow to go.
* Adds a UI widget to Metis (behind an `ingest` feature flag), where admins of the `triage` project can specify a directory and host (i.e. CAT vs IHG) to enqueue files for ingestion.

Note, I still have to make changes to the Chef roles to enable these ETLs.


graft:
Still paging through, but my first big question is - what is the value of scraping the CAT and maintaining a file list in a db? Why not simply have a view that lists the current CAT contents + an ingest file queue (i.e., only `should_ingest` items)? From the user's perspective it would be the same, and the only additional possibility the scraped list allows is being out of date.

coleshaw:
Hm, so I see what you are saying. On the surface, they would seem similar, and presenting a live view might make it more "up to date" for a user. However, if we run the ETL every 5-10 minutes, that seems like it should be sufficient. Especially since the CAT e-mail itself says "your data will be available in an hour (or up to 4 hours for Novaseq)"...so I assume most people aren't going to set a timer and check immediately when that hour is up. 

From a performance standpoint, I think having the files cached in our own database make the user experience much more responsive. There are currently 20k `fastq` files on the CAT, the majority of which belong to other projects (I assume). Thus running the `rsync` command does take a couple seconds (certainly longer than hitting our own database) -- I think there is the possibility that if you tried doing that "live" in response to a web request, that it won't be very responsive, especially if multiple people hit the page at the same time or someone types in lots of different directory names. Would you re-scrape the CAT every time someone typed in a new directory, and really make it a "live" query of the CAT? If you wanted to improve performance and wind up caching the first CAT scraping in the client or in Polyphemus, would you run into the same concern about the data being out of date?

I'm not sure that from the user's perspective, it is truly the same...

coleshaw:
I guess another minor point is that setting the scraping up as an ETL seems to lend itself to more future automation. If the CAT agress to put all our project files into say, a `data_library` sub-directory, then it would be simple to tweak the ETL to grab files from those known directories and stick them directly into the `ingest_files` table with `should_ingest: true` and skip the human-UI step. Doing the `rsync` request "live" could be moved in that direction as well, but that process seems to imply always having a human in the loop to flag `should_ingest`...minor point though, I think, since the `rsync` code would be easily moved into an ETL.

polyphemus/lib/etls/sync_cat_files_etl.rb
```diff
@@ -0,0 +1,107 @@
...
+
+  def remove_deleted_ingest_files(records)
+    current_cat_filenames = records.map do |record|
+      record.filename
+    end
+
+    ingest_files_for_host_query.where(ingested_at: nil)
+      .exclude(name: current_cat_filenames)
+      .all do |file|
+      file.delete
```
corps, line 104: Can we perform a softdelete or create a tombstone to record the action of files having been but being off the cat?

Tombstone approach might be easiest.

corps:
> I guess another minor point is that setting the scraping up as an ETL seems to lend itself to more future automation. If the CAT agress to put all our project files into say, a `data_library` sub-directory, then it would be simple to tweak the ETL to grab files from those known directories and stick them directly into the `ingest_files` table with `should_ingest: true` and skip the human-UI step. Doing the `rsync` request "live" could be moved in that direction as well, but that process seems to imply always having a human in the loop to flag `should_ingest`...minor point though, I think, since the `rsync` code would be easily moved into an ETL.

Yeah, I think this is kind of the point for me, too.
I do think it's possible to fetch the live data, but to be honest there is a non zero chance that will become out of date, too.  I don't want to rely too heavily on making live SFTP requests listing every result each time the user hits the webpage.  Timeouts I believe will be common -- with an ETL we can just retry and still keep a mostly up to date view working (afterall, there probably isn't much updates on the cat, the vast majority of syncs will be up to date anywho).

graft:
This is totally faboo. Still reading code; a few comments on the UI:
- Ingest host could be a dropdown
- Directory path could yield some kind of tree view (i.e. traverse the folder hierarchy). This might require some caching of the folder hierarchy after scraping.
I think these would make it easier for the user to understand what they ought to do. Without this I am left wondering about URL and format for operations that are actually constrained based on known data items. Also without this, browsing from the root folder (which I must know is named "SSD") is extremely painful as every single file is listed, slowing down my browser. Or else - is the assumption the user will receive, somewhere else, an exact string to paste in here?
- It appears to be impossible to select according to a file pattern, meaning I can only list a complete directory and get its entire contents? It would be nice to be able to select folders or a set of individual files (for example if one is feeling timid on dev and doesn't wish to download 80 fastq files from the CAT).

coleshaw:
Thanks, those are great suggestions.

I was kind of assuming that folks would get the direct path from their CAT e-mail, not browsing by hand. I feel like browsing would be super painful, because the vast majority of files don't belong in the Data Library, and the directory names are basically meaningless to me at least (outside of the leading timestamp). While putting ingest host as a dropdown makes a lot of sense, perhaps a help or placeholder would suffice for the directory path as a hint to the user?

Regarding individual files, I could do something for that. Bushra had indicated that she just downloads the entire directory and sorts out the files later, but perhaps I can change the UI to be a set of checkboxes with all selected by default? I think that would mostly be helpful for dev, since I think ingesting individual files in production would just result in more work overall ... folks should sort files out of folders using the triage process / metis client, at that point, instead of selecting individual files. I'm not sure in practice how often the CAT mixes files, and if we're asking them to isolate all data library files into our own folders, this shouldn't be an issue in production (and ideally this step gets automated away at some point).

If you'd like to test the SFTP ingest process on a single file on dev, you can also just change the flag in the database for a single record, and run `./bin/polyphemus etl sftp_ingest_metis_triage_files_etl run` to see what happens.

coleshaw:
> (for example if one is feeling timid on dev and doesn't wish to download 80 fastq files from the CAT).

Also, just wanted to note that clicking stuff in the UI doesn't cause files to automatically be downloaded from the CAT on dev -- the ETLs aren't scheduled to run on dev, so you have to run them manually. That means you can click as much as you want in the UI, and all that happens is that the `should_ingest` flag gets flipped on the server. Nothing is downloaded unless you run the ETL (obviously, I would manually unflip the vast majority of flags before doing so).

Even after I change that to checkboxes and you want to test on dev, you'll still have to run the ETL manually after using the UI to select file(s).

coleshaw:
I've updated with a flag for removed files in lieu of deleting the records, plus some additional constraints / input prompts for selecting the ingest host + directory (basically `select` boxes containing the valid values, instead of free-form text inputs). I've used checkboxes (with default all checked) for the resulting files, so users are free to unselect some if they wish. I've refrained from trying to do anything further with globbing or multi-directory search in the Metis ingest widget, because I'd like to automate that part away, anyways. I think most of the human intervention should be kept at the triage stage, not at the ingest stage, so I don't think it makes much sense to invest too much in complex ingest interactions...

PR #447, opened 2021-07-22
Fix datablock moving

graft:
This does 2 things:

1) Change the order of the file move operation so it happens before a change is saved to the database. If the move operation fails, this will prevent the temp-id of the file from being destroyed prematurely, so the job can re-run correctly.
2) Split the archive and checksum operations into separate commands. There is no reason for them to run together; my guess is glacier uploads are limiting and the checksumming could complete quickly if we let it run independently. Since we are running tape backup anyway, glacier uploads are less critical, it is more important that files be hashed and stored with their correct datablock efficiently.


coleshaw (approved):
Does this require corresponding changes to the Chef recipe, to set up a systemd timer for the Checksum command?

PR #451, opened 2021-07-23
tweak limit on sftp ingest to smaller batches

coleshaw:
This PR lowers the batch size of the ingest ETL from 20 to 1. The files are so large and take so long to get into Metis (the SFTP interface seems slow?), that I think it's worth chunking them into smaller batches so the ETL can update its state more frequently. Right now a restart to the ETL during an active run (i.e. caused by a deployment) causes the ETL state to be lost and start over again. So I'd like to save progress on every file, basically.


graft:
Is it possible instead to simply update the state after every file?

coleshaw:
Hm...let me find a small set of files and see...

PR #455, opened 2021-07-24
Swap order of fields in new project dialog

graft:
Another wee one to fix an issue Dan noted, where the order of fields in the "New Project" dialog on the Admin page was different from the project table above it. Seems trivial but since Bushra fell into the same error recently it seems worth repairing.


coleshaw (approved):
I've been bitten by this before, too. Great idea!

PR #459, opened 2021-07-28
Add the ability to follow defaultValue on debounced inputs

corps:
Test this by running `make -C vulcan start-storybook`, navigating to http://localhost:9700 after webpack build is done, and checking the workflow manager component.

1.  Reset and loading from file still works
2. Focus isn't randomly lost.

I was using the `key` prop to force remounts and reconsuming of the value, but this causes the browser to sometimes "lose" focus due to the way the dom works.  This alternative is more complex and requires explicit "following" of defaultValue, but it works.

The idea is that the inputs still debounce, but they optionally also track changes to defaultValue (resulting from forced setting of the value).  This change also uses the explicit debouncer constructor which also includes `reset()`, meaning that we can prevent accidental debounces from setting state after a component is dismounted, or settings changes.

### DIFF
--- a/etna/packages/etna-js/utils/debouncer.jsx
+++ b/etna/packages/etna-js/utils/debouncer.jsx
@@ -11,6 +11,7 @@ export class Debouncer {
     this.promise = new Promise((resolve) => {
       this.resolve = resolve;
     });
+    if (this.lastTimeout) clearTimeout(this.lastTimeout);
     this.lastTimeout = null;
     this.lastFire = Date.now();
     this.f = null;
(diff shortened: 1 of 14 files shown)


vulcan/lib/client/jsx/components/workflow/user_interactions/inputs/input_types.ts
```diff
@@ -81,7 +81,9 @@ export type WorkflowStepGroup = { label: string, steps: WorkflowStep[] }
 
 export type InputBackendComponent<Params extends {} = {}, Value = unknown, DataElement = unknown> = (p: WithInputParams<Params, Value, DataElement>) => React.ReactElement | null;
 export type WithInputParams<Params extends {}, Value, DataElement = unknown> = Params & {
-  onChange: (v: Maybe<Value>) => void; value: Maybe<Value>; data: DataEnvelope<DataElement> | undefined | null;
+  onChange: (v: Maybe<Value>) => void;
```
corps, line 84: No functional change, just unpacking the formatting.

etna/packages/etna-js/utils/debouncer.jsx
```diff
@@ -11,6 +11,7 @@ export class Debouncer {
     this.promise = new Promise((resolve) => {
       this.resolve = resolve;
     });
+    if (this.lastTimeout) clearTimeout(this.lastTimeout);
```
corps, line 14: `reset()` ought to prevent the last scheduled timeout from running anyways.

PR #460, opened 2021-07-30
Ingest from CAT to C4

coleshaw:
This PR adds a new "RemoteSSH" filesystem and an ETL to ingest flagged files from the CAT to C4. I'm not really sure how to add specs for this since it's basically SSH communication ... but I've tested it manually, and it seems to work well using the C4 service account that Harry had set up for us awhile ago.

I will set up the Chef roles with the required configs.

~~Note that I'm not working on the performance improvements for the SFTP -> Metis workflow that we discussed earlier this week. Using LFTP on C4 is super fast, and if that's the workflow that we want to support for triaging, then I don't think it makes sense to try and optimize the direct Metis ingestion workflow at this point.~~


coleshaw:
Close in favor of #474

PR #461, opened 2021-07-30
Magma query subqueries up tree

coleshaw:
This PR fixes an issue with Magma query with subqueries, to correctly work when going up the model tree, and adds specs for that scenario. I think this fixes a lot of the issues seen early in the week with 0 results returned (i.e. finding `sc_rna_seq` records where patient was treated with dexamethasone). The counts returned and UI table also seem to behave correctly with this fix...


corps (approved):
Looks good!

PR #464, opened 2021-08-02
fix issue with page bounds on subqueries

coleshaw:
This PR fixes the issue with page bounds and subqueries, where bounds were not correctly calculated on subqueries. This is because the `count_query` is distinct from the `base_query`, and so we need to include the subquery logic there as well (in `Magma::Question`). The user would see counts returned on the Query page, but the table pagination + rows rendered would not behave correctly ... i.e. searching patients with a specific demographic would result in varying number of rows rendered on the page because not every patient might have that demographic field (page 1 might have 0 rows, page 2 have 1 row, page 3 have 6 rows, etc.). This is because the bounds would be calculated on non-subquery results (i.e. bounds are `patient_1` to `patient_10`), but the actual, subquery results would be `[patient_2, patient_11, patient_23, etc.]`, leading to a mismatch in the UI.

### DIFF
--- a/magma/lib/magma/query/question.rb
+++ b/magma/lib/magma/query/question.rb
@@ -190,6 +190,7 @@ class Magma
 
       joins = @start_predicate.join.uniq
       constraints = @start_predicate.constraint.uniq
+      subqueries = @start_predicate.subquery.uniq
 
       joins.each do |join|
         query = join.apply(query)
@@ -197,6 +198,10 @@ class Magma
 
       constraints.each do |constraint|
         query = constraint.apply(query)
+      end
+
+      subqueries.each do |subquery|
+        query = subquery.apply(query)
       end
 
       query.distinct.select(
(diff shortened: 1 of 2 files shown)


corps (approved):
I do wonder what the performance impact of this might be, but it clearly needs to be done anyways.

I think at some point we may want to consider profiling magma a lot more on its performance areas.  My last profiling showed that there are vast quantities of small allocations that build up, and that the current ruby GC doesn't always recognize them for sweeping, causing a very large memory build up + an expensive collection when it finally catches up.
To fix this, we should consider how we can instrument (add some observable metrics) inside magma query code once I have magma metrics working correctly.

coleshaw:
That's a good point. I think we're hoping that by pushing the logic for some of these query conditions from Ruby code into SQL, we can get better performance, too, but definitely metrics and instrumentation can help identify areas for improvement.

One related issue I haven't tackled yet is moving the logic for `::first` into SQL. Currently handled in Ruby...

PR #467, opened 2021-08-03
Metis ingest etl performance

coleshaw:
This PR refactors the Metis ingest code to use `curl` directly in lieu of the `net-sftp` gem. When I test locally, it improves ingestion speed about 50x, so hopefully in production it will also show some improvement. 

This PR also makes a small change to update each ingested file's `ingested_at` field as the upload completes, instead of at the end of the batch.

Depending on how we end up hooking into the CAT / sequencing process, I'm sure this and / or #460 will require some minor tweaks.


graft:
Pretty cool! I didn't know curl could sftp... A 50x speed improvement is :+1:

coleshaw:
Close in favor of #474

PR #469, opened 2021-08-05
Timur query refactor

coleshaw:
This PR adds in some functionality to Timur Query page:

* Adds UI widget for toggling between `::every` and `::any` when adding Filters, so user can specify at each step of the path how they want the filters enforced.
* Toggling nested - flattened does not change the filter verbs.
* Fixes Magma query API to account for nested `::every` and `::any` subqueries.

### DIFF
--- a/magma/lib/magma/query/predicate/subquery_base.rb
+++ b/magma/lib/magma/query/predicate/subquery_base.rb
@@ -26,7 +26,7 @@ class Magma
 
     def subquery_filters(subquery_args, internal_table_alias, subquery_model)
       subquery_filters = []
-      while subquery_args.first.is_a?(Array)
+      while subquery_args.first.is_a?(Array) && !Magma::SubqueryUtils.is_subquery?(predicate, subquery_args.first)
         filter_args = subquery_args.shift
         subquery_filter = FilterPredicate.new(
           question: @question,
--- a/magma/lib/magma/query/predicate/subquery_filter.rb
+++ b/magma/lib/magma/query/predicate/subquery_filter.rb
@@ -33,7 +33,9 @@ class Magma
         nested_subqueries << create_nested_subquery(nested_args, subquery_model, internal_table_alias)
       end if has_nested_subquery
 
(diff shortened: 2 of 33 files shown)


graft:
Rather than the switch I might favor a dropdown/select with the items Any/Every, simply for aesthetic reasons - since all the other inputs in the row are selects I think it is easier to see the whole phrase if they have a consistent shape. It is a bit more annoying having to select, rather than just toggle-click, maybe the any/every dropdown could behave that way (although then the hinting is inconsistent).

It's not clear to me how flexible we intend this to become in the end - at present a number of queries are possible, but a number are not, and it is difficult to understand what is possible; e.g. I could at best make this query:
```
["patient",["::and",["demographic",["name","::equals","Gender"],"::any"],["demographic",["value","::equals","Female"],"::any"]],"::all",[["ipi_number"]]]
```
rather than the expected
```
["patient",["demographic",[ '::and', ["name","::equals","Gender"],["value","::equals","Female"] ],"::any"],"::all",[["ipi_number"]]]
```

This is like, does the boolean nest within the "any/every" or vice-versa?

coleshaw:
| Rather than the switch I might favor a dropdown/select with the items Any/Every
:+1: 

Hm, yeah, so for that specific case, I can add a spec for the API side (not sure if we have one, yet). I certainly think there is a lot of UI work around `::and` and `::or` to make those more usable, including for the use case you've outlined. But I think that's a lot of work (I have an idea of how it might look and behave in the UI), and I'd leave it for another PR.

In general, I'd prefer to let people use it and give us feedback on features or queries they want to build, but can't with the tool. As we try to cover 100% what you can do with the API, I think we risk making the UI complex enough that people don't want to use it or can't understand it. Some part of that can be mitigated by education on how to use the tool, but I'm not sure what the tipping point is where we say, "just use the API itself, it's not feasible to build this graphically". At least it seemed like the queries that Ravi, Gabi, and Bushra were building at the workshop could all be done with the current UI (sans bugs)...

PR #470, opened 2021-08-05
fix the 404 / invalid folder message in Metis UI

coleshaw:
This is a small PR to fix the "invalid folder" behavior in the Metis UI. Now, non-existent folders should show the below message, instead of what looks like an empty folder.

Screenshot from 2021-08-05 13-33-10


corps:
Looks good, thanks for the fix!

corps (approved):

PR #471, opened 2021-08-05
Fix redcap_loader bug involving ontologies

graft:
This fixes a bug involving the flat_record stuff, which stopped working due to a regression caused by the updates to the redcap entity iteration stuff. Now there is at least a test to cover this; also the specs are a little easier to make redcap fixtures. Magma fixtures (also required by the loader) are still in common, unfortunately.

### DIFF
--- a/polyphemus/lib/etls/redcap/lib/entity.rb
+++ b/polyphemus/lib/etls/redcap/lib/entity.rb
@@ -23,6 +23,10 @@ module Redcap
           :field
         end
       end
+
+      def flat_key?
+        false
+      end
     end
 
     def self.create(entity)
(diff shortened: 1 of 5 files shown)


graft:
* To clarify the nature of the bug: the redcap API itself has a bug, where if you request records in 'flat' mode, it will replace values with the corresponding ontology value (e.g., if you use SNOMED, it will return a SNOMED term string rather than the SNOMED id number). However, if you request records in 'eav' mode, it will replace not replace values. Therefore, in order to get the SNOMED terms (we cannot download the dictionary), we need to request the record both in 'eav' mode and in 'flat' mode and reconcile them. The flat records back from redcap appear to be grouped on all keys other than `field_name` and `value`; previously the index into this record was simply the record name, but now that we are iterating across several of the redcap entity keys, we need to index the flat_records differently so they can be looked up correctly.

coleshaw (approved):

PR #474, opened 2021-08-09
Ingest cat files magic oligo

coleshaw:
This PR implements the ingestion idea of a magic string (i.e. an oligo) attached to all sample identifiers submitted by D2B or genomics to the CAT. Files / paths with this oligo (set in a Chef configuration) will be scrapped from the CAT and set in a database. ETLs will be scheduled to download those files to both C4 and Metis. C4 will be used as an active triage area, whereas Metis will be used more for archival purposes, so we have a record of the original file available from the CAT.

This may result in some file / data-block duplication in Metis (depending on how the DS change files during the triage process), but we've decided that is an acceptable tradeoff at the moment.

This PR also removes the ingest "flagging" UI widget in Metis and corresponding Polyphemus API, as those are unnecessary for this workflow.

This PR supercedes #460 and #467.


etna/lib/etna/filesystem.rb
```diff
@@ -377,55 +377,151 @@ def exist?(src)
...
+        @ssh ||= Net::SSH.start(@host, @username, password: @password)
+      end
+
+      def mkdir_p(dir)
+        output = ssh.exec!("mkdir -p #{dir}")
+
+        raise RemoteSSHError.new("Unable to mkdir -p, #{output}") unless 0 == output.exitstatus
+      end
+
+      def lftp_get(username:, password:, host:, remote_filename:, &block)
```
corps, line 504: So I guess the lftp username/password is not necessarily the same as the ssh one?
Oof.  This one is interesting case.

coleshaw: Yeah, this is LFTP from the remote SSH host to a different server (i.e. from Polyphemus, SSH to C4, then from C4 use LFTP to fetch the file from CAT onto C4 -- so different credentials for both C4 and the CAT).

coleshaw: It does feel like the `Filesystem` abstraction may not be the right fit for this use case, so not sure if I should try to stream the file from CAT -> Polyphemus -> C4 (seems like an unnecessary network hop / bottleneck) like the other `Filesystem` impls (which seem to make more sense when one endpoint is Metis), or implement this as something different than a `Filesystem`. Thoughts, @corps ?

dtm2451:
Should we make the oligo something more meaningful than "oligo"?  I wonder if leaving it as "oligo" might actually be more confusing to teach people to add than say "dscolab", "dsco", or "datalib" which reflect that the data will come to us.

coleshaw:
Sure, we can definitely use something like one of those! Right now it should be just `oligo` for the tests and the Polyphemus config template -- the actual value will be set in the Chef role files, for both staging and production.

graft:
>This may result in some file / data-block duplication in Metis (depending on how the DS change files during the triage process), but we've decided that is an acceptable tradeoff at the moment.

I am pretty confident that for the vast majority of files this will not be an issue; fastq files (the bulk of our data) are almost always an immutable artifact. In the event that they are being routinely altered, as by a pipeline, it's a good habit to keep the raw data anyway. On the other hand, we will likely need to keep track of and mark some of these files for destruction in some cases - which actually seems easier if the data blocks are tracked in a database.

PR #475, opened 2021-08-09
Increase build stability, ensure that we only push net source changes…

corps:
…, even with sha injection.

So.  The main reason that everything deployed on every merge before came down to the way that I was injecting the `built-from-sha`, a file which is included in all images containing the last merged SHA so that deployments can include a message about their sha of origin.

This meant that essentially every commit resulting in a new SHA resulted in a new image digest, even if _that file_ was the only change itself.

The new approach, I label the images we built with the SHA that built it, too.  When building new images, I first build using the previous SHA, and check the resulting digest.  **If building an image using the previous SHA results in the same digest, the image does not need updating since only the SHA has changed**.  Similarly, **anytime a built image has a source change, its digest will be different even when the SHA is the same**, signalling that the image does need updating.

this technically requires two builds: a build with the previous SHA and then another build with the current SHA.  fortunately, as far as I have tested, this process is stable and does not add significant overhead.  It also, critically, _does not effect development_.  This image digest 2x build process only works when there is an external tag to push, locally this check is not made (because we prioritize file modified times, which is a safer faster check in most cases).

Furthermore, to closer match production, I would like to retire the psql 9 support in development, currently used by janus in development.  The trick, is that doing so will require a wipe of development janus data.

Before this PR gets merged, I'd like come up with a decent migration plan for this in development that I must think through.


deployer/Dockerfile
```diff
@@ -1,4 +1,5 @@
 FROM deployer
 
 COPY src/deployer.sh /bin/deployer
-RUN chmod +x /bin/deployer
+COPY docker/post-to-slack.sh /bin/post-to-slack
```
corps, line 4: post-to-slack used to be automatically injected into every image, but only etna and deployer images needed it.  Making it an explicit build requirement will reduce the amount of extra builds and deploys from changes to this simple script.

corps:
Ok, added the ability to migrate development janus databases as well.
After this lands, everyone will need to run ./bin/migrate_janus_psql.  This script keeps the old data around, so even if it fails there is a path to recovery.  I feel, 90% confident it should make the migration correctly first try.

deployer/Dockerfile
coleshaw: I plan to use this script in polyphemus, too, will this change affect that? Will I have to inject that script in the Polyphemus Dockerfile?

corps: This script is now built into etna-base, which polyphemus is built on, so it will still have this script.  The change is merely moving that explicitly into that dockerfile as opposed to "magic injection" in the build_image script.

coleshaw:
Looks good to me so far! Only thing for others, is I had to run the `./bin/migrate_janus_psql` command when all the services were down and not running.

PR #476, opened 2021-08-11
add slack notifications to ingest pipeline

coleshaw:
This PR adds some Slack notifications to the ingest pipelines, on the `data-ingest-ping` channel, so we can kind of track which files are available on C4, and when.


corps (commented):
I've been thinking of fixing this, but it seems that `post-to-slack`, and in general the SLACK tokens I generate and use with the webhooks, only support the original channel configured for that token, so the second parameter channel might not actually work correctly?

I'm looking for some help on improving our slack integration options, but I haven't had time to find the more appropriate solution.  Seems like maybe you have to have a different type of integration to have arbitrary channels in the paylod?

coleshaw:
Hm...it seems to work okay? I assume that's how the deployer is able to post to two different channels? I was also able to just manually post to two different channels, just by changing the channel name...

Now, the icon is static, which may throw folks off, but I think that's minor...

corps:
Oh okay, I may be remembering wrong.  Maybe I fixed it already and this was a past issue?

PR #477, opened 2021-08-11
Polyphemus etl view

graft:
This is a WIP; it is still in the sketchy phase while I have been sorting through ideas, but I think those are mostly concrete now. The outline of work is below in some detail.

The linchpin of this effort, I believe, is a usable configuration editor, which would allow the work of maintaining redcap loader configurations to actually move away from engineers to data scientists. As such, my main concern here has been how to build an editor that can be easily updated as the configuration is updated; essentially I want to be able to merely modify a configuration grammar and have the UI build from this grammar, with no project-specific UI code to maintain.

After much putzing around I have settled on using JSON schema to build this. (I took a look at Zod, which looks very nice, but in the end I think it is hard to walk away from the ancient and venerable, which JSON schema certainly is.) After trying to layout the redcap grammar I think JSON schema is not too bad for this task. Furthermore, there is a ridiculous amount of software built on top of JSON schema, including several automatic form generators to edit JSON configs using a schema. I haven't been able to find such a thing for other JSON validation systems. In particular I have my eye on https://github.com/rjsf-team/react-jsonschema-form, which is detailed in its treatment of JSON schema, provides react-ready widgets and has Material UI integration, all of which is very attractive. My hope is this will actually make an editor tractable and easy (after a little bit of schema writing). If not, I will chop this with an axe and just go for a text json editor, which at least will speed configuration cycles somewhat.

1. Add ETL endpoints
  - Jobs - the basic interface to running ETLs, each job exposes a config grammar related to a single ETL (does not require a db table)
  - Configs - the basic interface to running ETLs, consisting of (1) a set of parameters and (2) run instructions (requires a db table).

2. Add UI views
  - Project view
  - Add a new config
    - Set config name
    - Set job type
    - Set configuration
    - 
  - View/Edit config
    - List previous configs
    - Revert to a previous config version
    - Set run interval (disable, run once or run every Xs)
    - Remove config
    - Edit config

3. ETL job runner
  - job runner command consumes the job configs for the next job(s) to run
  - the job is run with the config provided
  - the job status and output are recorded in the config


coleshaw:
Nice, an auto-generated form would be awesome! I was also concerned that making an entire config UI would be a lot of work and difficult to maintain -- this would be super convenient.

corps:
Taking a closer look.

etna/packages/etna-js/utils/types.jsx
```diff
@@ -16,3 +16,5 @@ export const flatten = (array) => [...new Set([].concat(...array))];
 
 export const mapObject = (o, f) =>
   Object.assign({}, ...Object.keys(o).map((k) => ({[k]: f(k, o[k])})));
+
+export const getType = v => Object.prototype.toString.call(v).match(/^\[object (.*)\]$/)[1];
```
corps, line 20: This works, but have we considered `v.constructor.name` ?  Works as well without depending on regexing the inner name.  Super minor nit though.

corps (commented):
Yeah, looks good so far as a WIP.  The real question will be surrounding wether the form autogeneration works.

graft:
Okay, the latest commit uses the `json-editor` npm package, which is a pure JS implementation with no dependencies. Astonishingly, this works quite well to render a pretty decent form according to the schema.

However, with 3000 lines of JSON I end up with 76000 DOM nodes, which are all drawn at once. The upshot is that on my laptop the page takes about a minute to stabilize before you can edit. I'm not sure what the interactive experience is like in general, but it feels pretty unstable. A shame, because the form and interactions look pretty nice and would really get the job done.

I'm inclined to disfavor this because of the performance issues, which in the end become usability issues. I'm not sure I have many json-schema-form-generation bullets left to shoot here; I think the next alternative is likely building JSON-schema validation into CodeMirror, but I'm not sure if this easy enough for a data scientist to use.

coleshaw:
Could you break the form / schema into pages or sections? i.e. fill it out by REDCap form, or like 5 Magma attributes at a time, or otherwise paginate somehow?

graft:
Hmm, pagination would be tough because the JSON is so nested... perhaps I could present a static view with some hooks for editing sub-blocks, so only instantiating the editor in pieces. I will fiddle a bit..

polyphemus/lib/client/jsx/etl/run-pane.tsx
```diff
@@ -0,0 +1,125 @@
...
+  const reset = () => { setRunState(getRunState(run_interval)); setRunIntervalTime(getRunIntervalTime(run_interval)); setParams(params) };
+
+  const classes = useStyles();
+
+  return <EtlPane mode='run' selected={selected}>
+    <EtlPaneHeader title='Run'>
+      <Grid className={classes.runbar} spacing={1} container>
+      {
+        (newParams != params || (runState != run_interval && (runState != RUN_INTERVAL || runIntervalTime != run_interval))) && <React.Fragment>
+          <Grid item><Button onClick={ () => update({run_interval: runValue(), params: newParams}) }>Save</Button></Grid>
```
graft, line 76: This posts an update, which ought to reset the state (and thus make the save button disappear), but I can't quite make it work if you update params, probably because of the weird way I am initializing newParams.

graft:
I cannibalized the Polyphemus::Job class to run these EtlConfigs, which I am a bit sad about because the RedcapJob streaming stuff was so neat. However, it doesn't seem possible to maintain both methods of running the configuration, so I have taken down the jobs API (although not yet the etna command which invokes this API).

janus/lib/models/token.rb
```diff
@@ -123,6 +125,9 @@ def valid_task_token?(janus_user)
       return false if project_name == 'administration'
       return false if token_role == 'administrator'
 
+      # if the user is a supereditor, the request is honored
```
graft, line 128: This and related changes introduce the ability for a supereditor (or superuser) to make a task token for any project without having to be a member of that project. Ergo, Polyphemus has to be a supereditor now (i.e., in the administration group), but no longer has to be licensed to specific projects in order to run their jobs.

corps: At some point, we should consider refresh tokens that are still authorized and scoped by individual users.  My concern is the way that super editor contains the unique privilege of not only having access to everything, but effectively impersonating anyone as well.  That said, this works for now for sure.  Before we get to the point of refresh tokens we'll need to think the executor approach anyways.

corps:
Some broad questions while I'm trying to understand the scheduler,

How do you think your scheduler is going to synchronize across nodes?  If there are multiple schedulers running, how do they cooperate about starting and stopping jobs, or really keeping up with their state?   Is the intention that your scheduler only have one instance, or?  What is the integration of this with a broader pipeline look like?

How does your scheduler work when it needs to schedule work on a non local node?  If it needs to run non ruby code, say an arbitrary bash process, in an arbitrary image, will that image be required to maintain a polyphemus database connection to collect secrets and indicate running / failed states?  Will the executor do that and inject it into the environment?

If a job stalls or fails to start, does your scheduler plan to have things like auto retry built in?  Will display multiple runs and includes logs for historical failed runs?  Do owners get notified of failures or do they just need to check?

How will job dependencies be expressed?  If, say, the redcap job needs to be broken down into smaller parts to reduce fragility, how will data and scheduling work in that case?  (Example, we separate the redcap pull from the parsing from the magma push into separate jobs so that unreliability or issues in each does not break the entire etl and you can restart from a specific location).  I know you've expressed our etls as only being single node jobs, but I can think of many reasons why that won't be the case for very long.

What''s the story for making this compatible for our other existing etls, and swarm in general?

I ask because... I don't know how to review this code.  Is it throw away code?  If so, I don't have to nit pick it, it works to get the redcap etl out for now.  Is it code you intend to keep around and we're going to be supporting more long term?  Then I need to understand your bigger thought process.

You told me a few things:
1. That you've evaluated airflow and decided that it doesn't match our use case.
2. That you think you can fit the airflow as the scheduler backing your UI.

I need more details on both and what your plan is.

corps:
I guess, really... I feel like a lot of this code feels, unnecessary?  Incomplete?  A deadend.
Despite the massive amount of new code being introduced, and the massive amount of new pieces we will have to maintain, plenty of bugs we'll need to hunt down and features expanded, not a single piece of it feels... necessary, or really useful.

I feel like we could have just come away with a really nice form to help people edit the redcap modeling, and we plug that directly into the etl for now.  Then, we could have spent time evaluating a production grade tool (like ariflow, or something else, whatever), to fill in these gaps.

Instead, we've taken a long weaving path to... making something that doesn't clearly need to exist.  And we still, don't have a great story for non technical people to model redcap ingestion.

janus/lib/models/token.rb
coleshaw: Are we still giving super-editor privileges to non-DL folks? At some point I thought we gave Vincent some elevated privileges...?

polyphemus/lib/client/jsx/etl/run-pane.tsx
coleshaw: If you want `newParams` to be set to the `params` prop, perhaps you can initialize it to some empty value and then use `useEffect` to initialize it when params changes (which sounds like it might be what you want??). Something like:

```
const [newParams, setNewParams] = useState({});

useEffect(() => {
  setNewParams(params);
}, [params]);
```

It's probably not resetting because state won't re-initialize when `params` changes, it's not like a controlled input value.

polyphemus/lib/models/etl_config.rb
```diff
@@ -0,0 +1,128 @@
...
+    def validate_secrets(secrets)
+      etl_job_class.validate_secrets(secrets)
+    end
+
+    def validate_params(params)
+      etl_job_class.validate_params(params)
+    end
+
+    def run!
+      update(ran_at: DateTime.now)
```
coleshaw, line 42: Should this only be updated after successful runs? Or it's okay that a failure also updates the ran_at value?

polyphemus/lib/polyphemus/controllers/etl_controller.rb
```diff
@@ -0,0 +1,3002 @@
...
+      config: {},
+      secrets: {},
+      params: {},
+      run_interval: Polyphemus::EtlConfig::RUN_NEVER
+    )
+
+    success_json(etl_config.as_json)
+  end
+
+  def old_list
```
coleshaw, line 105: Is this basically the old config JSON, for reference? I don't see it being used anywhere, but just wanted to verify.

polyphemus/spec/spec_helper.rb
```diff
@@ -29,9 +29,11 @@
 RESTRICT_BUCKET = Polyphemus.instance.config(:metis)[:restrict_bucket]
 MAGMA_HOST = Polyphemus.instance.config(:magma)[:host]
 REDCAP_HOST = Polyphemus.instance.config(:redcap)[:host]
+JANUS_HOST = Polyphemus.instance.config(:janus)[:host]
 TEST_TOKEN = Polyphemus.instance.config(:polyphemus)[:token]
 
 PROJECT = "mvir1"
+REDCAP_TOKEN='09ED4B852506E81DC4E192061D602934'
```
coleshaw, line 36: I hope not an active token...?

polyphemus/lib/models/etl_config.rb
coleshaw: Do we notify folks of failed or succesful runs? How will people get feedback, other than refreshing the ETL page?

graft: Unfortunately refreshing is required. Ideas welcome to fix this.

polyphemus/spec/spec_helper.rb
graft: Random number

coleshaw:
So, some other thoughts...

I agree that the iterative process is particularly painful, and poor Bushra had to deal with me fumbling through it. I think JSON validation in the UI will help a bit, though I do wish we had some non-JSON way to edit the config. It's tough to keep all the pieces in your head when the config params are organized in a hash. If we can find some form-ish way to do that without it being custom for each ETL, that would be a big win and make it much more approachable.

How are you dealing with the need for Ruby models? Perhaps I missed it, but I didn't see anything for the user to set up the corresponding code for manipulating record names or other parameters. Does that still exist, and what is the thought on how new projects will set that up before users can run ETLs for other projects? Is that still in engineering-space, and somehow we have to disable REDCap loading for new projects until we set that up?

I'm glad you encrypted the secrets column. I am concerned that the secrets management will be brittle and leak somewhere, though. For example, we have censoring of the secrets not only in this PR, but also in the `etna` routes logger, and now those separate lists of secret keys have to be kept in sync (chef recipe and here). It seems easy to forget, especially if we add new loaders.

> I think this is exactly what this PR is; there is some fill-in glue right now that stands in for airflow, whatever, but it is not important. Beyond that I think the main thing I want to hold onto in the long-term is doing this work in user-space and not in engineering-space. The main utility in that regard is simply providing a UI for controlling/editing the project-specific configuration of ETLs (at least the Redcap loader) and attaching them to projects; what happens after the fact with this configuration (how they are executed) is less important to me, and consequently those parts of the code can (should) be jettisoned soon when we figure out a better way to do that.

If the main importance is to put the project config into user space, would it make sense to add that benefit while minimizing the glue code? i.e. allow config in a UI, but run synchronously via the gem or a UI button using the old, threaded method (the main difference being that no scheduler / saving of secrets / params required).

Or, if you feel like those glue parts are still important to improving the workflow experience, should we start working on a migration plan to "whatever the better way" is?

polyphemus/lib/models/etl_config.rb
coleshaw: I've baked in "send message to slack" as the default notification for errors in the other ETLs....might be an option to consider, you might be able to even mention (`@`) the user who ran the job, so they don't get lost in the notification channel.

graft:
> If we can find some form-ish way to do that without it being custom for each ETL, that would be a big win and make it much more approachable.

I agree, I would vastly prefer this. I did my best but couldn't make it work; a validated json is my stop-gap and is marginally better than the current situation. Perhaps Zach's idea of auto-generating the code for a form whenever the schema is updated can work.

> How are you dealing with the need for Ruby models? Perhaps I missed it, but I didn't see anything for the user to set up the corresponding code for manipulating record names or other parameters. Does that still exist, and what is the thought on how new projects will set that up before users can run ETLs for other projects? Is that still in engineering-space, and somehow we have to disable REDCap loading for new projects until we set that up?

It still exists - I think right now the loader will simply crash if you try to set one up on a project without ruby models setup. I don't have a good way to avoid this (i.e., it seems very hard to put into user space), as I don't have a good way to avoid the need for patch code in the loader in general. There might be a way to do this later if we are using Airflow, where users may define configs as python, but that might involve porting a lot of stuff into Python.
 
> I'm glad you encrypted the secrets column. I am concerned that the secrets management will be brittle and leak somewhere, though. For example, we have censoring of the secrets not only in this PR, but also in the `etna` routes logger, and now those separate lists of secret keys have to be kept in sync (chef recipe and here). It seems easy to forget, especially if we add new loaders.

I'm not pleased at all about having to do secrets this way; however, I'm not sure the redcap token is a secret that can be baked into a config (especially while they are user-specific), since they often need to be swapped in or out while running the loaders (e.g. the ipi loader, which runs across a dozen different redcap instances/tokens).

> If the main importance is to put the project config into user space, would it make sense to add that benefit while minimizing the glue code? i.e. allow config in a UI, but run synchronously via the gem or a UI button using the old, threaded method (the main difference being that no scheduler / saving of secrets / params required).
> 
> Or, if you feel like those glue parts are still important to improving the workflow experience, should we start working on a migration plan to "whatever the better way" is?

This idea didn't occur to me until later (just using the jobs API to run the loader from the UI), so I haven't fully digested it.   Running the job synchronously has also been a bit problematic in terms of timeouts; I'm not sure the method in this PR will avoid all of those timeout issues, but I was also hoping that making the execution asynchronous would free the workflow up to move to running in a real scheduler - that is, the ersatz scheduler here is a better dummy for how we want things to run.

The only real workflow disadvantage (other than potential timeout issues) I can see with this method compared to the current method is you couldn't run a job on an interval this way, but we haven't been running redcap loaders that way anyway, so perhaps this is a better path until we can use airflow to manage secrets, as avoiding the secrets issues seems worthwhile.

PR #479, opened 2021-08-11
modified style tag of the commit button

andresvicente00:


dtm2451:
Oooh!  interesting.  Thanks @coleshaw!  Let's me take a peak at this first, @andresvicente00, and then we can work on this piece next week.

Separately, I'd like you to 1) make some stylistic modifications to this code you changed.  If you look at the rest of the file, you'll notice that lines are normally indented, and outdented, incrementally based on their relation to lines above and below, but parts of your bit jump a round.  Let's try to follow the previous style.  I'd like you to just add some white space to do so, and make the code looke more like:

```
    <button
        tag1=......
        tag2={{
            background: ...,
            key2: ...,
            key3: ...,
            key4: ...}}
        tag3=......>
        Commit
    </button>
```

2) I also think that the button is perhaps even more bold now than we actually want (so you did too good a job in a way!).  Perhaps you could adjust or remove some of these style adjustments?  Then **`git pull`** to grab Cole's update to your branch. 
Then save your changes with a new commit (remember to `git add path/to/input_state_management.tsx` first, then it's `git commit -m "<description of change>"`), then `git push` back to here.

PR #482, opened 2021-08-12
Ingest tweak ssh host keys

coleshaw:
Small change to the ingest SSH host key script for Polyphemus, to delete any existing `known_hosts` file before appending new ones. This allows us to run the script on a systemd timer when deployed, to catch any new / changed SSH hosts. Current behavior is that the file ingest will fail (at least for Metis ingest) if the SSH host key is not recognized. I suspect the C4 ingest would also fail. 

I plan to run this infrequently compared to the ETL jobs, on the assumption that host keys won't change that often. If any key changes, the ingest ETLs will fail until the keys are updated again, but should pick right back up after they are updated.


corps (approved):

coleshaw:
Yeah, I agree, this makes it blindly trust whatever that DNS record is pointing to .... as an alternaive, do we just capture the keys and bake those into the image, and if one changes or we add a new host, just update the image?

corps:
> Well, I may wind up baking them into the image anyways. I'm having issues with getting the hosts file to actually persist or get created on deployments...

To persist, you'll need to add a volume to the app.  essentially something like this: `-v polyphemus-ssh:/root/.ssh`

coleshaw:
Ah...so even though the `systemd` job says it's running inside the `polyphemus_app` container, the file is still ephemeral because it's not a volume?

corps:
> Ah...so even though the `systemd` job says it's running inside the `polyphemus_app` container, the file is still ephemeral because it's not a volume?

The name given to the container is consistent, but the job is running with `--rm` which essentially removes the container on each run, destroying its ephemeral state.  This is ideal, because any expected persistence should be required to be declared (and thus explicitly trackable) rather than be implicit because the code writes to a location and hopes it remains.

PR #485, opened 2021-08-13
enables file delivery via xsendfile for Metis and Vulcan; removes fro…

coleshaw:
…m Magma

This PR re-enables xsendfile in dev environments for Metis and Vulcan, and removes it for Magma (I don't think Magma serves its own files? It always directs to a Metis download URL).

@corps , not sure if it was better to mount the `_app_fe` includes directory as I did here, or copy it in via some sort of Dockerfile command? Since only a couple apps use it, I just mounted it as a volume...


corps:
Thank you for digging into this.  This, works, but I worry if it's too specific?  Like it's not clear that only etna style apps will want to use apache frontends, and that only includes will be used as the hooks.  I'll jump into the branch real quick and adjust.

corps:
I mean, but you are right, I actually use this naming convention.  Honestly, my hope is eventually to fully retire our little apache instances running as frontends and find a much cleaner solution, it's all still not 100% yet clear.

PR #488, opened 2021-08-17
Ensure js routes are created in timur.

corps:
So,

The issue was that the timur image was not being created with public/js/routes.js, which depends on an invocation to timur the ruby app.

Before, this worked via build hook and was running successfully.  After changing how our build works, the environment variable it depended on had changed, but no tests existed to verify this integration.

Fix:  I added an integration test that fails with the current image, then adjusted the environment variable, and verified the integration test passes.


coleshaw (approved):
Seems to work for me locally, now.

corps:
Those test suites were nice and fast, yum

PR #489, opened 2021-08-17
paginate the results for metis files for magma records etls

coleshaw:
I think the ETL is failing on production because it's trying to link too many files at once, so the Metis response times out and thus Magma won't execute the update to `rna_seq`. This PR re-configures the ETL to return batches of Magma record_name / file combinations, in hopes that smaller batches will circumvent the Metis time outs.

### DIFF
--- a/polyphemus/lib/metis_file_for_magma_model_etl.rb
+++ b/polyphemus/lib/metis_file_for_magma_model_etl.rb
@@ -28,7 +28,8 @@ class Polyphemus
       metis_client: nil,
       magma_client: nil,
       file_name_globs: [],
-      metis_path_to_record_name_regex: nil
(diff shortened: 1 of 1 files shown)


corps:
Yes.  So in general, it's puma thread saturation.  Actually!  Let muse grafana and pull up a graph demonstrating the problem.

coleshaw:
> Yes. So in general, it's puma thread saturation. Actually! Let muse grafana and pull up a graph demonstrating the problem.

Ooo, visuals!

corps:
So, I confirmed this is not puma saturation, but likely time outs, instead.  Unfortunately, while metis isn't on the docker network, I'm not yet tracking timeouts / status returned from ruby, only visit counts. 

So, I'll think about collecting that metric soon.

PR #492, opened 2021-08-18
Ensure more correct build_image behavior, include system files in etn…

corps:
…a-apache by default.

This will further improve deployment stability by ensuring build_image always has the most up to date image to compare against by issuing a pull in the cases that it was not already available locally.

This also addresses the user issue for edge apaches by asking them to include the soon to be added user.system.include file which will downgrade them into the runner user correctly.

This is safe to deploy before the corresponding chef change because it will be a no opt when there is no system.include file in the image currently.

Following will be the appropriate changes on the chef repo.


bin/x-all-images
```diff
@@ -44,7 +44,7 @@ for image in $(ls $DIR/../*/Makefile | grep -v docker | xargs -n 1 dirname | xar
 done
 
 for image in $(ls $DIR/../*/*_app_fe/Makefile | grep -v docker | xargs -n 1 dirname | xargs -n 1 basename); do
-  echo "${IMAGES_PREFIX}$(basename $image)${IMAGES_POSTFIX}" >> $tmp
+  echo "${IMAGES_PREFIX}$(basename $image)_app_fe${IMAGES_POSTFIX}" >> $tmp
```
corps, line 47: oh whoops, I think I got this wrong, let me revert.

docker/archimedes-base/Dockerfile
```diff
@@ -38,21 +38,6 @@ RUN curl -o /tmp/dockerize.tgz -L $DOCKERIZE_URL && ( cd /usr/bin && tar xzf /tm
 RUN curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/install-poetry.py > /opt/install-poetry.py
 RUN python /opt/install-poetry.py --version $POETRY_VERSION
 
-# Install R from CRAN repo, to get latest
-RUN apt-key adv --keyserver keyserver.ubuntu.com --recv-key '0xE19F5F87128899B192B1A2C2AD5F960A256A04AF'
```
corps, line 42: R is not needed in archimedes right now, and recently we discovered that this package source fails to install due to some recent changes upstream.  Rather than get that working again, let's remove it for now since it is not even used.

coleshaw (approved):

coleshaw:
Hm...looks like the `rpy` dependency in Archimedes also needs to be removed...

corps:
> Hm...looks like the `rpy` dependency in Archimedes also needs to be removed...

Done, thanks!

PR #493, opened 2021-08-18
IPI etl rna seq attributes

coleshaw:
This PR adds an ETL to scan Metis for rna result TSVs (organized by plate). It parses that TSV and uploads the metrics / data to Magma.

This also moves the IPI renaming JSON file into it's own repository, much like the REDCap projects. No strict need for it, but basically just keeping that information tracked but separated.


polyphemus/Makefile
```diff
@@ -7,6 +7,12 @@ include ../make-base/stubs.mk
 	else \
 		touch .projects-mark; \
 	fi
+	git clone git@github.com:mountetna/renaming-projects.git lib/etls/renaming/projects || CLONE_FAILED=1
```
corps, line 10: I think you want this to happen before .projects-mark is touched, otherwise this might only attempt once, and on failure, not attempt again.

See how the `docker-ready` command depends on `.projects-mark` to run this target.  The touching of the file is what marks this as "done successfully"

corps (approved):

PR #501, opened 2021-08-19
Adding swarm agents and portainer stacks

corps:
So yeah, a little preview of the future of deployments.
Write a stack using docker-compose syntax.  Run it locally and verify it works.  Then use the portainer ui to deploy it into our cluster.  
When images update, they'll get automatically restarted.

No ssh.  No chef.  -Rollback built into the UI-.


coleshaw:
> Write a stack using docker-compose syntax. Run it locally and verify it works. 

Can you give an example of what that stack looks like?

Will that stack file include things like the systemd jobs, or updating configs (i.e. Chef roles)?

coleshaw:
And does portainer keep a log or history of such stack files? So I can see what other folks are doing for example, and learn how to deploy certain things?

PR #502, opened 2021-08-19
Add a bar_plot plotting function to archimedes, with accompanying vulcan UI

dtm2451:
I've prepped a python script that I'll have @andresvicente00 help me finish and turn into a function.

First we'll update this script so that it can make a relatively simple version of a bar plot from hard-coded data, then we'll add in some additional features, and then we'll convert all that into a function that will nolonger be hard-coded!

That'll probably be the scope of this PR, and we'll have a separate PR for linking the function in with a vulcan-UI.


dtm2451:
Should be good to go @andresvicente00 =)

PR #505, opened 2021-08-23
Metis database connections

coleshaw:
So, trying to improve Metis performance, I think the ETLs were creating contention for db connections because the `folder_path_calculator` class wasn't actually doing what it was intended to do :facepalm: . So this PR attempts to address some database / performance issues. I'm not sure this will actually improve behavior in production, since I haven't been able to duplicate the pool timeout issues locally, but I'll keep investigating.

Putting out the PR to get some initial reviews.

### DIFF
--- a/metis/lib/folder_path_calculator.rb
+++ b/metis/lib/folder_path_calculator.rb
@@ -17,7 +17,8 @@ class Metis
       @path_cache = {}
     end
 
-    def get_folder_path(folder)
+    def get_folder_path(folder_id)
+      folder = @folders_by_id[folder_id].first
       path = [folder.folder_name]
       folder_to_check = folder
       loop do
@@ -26,15 +27,17 @@ class Metis
 
         # Use the cached path value for the parent folder if it exists
         if @path_cache.has_key?(folder_to_check.folder_id.to_s.to_sym)
-          path.unshift(@path_cache[folder_to_check.folder_id.to_s.to_sym])
+          path.unshift(*@path_cache[folder_to_check.folder_id.to_s.to_sym])
           break
         end
 
-        # Add the parent folder to the list
+        # Add the parent folder to the path
         parent_folder = @folders_by_id[folder_to_check.folder_id].first
         path.unshift(parent_folder.folder_name)
         folder_to_check = parent_folder
       end
+      # Once we've determined the path to a folder, we'll add it to our cache
+      @path_cache[folder_id.to_s.to_sym] = path.dup
       path.join('/')
     end
   end
--- a/metis/lib/folder_path_calculator.rb
+++ b/metis/lib/folder_path_calculator.rb
@@ -36,7 +36,7 @@ class Metis
         folder_to_check = parent_folder
       end
       # Once we've determined the full path to a folder, we'll add it to our cache
-      @path_cache[folder_id.to_s.to_sym] = path.dup unless @path_cache.key?(folder_id.to_s.to_sym)
+      @path_cache[folder_id.to_s.to_sym] = path.dup unless @path_cache.has_key?(folder_id.to_s.to_sym)
       path.join("/")
     end
   end
--- a/metis/lib/folder_path_calculator.rb
+++ b/metis/lib/folder_path_calculator.rb
@@ -4,10 +4,9 @@
 
 class Metis
   class FolderPathCalculator
-
     def initialize(all_folders: nil, bucket:)
       all_folders = all_folders ? all_folders : Metis::Folder.where(
-        bucket: bucket
+        bucket: bucket,
       ).all
 
       @folders_by_id = all_folders.group_by { |fold| fold.id }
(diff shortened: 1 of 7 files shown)


metis/lib/metis.rb
```diff
@@ -13,7 +13,6 @@ class Metis
   def setup_db
     @db = Sequel.connect(config(:db))
     @db.extension :connection_validator
-    @db.pool.connection_validation_timeout = -1
```
coleshaw, line 16: @graft , I took this line out because, per my reading of [the documentation](https://sequel.jeremyevans.net/rdoc-plugins/files/lib/sequel/extensions/connection_validator_rb.html), using this setting may be less performant, in that a query must be executed each time to validate a connection in the pool? It also looks like we're not using the suggested coarse connection checkout in middleware. Testing this locally I can't really tell a difference in performance with this setting and without -- curious if you recall why you disabled the connection caching, originally? With the increased usage of Metis for ETLs and uploads, I feel like connections shouldn't expire, naturally, and forcing a validation each time may be unnecessary ...

metis/lib/server/controllers/metis_controller.rb
```diff
@@ -36,14 +36,15 @@ def require_folder(bucket, folder_path)
...
+      #   doing it in the database because there is no bulk way to do that,
+      #   and is recursive per file. Try to save queries / connections
       folder_path_calc = Metis::FolderPathCalculator.new(bucket: bucket)
 
       return [] unless files
 
       files.map { |file|
-        if file.folder
-          path = "#{folder_path_calc.get_folder_path(file.folder)}/#{file.file_name}"
+        if file.folder_id
```
coleshaw, line 46: My working assumption is that calling `file.folder` caused a database query, so each time a file was expanded to have it's path calculated, the database was hit. Using just the `folder_id` should circumvent that, and reduce contention for db connections...

metis/lib/folder_path_calculator.rb
```diff
@@ -17,7 +16,8 @@ def initialize(all_folders: nil, bucket:)
       @path_cache = {}
     end
 
-    def get_folder_path(folder)
+    def get_folder_path(folder_id)
+      folder = @folders_by_id[folder_id].first
```
coleshaw, line 20: Basically this method / class wasn't doing any path caching or folder caching...oops. :facepalm: Hopefully these changes fix that, and paths are cached in memory and the "bucket-wide" folder list is only fetched once from the database. Not sure how I can measure or see differences locally, but seems like this should be fixed regardless.

coleshaw:
If these changes don't help, I'm thinking to increase the number of connections in the pool (default is only 4), to say, 6 or 8? @corps , do you think the VM would support that? We seem close to some performance limit, so perhaps a small increase would improve things.

metis/lib/metis.rb
corps: Hmm.  I do think it's wise to leave this, tbh.  I don't think it effects our performance significantly and it could help reliability of the pool.  I just don't think this is causing a significant performance issue.

metis/lib/server/controllers/metis_controller.rb
corps: True.  I think this change will in fact help.

coleshaw:
In that case, should we implement the coarse checkout, or we should be okay leaving it as is?

corps:
> If these changes don't help, I'm thinking to increase the number of connections in the pool (default is only 4), to say, 6 or 8? @corps , do you think the VM would support that? We seem close to some performance limit, so perhaps a small increase would improve things.

I think it is very, very reasonable to increase the connection pool size, maybe up to 8 and see.

corps:
> In that case, should we implement the coarse checkout, or we should be okay leaving it as is?

Yeah, we could try doing that in the etna controller middleware?

corps:
Also, thinking about it more, based on the documentation,
we should increase the pool size to 16 to match the puma concurrency.  Basically every etna controller web request now will checkout the connection as it starts (which is fine, given that the vast majority of endpoints require a db connection), thus there needs to be an equal amount of connections as concurrency.

coleshaw:
Hm, okay, I can update the config for that. So then this:

>Note that setting up coarse connection checkouts reduces the concurrency level acheivable. For example, in a web application, using Database#synchronize in a rack middleware will limit the number of concurrent web requests to the number to connections in the database connection pool

should be okay if we match at 16?

corps:
> Hm, okay, I can update the config for that. So then this:
> 
> > Note that setting up coarse connection checkouts reduces the concurrency level acheivable. For example, in a web application, using Database#synchronize in a rack middleware will limit the number of concurrent web requests to the number to connections in the database connection pool
> 
> should be okay if we match at 16?

Yessir.  This is also -per process- so when we increase the number of web servers, or have a job running, they have their own independent pools.

Or, if you don't feel comfortable... we can remove the pool check altogether like you said, and check its impact.  My concern with doing that is that it may cause occasional errors when a connection is yielded that is actually reaped, but I don't _know_ in our case how often that occurs.

coleshaw:
I'm okay leaving it in for now, since things seem to run fine with it. Let's change things incrementally and see what helps. I'll put in this fix and add in coarse checking, then see what the behavior is like, then increase pool size. We can then remove the check as a third thing to try if needed.

PR #507, opened 2021-08-24
break up processed rna seq files etl

coleshaw:
This PR breaks up the one IPI ETL for rna_seq processed files into individual ETLs. The joined query was timing out on Metis (hitting the Apache timeout limit from edge-apache to metis_app_fe), so I think it makes sense to break it up and make each one more streamlined. 

I also paginated the metis find request when collecting all the files, so that a single request doesn't lock up a Puma thread / process beyond the timeout.

Note that this currently doesn't handle any auto-deletion of files on Magma (unless it's a file_collection being reduced in number of items ... should we also prevent that?). I think that's generally safer, since Magma is supposed to be the source of truth, and we probably don't want folks moving stuff or deleting stuff from Metis and then having that affect Magma records.


polyphemus/lib/etls/link_files_base_etl.rb
```diff
@@ -0,0 +1,103 @@
+require_relative "../metis_file_for_magma_model_etl"
+
+class Polyphemus::LinkFilesBaseEtl < Polyphemus::MetisFileForMagmaModelEtl
```
corps, line 3: :+1: 
Awesome!!! I like this alot.

corps (approved):

polyphemus/lib/metis_file_for_magma_model_etl.rb
```diff
@@ -48,51 +48,13 @@ def initialize(
       @record_name_gsub_pair = record_name_gsub_pair
       @limit = limit
 
+      @metis_file_limit = 200
+      @metis_files_for_cursor = {}
+
       super(
         cursor_group: EtlCursorGroup.new(file_cursors),
         scanner: HashScanBasedEtlScanner.new.start_batch_state do |cursor|
-          metis_request = Etna::Clients::Metis::FindRequest.new(
```
coleshaw, line 54: I didn't realize that a new Scanner was getting initialized for each batch of the cursor, so this expensive find request was getting run pretty often. I've moved that logic out of the scanner and into the ETL itself, so it can be cached per ETL run...

corps: If I have to insist, I think that it does worry me a bit that our ETL scanning is inefficient due to using find based logic.

I do think, eventually, we need to refactor the way we pick up files in such a manner that it is efficient to run.  While reducing how often the expensive query gets run is helpful, it may be symptom of the larger issue that if the find queries are slow, it probably will get much slower as more projects get added anyways.

I don't yet have any more useful thoughts atm other than this intuition.

coleshaw: Yeah, I went through several iterations for this base ETL / Cursor trying to find what made more sense. I think there are several factors at play, which perhaps require different approaches.

- file_collection attributes seem to require some find-type operation, because you need to get all the related files at the same time. I suppose you could try to do it from the Metis side with a less expensive, time-scan type find (checking updated_at), but then that creates added logic around adding or removing items from the current value of the attribute, which could get messy. Perhaps easier if you ignore removing files from the collection, but seems undesirable.
- For file attributes, you could do the ETL in the reverse fashion, perhaps using a time-based scanner against Metis (again, a less expensive find), and then link that into Magma. I think you'd still have to run Magma queries to make sure the records exist, but this could be better, especially now that the processed file ETLs are broken up by attribute.

Note that Metis find is scoped to a single bucket, so I don't think performance degrades as a function of number of projects. Rather performance should be negatively impacted by how many files / folders are in a given bucket...

coleshaw: For all the ETLs where we're scanning file contents ... I think you could do the same thing as for file attributes (Metis time-based scan and push results to Magma). But one issue I ran into when trying that method, was when doing the Magma record-exists checks, it felt like the etl `process()` call should only be called with valid data, instead of having to filter the data inside of `process()`...

corps: What about changing magma to return metis md5's?  Maybe by flag?
So you can magma records for updates, capturing the md5s, and instead of a generic find operation in metis, you can fetch the unique set of md5 files.  This is indexed by block, which can join inexpensively back to file by project name.  
Might warrant a fresh batch of endpoints, but I feel increasingly like ways of accessing metis by md5s maybe be important and useful.  From a security standpoint, this functionality is "internal only" and would either only work for system users or possibly use alternative internal authentication (something that may become more common over time, btw).

Btw, after we move our services to the swarm, internal mutual auth becomes 'free'.  This is built into the swarm mesh, we'd have unique urls we can reference services from each other that is inaccessible by the public web.

coleshaw: I'm not clear how accessing Metis MD5's from Magma would help in the case of these ETLs? For most of these ETLs, there is no existing Magma - Metis linkage yet, so there are no md5's to capture on the Magma side of things (at least in the first run of the ETL). Subsequent ETL runs could fetch the md5s from Magma, but wouldn't you still need to run some sort of Metis query to detect if the md5 changed, i.e. a new file was uploaded or replaced? I can see how accessing Metis by md5's could be very helpful, though.

Also, scanning for Magma record updates doesn't help, I think, because Magma record updates happen independently of Metis file uploads ... so I think you would miss a bunch of stuff? i.e. Arjun and Urmila upload a ton of Metis files....but they don't touch Magma, so there's no indication on the Magma side of things that there are new files on Metis...

corps: Right, sorry, I was thinking in the wrong direction.

What about this approach:
A metis time cursor is used to scan new file names and match them against a regex, recording somewhere else matched regexes and their theoritical magma record identifiers and matched patterns.

Another cursor in magma watches for new records to appear, and checks the above intermediate data for a match, then links once a magma record is created that hits a match.

The trick of course is that, currently, metis etls do not track deletes, but I think they ought to.  For now, we could get around this problem by just doing a batch query for files by id to see which ones actually exist at the time of linking.  This is still better than path based searches.

This also requires an intermediate data table, or putting the data into the hash cursor.

corps: You'll have to record the point at which a 'matching' record has been linked, so that unnecessary relinking does not occur.  All that state can be captured somewhere I think.

corps: Or, we just make path based find much faster :P

coleshaw: Hm...so basically do some intermediate processing, and build ETLs off of that? Similar to how we're doing an ETL sync from the CAT into a database table, then another ETL from that table to download files?

coleshaw: I'm not sure the Magma cursor would be able to just watch for new records (since that doesn't happen that often), but it could scan that intermediate data table, I think?

coleshaw: I was trying to avoid data base hits for each result and doing the path calculation in Ruby, but perhaps that's slow and I should just let the database handle it?

coleshaw: Just ran a small test locally...using database for path seems much slower than doing it in Ruby...

corps: Yeah, I think the recursive database function is going to be slow.

I have an idea, though.  Is it possible for the folder find method to use a parent id + a single regex?  So instead of materializing the entire path and doing recursive searches, you are required to first provide a parent folder id (or null for bucket folders) that narrows the search immediately, and then do a single regex scan on the (requires adding an index) folder / file name?  This pushes more logic on the client which must descend recursively with the regex, and it prevents regexes that span multiple folders, but it would be decidely less db impact.

corps: Actually, the index on the folder / file name may not be necessary immediately, each folder likely does not have _that_ many entries.  A full scan from the parent id is not too expensive likely.

coleshaw: Hm, is that equivalent to the client using the "list folder" endpoint (just list immediate children directories and files of a given folder) and having the client itself traverse the tree?

corps: This may actually 'unexpand' the query into an N * N + 1 query for recursive regexes, but I think it's not as bad as it seems at first glance.  You know, in fact, I bet we could use an etl to follow folder changes and 'predetermine' if they match a regex and are a valid root for future searches, meaning that finding files becomes using this ETL generated list of target folders and then regex searching files just inside that one folder.

Actually, hold on.  What about an etl that simply watches folder changes and 'marks' them as targets for linking?  Then, each directory is processed with a single, simple list folder + ruby regex.

corps: In other words, I feel we can greatly reduce the impact of these searches by doing some precomputation in an etl to greatly narrow the db search target down.

coleshaw: That is fairly slow from a shear number of requests, I think, since many of the folder structures are very branchy...each request is fast, though.

corps: So two etls:

1. etl determines 'target folders' whose paths match a specific pattern and puts that metadata somewhere.  attempts to link all folder contents when a target foldr is found
2. etl that follows file changes, and when a file change happens inside a target folder, attempt to link it.

This ensures regardless of the order (an existing folder is renamed / created that matches OR fild is added later to existing target folder) of operations, linking happens correctly, and with much more precomputational work being done.

coleshaw: Yeah, that makes sense. I think we'll get more benefit from doing some sort of pre-computation into an intermediate table or something, versus trying to squeeze a bit out of find...let me think about how to restructure the ETLs. I think there are weird edge cases around people renaming files or uploading updated files, but as long as we store the hash in the intermediate table, should be able to flag for re-linking.

PR #511, opened 2021-08-27
(WIP) Add a y-plot function to archimedes

dtm2451:
The next function that @andresvicente00 will help create!

Should end up somewhat similar to dittoSeq's `dittoPlot()` where violin plots and/or barplots and/or jittered points can be used to represent data such as:
image


dtm2451:
Function is built and connected with the UI and a workflow =).

But two new input-types are needed for the visualization UI setup in order to fully power this visualization.  Perhaps work for a future PR as this one is definitely at MVP-level (minimal viable product) currently, even is the `plots` input is hard-set.  They are:
1. defaultable version of `dropdownInput`: The `color_by` input would ideally be able to default to whatever the user gives to the `x_by` dropdown.  Internally, that behavior would follow if the input is either ignored or given the value "make" when `y_plotly()` is called in python. The current `dropdownInput` setup does not get a default, so need a defaultable version.
2. a multi-select option: the `plots` input takes in `["violin"]`, `["box"]`, or `["violin","box"]`/`["box", "violin]` (either of these last two would lead to the same output).  And in the future, I'd like to add "jitter" as an option as well. [`multiselect_string`](https://github.com/mountetna/monoetna/blob/master/vulcan/lib/client/jsx/components/workflow/user_interactions/inputs/multiselect_string.tsx) is a good candidate.

dtm2451:
On campus wifi issues, leading to an inability to stay connected to the cluster for more than 30 seconds, created some time for me to focus on these two extra bits.  Not perfect, but workable :man_shrugging:.

Would love a review via the the vulcan workflow!  To make it show up, you need to make an example workflow and give your dev profile access:
```
make up
make -C janus bash
○ → ./bin/janus add_project 'example' 'example'
○ → ./bin/janus permit developer@ucsf.edu example administrator
```
Then: https://vulcan.development.local/workflow/example/y-plot

coleshaw:
Looks nice, and works for me! Can I ask what the 0 and 1 mean in the x / y-axis selections? Those didn't seem to work for me, but the other options did generate some really cool violin plots!

dtm2451:
Thanks Cole!  The first 3 columns are actually straight from a umap workflow output.  O and 1 are umap embeddings. They're numeric and so they won't work when what's needed is discrete groupings.  You'd also get an error for using leiden, sample, or group for y_by/"Y-Axis Data" because that data needs to be continuous.

I suppose I should make proper error messages for all the Viz functions when data selections don't work.

PR #517, opened 2021-09-02
Polyphemus ipi rna seq create records refactor

coleshaw:
This PR refactors the IPI "create rna records" ETL to also use a simple updated_at find in Metis, instead of the slower / more onerous find previously used. The filtering is thus done in the ETL process itself instead of in Metis, so should cause less server load.


corps (approved):
Nice!!! This looks really simple and comes together nicely.  We can leverage this for even more automating pieces.

PR #526, opened 2021-09-07
Timur query slice intermediate column models

coleshaw:
This PR adds the ability to slice on intermediate models when selecting columns. I think this makes all of the example queries that folks have come up with, possible (still working on a document).

For example, if your root model is Patient, and you want to have a column for ScRnaSeq in your data frame, this may go through a Sample model:

Patient -> Sample -> ScRnaSeq

In order to have a "flat" data frame, you'll most likely want to "slice" out on a single sample (Patient -> Sample is one-to-many, as is Sample -> ScRnaSeq). For example, for each Patient, I want the Tumor Sample, and the First ScRnaSeq record. This would be something like:

```
["sample", ["tissue_type", "::equals", "tumor"], "::first", "sc_rna_seq", "::first", "tube_name"]
```

This PR adjusts the UI so that you can add that `["tissue_type", "::equals", "tumor"]` slice when selecting columns. Previously you could only add slices on the join model itself, which in the above example would be `sc_rna_seq`.


corps (approved):
Nice dude!  I'd say, the quality of this and our React code is going up, I can see your usage of hooks is working out really well.  Overall, I have no concerns, this works pretty well.

coleshaw:
Thanks, I need more practice for sure -- and would still like to get better at architecting this stuff...

graft:
This looks awesome, I can make any query I can think of so far.

However, If I setup as such: 
Screenshot_20210907_213335
when I add a second column from the same table, I get a duplicate dialog, which edits the two rows in tandem, seems like a bug.

coleshaw:
Ah, thanks for the bug report. I'll fix that.

graft:
I don't think you can solve that way as it is necessary to have multiple ways to slice the same column (e.g. to get treg live count and nktb live count).

coleshaw:
Ah, you mean that you actually have to be able to add in the join model multiple times, with independent behavior for each instance (and independent attributes selected, etc.)? Got it...

coleshaw:
Hm, that does seem tricky from a UX perspective. How do you uniquely identify which "column" select row you're editing at any given time, or how to line up the results in the data frame? Seems like a bigger refactor is required...let me think about it, perhaps I'll leave that for a subsequent PR if it seems like a significant chunk of work.

coleshaw:
I've also changed the UI and state around a bit to track individual columns, so that you can duplicate the model / attribute but with a different slice. This lets you formulate queries like:

```
["patient","::all",[["ipi_number",["sample",["tissue_type","::equals","Primary"],"::first","flow",["stain","::equals","treg"],"::first","population",["name","::equals","NK cells"],"::first","count"],["sample",["tissue_type","::equals","Primary"],"::first","flow",["stain","::equals","treg"],"::first","population",["name","::equals","CD45-"],"::first","count"]]]
```

To grab two different population count values as distinct columns, each one for a different population name.

coleshaw:
Some follow-on PRs for this work:

* Add an input box so you can edit the data frame's column names, with a default of `model.attribute`. This lets users avoid duplicate column headers, which would throw exceptions in R or Python.
* Hide the column slices / provide a "readable" summary of what the current slices are?

I'll probably merge in after #528 , since there will be some work required to resolve conflicts, because this PR moves stuff around.

corps:
This unfortunately produces conflicts but mostly due to test queries changing up with my last merge.  I'd pull master manually, and take your changes over mine, then adjust the test / update the snapshots.  Takes a bit.

coleshaw:
Cool thanks! I will do it manually as you suggest. I figured there would be a lot since I refactored a bit…

PR #527, opened 2021-09-07
Etna apps will run as non root, but make /tmp related folders writeab…

corps:
…le permissively.

Similar pattern should be used for any state folders.  Vulcan and metis state folders (/app/data) are handled specially.

### DIFF
--- a/docker/etna-base/Makefile
+++ b/docker/etna-base/Makefile
@@ -2,5 +2,6 @@ baseTag:=$(shell basename "$$(pwd)")
 fullTag:=$(IMAGES_PREFIX)$(baseTag)$(IMAGES_POSTFIX)
 
 image:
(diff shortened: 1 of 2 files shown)


docker/etna-base/Makefile
```diff
@@ -2,5 +2,6 @@ baseTag:=$(shell basename "$$(pwd)")
 fullTag:=$(IMAGES_PREFIX)$(baseTag)$(IMAGES_POSTFIX)
 
 image:
+	make -C ../etna-base-dev
```
corps, line 5: This also fixes an issue where etna-base could be invoked to be built before etna-base-dev, despite having it as a dependency.  I can't specify it as an item level dependency because it is a separate makefile al together, so this command sends a make invocation to it's file.

PR #528, opened 2021-09-09
Filtering handles all links correctly.

corps:
Address https://www.notion.so/ucsfdatascience/6bf73d7edfad4bd38a8527049c9f1510?v=b689c4e7890d4d2dbe3d8c6acb51a6ca&p=6d87fdcbba89412782c8b5d7d3897b5b

It is possible to query across many to one or one to one relationships using the `::any` clause.  The existing code would always insert such filter clauses at the end of the query, which would later get scrapped.   This approach simply validates that the anyModels lookup occurs for entries that exist (which themselves already are filtered by one to many relationships) before deciding which predicate fold is used.


timur/lib/client/jsx/utils/query_filter_path_builder.ts
```diff
@@ -29,34 +24,27 @@ export default class QueryFilterPathBuilder {
     let nestedFilterIndex: number = 0;
 
     this.path.forEach((modelName: string) => {
-      if (stepIsOneToMany(this.models, previousModelName, modelName)) {
```
corps, line 32: I suggest viewing this in github with `?w=1` so the whitespace diff of tabbing is dropped.
Essentially I drop the stepIsOneToMany conditional, then I pick the folding predicate clause based on entry in anyMap (essentially, non one to many relationships only have one valid predicate here).

The previous approach just appends the model (out of order) at the end, and not following the injection logic.

timur/test/javascript/utils/query_builder.spec.ts
```diff
@@ -3,40 +3,17 @@ import {QueryGraph} from '../../../lib/client/jsx/utils/query_graph';
...
-    revisions: {},
-    views: {},
-    template: require('../fixtures/template_victim.json')
-  },
-  wound: {
-    documents: {},
-    revisions: {},
-    views: {},
-    template: require('../fixtures/template_wound.json')
+    documents: {}, revisions: {}, views: {}, template: require('../fixtures/template_monster.json')
```
corps, line 6: sorry, autoformatter getting aggressive :P  maybe we could write a helper that generates these from the array of required files, sense they are all pretty much empty.  eh, anyways, works for now.

timur/test/javascript/utils/query_builder.spec.ts
```diff
@@ -51,86 +28,92 @@ describe('QueryBuilder', () => {
...
   function stamp(model_name: string, attribute_name: string) {
     return {
-      model_name,
-      attribute_name,
-      display_label: `${model_name}.${attribute_name}`
+      model_name, attribute_name, display_label: `${model_name}.${attribute_name}`
     };
   }
 
+  describe('for xcrs1 models', () => {
```
corps, line 35: I decided to use the actual xcrs1 model metadata here to write the test, so that it maps more cleanly to the original issue.  This pattern is called regression testing and it helps capture common patterns and prevent an existing bug fix from regressing.

timur/test/javascript/utils/query_graph.spec.ts
```diff
@@ -57,4 +58,19 @@ describe('QueryGraph', () => {
       ['labor', 'prize']
     ]);
   });
+
+  describe('for xcrs1 models', () => {
+    const models = require('../fixtures/xcrs1_magma_metadata.json').models;
```
corps, line 63: This behavior always worked, but I wrote the test as a first step to diagnosing the issue.  I left the test because, well, it's still a useful regression test and it runs fast, so.

coleshaw:
Nice, thanks!

coleshaw:
Did you mean to merge in this PR, or just close it? Still making changes?

coleshaw:
Just a note for the tests that failed because of a different query output array, I think the new format is also okay because the nested `::any` queries will return an equivalent result to not using `::any`, since they are one-to-one relationships.

corps:
@coleshaw updated with tests adjustments.
I'll verify in staging the query changes don't effect performance too much.  If it does, I'll investigate solutions.

PR #545, opened 2021-09-17
fix issue with bypassing validation

coleshaw:
This PR fixes the issue with magmaR bypassing validations that we noticed today at standup.

Basically, existing validations only work from the parent-direction down to the child, because they only check the key-value pairs in the update payload document.

But, that means an update to a "child" record itself (disconnected or connected, doesn't matter), would not have its identifier validated if the attribute were not duplicated in the payload, like:

```
"revisions": {
    "sample": {
        "IPIADR001.T1.12": {
            "patient": "IPIADR001"
        }
    }
}
```
Actually creates a record.

This PR adds an identifier check on the record_name, if that identifier attribute is left out of the update document.

### DIFF
--- a/magma/spec/validation_spec.rb
+++ b/magma/spec/validation_spec.rb
@@ -12,32 +12,18 @@ describe Magma::Validation do
     @validation_stubs ||= {}
   end
 
-  def stub_validation(model, att_name, new_validation)
-    validation_stubs[model] ||= {}
-    validation_stubs[model][att_name] = model.attributes[att_name].validation
-    model.attributes[att_name].validation = new_validation
-  end
-
-  def remove_validation_stubs
-    validation_stubs.each do |model,atts|
-      atts.each do |att_name, old_validation|
-        model.attributes[att_name].validation = old_validation
-      end
-    end
-  end
-
   context 'model record validations' do
     after(:each) do
       remove_validation_stubs
     end
     
-    it 'validates its own record name' do
+    it 'validates its own record name for update' do
       stub_validation(Labors::Monster, :name, {
         type: "Regexp", value: /^[A-Z][a-z]+ [A-Z][a-z]+$/
       })
 
       # fails
-      errors = validate(Labors::Monster, 'nemean lion', labor: 'Nemean Lion')
+      errors = validate(Labors::Monster, 'Nemean Lion', name: 'nemean lion', labor: 'Nemean Lion')
       expect(errors).to eq(["On name, 'nemean lion' is improperly formatted."])
 
       # passes
--- a/magma/spec/validation_spec.rb
+++ b/magma/spec/validation_spec.rb
@@ -43,6 +43,9 @@ describe Magma::Validation do
       # passes
       errors = validate(Labors::Monster, 'Nemean Lion', name: 'Nemean Lion', labor: 'Nemean Lion')
       expect(errors).to be_empty
+
+      errors = validate(Labors::Monster, 'nemean lion',  name: 'Nemean Lion', labor: 'Nemean Lion')
+      expect(errors).to be_empty
     end
   end
 
--- a/magma/spec/validation_spec.rb
+++ b/magma/spec/validation_spec.rb
@@ -26,6 +26,26 @@ describe Magma::Validation do
     end
   end
 
+  context 'model record validations' do
+    after(:each) do
+      remove_validation_stubs
+    end
+    
+    it 'validates its own record name' do
+      stub_validation(Labors::Monster, :name, {
+        type: "Regexp", value: /^[A-Z][a-z]+ [A-Z][a-z]+$/
+      })
+
+      # fails
+      errors = validate(Labors::Monster, 'nemean lion', labor: 'Nemean Lion')
+      expect(errors).to eq(["On name, 'nemean lion' is improperly formatted."])
+
+      # passes
+      errors = validate(Labors::Monster, 'Nemean Lion', name: 'Nemean Lion', labor: 'Nemean Lion')
+      expect(errors).to be_empty
+    end
+  end
+
   context 'attribute validations' do
     after(:each) do
       remove_validation_stubs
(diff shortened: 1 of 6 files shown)


magma/spec/validation_spec.rb
```diff
@@ -26,6 +26,26 @@ def remove_validation_stubs
...
+        type: "Regexp", value: /^[A-Z][a-z]+ [A-Z][a-z]+$/
+      })
+
+      # fails
+      errors = validate(Labors::Monster, 'nemean lion', labor: 'Nemean Lion')
+      expect(errors).to eq(["On name, 'nemean lion' is improperly formatted."])
+
+      # passes
+      errors = validate(Labors::Monster, 'Nemean Lion', name: 'Nemean Lion', labor: 'Nemean Lion')
+      expect(errors).to be_empty
```
graft, line 45: `errors = validate(Labors::Monster, 'nemean lion', name: 'Nemean Lion', labor: 'Nemean Lion')` should also pass so we have a way to repair broken validations for existing records.

graft:
It occurs to me that this change will break any edits to records with failing identifier validations, until the failing validation is resolved. This is possibly a good? in that it encourages the data to be more correct (i.e., in line with the current validation), but also bad, because it creates a barrier to improvement of other aspects of the record (i.e., filling in non-identifier attributes).

We might debate which way to fix this; if we wanted to continue to allow updates to a record with an invalid identifier, we might restrict this validation to record creation somehow (i.e., it will prevent a new record from being created but won't prevent updates to an existing record).

dtm2451:
> It occurs to me that this change will break any edits to records with failing identifier validations, until the failing validation is resolved. This is possibly a good? in that it encourages the data to be more correct (i.e., in line with the current validation), but also bad, because it creates a barrier to improvement of other aspects of the record (i.e., filling in non-identifier attributes).
> 
> We might debate which way to fix this; if we wanted to continue to allow updates to a record with an invalid identifier, we might restrict this validation to record creation somehow (i.e., it will prevent a new record from being created but won't prevent updates to an existing record).

Oooh this is a good point!  This change as-is might make it impossible for a normal user to fix the non-valid records.

A viable suggestion Saurabh, but I worry that the 'non-new' condition would be met for any 2nd upload, including unsupervised ETL runs.  Perhaps another option for allowing users to fix bad records would be to add an input, which defaults to on, that toggles whether validations are checked.  Drawback: this method would need to explicitly be hooked up to the Timur UI somehow in order to allow users to edit such records, rather than automatically letting them move forward :shrug:

coleshaw:
Why would you keep editing an invalid record instead of correcting it and then updating the now correct identifier?

graft:
If we introduce a new identifier validation which leaves existing
identifiers in an invalid state, we then have to at some point resolve
these bad identifiers. This might require some work and reasoning (i.e., it
might not be easily reparable). For example, in IPI we have decided we want
to name lymph node samples as .L\d+, but it might take some work to figure
out which of a set of T1, N1, T2, etc., become L1, L2, etc., as we went
through today in data integrity.  Before this confusion can be resolved,
however, we might have metadata to attach to each of these samples (e.g.,
the "T2" headshot), which we need to associate with the current record
before resolving its identifier. (maybe a confusing example, since the new
validation in this case is a superset of the old one, but hopefully the
point is clear).

graft:
> but I worry that the 'non-new' condition would be met for any 2nd upload, including unsupervised ETL runs.

I don't think this would be an issue; the first upload would attempt to create a new record and would fail the validation. The second upload would encounter the same situation (attempting to create a new, invalid, record) and would fail the same way.

coleshaw:
> If we introduce a new identifier validation which leaves existing
identifiers in an invalid state, we then have to at some point resolve
these bad identifiers. This might require some work and reasoning (i.e., it
might not be easily reparable). For example, in IPI we have decided we want
to name lymph node samples as .L\d+, but it might take some work to figure
out which of a set of T1, N1, T2, etc., become L1, L2, etc., as we went
through today in data integrity.  Before this confusion can be resolved,
however, we might have metadata to attach to each of these samples (e.g.,
the "T2" headshot), which we need to associate with the current record
before resolving its identifier. (maybe a confusing example, since the new
validation in this case is a superset of the old one, but hopefully the
point is clear).

I agree, I don't actually think that's a great example because it doesn't pinpoint the issue you're trying to raise. As you said, the new validation could co-exist with the samples created under the old validation in that case, so there isn't really any blocker to someone's workflow. Let's not mix data integrity with validations, since I think those are two separate issues.

The only way I see what you're saying actually happening (wanting to continue updating a record with an invalid record_name) is if a validation gets stricter -- has that ever happened, or do we imagine that happening? Seems like there would be a lot of other implications for existing data if we made record_name validations stricter (across all sample touch points, not just the data library), so not something that a project would do lightly, if ever, I feel. Seems like continuing to allow an invalid record name to exist goes against the whole idea of having validations, and why we've invested so much time into getting naming right?

coleshaw:
Also, I've added some specs to show this, but a user can edit an invalid record_name to fix it. But the current implementation won't allow a user to continue updating an invalid record without fixing the record_name first.

coleshaw:
>this method would need to explicitly be hooked up to the Timur UI somehow in order to allow users to edit such records, rather than automatically letting them move forward 

Unfortunately there actually is no way in Timur right now to edit record names ... so that would have to get added in, first. Though it is something we've talked about as a nice-to-have.

dtm2451:
> > this method would need to explicitly be hooked up to the Timur UI somehow in order to allow users to edit such records, rather than automatically letting them move forward
> 
> Unfortunately there actually is no way in Timur right now to edit record names ... so that would have to get added in, first. Though it is something we've talked about as a nice-to-have.

Unrelated to this PR, surely, but that's something we may want for HuMu soon.  Many of the biospecimen records were created before modeling/naming were finalized and there's a bunch with the CyTOF assay `-CY1` tacked on.  A good test case as there are also properly named records, created from the scRNAseq side, so some matching (merge these records?) / some not!

coleshaw:
Okay, this should only check the record name on creation, which  means that you can update records with invalid record_names, but when you try to change it to a new record name, validation will run against the new record name. So we can keep updating "invalid" records, as well as fix invalid record names. I think this meets both goals.

Issue #541, opened 2021-09-17
".1"/".2" appended to identifier names in some magmaR updates


dtm2451:
Just a placeholder for an issue mentioned in standup for now.  Will add a better description...

Issue #431, opened 2021-06-23
creating new orphan records via /update returns "successful" but does (?)nothing


dtm2451:
At least I think it does nothing.

Bushra and I were trying to link some sc_rna_seq data inn magma via magmaR earlier.  I only noticed at the end that the target records and parent records for those samples did not exist.  (So Bushra is going to work on the upstream sample/patient info before we try again.)

But this issue is about how the GET status returned was "successful" even though I don't believe anything was actually done?  Either nothing was done, or we created orphan records, but either way, perhaps a different status code should have been returned?

This was one of the three updates we'd tried:
```
{
   "project_name":"ipi",
   "revisions":{
      "sc_rna_seq":{
         "IPICRC136.N1.scrna.live":{
            "raw_counts_h5":{
               "original_filename":"raw_feature_bc_matrix.h5",
               "path":"metis://ipi/data/single_cell_GEX/processed/5prime_GEX/CRC/IPICRC136.N1.scrna.live/raw_feature_bc_matrix.h5"
            }
         }
      }
   }
}
```

graft:
Your record is there, you can see it on Timur if you click the 'show disconnected records only' filter option. This is the intended behavior; the data inserts correctly, so you get a 200. The data can also be queried out successfully, with the correct query params. However, since it is attached to disconnected data, it won't show up in your typical query. If the data is attached to the graph again, then it should appear in a regular search once more.

dtm2451:
roger roger!  Sorry that you probably told me this before!

I'm going to leave this open for now as a reminder that I should update `magmaR::updateValues()`'s record summarization to match.

dtm2451:
Issue "solved" by adjusting the update-summaries to call records as "NEW (or orphan)" in #549.

PR #549, opened 2021-09-21
Add 'updateFromDF()' to magmaR

dtm2451:
Adds a new convenience function to magmaR that essentially mimics `etna administrate models attributes update_from_csv`, just in R.

I put it together after discovering a wonderfully relevant chunk of code from a past `magmaR::update()` upload.

Not the highest priority to finish, unless a user requests such functionality.

To Do:
- [x] Add to the Upload vignette
- [x] Unit tests


etna/packages/magmaR/R/update.R
```diff
@@ -106,7 +107,7 @@ updateValues <- function(
         num_new <- length(rec_names_new)
         
         ### Summarize for NEW records
-        cat("For model \"", modelName, "\", this update() will create ", num_new, " NEW records:\n    ",
+        cat("For model \"", modelName, "\", this update() will create (or update) ", num_new, " NEW (or orphan) records:\n    ",
```
dtm2451, line 110: Addresses #431 by adjusting the 'NEW' records section of the update summary to report that they could also be orphan records.

coleshaw:
Do you think this addresses the mystery issue with the extra `.1`, `.2`, etc. records being created?

Perhaps @graft is the most qualified, since he's worked with R more?

graft:
> Do you think this addresses the mystery issue with the extra `.1`, `.2`, etc. records being created?
> 
> Perhaps @graft is the most qualified, since he's worked with R more?

I don't think this addresses those issues? Or at least I don't see how. But @dtm2451 closed #541 so maybe it does?

dtm2451:
I think those were created from an improper input where that was the actual request.  That + plus Bushra and I must have ignored the Summary of the update that was about to be run, as there isn't much manipulation after the stage where records to be updated are summarized.  Certainly no name-setting where ".1"/".2" uniquefication would happen.

I've not found anywhere in the `updateValues()` function where something like that would have happened, so I'm relatively confident it was upstream of magmaR.  Sorry, as I could have commented it, but that's why I closed #541.

graft (approved):

dtm2451:
:+1: Beginning the process of re-submitting to CRAN.  I'll wait to merge until the first round of checks are done.

dtm2451:
Submitted to CRAN

PR #552, opened 2021-09-24
Production deploy

corps:


dtm2451 (approved):
For vulcan:
- pushes ability to use table data in project-definitions, but I didn't actually add any of these. So no user-visible changes.

PR #553, opened 2021-09-24
Timur query params in url

coleshaw:
Sorry, this is a really messy PR....it includes the following changes:

* Breaks apart the single query context into three separate contexts, in an attempt to reduce unnecessary re-renders (mild impact).
* Refactors Magma reducer to bulk update model templates and documents instead of per-model, in an attempt to reduce unnecessary re-renders caused by the fetching of IDs for the identifier search box (high impact).
* Adds in URL query params that sync with the query builder state. If you refresh the page, the query still exists. As you edit your query, the URL updates as well.
* Adds in two new buttons. One for resetting the form to blank, another to copy the URL so you can share with other people.
* General refactor of components to reduce usage of contexts in favor of shallow prop-passing.
* Remove some Grid components where unnecessary.
* Updates to webpack 4 (I needed this to profile the app in the browser).


corps (approved):
In general, I approve of the approach.

timur/lib/client/jsx/contexts/query/use_uri_query_params.tsx
```diff
@@ -0,0 +1,73 @@
...
+    Object.entries(columnState).forEach(([key, value]: [string, any]) => {
+      searchParams.set(key, JSON.stringify(value));
+    });
+
+    if (search !== searchParams.toString()) {
+      history.pushState({}, '', `${pathname}?${searchParams.toString()}`);
+    }
+  }, [rootModel, search, whereState, columnState, pathname]);
+
+  // Set current state to reflect query params only on component load
```
corps, line 50: Nice.  I think this works pretty well.

timur/lib/client/jsx/contexts/query/query_where_context.tsx
```diff
@@ -0,0 +1,102 @@
...
+  orRecordFilterIndices: [] as number[]
+};
+
+const defaultQueryWhereState = {
+  ...defaultQueryWhereParams
+};
+
+export type QueryWhereState = Readonly<typeof defaultQueryWhereState>;
+
+export const defaultQueryWhereContext = {
```
corps, line 16: Yeah, breaking this out makes some nice sense.  It is a bit more verbose to do so, but one thing I like about this approach is how easily you can go from "one big umbrella context" to get started, and then decompose into smaller systems as you build out the system and find the best places to get granular.  Context end up being fairly decomposible.

The default context bit is nice because it also gives obvious stubs for writing tests on highly integrated parts.

PR #556, opened 2021-09-28
work around error raised by autocomplete about uncontrolled / controlled

### LINKED ISSUE #554: Query: column attribute appears blank though has selected value in builder
bug with deleting a column "blanks" out the attribute selector for other columns?

coleshaw:
This PR should close #554. Turns out the Autocomplete component was throwing an error when you deleted a column, because for a brief moment in time, the selected `value` would be invalid, since the column model may have changed. Related to this [GH issue](https://github.com/mui-org/material-ui/issues/18173).

So this PR just sets to `null` when no valid match is found, which allows Autocomplete to behave properly when the attribute choice set lines up with the right column definition.


corps:
Makes sense, thanks for hunting this down.  `undefined` vs `null` strikes again.  This is why I am fan of `Optional` / `Maybe` type wrappings.

corps (approved):

PR #559, opened 2021-09-29
Zc/linker etl refactor

corps:
GOAL:
To condense the many separate linking and attribute processing metis ipi etls into a single consistent configuration that can scale to other projects easily.  

NON GOALS (for this PR):
I'm trying to cut scope down and release incrementally, so in this PR I am just focused on the refactoring for the existing ipi work.  My single cell linking work will come after this leaves and is proven safe in production, since it builds on the same concept.

I'm also focused on first collapsing code where I can.  The question of how to extract all of this configuration into fully generalized parts is open question, but actually I can already see some paths emerging (I think a lot of regexes, buckets, and projects configured, plus possible project specific engineering subclasses for legacy linking).  We'll have to experiment more on this, but overall it atleast makes rolling this out to other projects easy from the engineering side.

There's also a separate chunk of work related to tracking metrics RE how much of each record is actually linked and accounted for by attribute.  I decided to keep that out of this PR and fold it in after as well.

Still working on moving over the tests and adjusting them.  Fortunately, most of the tests actually do work as is, but just need movement into the aggregate test file and some slight setup / teardown adjustment.

Have confirmed locally and remotely this approach is sound, but still working out the tests and kinks.


polyphemus/lib/etls/ipi/ipi_etls.rb
```diff
@@ -0,0 +1,61 @@
...
+require_relative './file_processors/ipi_rna_seq_attribute_processor'
+require_relative './file_processors/ipi_rna_seq_matrix_processor'
+require_relative './linkers/ipi_rna_seq_processed_files_linker'
+require_relative './linkers/ipi_rna_seq_raw_fastq_files_linker'
+require_relative './record_creators/ipi_rna_seq_and_plate_record_creator'
+require_relative '../linker_processor'
+
+class Polyphemus
+  module Ipi
+    class WatchFoldersConfig < Polyphemus::ProjectWatchFoldersConfig
```
corps, line 13: So here's a good place to start to understand this PR.  My goal was to have one file be the source for all metis oriented processing for the project.  The syntax and configuration is simple and consolidated. There are actually 3 separate etls that will run per project; one for watch folders, one for individual metis file changes, and the other propagating updated_at.

Each of these etls is idempotent and safe to reset.  In general, to reprocess linking, one only needs to reset the watch folders etl since it will process every matching folder's contents in bulk.

polyphemus/lib/etls/linker_processor.rb
```diff
@@ -0,0 +1,20 @@
...
+    @linker = linker
+    @model_name = model_name
+  end
+
+  def process(cursor, files)
+    if cursor[:project_name] != linker.project_name || cursor[:bucket_name] != linker.bucket_name
+      raise "Invalid linker configuration: input cursor project and bucket do not match linker specification."
+    end
+
+    puts "linking #{files}"
```
corps, line 14: whoops, need to clean up spare puts

etna/lib/etna/clients/metis/models.rb
```diff
@@ -281,6 +281,13 @@ def initialize(raw = {})
           @raw = raw
         end
 
+        def with_containing_folder(folder)
```
corps, line 284: This simply allows translating from the "hide_paths: true" to full path folder objects.
It is amusing, but full path resolution of a file is an expensive operation just due to the potential recursive aspects that make single scan lookups hard (won't know what precise order you will find parent folders in before you look them up).  

Incidentally, the attribute processors will no use watch folders and not use hide_paths: true, and I believe this further increases performance.

polyphemus/lib/etls/add_watch_folder_base_etl.rb
```diff
@@ -1,48 +1,86 @@
...
+        end
+      end
+    end
+  end
+
   def process(cursor, folders)
     # We'll need to filter out the folders based on folder_name here, using the
     #   supplied regexes
-    logger.info("Found #{folders.length} updated folders: #{folders.map { |f| f.folder_path }.join(",")}")
+    logger.info("Found #{folders.length} updated matching folders: #{folders.map { |f| f.folder_path }.join(",")}")
```
corps, line 49: oh you know what, I can revert this.  I had previously tried making filtering a step up from process function but changed gears, so this is an arifact.

coleshaw:
Looks neat, the config abstraction looks pretty useful.

corps:
This is at last done and ready for review, does not include single cell stuff yet.  jsut the refactor for the etl configuration that exists.

coleshaw (approved):

PR #565, opened 2021-09-30
update query operators for strings, add ::notin verb to query api

coleshaw:
This PR adds some of the "numeric" comparisons for string operators to the query UI, to handle numberish values stored in string attributes, i.e. like some clinical data.

Adds `::notin` for numbers, fixes how number values get converted to floats in the builder.

Minor update to column display labels, so if the user types something in before selecting a model, the label won't get wiped out.


graft:
Hmm, no harm in the alias, but why not just use `::not`?

coleshaw:
Yeah, good question! I couldn’t figure out how to differentiate in the UI when to send an array back (split a comma-separated string like for ::in and ::slice), and when to send a single string back, if using ::not for both cases. Limited by the fact that I chose to do detection of splitting of comma-separated string operands based on the Magma predicate itself (::in) instead of the user readable label (In). 

Smaller reason is I think it also makes the API a bit more explicit. Like “notin” is opposite of “in” and both take arrays. “Equals” is opposite of “not” and both take strings”.

PR #568, opened 2021-10-01
Etl view fixes, part 2

graft:
More bug fixes:

1) Apparently it does not suffice to require `active_support`, you must require `active_support/all` to get core extensions - although I'm not quite sure why this doesn't seem to be an issue in development, where I need not require active_support at all and the command runs fine. Somehow I feel like this will all end up being a bad idea just to get `#symbolize_keys`.
2) A redcap loader bug.


coleshaw:
> Apparently it does not suffice to require active_support, you must require active_support/all to get core extensions - although I'm not quite sure why this doesn't seem to be an issue in development, where I need not require active_support at all and the command runs fine. Somehow I feel like this will all end up being a bad idea just to get #symbolize_keys.

Oh darn, I wonder if this is because the Magma version is pinned to 4 something, perhaps they changed the internal imports? Good to know for when we upgrade the other version, I guess. It is quite a lot just to get `symbolize_keys`, or we could go through and swap that out...

coleshaw (approved):

PR #570, opened 2021-10-01
Timur query expand matrices without slice

coleshaw:
A couple of small UI changes around matrices, based on feedback from Dan and Saurabh:

* Matrix attribute columns without slices will expand with all the possible gene names (to make this line up better with the "Expand matrices" option). Previous behavior was that they would not expand without a slice, ignoring the "Expand matrices" toggle, which was counterintuitive.
* As a result, and the challenge of rendering tens of thousands of columns, the UI data frame will only render the first 10 columns of any data frame. A message will appear in this case to advise the user that they should download the TSV for the full results. We could adjust this if it seems necessary, but on my monitor I can only see 5-ish columns at a time, anyways.


timur/lib/client/jsx/components/query/query_table.tsx
```diff
@@ -14,6 +16,10 @@ import {QueryTableColumn} from '../../contexts/query/query_types';
 const useStyles = makeStyles({
   table: {
     minWidth: 650
+  },
+  columnWarning: {
+    color: 'darkred',
```
dtm2451, line 21: a lighter shade of red would stand out more to me.  I understand that red is super bright to color-typical individuals, but dark red can blend in with black for me!

dtm2451:
Looks GREAT!  But the one nit I have is that I didn't actually notice the warning about columns being absent until I started through the code and noticed that I should look for one.  Maybe if it was a lighter red, or boxed, or if it popped up on the right side closer to the `query` button that I had just hit in order to render the data (as my eyes were already over there)?

coleshaw:
Thanks for the feedback! I will take a look at moving it on the page -- have updated the color and added more words and *** to try and make it stand out more.

dtm2451:
Yup much more noticeable now!  You could leave it where it's at... I was just tossing out ideas.  Up to you.

dtm2451 (approved):

PR #571, opened 2021-10-02
show vulcan + polyphemus in hamburger

graft:
What it says, adds polyphemus + vulcan to the hamburger.


coleshaw (approved):
Looks good -- do you also need to update the Chef recipes? Those hosts might already be included in the default config, but didn't check...

PR #572, opened 2021-10-03
Use the root view for Vulcan

graft:
A small one which adds the default root view to Vulcan and splits the workflows by project, similar to the way other etna apps are setup. This has the disadvantage of disappearing a "see all the workflows even if you can't run them" view, which seems ok to me, but I solicit your opinions on the tradeoff.

### DIFF
--- a/vulcan/lib/client/jsx/components/dashboard.jsx
+++ b/vulcan/lib/client/jsx/components/dashboard.jsx
@@ -7,88 +7,53 @@ import {pushLocation} from 'etna-js/actions/location_actions';
 import {VulcanContext} from '../contexts/vulcan_context';
 import Card from '../components/dashboard/card';
 import {workflowName} from "../selectors/workflow_selectors";
-import ReactModal from "react-modal";
 import SelectInput from 'etna-js/components/inputs/select_input'
 
-const modalStyles = {
-  content: {
-    top: '50%',
-    left: '50%',
-    right: 'auto',
-    bottom: 'auto',
-    marginRight: '-50%',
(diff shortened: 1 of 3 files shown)


dtm2451 (commented):
On a basic level, I think it can make sense to break things down this way.  And it works for me in testing.

I wonder if we do want this however due to the future plan of having cross-project analysis, as I'm not sure how that fits into this structure.

I'm not sure.  Maybe this really is the way to go.

As a potential alternative method, though does nothing to fix my cross-project application concern, we could give an optional project selection piece on the same page where all workflows are shown?  And if the user clicks this, then they only see workflows available to the selected project.

corps:
I'm ok with this.  Even cross project analysis will be done, I think, in the context of a project.  Another words, you'll likely start by picking the project, and then further selecting the sources from other, other projects.  No context will be without a project, even cross project analysis.

corps (approved):

graft:
Yeah I agree with Zach, a "cross-project" analysis is just another kind of project with a defined data set, the only difference is the data source is other library projects. Also I think we are best off developing this concept later, once we have within-project analysis worked out.

>As a potential alternative method, though does nothing to fix my cross-project application concern, we could give an optional project selection piece on the same page where all workflows are shown? And if the user clicks this, then they only see workflows available to the selected project.

We could make the old dashboard available at `/workflows` this way, but I'm not sure I see the utility of this all-projects view, really, other than fostering a grass-is-greener, i-wish-i-had-that kind of sentiment amongst users.

PR #575, opened 2021-10-05
Polyphemus batch cat download notifications

coleshaw:
This PR pulls out the ingest Slack notifications to separate ETLs, so that we can batch the ingested files together into a digest and run at separate times (i.e. once a day).


etna/lib/etna/clients/metis/workflows/metis_upload_workflow.rb
```diff
@@ -110,8 +110,6 @@ class Upload
 
           def initialize(source_file: nil, next_blob_size: nil, current_byte_position: nil)
             self.source_file = source_file
-            self.next_blob_size = next_blob_size
```
corps, line 113: is there a reason for this deletion?

polyphemus/lib/etls/slack_notification_module.rb
```diff
@@ -0,0 +1,10 @@
+require_relative "../helpers"
+
+module Polyphemus::WithSlackNotification
+  include WithLogger
+
+  def slack(title, channel, msg)
```
corps, line 6: Interestingly, I implemented something similar, too, so I'll fold in my PR with yours after.

etna/lib/etna/clients/metis/workflows/metis_upload_workflow.rb
coleshaw: It seems redundant with the subsequent two lines of code?

corps: oh, hmmm, I see.

corps (approved):

PR #578, opened 2021-10-06
Add updated at indexes to metis files and folders

corps:
Somehow, I had the delusion that folders and files already had an index on updated_at, but I think I was confused with magma and the magma etl.

I have high confidence that this is the source for the find api's high response times, and this simple fix should bring them back down.  All etl workflows are consuming in order of updated_at, and having these columns unindexed is forcing very large sorts in memory for every single batch.


coleshaw:
Ah, good find!

corps:
I'm ok with either strategy?  The one index to be very careful about adding without waiting for use case is for string columns, but numbers and dates is cheap.

I think in this case I assumed at the time of building etls that this particular index.

PR #579, opened 2021-10-06
remove duplicate slack notification module; update all uses of the no…

coleshaw:
…tifier

This PR removes the duplicate module functionality in favor of the one in `helpers.rb`, with `Shellwords.escape()`. Also updates any usage of the slack post script to use this module.

Also proposes a fix for the ETLs that require downloading a file. Might take a performance hit, but because download URLs are signed by Metis, I think we may have to accept the tradeoff?


polyphemus/lib/etls/add_watch_folder_base_etl.rb
```diff
@@ -140,7 +140,6 @@ def folders_files(cursor, folders)
     find_request = Etna::Clients::Metis::FindRequest.new(
       project_name: cursor[:project_name],
       bucket_name: cursor[:bucket_name],
-      hide_paths: true,
```
coleshaw, line 143: I think these changes are required to allow the matrix / attribute processors to work. They have to download files, which means you can't construct the download URL by hand from a known folder_path (has to be signed by Metis).

corps: ahhh shoot.  ok, sure.  this is fine.  I totally missed this while testing yesterday because I was cofused on the linker.

PR #586, opened 2021-10-07
Mount the vulcan cache in archimedes instances

dtm2451:
Pairing with @graft, we discussed an idea for sharing data between workflows which we think is fairly simple, and which can be made safe if we set up vulcan to rely on 'viewer-level-only + non-PHI' task tokens.  We could then, safely, make the project's cache available, read-only, to the archimedes instances created for running workflow steps.  And then, sharing data across workflows could be as simple as having a `copy url` button next to the data output of workflow 1 (eg: the umap_workflow_anndata.h5ad of the UMAP workflow) which a user would then provide as a string input to workflow 2 (eg: the dge workflow).

This cache sharing idea is exemplified in the current commits, although it doesn't quite work.

### DIFF
--- a/vulcan/lib/orchestration.rb
+++ b/vulcan/lib/orchestration.rb
@@ -102,6 +102,8 @@ class Vulcan
         "-v",
         "/var/run/docker.sock:/var/run/docker.sock:ro",
         "-v",
+        "/app/data/built/#{session.project_name}:/app/built:ro",
+        "-v",
         "#{Vulcan.instance.config(:archimedes_exec_volume)}:/archimedes-exec",
         Vulcan.instance.config(:archimedes_run_image),
         "poetry",
--- a/vulcan/lib/server/workflows/scdge.cwl
+++ b/vulcan/lib/server/workflows/scdge.cwl
@@ -48,7 +48,7 @@ steps:
     run: scripts/dge_grab_obs.cwl
     label: 'Extract metadata Options'
     in:
-      scdata.h5ad: mockSetup/scdata.h5ad
+      data_url: 1_Data_Selection__vulcan_cache_url
     out: [metadata]
     
   setDEMethods:
--- a/vulcan/lib/server/workflows/scripts/dge_grab_obs.py
+++ b/vulcan/lib/server/workflows/scripts/dge_grab_obs.py
@@ -1,7 +1,11 @@
 from archimedes.functions.dataflow import input_path, output_path
 from archimedes.functions.scanpy import scanpy as sc
+from archimedes.functions.environment import project_name
 
-scdata = sc.read(input_path('scdata.h5ad'))
+data_target = "/app/built" + input_path('data_url').split(project_name)[1]
+
+# scdata = sc.read(input_path('data_url'))
+scdata = sc.read(data_target)
 metadata = scdata.obs
 
 metadata.to_json(output_path("metadata"))


corps:
Questions:

Are we concerned about security?  If the entire built directory is mounted, even read only, are we ok with all scripts having access to all project data?  Right now, this isn't even project scoped.

How does archimedes know how to correctly read or manipulate this directory?  Will logic from the ruby side be converted?

Will this, translate to local running of archimedes well?  Will local versions also have and manage a local cache?

Basically, it feels like the vulcan cache logical is moving into archimedes, which is fine, I'm just curious what the full fledged plan or thought is.  This alone raises red flags, and I'd definitely caution merging this alone as is.

corps:
Ah, ok, I see the project scope.  Hmm, is there any guarantees that project_name isn't malformed to allow for remote access of the file system?  I'd just want to double check that attack vector.

dtm2451:
Beautiful.  I didn't even get a chance to message in #data_library_engineering to introduce this idea before you shared your thoughts!  We want these thoughts!  **This is only an idea, and if it's a bad idea, then we shouldn't merge.**

(Also, if we do think it's workable, I would need your help @corps because the current code for mounting the built directory doesn't work.)

Let me try to answer these questions:
- scoping: it would be project scoped, so only the current project's built directory should be made available to the runner.

> How does archimedes know how to correctly read or manipulate this directory? Will logic from the ruby side be converted?
>
> Will this, translate to local running of archimedes well? Will local versions also have and manage a local cache?

Our idea was that the cache would remain maintained entirely by vulcan. The archimedes runner would get just a read-only view of that cache.  Nothing should ever need to be sent back to this cache via archimedes.  The file would be read in via this path, but any transfer out of the runner would still be via an `output_path('input_name')` or related route.  I don't think there's ruby logic to duplicate then?

> Hmm, is there any guarantees that project_name isn't malformed to allow for remote access of the file system? I'd just want to double check that attack vector.

The path used is our own project scoping.  We read in the project_name from the environment on the `docker_run` side, so I'm not sure of attack paths there, but I do want to consider this before potentially creating a security issue.

graft:
> Actually, I'm just going to take a step back from this. I trust yall think this out, but I'm taking myself off the review here.

Quite the opposite, please bring up any critiques you can think of. I think this should be okay, since it is project-scoped and will eventually be a shared binary cache, but I'm not sure what we might have missed here.

coleshaw:
General questions ...

* I thought the cache was (or will be) temporary -- is there an assumption that the user runs the first workflow and then immediately goes onto the second workflow, so the URL they pass in is "fresh" and you know the data exists in the cache? Once the cache becomes temporary, what happens if someone saves that URL and tries to use it after the file expires?  I think we may have talked about tradeoffs with storing things in Metis versus linking to the Vulcan cache directly versus uploading a file, but I kind of forget the details...
* Why mount the entire directory for a project instead of the single file that the user passes in? Seems a bit much and perhaps helps with the security aspect. Is there a way to just bind the one file to Archimedes using the [current input hooks](https://github.com/mountetna/monoetna/pull/586/files#diff-03522728411876fd232c24f006894b029b53abe0f88d2a503fa65f8a4aec7072R122-R124)?

graft:
Great questions - lemme see if I can sum up our reasoning here. First, some observations:
- Vulcan users will mostly be viewers of a project, especially in the era when most projects are public.
- Only editors of a project can upload to Metis.
- Managing another upload mechanism and file store independent of Metis is a bunch of extra work.
- As soon as we allow any users to deposit data, we have some responsibility to ensure that the data can actually be stored, e.g. it is not PHI. Mostly we ensure this through setting the editor role for a project, having validated/anonymizing "safe" routes for accepting e.g. clinical data. All of this evaporates if anyone can upload an XLS of clinical data.
- Eventually the Vulcan cache will be a shared artifact for the project, not locked to a user or session.
- The Vulcan cache should remain free of PHI, so that everyone can peruse it safely.

Given all of these things, it is clear that vulcan jobs should run with a non-PHI, viewer task token (not currently issued, but easy to add). It's also clear that we can't allow vulcan users to download data from one workflow locally and upload it into the next, because we simply can't allow user uploads by project viewers. Since we want the Vulcan cache to be safe to share amongst project viewers, we can't allow any viewer to inject potentially harmful data; any data that is used is hopefully coming from Magma, where we can scrub it clean, or if it is in some metis-accessible project path, at least we know an editor has placed it there presumably responsibly.

However, we DO want vulcan users to be able to share data between workflows. How can we do it?

If the cache is safe, and we want vulcan users (i.e., viewers of the project) to be able to see it all, we can also use cache locations to share data between workflows. This has the additional benefit that we don't have to move data from or copy it to anywhere to make it available to other Vulcan workflows. 

The usage presented in this PR is pretty rudimentary (just passing a location string as an input and mounting the cache), but hopefully this can be refined into a useful entity. The implementation is maybe entirely wrong, and there may be some CWL metaphor for this I am missing; this method was intelligible to us, so we tried it out. 

Some particulars:
- Why mount the entire directory?
The path as I understand it would be: the user completes workflow A, and takes a reference to a generated file output from there to workflow B, where they enter it as a primary or step input. A subsequent step has to take this input string and read the file from that cache location. Currently the archimedes runner wires input paths into the docker container, but I wasn't sure how to tell it that it should wire a given input based on the location in its contents instead (i.e., dereferencing the cache file to get another cache URL to mount); that seemed unsafe as then we would have to be meticulous about ensuring we weren't mounting an unsafe path. However, given the above assumptions, mounting the entire cache for a project IS safe, so we could just do that, then read the file within the step based on the input string containing the location reference.
- The cache is (or will be) temporary
It is true that the cache is ephemeral, but: 1) the cache is populated as a result of calculation, which may hopefully be re-run if necessary, 2) we can be less zealous about deleting useful locations (e.g., count references before deleting), and 3) we can remember enough about the previous cache state to help the user manage the situation of a no-longer-there cache location. Hopefully, also, most of this usage is "hot" usage of the kind you described above, e.g., I made a dataframe and now want to take it to a plotting workflow, so I can be sure that my piping hot cache reference is still good.
- Won't people have trouble sorting opaque cache URLs?
Could be an issue, we might try a number of ways to alleviate this pain. E.g., instead of copying a reference to your clipboard, save references to a bookmark list you can write your own labels for; we could make the URLs less opaque by including a few more metadata items in it (like the name/step of the workflow that produced it).

graft:
> Just to clarify, this doesn't mean that we also offer **PHI**, viewer task tokens?

Currently the task token degrades permissions down to editor level, but if you have the "restricted" bit set (which we use for flagging PHI, e.g. with the COMET waiver stuff), this simply gets passed through (i.e., if you have `A:my_project` permission you end up with a `E:my_project` task token); instead, we want to only allow `v:my_project` task tokens, so no vulcan process can access `restricted` data.

>I'm not sure that mounting the entire project's cache absolves us from:
> > we would have to be meticulous about ensuring we weren't mounting an unsafe path

I think this would be okay, because things like /etc/passwd would not be mounted in the container; if the only things that are mounted are safe, the user can safely try to dereference any location without fear they will be reading something inappropriate. The same is not true if we give them the ability to instruct what is mounted, whereby they could get an unsafe file mounted into the sandbox.

dtm2451:
Deprioritizing and putting a pause on this idea for now.

We decided, in Friday planning today, that it does make sense to tack the differential expression steps onto the end of the UMAP workflow, and in that case, there is no immediate need for building this data sharing pathway.

PR #588, opened 2021-10-08
IPI Single Cell Processed Pool

corps:
This introduces the ability to process and link single cell pool data attributes.

It is missing a few computed attributes, and the raw fast q linkage which requires a change in magma attribute to file collection.  It also is missing non pool data, and the prometheus metrics.  But, this core bit works (tested in production) so wanted to ship this part first, to be followed up on.

### DIFF
--- a/polyphemus/lib/etls/ipi/ipi_etls.rb
+++ b/polyphemus/lib/etls/ipi/ipi_etls.rb
@@ -7,12 +7,45 @@ require_relative './linkers/ipi_rna_seq_processed_files_linker'
 require_relative './linkers/ipi_rna_seq_raw_fastq_files_linker'
 require_relative './record_creators/ipi_rna_seq_and_plate_record_creator'
 require_relative '../linker_processor'
+require_relative '../shared/single_cell/single_cell_linkers'
 
 class Polyphemus
   module Ipi
+    # Broken out to simplify tests
+    module SingleCellLinkers
+      def add_single_cell_linkers!
+        process_watch_type_with(
+          bucket('data')
+            .watcher('single_cell_pool_processed')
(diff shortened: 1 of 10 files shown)


polyphemus/lib/etls/metis_files_linker_base.rb
```diff
@@ -156,15 +158,14 @@ def organize_metis_files_by_magma_record(
     path_regex:,
     record_name_gsub_pair: nil
   )
+    logger.info("Files #{metis_files}")
```
corps, line 161: Will remove

polyphemus/lib/etls/ipi/ipi_etls.rb
```diff
@@ -7,13 +7,46 @@
...
 
 class Polyphemus
   module Ipi
+    # Broken out to simplify tests
+    module SingleCellLinkers
+      def add_single_cell_linkers!
+        process_watch_type_with(
+          bucket('data')
+            .watcher('single_cell_pool_processed')
+            .watch(/^single_cell_GEX\/.*\/[^\/]*POOL[^\/]*\/.*$/),
```
coleshaw, line 20: For other projects, there wind up being like four or more different kinds of single_cell analysis. (`single_cell_GEX`, `single_cell_BCR`, `single_cell_TCR`, and `single_cell_CITE` .. perhaps with ATAC later?). Would you make different watchers in that case (line 19) to go along with the different regexes?

graft: It might be best to make the identifier validation for the single-cell model in magma/ipi the canonical place to retrieve your regexp (and set that regexp correctly in Magma if it is not), so the two do not go out of sync.

PR #591, opened 2021-10-08
Move Etna request logging from Route to Controller

coleshaw:
In preparation of doing date-shifting in Magma itself, this PR moves the request logging from the Route definition into the Controller. This allows the Magma update controller to add in additional keys to redact from the log -- so my thought (not fully formed yet, though), is that once we define a "shifted date" type attribute (and possibly a "shifted age" type dictionary entry?), the update controller can find all of those for a given project and redact those keys, too, so we're not accidentally storing the raw, unshifted data? Hopefully that seems reasonable and makes sense.

Current logging behavior for all apps and routes should be the exact same as now, except we might see some additional logging for routes without the `#action` parameter in the route definitions. Typically things like root path requests, where the server is just rendering an ERB template.

### DIFF
--- a/etna/lib/etna/controller.rb
+++ b/etna/lib/etna/controller.rb
@@ -106,7 +106,11 @@ module Etna
         [ key, @censor.redact(key, value) ]
       end.to_h
 
-      log("User #{@user ? @user.email : :unknown} calling #{self.class.name.sub("Kernel::", "").sub("Controller", "").downcase}##{@action} with params #{redacted_params}")
+      log("User #{@user ? @user.email : :unknown} calling #{controller_name}##{@action} with params #{redacted_params}")
+    end
+
+    def controller_name
+      self.class.name.sub("Kernel::", "").sub("Controller", "").downcase
     end
 
     def success(msg, content_type='text/plain')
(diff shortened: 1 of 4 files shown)


graft:
Could you clarify how you see this working for the `/update` controller? Would it override `log_request`? This might work okay in that instance but it could be a little confusing when you have multiple actions in one controller. You could instead allow the endpoint to place a patched params hash into `request.env['etna.redacted_params']`; then when the controller action or block returns, you could `ensure; log_request; end` and have log_request look for `redacted_params` if it exists, so each action could handle its own redaction separately.

What a pain all of this log-redaction is, wish it were avoidable somehow...

coleshaw:
> Could you clarify how you see this working for the /update controller?

I was actually thinking of overriding `redact_keys`, like [here](https://github.com/mountetna/monoetna/pull/591/files#diff-b5b66a8d168b6f05f0187f0f97f038d21d9a0f344519d5eca2addb9da9bdb634R36). There's only one action per Magma controller, so I didn't think it would be an issue, but that's a good point that it would affect all routes in a controller, if we ever  have that use case. I can fiddle around with your suggestion and see how that works.

We could also just not log the parameters in requests 😝. I'm not sure how extensively folks rely on the params log for debugging, especially since the params are compacted so we lose information anyways 🤷 .

PR #596, opened 2021-10-13
Magma model date shift flag

coleshaw:
This PR adds a flag on Magma model definitions, for `date_shift_root`. The idea is that each project can have at most one model set as the date_shift_root, whose identifier the records will use to consistently shift dates. This flag can be set when you add a model or update an existing model, and the action validations verify that, for a given project, at most one model has this flag set. Generally it will be the patient or subject model.

### DIFF
--- a/polyphemus/lib/etl.rb
+++ b/polyphemus/lib/etl.rb
@@ -116,15 +116,13 @@ class Polyphemus
         logger.info("Attempting to process")
         if batch.empty? || process(cursor, batch) == :stop
           logger.info("No more work found, stopping.")
-          return false
+          next false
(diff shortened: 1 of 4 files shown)


magma/lib/magma/model.rb
```diff
@@ -209,5 +222,20 @@ def json_document(attribute_names = nil)
           model.identity => identifier
       )
     end
+
+    def date_shift_root_record
```
coleshaw, line 226: Hm...going to merge this PR in, but as I started working on the shifted_date_time attribute, I think this will morph and probably not be here.

PR #598, opened 2021-10-13
Adding validate command to metis client

corps:
As per arjun's request, when a file is not archived, _he prefers an error over just output_.  In other words, you can see the validate command as: `a command that attempts to assert a file has been archived.  if that command succeeds, it means that the file has been confirmed to be archived.  If it fails, it means we could not confirm the file be archived.`


graft:
Since this is just using the list api for a particular file, rather than making a single unique command to do this limited check I would favor supporting this usage via a --long flag to the ls command, which could list all of the information in the file listing in some parseable format.

arkal:
i tested this and the ShellError doesn't really abort the shell raising a non-zero exit code (see below). So what it's basically doing now is in fact a text output telling me if the file was archived or not. This is acceptable behavior too, and in this case i'd second @graft's suggestion of making this an addon to `ls --long`

```
arkal:tmp arjunarkalrao$ cat foo.sh 
project xvir1

validate data/single_cell_GEX/processed/XVIR1-POOL-SCG47/cellranger/possorted_genome_bam.bam
validate data/single_cell_GEX/processed/XVIR1-POOL-SCG48/cellranger/possorted_genome_bam.bam
validate data/single_cell_GEX/processed/XVIR1-POOL-SCG49/cellranger/possorted_genome_bam.bam
validate data/single_cell_GEX/processed/XVIR1-POOL-SCG50/cellranger/possorted_genome_bam.bam
arkal:tmp arjunarkalrao$ ~/src/metis/bin/metis_client < foo.sh 
> project xvir1
xvir1:> 
xvir1:> validate data/single_cell_GEX/processed/XVIR1-POOL-SCG47/cellranger/possorted_genome_bam.bam
/Users/arjunarkalrao/src/metis/bin/metis_client:340: warning: URI.escape is obsolete
Archived
xvir1:> validate data/single_cell_GEX/processed/XVIR1-POOL-SCG48/cellranger/possorted_genome_bam.bam
/Users/arjunarkalrao/src/metis/bin/metis_client:340: warning: URI.escape is obsolete
Archived
xvir1:> validate data/single_cell_GEX/processed/XVIR1-POOL-SCG49/cellranger/possorted_genome_bam.bam
/Users/arjunarkalrao/src/metis/bin/metis_client:340: warning: URI.escape is obsolete
File is not archived!
xvir1:> validate data/single_cell_GEX/processed/XVIR1-POOL-SCG50/cellranger/possorted_genome_bam.bam
/Users/arjunarkalrao/src/metis/bin/metis_client:340: warning: URI.escape is obsolete
File is not archived!
xvir1:> 
arkal:tmp arjunarkalrao$ echo $?
0
```

metis/bin/metis_client
```diff
@@ -651,7 +652,8 @@ EOT
               d.strftime("%b %d  %Y") :
               d.strftime("%b %d %H:%M")
           },
-          file[:file_name] || "#{file[:folder_name]}/"
+          file[:file_name] || "#{file[:folder_name]}/",
+          is_file ? (file[:backup_id] ? "archived" : "unarchived") : "n/a"
```
graft, line 656: Might want to make this not-the-last column since they just seem to be space-separated and file/folder names can contain spaces. Also looks like you need to add columns below to the bucket map and the width computation.

corps:
Looking at this again.

PR #600, opened 2021-10-18
Redcap form

graft:
This PR adds a "form" view on top of the json editor for the Redcap loader. The JSON editor is still available if need be.

The main issues with the form, aside from its exceeding complexity, are the performance issues resulting from its size and the number of widgets being drawn. I've attempted to mitigate this with pagination and tabs, it remains to be seen if this is sufficient.

The interface is busy, and likely contains many bugs; a lot of play-testing is required. While busy, the interface is highly constrained (the only significant constraint it is lacking is data from the redcap template), and I'm optimistic it will at least make it easier for anyone to configure a fresh loader, although it remains to be seen if it is still too cumbersome to be useful.


graft:
>     * When I add a new script, it doesn't get formatted like the ones that come from the original JSON. The original ones show up in these nice chips, but the new one (first one) looks just like a raw text input field. Should they render the same?

Confusingly, there is a "short" and "long" form for this. By default it allows you just to enter a string redcap field (e.g. `{ magma_attribute_name: "redcap_field_name" }`, if you don't want to do any complicated filtering, matching, use of labels, etc. If you want to do something else, it allows a hash, e.g.: `{ magma_attribute_name: { redcap_field: "redcap_field_name", value: "label" } }`.   The string shorthand is exactly equivalent to `{magma_attribute_name: { redcap_field: "redcap_field_name", value: "value" } }`. :shrug: if this is unnecessarily complex or if it makes anything easier.

>     * When I try adding an "exists" condition to the one I created, it explodes the text input (perhaps due to above issue?). It does save, but it winds up duplicating the script, so now I have a text-version "gender" as well as a properly formatted "gender" that includes an "exists" condition.

Fixed, this was because the above is unnecessarily complex!

>     * The change to `fetch` in the Magma actions breaks Timur Query, because query uses `getAnswer` directly instead of going through an action. I see now that there is `requestAnswer`, so we should probably migrate Query to use that, instead. Could you patch that in this PR? If you don't have time this week, do you mind if I push commits to this branch? The change would be in `timur/lib/client/jsx/components/query/query_use_results_actions.tsx`.

Also fixed. Hopefully there aren't any more of these lurking out there outside of tests.

coleshaw:
@graft , were you meaning to push updated commits?

PR #601, opened 2021-10-18
Timur query bug database error

coleshaw:
This PR addresses two issues brought up from user testing on Friday:

* Database errors were not being returned as a list, which meant they were renderd in the UI as individual letters.
* Filtering on numeric strings (i.e. age in a demographic table) broke in production because some demographic fields have mixed lettters + numbers. This PR updates the regex to only accept purely numeric-like strings. Previously the constraint would allow entries like "Uberon:1101" to pass through to the `>` comparison, which would then fail because the DB could not cast that to a number.

### DIFF
--- a/magma/spec/query_spec.rb
+++ b/magma/spec/query_spec.rb
@@ -1114,9 +1114,16 @@ describe QueryController do
 
       expect(json_body[:answer].map { |a| a.last }).to eq([@lion_difficulty.id, @stables_difficulty.id])
       expect(json_body[:format]).to eq(['labors::characteristic#id', 'labors::characteristic#id'])
+
+      query(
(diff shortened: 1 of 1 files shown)


graft (approved):
:+1: always confusing to me whether we should support scientific notation also...

coleshaw:
> +1 always confusing to me whether we should support scientific notation also...

Do the "integer" and "float" columns in Postgres accept scientific notation?

graft:
> Do the "integer" and "float" columns in Postgres accept scientific notation?

Hmm, float is inherently a kind of scientific notation (i.e. sign, mantissa, exponent) so I'd say yes - I was imagining the user might try to use scientific notation in their query - seems like Postgres likes `SELECT 6.02e23` just fine anyway.

PR #603, opened 2021-10-19
Polyphemus redcap set record id field

coleshaw:
This change goes along with mountetna/redcap_projects#36, which I think can be merged and deployed first.

This PR implements a change to allow REDCap project loaders to generate identifiers from a REDCap field other than the `record` value given in REDCap. An example project is AutoIPI.

One change when creating the configs: it does require the person doing the configuration to include the alternate field name in every model config, so that it is available in the record. For example, populating a demographic table attached to subject, would require that every script includes `"subject": "my_project_identifier_field"`. This could get cumbersome with a ton of scripts, but perhaps alleviated by the "copy script" functionality in the new UI form (#600).

If we would like to constrain it more, I think we could also just add some sort of configuration parameter to the model config, like:

```
"subject": {
  "each": [],
  "scripts": [],
  "identifier_field_name": "my_project_identifier_field"
}
```

However, if we think that projects might need code to manipulate one or more REDCap fields into a single Magma record name, perhaps it's best to leave that in the Ruby model definition and not put too many constraints or assumptions into it. The current way this is implemented, the Ruby models can use whatever someone puts into the scripts. Unfortunately, it kind of leaves things in the hands of 1) whoever writes the Ruby model definitions, and 2) whoever writes the config. Kind of brittle since it's a human-communication process at that point, so open to suggestions on how to make this more robust.

As a note, I don't think we need to send the full record to `offset_id()`, as long as the identifier itself includes the alternate REDCap field's value.

### DIFF
--- a/polyphemus/Gemfile.lock
+++ b/polyphemus/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: /etna
   specs:
-    etna (0.1.41)
+    etna (0.1.42)
       concurrent-ruby
       curb
       jwt
@@ -185,4 +185,4 @@ DEPENDENCIES
   yabeda-puma-plugin
 
 BUNDLED WITH
-   2.2.25
+   2.2.27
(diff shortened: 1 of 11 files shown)


coleshaw:
Ah, good point about just `subject: "autoipi_id"`!

I'll work on updating the back-end stuff, and work off of your branch for the front-end form widget.

Issue #615, opened 2021-10-27
UI Steps not always marked as requiring user input to be run


dtm2451:
Re: Discussion in #613, certain stepUserInput workflow steps don't seem to be recognized as denoting `bufferedSteps` / steps that require user choices be `committed` before the step can be performed.

Seems to affect at least select-autocomplet & nested-select-autocomplete, but maybe also others.

Comments from Cole, then Dan:
> > Cool, that works!
> > So I got it into a funky kind of state ... not sure if it's directly related to this PR (might be a separate issue), but I think what I did was:
> > 
> > 1. I ran the workflow on IPI, selecting all the Kidney records.
> > 2. While the "import into scanpy" step was still running, the Batch Correction UI widget appeared and let me select something. I could not commit. But I selected something.
> > 3. Once the two intermediate steps ran and it got to the "Batch Correction, Select batch_by" step, the UI box was re-highlighted. I clicked commit and then run.
> > 4. Nothing happened, the commit button disappeared, and I could not run again.
> > 5. I had to change my "batch correction" selection, re-commit, and then I could re-run.
> > 
> > I think the breaking behavior was being able to select an option before actually getting to the step in the workflow. Should that either 1) not let me select, or 2) remember what I selected?
> 
> Hmmmm when I try to recreate this myself, I don't have the ability to commit at step 3, though I do get to hit the run button, and I can hit that button all that I want to little effect until after I make any change in the batch-by input. I also notice that the "select_batch_options" step was actually missing from the `bufferedSteps` prop of the Vulcan State until after I made a change in that input. Seems that the same goes for the color-by selection step later on!
> 
> Not an issue with the current PR as I can recreate from master. Perhaps an issue related to (nested-)select-autocomplete??

Issue looks like:
![commit-run-issue](https://user-images.githubusercontent.com/38870347/139151240-a5fe0480-2ca2-4dea-9d75-540d30417b5b.gif)

PR #616, opened 2021-10-27
Fix for stepUI vulcan steps not being recognized as bufferedSteps

dtm2451:
Addresses #615.

Updates both select-autocomplte and nested-select-autocomplete to utilize the `useSetsDefault` method for setting their initial value.  That function sets a default value, and also calls `onChange(<the-default>)` which will then trigger the actions for the input to be considered as a bufferedStep.

For nested-select-autocomplete, rather than do a major refactor, I had to change the initial value to be '', the empty string.  The associated validation (NotEmptyValidator) is configured to reject this all the same, and visually, there is no difference to the user.


dtm2451:
Ah nice!  Thanks.  Seems to work.  I just did a search for "DropdownAutocomplete" across our entire monoetna code base and it only yielded the definitions within etna and our vulcan UIs that use it.  So I'll double-check the other vulcan UIs, but then I think it's safe to merge!

PR #619, opened 2021-10-29
Timur query filter clauses

coleshaw:
This PR updates the Query UI and fixes some bugs:

* Can add individual "clauses" to each Where Filter. This should address the use cases brought up by @bsamad and @graft , with regards to aggregating multiple clauses in a single filter (useful for tables as well as bundling child-model constraints). For example, filtering on demographic age > # (or even within a range), or filtering on a set of patients with primary samples and specific compartments for bulk data.
* Fixes a bug with subqueries and subselects, when adding an `::any` or `::every` selector inside of a filter, that was exposed by the above request.


timur/lib/client/jsx/components/query/query_filter_control.tsx
```diff
@@ -1,219 +1,183 @@
...
 
-  // When the operand value changes, follow it
-  useEffect(() => {
-    if (filter.operand !== previousOperandValue) {
-      debouncer.reset();
-      setOperandValue(filter.operand);
-      setPreviousOperandValue(filter.operand);
-    }
-  }, [filter.operand, debouncer, previousOperandValue]);
+  const modelChildren = useMemo(() => {
```
coleshaw, line 112: @graft , I made an assumption here that clauses can only be appending to either the selected filter's model, or it's children models. Because if you go deeper (or elsewhere in the graph), I'm not sure how you would handle the any / every selection at each step in the UI, and it wasn't clear from your use case if that would ever be desireable versus just adding filters at the top level?

I guess one alternative is to allow users to just go "down" the graph, like with slices (i.e. timepoint -> immunoassay -> analyte in COMET), which would be like 2 any / every selectors but without the ability to add slices for each model (nesting all these would get confusing, I think). We would also have to check that the API itself would construct the sub-selects correctly for deeper filters.

PR #622, opened 2021-11-01
Magma dry run for update

coleshaw:
This PR introduces a `dry_run` flag for Magma's `/update` endpoint, where it will return the payload as it would execute, but without making any actual database changes. This will be important for REDCap loading, where users will want to preview the Magma updates (and we store the preview data in Polyphemus logs, in order to show it in the config UI), but we want users to only see the date-shifted results.


graft:
Looks nice! We might consider later moving the stats-computation stuff that currently runs in the redcap loader (e.g. number of records blanked in tables) into the update endpoint (like return `statistics: { blanked: 'etc.' }` for each model when you `--dry-run`).

dtm2451:
Ooooh!  I might want to be able to piggy-back off of this for some magmaR summarization too!  It could potentially simplify some code I already built on that side, but I'll have to think about it more.

Anyway, I've tested via a modified magmaR and it looks good to me!  (The documentation update may take longer than the code update if I do expose this option to magmaR users lol.)

coleshaw:
Hm, interesting. Do those statistics make sense in a general way (i.e. blanked records), or are they REDCap loading specific? We would also have to move the concept of "strict" updates into Magma, in that case.

dtm2451:
> Hm, interesting. Do those statistics make sense in a general way (i.e. blanked records), or are they REDCap loading specific? We would also have to move the concept of "strict" updates into Magma, in that case.

Not sure I have the expectation of where the concept os "strict" updates is handling within my own picture of our infra, so definitely coming from a more naive spot.  But a few examples of stats that could be helpful in some scopes: a # blanked samples (/newly-orphaned?), # de-orphaned maybe too, # of new records (which I understand may always be zero for ETL-loads, but wouldn't be for magmaR), summary of # new attributes filled in for first time (versus literally updated) for X # of records of model Y, are some statistics that I think could help confirm correctness of update submissions for magmaR users and could be nice for polyphenmus too!

coleshaw:
A "strict" update is a concept we've been using with REDCap loading, where sometimes records or data are taken out of the project, and so you want to clear out "stale" Magma records... instead of just updating by adding new data, also remove old data.

PR #625, opened 2021-11-01
Vulcan Save/Open bug fix

dtm2451:
Currently, the system always thinks the user is opening a json pointing to a different project-workflow combination because of an incorrect variable.  This PR fixes that.


corps (approved):

dtm2451:
Additionally: 2nd commit allows users to use the save functionality even in "the middle" of the workflow.  Major use case: a user has made a umap, but aren't ready to run DGE yet. they still want to be able to save how they got where they are in case their browser crashes, or in case they want to share to a collaborator.  Dev use case: we want to test a component within a workflow to check its behavior while steps are running, and we want to be able to revert back to the stage where the input was *ready* to be filled, but not-actually-filled-in yet.

Issue #632, opened 2021-11-02
dropdown-autocomplete could use options scroll


dtm2451:
For cases where there are more than 10 or so options, it is currently impossible to see that there are more options available than the ones initially displayed, or even know what those options might be without typing A-Z as a first letter.

Would improve the color-by and batch-by steps of vulcan scRNAseq!

dtm2451:
Introduced with the MaterialUI upgrade in #754

PR #633, opened 2021-11-03
scRNAseq- populate color-by options from the data object

dtm2451:
Simplifies the code & allows access to coloring by 'n_counts' and other calculated QC metrics, in addition to all the bits pulled from magma.  Also better integrates the options that users see in the color-by selection input with what actually will exist in the object they might download.

PR also:
1) increases the number of items displayed by nested & non-nested autocomplete dropdown inputs so that users can actually see the many options they will have.  Stopgap for #632, but a cursor-based scrolling mechanism could be a target instead (perhaps in the future).
2) Increases the time that individual steps are allowed run for from 5min to 15min
3) combines two of the previous cwl steps into one in order to avoid metis file urls being cached as they would go stale within 24hours.
4) Moves trimming of project_settings data to happen within that initial step, rather than caching the setup for all projects and requiring steps that used it to do `input_json("project_data")`**`[project_name]`**


dtm2451:
Turned into a bit of a smorgasbord of mini pre-release bits

PR #637, opened 2021-11-04
Improve linking performance with more strict regex.

corps:
Better performance for linking by having a stricter regex.
Current issue is that the folder regex looks for pretty much ANY folder underneath a processed record folder, but that is apparently a lot of extraneous stuff.

This restricts the watch folder regex to at most, one folder deep (which matches all known cases in the pipeline).

### DIFF
--- a/polyphemus/lib/etls/ipi/ipi_etls.rb
+++ b/polyphemus/lib/etls/ipi/ipi_etls.rb
@@ -17,7 +17,7 @@ class Polyphemus
           process_watch_type_with(
             bucket('data')
               .watcher('single_cell_pool_processed')
-              .watch(/^single_cell_[^\/]*\/processed\/.*\/[^\/]+POOL[^\/]+\/.*$/),
+              .watch(/^single_cell_[^\/]*\/processed\/.*\/IPIPOOL[^\/]+(\/[^\/]*)?$/),
             Polyphemus::LinkerProcessor.new(
               linker: single_cell_pooled_linker,
               model_name: 'sc_rna_seq_pool'
@@ -33,7 +33,7 @@ class Polyphemus
           process_watch_type_with(
             bucket('data')
               .watcher('single_cell_processed')
-              .watch(/^single_cell_[^\/]*\/processed\/((?!POOL).)*$/),
+              .watch(/^single_cell_[^\/]*\/processed\/.*\/IPI((?!POOL)[^\/])+(\/[^\/]*)?$/),
             Polyphemus::LinkerProcessor.new(
               linker: single_cell_non_pooled_linker,
               model_name: 'sc_rna_seq',
@@ -47,9 +47,7 @@ class Polyphemus
       end
 
       def single_cell_non_pooled_linker
-        SingleCellLinker.new(
-          record_name_regex: /.*\/(?<record_name>IPI[^\/]*)\/.*$/
-        )
+        SingleCellLinker.new
       end
 
       def single_cell_pooled_linker
@@ -115,10 +113,13 @@ class Polyphemus
 
           matched_folders = folders.map do |folder|
             record_name = @linker.record_name_by_path(
-              folder.folder_path + '/',
-              @linker.record_name_regex
+              folder
             )
-            [folder, record_name] if record_names.include?(record_name)
+            if record_names.include?(record_name)
+              [folder, record_name]
+            else
+              nil
+            end
           end.select { |f| f }
 
           matched_folders.each do |folder, record_name|
@@ -147,6 +148,7 @@ class Polyphemus
           super(
             project_name: 'ipi',
             bucket_name: 'data',
+            record_name_regex: /.*\/(?<record_name>IPI[^\/]*)\//,
             **kwds
           )
         end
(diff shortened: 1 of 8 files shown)


polyphemus/lib/etls/ipi/ipi_etls.rb
```diff
@@ -17,7 +17,7 @@ def add_single_cell_linkers!
           process_watch_type_with(
             bucket('data')
               .watcher('single_cell_pool_processed')
-              .watch(/^single_cell_[^\/]*\/processed\/.*\/[^\/]+POOL[^\/]+\/.*$/),
+              .watch(/^single_cell_[^\/]*\/processed\/.*\/IPIPOOL[^\/]+(\/[^\/]*)?$/),
```
corps, line 20: Realized that all ipi record folders include IPI and that it improves accuracy of regexs.

polyphemus/lib/etls/ipi/ipi_etls.rb
```diff
@@ -47,9 +47,7 @@ def add_single_cell_linkers!
       end
 
       def single_cell_non_pooled_linker
-        SingleCellLinker.new(
-          record_name_regex: /.*\/(?<record_name>IPI[^\/]*)\/.*$/
-        )
+        SingleCellLinker.new
```
corps, line 50: This becomes much simpler once I refactored the ipi single cell linker's record name regex to be shared.  In truth, pooled and non pooled linkers are the same, but I kept this separate for now in case other adjustments become obvious.

polyphemus/lib/etls/ipi/ipi_etls.rb
```diff
@@ -33,7 +33,7 @@ def add_single_cell_linkers!
           process_watch_type_with(
             bucket('data')
               .watcher('single_cell_processed')
-              .watch(/^single_cell_[^\/]*\/processed\/((?!POOL).)*$/),
+              .watch(/^single_cell_[^\/]*\/processed\/.*\/IPI((?!POOL)[^\/])+(\/[^\/]*)?$/),
```
corps, line 36: The main fix: previously this would expand ANY folder underneath the IPI record folder, and cause a lot of unnecessary processing that didn't catch anything.

All relevant IPI record folders may either contain files directly in them, or one folder deep, no more.

polyphemus/lib/etls/ipi/ipi_etls.rb
```diff
@@ -115,10 +113,13 @@ def process(cursor, folders)
 
           matched_folders = folders.map do |folder|
             record_name = @linker.record_name_by_path(
-              folder.folder_path + '/',
```
corps, line 118: the record name regex works well for folders --- since file names themselves can also often match the ipi record regex, but are not identifiers non the less.

To make this system work better, record_name_by_path accepts either folders or files and changes its logic to better match based on the target.  See other usages in this PR.

polyphemus/lib/etls/ipi/linkers/ipi_rna_seq_files_linker_base.rb
```diff
@@ -3,9 +3,8 @@
 
 class Polyphemus::IpiRnaSeqFilesLinkerBase < Polyphemus::MetisFilesLinkerBase
   def initialize(project_name: "ipi", bucket_name:, attribute_regex:, record_name_regex:)
-    super(project_name: project_name, bucket_name: bucket_name)
+    super(project_name: project_name, bucket_name: bucket_name, record_name_regex: record_name_regex)
```
corps, line 6: Making record_name_regex an attribute of the base linker really makes things simpler in the end, rather than this subclass.  See other places where it disappears since it is now guaranteed in the constructor.

polyphemus/lib/etls/metis_files_linker_base.rb
```diff
@@ -161,8 +162,15 @@ def should_skip_record?(record_name)
     false
   end
 
-  def record_name_by_path(path, regex)
-    match = path.match(regex)
+  def record_name_by_path(file_or_folder)
+    if file_or_folder.is_a?(Etna::Clients::Metis::File)
+      match = file_or_folder.file_path.match(@record_name_regex)
+    elsif file_or_folder.is_a?(Etna::Clients::Metis::Folder)
+      match = (file_or_folder.folder_path + '/').match(@record_name_regex)
```
corps, line 169: So here's the trick -- the record_name_regex, to differentiate between folder segments and filenames (basenames), requires a '/'.  This helps prevent us from picking up files that happen to match record names but are not records.  In the case of a folder, when we're processing just the folder name itself, we artificially add the ending slash so the regex understands that it is a directory path segment and not a filename.

coleshaw (approved):

polyphemus/lib/etls/ipi/ipi_etls.rb
dtm2451: These allow additional characters between "IPI"/"IPIPOOL" and the next "/" correct?  I'm looking at folder structure and trying to double check what I can.

polyphemus/lib/etls/ipi/ipi_etls.rb
dtm2451: Connected with this comment below:
> These allow additional characters between "IPI"/"IPIPOOL" and the next "/" correct? I'm looking at folder structure and trying to double check what I can.

dtm2451 (approved):

polyphemus/lib/etls/ipi/ipi_etls.rb
corps: Yes.  It requires only that the identifier start with IPI and IPIPOOl.  Tests cover this as well.

PR #638, opened 2021-11-04
Polyphemus redcap config UI stale

coleshaw:
I was running into an issue with stale properties re-appearing across different scripts (see below)

redcap_edit_behavior

This PR addresses that by using some hooks to try and update the callback functions. I think the issue was that the callbacks were defined with a certain `config` value, and they wouldn't update when the config updated. This change insures that the update callbacks will also update when `config` changes.

I'd like to add some tests around the components and possibly break them apart a bit, so might add more to this PR.


polyphemus/lib/client/jsx/etl/redcap-form.tsx
```diff
@@ -289,6 +289,12 @@ const RedcapAttribute = ({att_name, attribute_value, update}:{
 
   let valueComponent;
 
+  const handleUpdateValue = useCallback((field_name: string, newValue: any) => {
```
corps, line 292: I was thinking that, and this doesn't have to fit into this PR, we might want to consider creating some custom hooks that handle some more of these cases where we're essentially wrapping existing hooks with a bit of very generic logic.

Something like, `useStateObjectWithFilter` ?  Especially since callbacks around other hooks gets kinda verbose doing over and over again.

coleshaw: Ah, hm, let me see if I can wrap my head around what that would look like. I think I see what you mean, that these are kind of all thin wrappers around `update()` (which, at the end, is a `setState` type call).

corps (approved):

PR #621, opened 2021-10-30
expand DGE documentation within the vulcan scRNAseq workflow

dtm2451:
Adds a help-doc to the DGE-setup step and also adds a, large, devoted section to the main scRNAseq vignette.

Could use some extra eyes for typo-catching and notes on whether the explanations make sense!

(View them from within vulcan.dev.local!  They're just ugly strings within the cwl and metadata.json)

PR #640, opened 2021-11-05
Production Deploy

dtm2451:
Deploys the most recent Vulcan updates.  #633 and #621, as well as a change by @corps


dtm2451:
Was just gonna push this through so it'd be in before tomorrows standup, thinking it was only my bits, but there's a bit more.  Will await @corps approval.

corps (approved):

PR #641, opened 2021-11-05
Redcap form fixes

graft:
A few fixes to the redcap form:

(1) resolves an issue where the reset button did not clear the value property components correctly. This might be a separate issue from the one in @coleshaw's PR, not clear.

(2) adds the current commit to the revision history, an oversight that has been annoying me for a while.


coleshaw:
Hm, not sure if the reset button issue would be related. I didn't click it when I saw the issue I had, so might be different...?

coleshaw (approved):
Looks good to me!

PR #644, opened 2021-11-09
fix issues with etna and archimedes release?

coleshaw:
@corps , I ran into issues updating my branch against some of your build system fixes (https://github.com/mountetna/monoetna/runs/4155444815?check_suite_focus=true). These changes to the build scripts seem to fix the issues when I test locally, but not sure if they are aligned with the overall build process? I think the `archimedes` one was a simple Make dependency change, but not sure about the `etna` changes. Would love your input on this.

### DIFF
--- a/docker/etna-base-dev/entrypoints/development.sh
+++ b/docker/etna-base-dev/entrypoints/development.sh
@@ -26,6 +26,16 @@ if [ -n "$UPDATE_STATE" ]; then
 eval "${app_name_capitalized}_ENV=test ./bin/${APP_NAME} migrate"
 fi
 
+if [ -n "$FULL_BUILD" ]; then
+  bundle install -j "$(nproc)"
+fi
+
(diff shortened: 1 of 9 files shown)


docker/etna-base-dev/entrypoints/development.sh
```diff
@@ -26,6 +26,16 @@ if [ -n "$UPDATE_STATE" ]; then
 eval "${app_name_capitalized}_ENV=test ./bin/${APP_NAME} migrate"
 fi
 
+if [ -n "$FULL_BUILD" ]; then
```
corps, line 29: I'm not sure this is the right fix, development.sh invokes `build.sh` which should be running this, which makes me think there is a reason why I don't do this.

I think the correct solution may be to remove the RELEASE_TEST instead from the run-image-test weirdly.

coleshaw: Hm...k, let me try that...

PR #649, opened 2021-11-11
Divide primary input cache

graft:
This attempts to break the primary input cache into a bunch of smaller caches so that changes to individual inputs don't force the whole workflow to rerun.

I think this works? There is some strange lag going on in the UI between when you click "Commit" and when the "Run" button turns on that I am suspicious of.


coleshaw:
It also seems like there is more lag between clicking "Run" and when the POST call returns, and the button changes to "Stop"?

coleshaw:
When testing locally, it looks like ~2 seconds to go from commit -> Run, due to checking the status of the workflow. It took about 10 seconds to go from Run -> lightning (actual running)

coleshaw:
When testing this, I seem to get a high number of these errors: `TypeError: NetworkError when attempting to fetch resource.`, though the workflow seems to be running on the server-side and refreshing fixes the UI to match current status. Perhaps some timeout that's causing this on the front-end only? Or some poll that doesn't get cancelled?

coleshaw:
Outside of the two issues I ran into, the functionality of not restarting the entire workflow seems to work well!

vulcan/lib/server/models/workflow.rb
```diff
@@ -38,24 +38,36 @@ def as_steps_json(name)
         }
       end
 
+      def step_key(step)
```
corps, line 41: Super minor nit -- I believe this step variable is actual a step_input, specifically, right?  Super minor.

vulcan/lib/server/models/workflow.rb
```diff
@@ -64,6 +76,10 @@ def source_as_string(source)
     end
 
     class WorkflowInputParameter < Cwl
+      def source
+        [ :primary_inputs, self.id ]
```
corps, line 80: Ahh, got it, so this basically makes that interface consistent.  Makes sense.

corps (approved):
Nice, simple changeset.  I appreciate that it's long studied and short in length, I get it was probably hard to find the right cut.  I can't explain right now what the network errors are, I don't think they are related to your changeset though.

dtm2451 (commented):
Seems to work as expected, :+1:

> When testing locally, it looks like ~2 seconds to go from commit -> Run, due to checking the status of the workflow. It took about 10 seconds to go from Run -> lightning (actual running)

I notice the same.  I wonder if this lag will introduce a new confusion point for our users in a state where we didn't need to care about appearances previously... during the 2 secs between clicking `Commit` and seeing the gray `Run` button become enabled, it switches to a `Stop` button during the orchestrator's calculation period.  I think remaining a grayed out / disabled `Run` button during this period is better.

Unfortunately, further, if I click the `Stop` button during the same period, the action taken seems to be as if I'd clicked `Run` button becuase the next chunks of work get run (hammer electrified and all).  Seems out-of-scope here, but two things that can be fixed together an upcoming PR.

vulcan/lib/orchestration.rb
```diff
@@ -57,9 +57,11 @@ def load_json_inputs!(storage)
 
       # Ensure that the primary inputs, at least, have been loaded so that the status endpoint can report
       # meaningful steps running.
-      pi = build_target_for(:primary_inputs)
-      if pi.should_build?(storage)
-        run!(storage: storage, build_target: pi, token: nil)
+      workflow.inputs.each do |input|
+        pi = build_target_for(workflow.step_key(input))
+        if pi.should_build?(storage)
+          run!(storage: storage, build_target: pi, token: nil)
```
corps, line 63: This is probably the performance hit that causes the commit and run to perform slow.
Is it possible to change run! to execute multiple primary input file loads together?  Or at least share some of the overhead at a lower cost?

PR #650, opened 2021-11-12
Zc/load application secrets from file

corps:
This is a minor adjustment ahead of polyphemus migration to swarm infra.

Goal: make it easy to compose configuration without chef (for instance, base staging config gets merged with metis specific config), and inject <systems> secrets into config at process time (these being different from airflow secrets known to users, this is more like database passwords and rollbar tokens, etc).

Considered these options
1.  Allowing multiple yml files to be merged -- more or less what this solution is, but I found environment variables pointing to files to consistently be the easiest option.
2.  Script the put together configs similar to chef -- more moving parts that gets tricky to maintain.  From vulcan and metis staging migrations, this path seemed least resistant but supporting all kinds of config.yml value injections we'll need.

### DIFF
--- a/etna/lib/etna/application.rb
+++ b/etna/lib/etna/application.rb
@@ -50,6 +50,21 @@ module Etna::Application
     Etna::Application.register(self)
   end
 
+  # a <- b
+  def deep_merge(a, b)
+    if a.is_a?(Hash)
+      if b.is_a?(Hash)
+        b.keys.each do |b_key|
+          a[b_key] = deep_merge(a[b_key], b[b_key])
+        end
+
+        return a
+      end
+    end
+
+    a.nil? ? b : a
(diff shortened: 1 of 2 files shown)


coleshaw:
Looks good to me. For my own knowledge, will these environment variables and files be managed in Portainer, once we move to swarm?

coleshaw (approved):

corps:
> Looks good to me. For my own knowledge, will these environment variables and files be managed in Portainer, once we move to swarm?

Yes.  The environment variables will be defined in the swarm yaml files which will exist both in the repo and in the portainer UI (the portainer UI will sync chagges back to the repo so you can still use it as the main driver).  The files will be managed through portainer (or the docker swarm command line) and attached to services.

PR #653, opened 2021-11-16
Vulcan input drawer

graft:
A mostly-cosmetic change that hides the primary inputs from view in a drawer unless the user specifically wants to go and muck with them. Ostensibly this means a workflow ought to run with the baked-in defaults.

I also changed the step input groups to use a similar material-ui interface for consistency, which has the benefit of making the input feed more compact and gives some nice transitions. So far I haven't touched the actual inputs, just their containers.

### DIFF
--- a/vulcan/lib/client/jsx/components/workflow/session/primary_inputs.tsx
+++ b/vulcan/lib/client/jsx/components/workflow/session/primary_inputs.tsx
@@ -22,12 +22,14 @@ import {makeStyles} from '@material-ui/core/styles';
 
 const useStyles = makeStyles((theme) => ({
   card: {
-    borderRadius: 0
+    borderRadius: 0,
+    border: '1px solid #eee',
   },
   header: {
-    padding: '5px 5px 5px 10px'
+    padding: '5px 10px'
   }
 }));
+
 export default function PrimaryInputs() {
   const {commitSessionInputChanges, dispatch} = useContext(VulcanContext);
 
@@ -89,8 +91,8 @@ function PrimaryInputsInner() {
             <Typography variant='h6'>Inputs</Typography>
           </Grid>
           <Grid item>
-            <IconButton onClick={ () => setOpen(!open) }>
-              { open ? <ExpandLessIcon/> : <ExpandMoreIcon/>}
+            <IconButton size='small' onClick={ () => setOpen(!open) }>
+              { open ? <ExpandLessIcon fontSize='small'/> : <ExpandMoreIcon fontSize='small'/>}
             </IconButton>
           </Grid>
         </Grid>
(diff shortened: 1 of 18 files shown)


vulcan/lib/client/jsx/components/workflow/steps/step_icon.tsx
```diff
@@ -0,0 +1,62 @@
...
+  },
+  [STATUS.PENDING]: {
+    color: 'gray'
+  },
+  [STATUS.ERROR]: {
+    color: 'red'
+  }
+}));
+
+type IconClass = typeof CheckIcon;
```
graft, line 34: I'm pretty sure this is the wrong way to do this, but :shrug: what the right way is.

vulcan/lib/client/jsx/components/workflow/steps/step_icon.tsx
```diff
@@ -0,0 +1,62 @@
...
+  step: WorkflowStep | WorkflowStepGroup;
+}) => {
+  let {state} = useContext(VulcanContext);
+  const {workflow, status} = state;
+  if (!workflow) return null;
+  const statusStr = statusStringOfStepOrGroupedStep(step, workflow, status);
+  const label = labelOfStepOrGroupedStep(step);
+  let icon_config = icons[statusStr] || icons[STATUS.PENDING];
+
+  let IconComponent:IconClass = STATUS.RUNNING === statusStr ? AnimatedClock : icon_config.icon;
```
graft, line 54: The AnimatedClock is slightly different and ought to be upgraded here for consistency...

vulcan/lib/client/jsx/components/workflow/steps/step_icon.tsx
coleshaw: Perhaps the AnimatedClock should be wrapped in `SvgIcon` (https://v4.mui.com/api/svg-icon/) instead, and in which case the IconClass is just `typeof SvgIcon` ?

vulcan/lib/client/jsx/components/workflow/session/primary_inputs.tsx
```diff
@@ -59,19 +78,39 @@ function PrimaryInputsInner() {
...
-          );
-        })}
-    </div>
+    <Card className={classes.card}>
+        <Grid className={classes.header} container alignItems='center' justify='space-between'>
+          <Grid item>
+            <Typography variant='h6'>Inputs</Typography>
+          </Grid>
+          <Grid item>
+            <IconButton size='small' onClick={ () => setOpen(!open) }>
```
coleshaw, line 94: super, super-nit: On the other drawers, it seems like clicking on the heading itself seems to open / close the drawer (not just the icon). Should we be consistent and put this onClick handler here on the grandparent `Grid`?

coleshaw (approved):
Looks nice, definitely saves space and is less intimidating upon page load!

dtm2451:
Hmmm.  I like this visually, but a couple nits:
1. Perhaps, as Cole suggested already, "Inputs" should have an adjective in it.  I'd be a fan of "Primary" more than "Advanced" because of #2
2. I'm **not sure** but there **may** be workflows where we'd want users to see the primary inputs expanded by default.  Idk if I think it's a good idea to bake in the expectation that all workflows will be set up with viable primary input defaults.  Probably okay but I'd like to ponder on it.

PR #654, opened 2021-11-16
Query help doc

coleshaw:
This PR adds the query documentation to the mountetna.github.io site, as well as makes the Timur help link context-aware, so it jumps to the right section of the github.io page based on your current Timur section.


coleshaw:
@graft , can you take a look in your browser and see if this looks better? HTML embedded into the Timur page.

graft:
A few small nits when I render this - the headings for some sections look minuscule (smaller than the text) in some sections (e.g. `Nesting and Expanding Matrices`). You could maybe move the query guide to a sub-page with a link and upgrade its heading levels? Also there appear to be some data missing from the two tables in the appendix in the "Attribute Type" column.

PR #657, opened 2021-11-17
Zc/handle mvir1 cascade better

corps:
So,
In the case of patients who are redacted purely for waiver (not denial of consent), we no longer simply move their data with the recursive move.  Instead, we implement an algorithm where the files effectively are recursively copied, but that 'empty marker files' are left behind in their place merely to indicate their data's presence in the system, to simplify upload handling for @arkal.

This happens atomically thanks to the ordering of uploads -> file overwritting.

For patients who are actively no consent, the old behavior remains: purge actively from the system.

A couple of generally useful workflows came as a result of this, a recursive metis walker workflow and a diff generating workflow.


etna/lib/etna/clients/metis/client.rb
```diff
@@ -24,8 +24,13 @@ def list_all_folders(list_all_folders_request = ListFoldersRequest.new)
       end
 
       def list_folder(list_folder_request = ListFolderRequest.new)
-        FoldersAndFilesResponse.new(
-          @etna_client.folder_list(list_folder_request.to_h))
+        if list_folder_request.folder_path != ''
```
corps, line 27: This also fixes a bug that hadn't been really tested -- 'listing' the root folder is actually backed by the bucket_list action, but code paths that accept arbitrary folder_path values won't see this on their own, so it is necessary to catch and handle it better here.

etna/lib/etna/clients/metis/models.rb
```diff
@@ -50,7 +50,10 @@ def to_h
         end
       end
 
-      class ListFolderRequest < Struct.new(:project_name, :bucket_name, :folder_path, keyword_init: true)
+      class FolderRequest < Struct.new(:project_name, :bucket_name, :folder_path, keyword_init: true)
```
corps, line 53: Just acknowledges that many FolderRequest types have the same fields and an explicit super class makes it a little bit safer to share a request object instantiated once and used in separate operations.

etna/lib/etna/spec/vcr.rb
```diff
@@ -1,9 +1,27 @@
+require 'cgi'
 require 'webmock/rspec'
 require 'vcr'
 require 'openssl'
 require 'digest/sha2'
 require 'base64'
 
+def clean_query(json_or_string)
+  if json_or_string.is_a?(Hash) && json_or_string.include?('upload_path')
```
corps, line 9: upload paths have non deterministic urls thanks to Nonce, so in test handling I created this helped to make it easier to clean them out from those embedded urls by replacing them with the consistent string 'etna-signature'.  This is opt-in helper function and not default behavior.

etna/lib/etna/spec/vcr.rb
```diff
@@ -30,6 +48,10 @@ def request_1.path
       end
     end
 
+    c.register_request_matcher :try_uri do |request_1, request_2|
+      clean_query(request_1.uri) == clean_query(request_2.uri)
```
corps, line 52: For VCR matching, however, it is important to clear non deterministic url components in all cases.

etna/lib/etna/spec/vcr.rb
```diff
@@ -49,7 +71,7 @@ def request_1.path
       end
     end
 
-    # c.debug_logger = File.open('log/vcr_debug.log', 'w')
+    c.debug_logger = File.open('log/vcr_debug.log', 'w')
```
corps, line 74: Whoops, will add this back in.

coleshaw:
Is the idea that if the ETL runs during the middle of an upload, and a patient's folders need to be moved, the empty file markers preserve the folder structure, so that the target folder path will never disappear from underneath an active upload?

Will this run into release conflicts if a patient has marker files in the `processed` bucket, with restricted files in the waiver bucket, and then later their data is released? I think the ETL might try to copy over the files from the waiver bucket into the processed bucket, and then throw exceptions because a file exists at the same path?

polyphemus/lib/metis/mvir1_waiver.rb
```diff
@@ -25,13 +25,125 @@ def pool_search_string(pool_name)
...
+    io.rewind
+
+    uploader = Etna::Clients::Metis::MetisUploadWorkflow.new(
+      metis_client: @metis_client,
+      project_name: @project_name,
+      bucket_name: to_bucket,
+      metis_uid: '7af8e4be837744ce791c6239d00afefd',
+    )
+
+    uploader.do_upload(
```
coleshaw, line 110: Some of these files can be quite large ... does this mean they will all be streamed down / up from Polyphemus, instead of using the Metis `copy` API? Will that be efficient if there are a significant number of large files?

coleshaw:
I wonder if the way that you're copying files by streaming actually makes my empty marker question irrelevant, because Metis will just treat the stream as a new file and overwrite the marker. However, in that case I guess my questions would still be about efficiency of streaming through Polyphemus, but also then, should the `release_patient_data` method use the new `try_copy` method instead of the previous, recursive copy method?

corps:
> Is the idea that if the ETL runs during the middle of an upload, and a patient's folders need to be moved, the empty file markers preserve the folder structure, so that the target folder path will never disappear from underneath an active upload?

The idea is that if a file is left behind from an initial redaction, no upload should occur to the original bucket due to arjun's upload logic he runs. (if file exists, do not reupload).
This is a the compromise on true transactional operations which is a beast to build correctly.

> 
> Will this run into release conflicts if a patient has marker files in the `processed` bucket, with restricted files in the waiver bucket, and then later their data is released? I think the ETL might try to copy over the files from the waiver bucket into the processed bucket, and then Metis will throw exceptions because a (marker) file exists at the same path?

So, the e2e actually validates that this ok, since it copies from one bucket (with markers left behind) and then copies back into it (overwriting the markers).  Uploads do overwrite existing files EXCEPT in the case of read only (which is not the case here).

corps:
> I wonder if the way that you're copying files by streaming actually makes my empty marker question irrelevant, because Metis will just treat the stream as a new file and overwrite the marker. However, in that case I guess my questions would still be about efficiency of streaming through Polyphemus, but also then, should the `release_patient_data` method use the new `try_copy` method instead of the previous, recursive copy method?

Actually, you do raise a point that I need to improve the logic so that it doesn't copy over marker files and overwrite real files, whew!

The efficiency is not great but the intention was the new code path of streaming only triggers for initial waiver patients ONCE, and not anymore after.  So that would be acceptable.  I need to improve the logic to make sure that works sanely.

corps:
In a sense, the mvir1 waiver logic /should/ be an etl that processes changes to magma and metis, but right now it fires continuously, which is a bit odd.

But yes, the performance post adding the "don't copy over markers" shouldn't be so bad other than the diff calculations.  Hmm.  I need to think more on this in truth.

coleshaw:
> The idea is that if a file is left behind from an initial redaction, no upload should occur to the original bucket due to arjun's upload logic he runs. (if file exists, do not reupload).

Oh, I got it! Thanks for clarifying. The marker files are so when he re-runs his upload script, it "realizes" that it has already uploaded the files and doesn't try to re-upload them now that they've been moved.

coleshaw:
> it doesn't copy over marker files and overwrite real files, whew!

Oh yeah, that would be unfortunate...

PR #658, opened 2021-11-18
Fix validation for scRNAseq batch- and color-by

dtm2451:
Adds a `*Strongly*NotEmptyValidator` as an option for user input validation which additionally disallows empty strings.  Addresses a bug report by Alexis in which he didn't get a umap.  It was actually because he hadn't picked what to color his umap by.

The possibility to bypass this selection would have been introduced when Cole and I fixed state update bugs for select-autocomplete (batch-by) and nested-select-autocomplete (color-by) by swapping their initial state to default to `""` instead of `null`. 
Current validator complains on `null` but allows `""`.  For some other input types, that's necessary, so this PR adds a new validator that complains on both, and applies it to both these input types.


graft:
For clarity maybe these should be NotEmptyValidator and NotNullValidator...

dtm2451:
> For clarity maybe these should be NotEmptyValidator and NotNullValidator...

The 'basic' version checks more than just null.  Otherwise, I would definitely agree with you.  Actual workhorse is:
https://github.com/mountetna/monoetna/blob/064613f3a34f1afd0e3f0391396041b4a68632b0/vulcan/lib/client/jsx/selectors/workflow_selectors.ts#L125-L133

coleshaw:
:+1: Nice catch!

corps:
Excellent work!

corps (approved):

PR #659, opened 2021-11-18
scRNAseq - improve DGE outputs

dtm2451:
Makes 2 changes:

- eliminates export of a single, confusing, column of the scanpy dge calculation.
- (Majority of code changes) Updates the code for the preliminary DGE (the one that's run early in order to power cluster annotation) to capture improvements I'd put within the code that behind the later DGE step.

EDIT, also:

- unifies the pval calculation method that's selected by default for the two different DGE steps
- fixes a bug in the filtering of the full DGE results that led to an increase in this data's size, rather than the expected decrease.


dtm2451:
Would be great if I could import the data collection function from one script into the other instead of recreating it, but imagine that might be hard to square with the linter. @corps?

vulcan/lib/server/workflows/scRNAseq.cwl
```diff
@@ -90,7 +90,7 @@ inputs:
     doc: 'A set of strings, separated by commas, for which gene symbols starting with these strings should NOT be shown in the umap overlay. Not case-sensitive, so the same strings will work for both human and mouse ribo & mito genes. Note: this does not affect the full differential expression table that can be downloaded.'
   6_Initial_Cluster_DGE__dge_method:
     type: string
-    default: 'wilcoxon'
+    default: 't-test'
```
dtm2451, line 93: unifies the default here with the default of the later, full, DGE

vulcan/lib/server/workflows/scripts/dge_filter.py
```diff
@@ -21,12 +21,14 @@
 # min_pct
 min_pct = float(input_var('min_pct'))
 # Algorithm choice: OR filter, a.k.a. either group meets the cutoff
-pct_cols = [s for s in DEdf.keys() if (s=="pts" or s.startswith("pts_"))]
-pct_col1 = list(DEdf[pct_cols[0]])
-pct_col2 = list(DEdf[pct_cols[1]])
-idx_passed = [
-    i for i in list(DEdf_filt.index) if ( pct_col1[i] >= min_pct or pct_col2[i] >= min_pct )]
-DEdf_filt = DEdf_filt.loc[idx_passed,:]
```
dtm2451, line 29: Instead of decreasing the number of rows, this often led to a massive increase!  Pandas is confusing and I'm not sure exactly how this worked... but fix below

vulcan/lib/server/workflows/scripts/dge_filter.py
```diff
@@ -21,12 +21,14 @@
...
+pct_cols = [s for s in DEdf_filt.keys() if (s=="pts" or s.startswith("pts_"))]
+pct_col1 = list(DEdf_filt[pct_cols[0]])
+pct_col2 = list(DEdf_filt[pct_cols[1]])
+pct_passed = list(map(
+    lambda i: (pct_col1[i] >= min_pct or pct_col2[i] >= min_pct),
+    list(DEdf_filt.index)
+))
+DEdf_filt = DEdf_filt.loc[pct_passed]
```
dtm2451, line 31: Now I create a boolean list that's the same length as there are rows in the df, then use this variable with the same subsetting method as in all the other subsetting lines above.

PR #661, opened 2021-11-18
Zc/raw fastq ipi linking2

corps:
Approach to the fastq migration.

Approach:
I will create new attributes, link them, validate their values are correct, then perform the rename.
Current command here was added to migrate the existing empty value columns, but my approach for the attributes that have values will be to, again, first create the new attribute, complete the linking, then execute the rename.  I don't intend to keep the code that performs the migration checked into master, and will simply use the local polyphemus command as 'scratch'.

### DIFF
--- a/polyphemus/lib/etls/ipi/ipi_etls.rb
+++ b/polyphemus/lib/etls/ipi/ipi_etls.rb
@@ -11,6 +11,26 @@ require_relative '../shared/single_cell/single_cell_linkers'
 
 class Polyphemus
   module Ipi
+
+    module RecordNameDemangler
+      def corrected_record_name(record_name)
+        if record_name.include?('POOL')
+          record_name.gsub(/_/, '.').sub(/^([^.]*\.[^.]*\.)(.*)$/) do
+            first = $1
+            second = $2
+            first + second.downcase
+          end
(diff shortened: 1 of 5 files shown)


corps:
Migration is complete now, all that remains is to get this into master and have it running constantly.

polyphemus/lib/etls/ipi/ipi_etls.rb
```diff
@@ -11,13 +11,74 @@
...
+        end
+      end
+
+
+      SINGLE_CELL_ROOT = /^single_cell_[^\/]*/
+      POOL_CONTAINER = /.*\/IPIPOOL[^\/]+(\/[^\/]*)?$/
+      NON_POOLED_CONTAINER = /.*\/IPI((?!POOL)[^\/])+(\/[^\/]*)?$/
+    end
+
+    module RawFastqLinkers
```
coleshaw, line 39: Super nit: I think there are also raw fastq files for other (non-single-cell) assays, like whole exome sequencing and bulk RNA sequencing. Not sure if here you want to specify this module as `SingleCellRawFastqLinkers`?

coleshaw (approved):

PR #664, opened 2021-11-30
Better tracking of changesets in CI

corps:
I made a change to ensure that in CI, only projects that "have a recent change" actually get built and tested to continue to improve branch and CI performance.

this works.  It works so well, it demonstrated that the concept of "have a recent change" was previously ill defined.

Previously, I relied on search git history for commits that belong to GithubAction, which are commits created on master while syncing subtrees.

The problem is that the way sync subtrees writes commits is, not nearly as neat as I thought it would be.  It interweaves subtree merge commits across repos in an inconsistent way which made this definition of 'have a recent change" unreliable, and sometimes CI wasn't building and releasing images of things that actually changed.

To fix this, I've introduced two new branches `staging-build` and `production-build` which help track "build state" in a more sane way.  (I could have used tags, but it resulted in more complex git fetch commands than branches).

1.  Branches build by comparing themselves against master.  Ideally, this would check by comparing against the pull request target instead of master, but I haven't worked this out quite yet.  This still works in 99.999% of cases.
2.  Master builds by comparing itself against staging.  It then pushes itself to staging, which triggers automatically
3.  Staging builds by comparing itself against staging-build.  It then pushes itself to staging-build.  The push to production is still a PR.
4.  Production builds by comparing itself against production-build.  It then pushes itself to production-build.  production-build is a terminal point that merely tracks state of what has been successfully built and pushed.

### DIFF
--- a/.github/workflows/branches.yml
+++ b/.github/workflows/branches.yml
@@ -43,4 +43,6 @@ jobs:
           CI_SECRET: ${{ secrets.CI_SECRET }}
           IMAGES_POSTFIX: :master
           TARGET_ROOT_BRANCH: master
-        run: make -C ${{ matrix.test-app }} release
+        run: |
+          git fetch origin master
+          make -C ${{ matrix.test-app }} release
--- a/.github/workflows/branches.yml
+++ b/.github/workflows/branches.yml
@@ -42,7 +42,7 @@ jobs:
           IS_CI: 1
           CI_SECRET: ${{ secrets.CI_SECRET }}
           IMAGES_POSTFIX: :master
-          TARGET_ROOT_BRANCH: master
+          TARGET_ROOT_BRANCH: origin/master
         run: |
           git fetch origin master
           make -C ${{ matrix.test-app }} release
--- a/.github/workflows/branches.yml
+++ b/.github/workflows/branches.yml
@@ -42,4 +42,5 @@ jobs:
           IS_CI: 1
           CI_SECRET: ${{ secrets.CI_SECRET }}
           IMAGES_POSTFIX: :master
+          TARGET_ROOT_BRANCH: master
         run: make -C ${{ matrix.test-app }} release
--- a/.github/workflows/branches.yml
+++ b/.github/workflows/branches.yml
@@ -6,11 +6,11 @@ on:
       - '*/*'
       - '*/*/*'
       - '!master'
-      # The swarm stacks branch is for keeping up to date with swarm stack configuration,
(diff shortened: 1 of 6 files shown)


corps:
Fixing breakage, it's just related to github action's checkout not always capturing all history.  Arg.

.github/workflows/branches.yml
```diff
@@ -6,11 +6,11 @@ on:
       - '*/*'
       - '*/*/*'
       - '!master'
-      # The swarm stacks branch is for keeping up to date with swarm stack configuration,
-      # thus running CI on it would be a major waste.
       - '!swarm/stacks'
       - '!production'
       - '!staging'
+      - '!production-build'
```
corps, line 12: Don't trigger the branch workflow on these specially named branches.

.github/workflows/branches.yml
```diff
@@ -37,9 +37,12 @@ jobs:
       - name: Run ${{ matrix.test-app }} test suite
         id: app_first_test_run
         env:
+          # Runs tests and builds images, but does not push (PUSH_IMAGES not set to 1)
           IMAGES_PREFIX: etnaagent/
           IS_CI: 1
           CI_SECRET: ${{ secrets.CI_SECRET }}
           IMAGES_POSTFIX: :master
+          TARGET_ROOT_BRANCH: master
```
corps, line 45: When comparing a branch for changes made, use master as the base for the diff.  TODO: make this relative to pull request target.

.github/workflows/production.yml
```diff
@@ -4,24 +4,32 @@ on:
...
-      - name: Test + Push Release
+      - name: Build & Push production tag images to dockerhub
         env:
           IS_CI: 1
           PUSH_IMAGES: 1
           IMAGES_PREFIX: etnaagent/
           IMAGES_POSTFIX: :production
           NO_TEST: 1
-        run: make -j 4 release
+          TARGET_ROOT_BRANCH: production-build
```
corps, line 30: Production builds what is different from the last successfully pushed 'production-build' point.

docker/build_support/find-oldest-branch-root.sh
```diff
@@ -2,5 +2,11 @@
 
 set -e
 
-git log --pretty='format:%h %an' | grep GithubAction | head -n 1 | cut -d' ' -f1
+if ! [[ -z "$TARGET_ROOT_BRANCH "]]; then
```
corps, line 5: Rather than history manipulation, require explicit definition of the root branch based on the workflow in CI.

PR #668, opened 2021-12-02
Magma query tsv download from the server

### LINKED ISSUE #669: query#action logs show a truncated log line
Magma log output includes the regular Etna controller logging, but the log for the query endpoint does not seem to show the complete input. E.g. this line: `calling query#action with params {:query=>"[\"::first\", \"::identifier\"]", :project_name=>"mvir1"}` which is evidently the final clause of a valid query, but not the entire thing. Perhaps a result of recent changes to log redacting?

coleshaw:
This PR adds a "tsv" format option to the query API on Magma, which means that Timur Query UI can download TSVs directly from the server instead of generating them in the user's browser. This is much more performant and allows Query UI users to download large gene expression TSVs from Query directly. The current functionality only really works for small, non-gene-expression-size TSVs.

The code feels like it could be tightened up with better abstractions, but ... here it is. It generates the TSV based off of the final attributes that the user requested as part of the query. It ignores all the intermediate data that would appear in the JSON version.

For example, if you query `["patient", "::all", ["sample", "::all", "rna_seq", "::all", "tube_name"]]`

the JSON would include something like

```
[
  [ IPIADR001,
    [
      [ IPIADR001.T1,
        [
          [ IPIADR001.T1.rna.live, IPIADR001.T1.rna.live ],
          [ IPIADR001.T1.rna.stroma, IPIADR001.T1.rna.stroma ]
        ]
     ]
  ],
]
```

The TSV will skip the intermediate "sample" data and instead just include the patient ipi_number and rna_seq tube_name data (IPIADR001 and IPIADR001.T1.rna.live).

Should work for matrices, expanded / non-expanded, and also allows the user to rename the TSV columns. No functionality to transpose the TSV for you yet, like in Search, but can add that in a subsequent PR. The transpose functionality exists on the server-side, but there is no UI widget for it.

Also closes #669 .


magma/development/initialize-projects
```diff
@@ -12,6 +12,7 @@ if [ "${MAGMA_ENV}" == "test" ]; then
 PGPASSWORD=password psql -h magma_db -U developer -d magma_development -c "DROP DATABASE magma_test;"
 PGPASSWORD=password psql -h magma_db -U developer -d magma_development -c "CREATE DATABASE magma_test;"
 PGPASSWORD=password psql -h magma_db -U developer -d magma_development -c "GRANT ALL PRIVILEGES ON DATABASE magma_test TO developer;"
+PGPASSWORD=password psql -h magma_db -U developer -d magma_test -c "CREATE SCHEMA labors;"
```
coleshaw, line 15: I seemed to need this locally to run tests in the Magma container...so just adding it to the script for convenience.

magma/lib/magma/server/query.rb
```diff
@@ -11,7 +11,7 @@ def action
         return success(Magma::Predicate.to_json, "application/json")
       end
       question = Magma::Question.new(
-        @project_name, @params[:query],
+        @project_name, @params[:query].dup,
```
graft, line 14: Will this only be a shallow copy and so vulnerable to internal arrays being shifted?

graft (approved):
Looks good to me. I like the use of 'transpose', is this only available in TSV mode?

The most difficult part of this seems to be the logic for collapsing deeply nested paths, which I'm impressed you figured out rather than just punting and demanding a simpler format when making TSVs. But I'm not quite sure I understand how that behaves, with e.g. '::all' of a table attribute or a matrix `::slice`.

coleshaw:
> Looks good to me. I like the use of 'transpose', is this only available in TSV mode?

Yes, at the moment it is only useful with the TSV format. I'm not sure how you would transpose a nested JSON response....what would that look like, you think?

> But I'm not quite sure I understand how that behaves, with e.g. '::all' of a table attribute or a matrix ::slice.

So I think the logic for this is right, but as you said, it's tricky, so perhaps I missed something. Both of these are handled slightly differently through a combination of [calculating the attribute "leaf" in each format path](https://github.com/mountetna/monoetna/pull/668/files#diff-4135f9f0debfa86916f2225b305ac06d1e3d9de3a16ddb633b77fa42120bfe3aR40), [dig_flat](https://github.com/mountetna/monoetna/pull/668/files#diff-683ba8d6feb37bc1c364d74eee6fa0727d7b474f93b78640a05096445ae500a0R171) and [path_to_value](https://github.com/mountetna/monoetna/pull/668/files#diff-683ba8d6feb37bc1c364d74eee6fa0727d7b474f93b78640a05096445ae500a0R46). I'll do my best to explain...

The first assumption is that the format matches the answer in terms of array structure and nesting -- this generally is true, **except** in the two cases you mentioned, `::all` and matrix data (with or without `::slice`). `path_to_value` finds the array-index path to get to the given model + attribute.

In the case of `::all`, the `answer` portion will have an array of values at some index in the path, whereas with `::first`, the index-path lines up exactly. Example:

Query like `{"project_name": "ipi", "query": ["patient", "::all", [["sample", "::first", "rna_seq", "::first", "tube_name"]]], "page_size": 2, "page": 1}`

```
{
    "answer": [
        [
            "IPIADR001",
            [
                null
            ]
        ],
        [
            "IPIADR002",
            [
                "IPIADR002.T1.rna.stroma"
            ]
        ]
    ],
    "type": "Array",
    "format": [
        "ipi::patient#ipi_number",
        [
            "ipi::rna_seq#tube_name"
        ]
    ]
}
```

`path_to_value` on `ipi::rna_seq#tube_name` returns `[1, 0]`. For each element in the `answer`, then, we'll look at `[1, 0]`. In this case it's a string (or `null`), so we know that's the "answer" for each record (`null` and `IPIADR002.T1.rna.stroma`). The reduction has happened already.

Whereas with `{"project_name": "ipi", "query": ["patient", "::all", [["sample", "::all", "rna_seq", "::all", "tube_name"]]], "page_size": 2, "page": 1}`

```
{
    "answer": [
        [
            "IPIADR001",
            [
                [
                    [
                        "IPIADR001.T2",
                        []
                    ],
                    [
                        "IPIADR001.N2",
                        []
                    ],
                    [
                        "IPIADR001.N3",
                        []
                    ],
                    [
                        "IPIADR001.N1",
                        [
                            [
                                "IPIADR001.N1.rna.live",
                                "IPIADR001.N1.rna.live"
                            ]
                        ]
                    ],
                    [
                        "IPIADR001.T5",
                        []
                    ],
                    [
                        "IPIADR001.T6",
                        []
                    ],
                    [
                        "IPIADR001.T4",
                        []
                    ],
                    [
                        "IPIADR001.T1",
                        [
                            [
                                "IPIADR001.T1.rna.tumor",
                                "IPIADR001.T1.rna.tumor"
                            ],
                            [
                                "IPIADR001.T1.rna.live",
                                "IPIADR001.T1.rna.live"
                            ],
                            [
                                "IPIADR001.T1.rna.stroma",
                                "IPIADR001.T1.rna.stroma"
                            ],
                            [
                                "IPIADR001.T1.rna.tcell",
                                "IPIADR001.T1.rna.tcell"
                            ],
                            [
                                "IPIADR001.T1.rna.myeloid",
                                "IPIADR001.T1.rna.myeloid"
                            ],
                            [
                                "IPIADR001.T1.rna.test",
                                "IPIADR001.T1.rna.test"
                            ]
                        ]
                    ]
                ]
            ]
        ],
        [
            "IPIADR002",
            [
                [
                    [
                        "IPIADR002.T1",
                        [
                            [
                                "IPIADR002.T1.rna.stroma",
                                "IPIADR002.T1.rna.stroma"
                            ],
                            [
                                "IPIADR002.T1.rna.live",
                                "IPIADR002.T1.rna.live"
                            ],
                            [
                                "IPIADR002.T1.rna.tumor",
                                "IPIADR002.T1.rna.tumor"
                            ]
                        ]
                    ]
                ]
            ]
        ]
    ],
    "type": "Array",
    "format": [
        "ipi::patient#ipi_number",
        [
            [
                "ipi::sample#sample_name",
                [
                    "ipi::rna_seq#tube_name",
                    "ipi::rna_seq#tube_name"
                ]
            ]
        ]
    ]
}
```

`path_to_value` for `ipi::rna_seq#tube_name` returns `[1, 0, 1, 0]`. Note that the first `[1, 0]` step is a reduction step (`::all` on `sample`), and in the `answer`, we can see that instead of having a `sample_name` string as `format` indicates, we have an array. So `dig_flat` reduces in this case by iterating through each element of that inner array and grabbing the values from `[1, 0]` (the second set of `[1, 0]` in the `path_to_value`) from each internal element. For internal elements that have data, we'll see that `[1]` is actually an array as well, not a string (the second `::all` for `rna_seq`), so `dig_flat` will further reduce and in those cases will grab the values from index `0` of the innermost arrays (the last `0` in `path_to_value`). Perhaps that method should be called `dig_reduce`...

Note that it is possible this may break for `file_collection`, but it seems to work (in specs) because the entire array of file URLs is just essentially replicated, I think.

The logic for extracting matrix data in `dig_flat` is nearly identical to non-matrix, non-all data. What makes the difference is the determination of the "column" name in the `format` (which in turn defines the `path_to_value`).

For example, with a query like `{"project_name": "ipi", "query": ["rna_seq", ["tube_name", "::equals", "IPIBLAD005.N1.rna.live"], "::all", "gene_tpm", "::slice", ["ENSG00000000003"]], "page_size": 2, "page": 1}` (note -- this is dev data, so no clue if the answer data is correct):

```
{
    "answer": [
        [
            "IPIBLAD005.N1.rna.live",
            [
                44.68
            ]
        ]
    ],
    "type": "Array",
    "format": [
        "ipi::rna_seq#tube_name",
        [
            "ipi::rna_seq#gene_tpm",
            [
                "ENSG00000000003"
            ]
        ]
    ]
}
```

The determination of the query "columns" [ignores the inner-array listing out the matrix slices / columns](https://github.com/mountetna/monoetna/pull/668/files#diff-4135f9f0debfa86916f2225b305ac06d1e3d9de3a16ddb633b77fa42120bfe3aR48), and they are only injected as headers [if necessary](https://github.com/mountetna/monoetna/pull/668/files#diff-683ba8d6feb37bc1c364d74eee6fa0727d7b474f93b78640a05096445ae500a0R87). So the matrix column in this case is `ipi::rna_seq#gene_tpm` **not** `ENSG00000000003`, and the `path_to_value` is thus calculated as `[1, 0]` for the matrix. This winds up matching the path in `answer`, except in Ruby we have to serialize the [MatrixValue](https://github.com/mountetna/monoetna/pull/668/files#diff-683ba8d6feb37bc1c364d74eee6fa0727d7b474f93b78640a05096445ae500a0R198) instance.

magma/development/initialize-projects
corps: So, the way it works now is that it scans the config.yml for the test section and initializes schemas for projects  based on entries in that config.  Can you verify if that is happening for you locally?  In the same file, the for loop.

corps: Oh you know, I think this doesn't work for test, only development at the moment.  Maybe we could change the for loop to do a similar initialization for test section?

coleshaw: It doesn't happen when I run `make -C magma shell`:

```
+ for hook in /app/development/*
+ '[' -x /app/development/initialize-projects ']'
+ /app/development/initialize-projects
DROP DATABASE
CREATE DATABASE
GRANT
+ '[' -n '' ']'
+ mkdir -p /app/data/uploads
+ mkdir -p /app/data/data_blocks
+ exec bash

 2021-12-03 19:15:06 ⌚  0a93cbb12867 in /app
○ → bundle exec rspec spec/query_spec.rb 

An error occurred while loading spec_helper.
Failure/Error: Magma.instance.db[:attributes].truncate

Sequel::DatabaseError:
  PG::UndefinedTable: ERROR:  relation "attributes" does not exist
```

Perhaps the $UPDATE_STATE environment variable isn't getting set correctly? Would a better fix be to add that variable to Magma's `docker-compose.yml`, like `    'UPDATE_STATE': '1'` under the app?

coleshaw: If I update Magma's `docker-compose.yml`, I get the expected output:

```
+ for hook in /app/development/*
+ '[' -x /app/development/initialize-projects ']'
+ /app/development/initialize-projects
DROP DATABASE
CREATE DATABASE
GRANT
Your Gemfile lists the gem curb (>= 0) more than once.
You should probably keep only one of them.
Remove any duplicate entries and specify the gem only once.
While it's not a problem now, it could cause errors if you change the version of one of them later.
Migrating to latest, globally
Initializing ./spec/labors/...
```

PR #670, opened 2021-12-03
Magma query attribute distinct

coleshaw:
This PR adds a `::distinct` verb to the model predicates, to allow you to grab all distinct values of an attribute. i.e. `["demographic", "::distinct", "name"]` returns the existing table `name` entries.

This is in preparation of simplifying the Query UI to pre-populate some clause fields as dropdowns instead of blank text fields.

Note, this doesn't work for foreign key relationships or JSON entries (i.e. files)....not sure if that's okay? We could get the files working by converting the columns to JSONB instead of JSON, but in reality all the file attributes are unique and distinct anyways, so I don't think that's really in our use case.

### DIFF
--- a/magma/lib/magma/query/distinct.rb
+++ b/magma/lib/magma/query/distinct.rb
@@ -1,13 +1,11 @@
 class Magma
   class Distinct
-    attr_reader :table_alias
-
-    def initialize(table_alias)
-      @table_alias = table_alias
+    def initialize(column_name)
+      @column_name = column_name
     end
 
     def apply(query)
-      query.distinct.unordered
+      query.distinct.order(@column_name)
     end
   end
 end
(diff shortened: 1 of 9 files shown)


corps:
I agree with your assumptions.

corps (approved):
Let's make sure the client intelligently does not add distinct to json / file columns since doing so likely results in double the performance cost for them with very little benefit (forces one scan to collapse the distinct before the scan to select even occurs, so may as well just grab the whole set in that case)

coleshaw:
On the client side, I was planning to leave the input as a text field except for table `name`s, so hopefully will avoid that type of double-distinct issue...

graft:
This is neat, I also wrote about using this in the "map features" page on notion (task 3). There I posited this would only be useful for string attributes, so it might be useful to restrict to that.

coleshaw:
Great point, that makes sense that it might only be useful on string attributes.

PR #672, opened 2021-12-04
Metis recursive delete

graft:
Adds recursive deletion to the API, UI and metis_client.

I contemplated, but did not include here, saving the deleted contents into a deletion log table. It seems like if we are allowing wholesale deletes of trees, but saving the contents, that an undelete function should be in order. TBD.


coleshaw (approved):
Nice! Works for me in both the UI and the client. I think a lot of folks will find this super-useful!

PR #673, opened 2021-12-06
starting vulcan docs

graft:
Ibid


mountetna.github.io/vulcan.md
```diff
@@ -0,0 +1,117 @@
...
+and the Vulcan Workflow UI will appear with a new workflow session.
+
+The interface is divided into three columns. On the left, the user may
+configure primary inputs, or, once the workflow is running, UI query inputs
+created by the workflow. Outputs, such as plots or download links, appear in
+the second column. The final column shows the steps in the workflow in the
+order they will run (approximately, since the workflow may branch).
+
+The user may change inputs as they see fit; changes will not be persisted to
+the session until the user commits their edits to each required input. Once the
```
coleshaw, line 112: Would it be helpful to explain the `commit` button here? I think we had talked about changing how that interaction would happen (commit vs run), not sure if we reached a concrete conclusion...

PR #676, opened 2021-12-06
Swarm/stacks

corps:


coleshaw:
Hm, should `swarm/config-scripter/node_modules` be in `.gitignore`? I'm suprised our top-level `.gitignore` doesn't include a `node_modules` entry...

PR #678, opened 2021-12-07
scRNAseq - fix bug in dge_filtering script

dtm2451:
Fixes a bug at the "Filter the DGE Output" step of the scRNAseq workflow:
```
Python error while executing script: Traceback (most recent call last):
  File "/archimedes-exec/tmp5umht97s/script.py", line 27, in <module>
    pct_passed = list(map(
  File "/archimedes-exec/tmp5umht97s/script.py", line 28, in <lambda>
    lambda i: (pct_col1[i] >= min_pct or pct_col2[i] >= min_pct),
IndexError: list index out of range
```
Caused by creating indexes from a pandas DataFrame in a non best practice way.  PR switches it over to a safer method.


vulcan/lib/server/workflows/scripts/dge_filter.py
```diff
@@ -26,7 +26,7 @@
 pct_col2 = list(DEdf_filt[pct_cols[1]])
 pct_passed = list(map(
```
corps, line 27: Hold on, this all seems very suspicious.  I think your data types and your thinking about index and keys has been distorted.


I don't clearly understand how `DEdf_filt.index` or its length is supposed to correlate to the length of further mapped and filtered individual columns.  (Seems like the filtering can ensure that the indexing from the original does not match the inner).  But I think the following code is achieving the semantic goal: apply a predicate to all pairs by ordering, using `zip` and no index at all.


Let's try something simpler: 

```
pct_passed = [pct1 > min_pct or pct2 > min_pct for pct1, pct2 in zip(pct_col1, pct_col2)]
```

PR #679, opened 2021-12-07
Pull redcap projects into monoetna

corps:
Goal:
  Make it possible, -in the interim-, to deploy a polyphemus stack into the newvm without wiring up ssh keys and syncing across docker volumes.  The image builds and deploys, by default, with the latest existing etls.

Followup:
*  Change chef to no longer checkout and mount these in the old vm.  This will be safe -after- this PR hits production and is confirmed to be working.
 *  Begin the polyphemus-staging rollout to swarm.  
 *  Begin integrating first pass alternative with airflow, one with immediate turn around.
 *  Validate this works in staging.
 *  Rollout to production.


coleshaw:
Is this directory `polyphemus/lib/etls/renaming/projects` supposed to be a git subproject? Or just copied over like the redcap project?

PR #680, opened 2021-12-07
Timur query populate table names

coleshaw:
This PR updates the Timur Query UI with some small tweaks:

* Uses the `etna-js` Magma action `requestAnswer` instead of the raw `magma_api` `getAnswer` call. This standardizes all the calls to `requestAnswer` so shouldn't run into refactoring issues like before.
* Fixes a bug with matrix attributes not rendering correctly in the Query data frame, when in "nested" mode. Now it works like the Seach page, where a dialog appears with the right matrix data, if in nested mode (with or without slices).
* Pre-populates any `string` attribute operands in Query with a combobox input, where you can see the list of current, distinct values in Magma for a field, but you can also type in your own input (i.e. for an `::in` or `::notin` operator). This is particularly useful for table `name` attributes, though may be confusing for things like table `value` or `notes`. We can restrict if we want, but starting out with all `string` attributes. These operands are used in the Filters and the column Slices.


etna/packages/etna-js/actions/magma_actions.js
```diff
@@ -68,7 +70,7 @@ export const consumePayload = (dispatch, response) => {
 };
 
 const showError = dispatch => response => {
-  if ('error' in response) {
+  if (_.isObject(response) && 'error' in response) {
```
corps, line 73: Curious -- why was this change necessary?  would response && 'error' in response suffice?

etna/packages/etna-js/actions/magma_actions.js
```diff
@@ -357,8 +359,10 @@ export const requestAnswer = (question, callback = null) => dispatch => {
       if (callback) callback(response);
       else return response;
   }).catch( error => {
-    console.log(error);
-    throw new Error(error);
+    return Promise.resolve(error).then((e) => {
```
corps, line 362: Also very confused about this -- why is this extra layering necessary?

timur/lib/client/jsx/components/query/query_use_query_clause.tsx
```diff
@@ -30,33 +38,40 @@ const useQueryClause = ({
...
+
+      return template.attributes[
+        clause.attributeName
+      ].attribute_type.toLowerCase();
     }
-    return 'text';
+    return '';
   }, [clause.attributeName, clause.modelName, graph]);
 
+  const fetchDistinctAttributeValues = useCallback(() => {
```
corps, line 50: My general concern here is cancellability.

Is it possible for fetchDistinctAttributeValues to be called multiple times before completion?  say a user pressing the button multiple times or navigating to and from the page?

Currently, it seems like this code would introduce a race condition in the above case -- that is, the order in which the .then operators get invoked after many calls to the original callback is non deterministic and likely to lead to confusion.

I have some tools in etna already to address this: `Cancellables` in `cancellable.jsx`.  But it's only a problem if the answer to #1 is yes.

corps (approved):

etna/packages/etna-js/actions/magma_actions.js
coleshaw: The TSV response from Magma (for `/query` TSV download) is just a string, so `'error' in response` throws an exception.

etna/packages/etna-js/actions/magma_actions.js
coleshaw: I actually might have added this before making the change to `showError` above for the TSV download, when trying to figure out why the TSV download was throwing an exception. Might not be necessary -- will check.

PR #684, opened 2021-12-09
Metis recursive ls and hash

graft:
Using @ravipatel4's [code](https://github.com/ravipatel4/metis/commit/b1680f5db3b524c97343ac1cefc62283561ae161), this adds recursive ls. I refactored a bit so that instead of buffering (which might be a while for a very large listing) it streams a bit, and is hopefully more interruptable.

Does not seem to work in the project root, thank goodness. Niggling remainder: I think `ls -l -r` should probably list relative paths and not absolute ones, as `ls -l` does.

Also makes the file hash visible in ls -l, which makes it pretty fat, but more useful.


coleshaw:
I seem to get stuck into an infinite loop when I run this locally ... ? Perhaps some weird artifact of my local structure? But I'm pretty sure I have no loops in my Metis directory structure... Does that happen for anyone else?

coleshaw:
Also, is it your expectation that this lists all sub-folders recursively as well? I think the original did not.

graft:
The recursion was already there, I only added printing out folders as well. I fixed a bunch of issues. The main remaining one is that now that the ls does not buffer, different calls to `long_display` result in different column widths. Dunno what to do - fixed width columns?

dtm2451:
You could reprint the header each time?

coleshaw:
Do you mean some output like this?

```
developer@ucsf.edu  Nov 02  2020 n/a n/a TestBucket/test_copy/
developer@ucsf.edu  Nov 02  2020 n/a n/a TestBucket/test_copy/IPIADR001/
developer@ucsf.edu    Feb 04 21:39        n/a                                   n/a TestBucket/test-files/
developer@ucsf.edu 85 Nov 08 16:54 unarchived temp-eec9cc6eaa76740db165cab9fec743e7 TestBucket/test-files/test_file.html
developer@ucsf.edu    Feb 04 21:39        n/a                                   n/a TestBucket/test-files/subdir1/
developer@ucsf.edu 13 Nov 08 16:53 unarchived temp-81a1ef0c0802a3ef19040dc3a4cb9d50 TestBucket/test-files/subdir1/test_file.js
```

I think there's a stray data point getting in there somehow -- note the extra `85` and `13` values in column 2 of the two "longer" rows...

coleshaw:
Oh, I see, the lack of buffering means that you can't predict future column widths, for nested files...but since you do know the minimum possible widths for those `n/a` columns, fixed column width might make sense...

corps (approved):

PR #689, opened 2021-12-10
Link Timur-Query to Vulcan-Viz via queryTerms string

dtm2451:
Makes use of magma/query's the new format="tsv" option, and extends this into magby.query() to power the data pull step of ready-made visualization workflows. Explicitly adds Ipi to the projects for these workflows as well.

2 open Qs (at least) before merge:
1. Should I add other projects as well, because Timur-Query is widely available, so these workflow could be made available to all projects really.
2. How to set up magby with poetry in order to have it build from files properly after changes are made.  There's an open issue in poetry regarding exactly this type of use-case.  Basically, poetry fails to use the updated code and instead works off of a previous cache.  The fix Zach gave me when we paired was to delete the contents of the previous cache, and then create a soft link to the /magby/magby/ folder from within the vulcan image. But... that's not necessarily the right long-term solution.


etna/packages/magby/magby/Magby.py
```diff
@@ -74,19 +74,24 @@ def retrieve(self,
     def query(self,
               projectName: str,
               queryTerms: List,
+              format: str = "json",
               **kwargs) -> Dict:
         '''
         Performs an expressive query to Magma.
         :param projectName: str. Project name in Magma
         :param queryTerms: List[str]. Query terms to magma. See Magma documentation
+        :param formt: str. "json" or "tsv". Adjusts the format of the returned data. See Magma documentation
```
dtm2451, line 83: @coleshaw, please correct if needed

coleshaw: "json" won't actually do anything. The backend will currently default to JSON response unless `format=tsv` is passed in.

dtm2451: gotcha!  Adjusted here.  We should add this and the distinct change to the mountetna.github.io query docs.

coleshaw: Oh thanks, good reminder! I'll put in a PR to update the query docs for the wiki.

coleshaw:
Regarding other projects, I think that's a larger group / Gabi question. If I recall correctly, I think we were going to roll out Vulcan slowly since we weren't sure how it would handle the load for all projects?

coleshaw:
Hm, regarding Poetry install, is that an issue only in development, or does that affect the production images, too? I haven't played around with Magby, so not familiar with the workflow.

dtm2451:
> Regarding other projects, I think that's a larger group / Gabi question. If I recall correctly, I think we were going to roll out Vulcan slowly since we weren't sure how it would handle the load for all projects?

Yea, neither of these Qs are really something I expected to answer on my own / get answered via random comments in the PR lol.  I should bring this Q up on Friday's planning meeting!

dtm2451:
> Hm, regarding Poetry install, is that an issue only in development, or does that affect the production images, too? I haven't played around with Magby, so not familiar with the workflow.

I... am not sure.  Certainly an inssue for development, but Idk about production.  @corps made a comment recently about how our deploys go pretty heavy on rebuilding caches, but I'm not sure if that would include rebuilding the poetry cache.

The expected workflow: We have Magby loading into archimedes from files via poetry, and so if we made a change to a .py script in Magby, that change should be incorporated in the next archimedes build.  I trigger archimedes builds in dev with `make bash -C vulcan`.  Should be simply that.  But: even though poetry clearly notices the file change, and re-installs all packages from scratch, a cache is not updated and previous magby code is what ends up in error messages =/.

corps:
The poetry install issue right now is only an issue in development.  Essentially, when poetry "installs" a path dependency for the first time, it _copies_ that dependency into the virtual env project directory, and refuses to release it without uninstalling all over again.

Poetry acknowledges this issue but there is as of yet a general solution to "let me develop against the source code of a dependency" option that works well.

dtm2451:
When I do merge this in, (and also any future PR involving magby updates, until a valid fixes gets into poetry)...

I can ping with the code to 1) either clear this poetry cache to force a proper build in our dev environments, or 2) make a soft link that skirts the issue, as Zach helped me to set up.

Can also put that in PR headers so we have it when testing!

dtm2451:
Seems open Qs mentioned from the onset are satisfied for the purposes of this PR.

I'll see about polishing the visualization vignettes and bit or any other small bits, then will aim to release!

PR #696, opened 2021-12-16
add documentation about new query params

coleshaw:
Small PR to add documentation about new `/query` params for Magma.

### DIFF
--- a/mountetna.github.io/magma.md
+++ b/mountetna.github.io/magma.md
@@ -323,9 +323,9 @@ The Magma Query API lets you pull data out of Magma through an expressive query 
 - `project_name` - the project key
 - `query` - An array of query predicates (see below for details)
 - `format` - Optional. If `tsv` is provided, a TSV is returned instead of a JSON paylod. By default, a JSON payload is provided.
-  - `user_columns` - If `format=tsv`, can provide an array of string values that act as column labels. The length of this array must match the number of requested columns.
-  - `expand_matrices` - If `format=tsv`, boolean value to expand matrix attributes into individual columns, with one matrix data point per column.
-  - `transpose` - If `format=tsv`, boolean value to transpose the resulting TSV. This means that rows become columns, and vice versa.
+- `user_columns` - Ignored unless `format=tsv`. Can provide an array of string values that act as column labels. The length of this array must match the number of requested columns.
+- `expand_matrices` - Ignored unless `format=tsv`. Boolean value to expand matrix attributes into individual columns, with one matrix data point per column.
+- `transpose` - Ignored unless `format=tsv`. Boolean value to transpose the resulting TSV. This means that rows become columns, and vice versa.
 
 A general form of the query is:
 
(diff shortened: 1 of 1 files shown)


dtm2451:
Very helpful for me re clients updates and Vulcan-Viz =).

Relatedly, but separate from this PR perhaps:
- I wonder if the column names array could be added to the Query page below where the queryTerms are currently displayed?
- I should add primary inputs to the Viz workflows for `user_columns` string, `expand_matrices` boolean, and perhaps `transpose` boolean.) Perhaps within #689
- And for when we build a better path toward Query -> Vulcan Workflow jumpstarts, we'll want to fill in all these inputs, not just the queryTerms!

mountetna.github.io/magma.md
```diff
@@ -322,8 +322,12 @@ The Magma Query API lets you pull data out of Magma through an expressive query
 
 - `project_name` - the project key
 - `query` - An array of query predicates (see below for details)
+- `format` - Optional. If `tsv` is provided, a TSV is returned instead of a JSON paylod. By default, a JSON payload is provided.
+  - `user_columns` - If `format=tsv`, can provide an array of string values that act as column labels. The length of this array must match the number of requested columns.
+  - `expand_matrices` - If `format=tsv`, boolean value to expand matrix attributes into individual columns, with one matrix data point per column.
+  - `transpose` - If `format=tsv`, boolean value to transpose the resulting TSV. This means that rows become columns, and vice versa.
```
dtm2451, line 328: I wonder if the extra indentation here might be confusing to users about whether these inputs are provided the same way as other inputs?  Perhaps instead of indenting over and the text "If `format=tsv`", could keep same indentation and change text to "Ignored unless `format="tsv"`".  Or keep the extra indents but *explain the indent* via a title level "Additional inputs ignored unless `format="tsv"` + then the requisite extra indent level for these inputs.

dtm2451 (commented):
Perhaps an code example using all/some of these new inputs too?

coleshaw:
> I wonder if the column names array could be added to the Query page below where the queryTerms are currently displayed?

Sorry, I'm not sure what you mean by this? The array is editable currently as individual inputs in each "Column" row, but do you mean like the columns in the preview data frame should be editable, instead?

dtm2451:
> > I wonder if the column names array could be added to the Query page below where the queryTerms are currently displayed?
> 
> Sorry, I'm not sure what you mean by this? The array is editable currently as individual inputs in each "Column" row, but do you mean like the columns in the preview data frame should be editable, instead?

Nope, I mean something that would be very similar to the current queryTerms display.  Perhaps another code-chunk-looking bit below the queryTerms one (and likely also add labels for both). Like the queryTerms display, this would not be edit-able, just presented.  End goal being the exact same I think: In addition to the queried data itself, to presenting the intermediate information of what went/would go into the `magma/query`.

Use case: Like how I already hijacked the queryTerms display to get the Vulcan Viz queries started, such a display could be copy&pasted/etc. to inform the same queries, where requested column names would be carried over as well!

dtm2451:
*Anywhere I've said queryTerms, that's an artifact of a name-change in magmaR that I'd like to revert.  Refers actually the *`query` input!

coleshaw:
Oh, I see, just a read-only version of the `user_columns` that will get sent to the server! Thanks for the clarification!

PR #697, opened 2021-12-17
separate output logs endpoint from etl_config

graft:
A little quality-of-life one just separating out a logging (output) endpoint for the Polyphemus ETL config UI. In the absence, some processes generate large logs which results in a very large loading time for the config list. This splits to loading in the background (though all the data is still loaded at once, not when you click through to individual panes), hopefully making this easier.

Not done here is some indication that Polyphemus is loading during this process. There is no throbber for Polyphemus, and I wish there were a more generic Etna interface to track the getting of data (possibly hooking into json_get, etc.).


coleshaw (approved):
I agree that some indication that a giant log string is loading would be nice, but at least this breaks up the data loading...

PR #698, opened 2021-12-20
Metis cache image thumbnails

coleshaw:
This PR removes the old Metis `assimilate` code, and adds in some thumbnail generation and caching via `libvips`. I will set up the thumbnail generation to run every minute like the `archive` and `checksum` jobs. Also adds in the thumbnail preview to the Timur UI component, with a simple "zoom" button to go between a small (240px wide) image and a medium-sized image (640px wide).

Additional actions:
* Update chef role with `thumbnail_path` param for Metis in `config.yml`.
* Update chef recipe for Metis to schedule the `generate_thumbnails` job.


graft:
One thing I might worry about is the startup cost of running the thumbnailer across the few-hundred-thousand odd files we already have. Do you think, before this is run for the first time, we could somehow prepopulate `use_thumbnail: false` for at least a large proportion of files?

coleshaw:
Hm, so we could write a "set has_thumbnail" command to detect mimetypes of each data_block and set the `has_thumbnail` flag accordingly...perhaps I could set that up as a separate command and run that before scheduling the `generate_thumbnails` job. Would that be okay? That could also just run at a more frequent interval than generate_thumbnails, since I'm assuming most files aren't images (i.e. set has_thumbnail runs every minute, but generate_thumbnail runs every 10 minutes).

coleshaw:
One issue with trying to have a different command just to set the flag is that we would have to hardcore or determine the image mime types we want to thumbnail…

coleshaw:
Could you clarify what kind of startup cost you're concerned about? At 11 data blocks per minute (roughly...assuming most will be no-thumbnail and rarely we'll need to generate a thumbnail, which takes ~22s for a 1.6GB file, <1s for small files), all the blocks would be processed in a couple weeks. I don't think that's too bad as far as a time startup cost...

coleshaw:
I've added a command that we can manually run that will flag non-image blocks based on mimetype. It does rely on a hardcoded set of types, though (jpeg, png, tiff, and svg).

graft:
I guess I wanted to avoid waiting weeks for the initial calculation, but i suppose you are right, seems harmless enough. I was imagining a less exact filter to debulk the problem (like exclude all files less than 1kb), but it is probably not necessary as you say.

coleshaw:
Hm, a different issue is that `libvips` doesn't recognize czi images. I will see if there is some way to generate thumbnails for those...

coleshaw:
Perhaps a different (quicker / more efficient) approach would be for the command to only inspect "image" attribute types and then drill down to those data blocks?

graft:
Those are 3d image stacks, less straightforward what a "thumbnail" is in that context. I think we'll need some sort of other viewer if we want to look at those. Doesn't seem to be much javascript CZI support available, however.

coleshaw:
You can convert them to TIFFs: https://allencellmodeling.github.io/aicsimageio/

Not sure which stack they extract, but it does seem to generate a viewable image locally.

corps:
To clarify -- is the thumbnail path supposed to be a permanent data mount?  Should this be part of the metis nfs data mount?  is this just a local mount?  How big do we expect the data to get?  Just curious.

coleshaw:
Hm, good questions. I was planning to map it to the NFS mount, but not have it backed up to tape, since we can regenerate them if needed. Images are a small subset of files, and individual thumbnails aren't that big, so I don't expect it to take up a significant amount of size on the mount. Those could be wrong assumptions....

corps:
> Hm, good questions. I was planning to map it to the NFS mount, but not have it backed up to tape (like uploads), since we can regenerate them if needed. Images are a small subset of files, and individual thumbnails aren't that big, so I don't expect it to take up a significant amount of size on the mount. Perhaps in the range of MBs, accumulated across all data blocks. Those could be wrong assumptions....

I think if we discover they are cheap to regenerate, we can honestly just use a docker volume with a name.  In this manner it works more like a 'boring cache' that is mostly available, and we're not bothering with the overhead or storage in the nfs.  I'd just be curious to measure the net size after things run.

It might honestly be a super easys, super cheap operation and if that's the case we don't have to over think it.

coleshaw:
Sure, I can map them to a named docker volume. The "expensive" part of it, it seems, is determining what is an image. `ruby-vips` doesn't seem to include a list of mimetypes that it supports, so to avoid manually crafting / maintaining the list, I'm testing every single data_block. We could make the thumbnail generation cheaper by assuming a list of mimetypes and storing that for each data_block when it's created, or using some other type of heuristic like Saurabh suggested (files above a certain size).

PR #699, opened 2021-12-22
Production deploy

coleshaw:
I think this deploy:

* Splits out polyphemus logs
* Adds in Metis thumbnails
* Some portainer fixes (?)

Seeing if I can start the thumbnail generation before holidays...


corps:
This is safe.  In general the portainer yml files and scripts **do not deploy automatically** from git.  They are manually applied.

Images = auto deployed
Infra Yml = manually deployed

corps (approved):

PR #704, opened 2022-01-03
Metis move folder new bucket UI

coleshaw:
This PR adds a UI dialog to the folder controls in Metis, to let users move a folder to a different bucket. Though the API is the same as `rename folder`, I added a new UI dialog because there is an additional input text field for the `bucket_name`. Opening up this control in the UI because I was asked to move a folder before break, and so it seems like it may be useful for general users to do as well.


graft:
Looks good to me. A few notes:
- This also allows moving the folder within the bucket, so it should perhaps simply say "Move folder".
- In the future I might wish for a little more constraint in the bucket and folder selection (e.g., dropdown for bucket + some kind of typeahead for folder); it seems we are in want of a generic "select bucket + folder" widget

graft:
Hmm, a few issues:
- The current view is not updated when the operation succeeds (i.e., the moved folder remains until I refresh)
- You are specifying the final destination folder path, not the destination (parent) folder. This is not obvious
- Allowing the former also allows renaming the final folder_path, again not obvious

In the shell move + rename are both handled by `mv`, but in most UIs I think "move folder" and "rename folder" are decoupled, so it might make sense to specify only the destination (parent) folder (although this would create some confusion for empty string => root folder destination).

coleshaw:
> The current view is not updated when the operation succeeds (i.e., the moved folder remains until I refresh)

Hm, I don't see that behavior when I test. Your FE container is refreshed and running the latest code? Let me try again...

> You are specifying the final destination folder path, not the destination (parent) folder. This is not obvious
> Allowing the former also allows renaming the final folder_path, again not obvious

This is the exact same behavior as "Rename folder". Perhaps there is better wording that could be applied to both dialog prompts?

> This also allows moving the folder within the bucket, so it should perhaps simply say "Move folder".

Hm, good point. Does this make "Rename folder" redundant? Or it's okay to leave that in since it saves an input box?

graft:
Hmm, "Rename Folder" does: `new_folder_path: filePath(current_folder, new_folder_name)`, meaning it does not actually let you change the path of the folder, only its name in its current location (although the endpoint allows you to change the path, and now the bucket as well). I would say that "Move Folder" (without renaming) and "Rename folder" (without moving) are the two actions we should include in the UI, consistent with other file/folder UIs.

coleshaw:
With rename folder, you can actually move the folder, too, if you include a path via an existing folder in the `new_folder_name`...i.e. to rename `folder/` and move it at the same time with the UI "Rename folder" widget, you can specify a `new_folder_name` like `existing_sibling/new_folder`. It is buggy in the UI, and it does only work into sibling folders, but "Rename" does allow you to do a limited move in the UI, it's not move-less. Note the UI issue is similar to the issue you raise dabove, where it doesn't disappear from the view until you refresh, even though it has been moved.

So should "Rename folder" be stripped of it's "move" capability, if you want to keep "Move folder" as rename-less? Or we could also consolidate, like the command-line client, and pre-populate bucket / folder_path text inputs?

coleshaw:
> consistent with other file/folder UIs.

Note that you can also "move" files via the "Rename file" UI widget by providing a path, like I mentioned above for "renaming" folders (i.e. `test.txt` -> `sibling_folder/test2.txt`) ... so right now allowing you to move + rename in a single step is actually consistent with the other UI behaviors.

coleshaw:
> The current view is not updated when the operation succeeds (i.e., the moved folder remains until I refresh)

Also, can you please verify this happens on a fresh `make -C metis restart` and hard refresh of your browser? I can't duplicate the behavior on my end with this branch, and I see the moved folder removed from the view. Silly request, but could you be more explicit in the steps / folder structure that you used, when you saw this behavior? I'll try to duplicate the behavior, but so far I've done the following, with the moved folder always disappearing from the UI:

* root folder in bucket 1 moved to root folder in bucket 2
* root folder in bucket 1 moved to child folder in bucket 2
* child folder in bucket 1 moved to root folder in bucket 2
* child folder in bucket 1 moved to child folder in bucket 2

All without renames (just moves).

coleshaw:
Oh! Perhaps I've figured out the "does not leave the UI" issue -- does this only happen for you when you move within the same bucket?

graft:
Yes, just tested and was about to report exactly this: if i move within the same bucket, the folder is not removed. Actually, looking at the stack, the folder IS removed, but then is added back at its new location. I think the way the `directory` store is written, it expects the store to only contain the contents of the current directory, so if the new folder is added to the directory store, it shows up. I don't understand why giving a different bucket_name makes it invisible though...

graft:
> > 
> 
> Note that you can also "move" files via the "Rename file" UI widget by providing a path, like I mentioned above for "renaming" folders (i.e. `test.txt` -> `sibling_folder/test2.txt`) ... so right now allowing you to move + rename in a single step is actually semi-consistent with the other UI behaviors.

Yeah I would consider this to be a bug. What I mean by "consistent with other UIs" is, e.g., the finder on my machine, Dolphin, which very cleanly separates "move" from "rename", i.e. you cannot "rename" a file into a sibling folder (in fact in Unix since / is a legal filename character it just would name the file `sibling_folder/test2.txt` in place). Metis UI ought to be checking that the folder_name is valid (i.e., excludes `/`) before sending it.

coleshaw:
>  I don't understand why giving a different bucket_name makes it invisible though...

Shouldn't that be the intended behavior? If you move the folder to a different bucket, it is removed from your current view?

coleshaw:
K, so I've changed the behavior to cleanly separate "Rename folder" and "Move folder", with a default bucket for "Move" to be the current bucket. I've also done the same for files (there was no "Move file" before). Should fix the UI-ish bugs around the row not disappearing as well.

PR #706, opened 2022-01-06
Timur query expose user columns

coleshaw:
This PR exposes the `user_columns` values in the query preview CodeMirror pane in Timur Query, to help with migrating the information into Vulcan. Also includes some Vulcan plot fixes from @dtm2451


timur/lib/client/jsx/components/query/query_results.tsx
```diff
@@ -81,6 +83,12 @@ const QueryResults = () => {
     setPage(newPage);
   }
 
+  const userColumnsStr = useMemo(() => {
+    return JSON.stringify(userColumns(columns));
+  }, [columns]);
+
+  const codeMirrorText = `query: ${queryString}\\nuser_columns: ${userColumnsStr}`;
```
dtm2451, line 90: extra `\` here for the `\n` causes it all to render on one line for me.

coleshaw: Oh phooey, thanks, I'll take another look. Without the extra `\` I was seeing funky test behavior. :stuck_out_tongue_closed_eyes:

timur/lib/client/jsx/selectors/query_selector.ts
```diff
@@ -234,3 +234,13 @@ export const queryColumnMatrixHeadings = (column: QueryColumn) => {
...
 };
+
+export const userColumns = (columns: QueryColumn[]) => {
+  let columnLabels = columns.map(
+    ({display_label}: {display_label: string}) => display_label
+  );
+
+  // We need to duplicate the identifier column when renaming,
+  //   since that is provided as the root of the question.answer.
+  return [columnLabels[0], ...columnLabels];
```
dtm2451, line 245: I'm still on the fence about on this duplicated identifier column.  In my testing, the column names do seem to get uniqued (with a ".1" added to one of them), so there shouldn't be downstream issues with say pandas figuring out which column to target.  But I do think users might get confused.

Fully agree with the tendency to introduce as few special cases as necessary.  But I think maybe making the model identifier get added only when no other columns are (yet) requested, as we talked about in slack, is indeed worth doing.  That would square the raw output of the query(..., format="tsv") call for vulcan with what's displayed in the Query page while removing the seemingly unnecessary extra column.

coleshaw: So doing what we discussed in Slack in the Query UI is okay (even with treating the identifier column as special), and I had it working. However, the challenge I started running into, and the reason I backed off, is that you can't actually make a request with no columns to `/query`, and I'd have to think about how to make the TSV itself not have duplicate columns.

For example, if you wanted to get just a list of `rna_seq` tube names in the system, you can't do (though perhaps you should be able to?):

```
["rna_seq","::all",[]]
```

In which case you have to do:

```
["rna_seq","::all",["tube_name"]]
```

But that leads to a TSV with duplicate columns because of the way that the Magma question returns data. So it's not yet clear to me that there's a clean, generic solution to the issue on the server-side to clean up the TSV itself. I feel like it's simpler in the UI because the UI knows what the root identifier is and controls how things are presented and sent to the `/query` API, so it can hide that first column and inject it if needed.

For example, what if someone requests the identifer (using the query API), but not in the "first" column?

```
["rna_seq","::all",[["gene_tpm","::slice",["ENSG00000000003","ENSG00000000005"]], ["tube_name"]]]
```

What is the expected behavior here, for a TSV? The raw query API knows nothing about the intent of the first column, and has to infer it...which I guess we could do, but this is where I don't particularly like all the "first column do something special" logic spreading out...would it be appropriate for Vulcan itself to just take out the first column of a TSV? But Vulcan also can't assume anything about the shape of the data frame...

coleshaw: I think what it comes down to, is that we know this use case of the first column being the identifier is true for the Timur Query UI situation, and I have the logic to make that work. I could migrate that to the server-side, and make it generally available through the TSV. However, what I'm not confident about is that this use-case applies to all users and is generalizable enough that the server should always behave that way. Perhaps you could help me understand if that is true or not? If it is true and this kind of use case should apply to all users of the TSV query endpoint, then I can just migrate that logic to the server-side, to treat the first column uniquely. And if that's not the case, perhaps we can come up with some rules or option params to apply it....but would have to better understand the use cases.

dtm2451: Ahhhh, server-side, I don't think we need to make a change.  My suggestions have all be scoped to the Query-UI and Query-UI -> Vulcan uses.

Here's what I think for server-side:  If a user does have the impetus to understand magma enough to build and adjust their own query inputs from scratch, I don't think we want to make that any more complicated than it already is.  I can imagine a use-case where it might make sense to intersperse the identifier column every 10 columns or so.  Commonly? no, but why not allow it for a simple way to make viewing the data in a simple text editor, rather than needing to open a table-viewer and freeze the first column.  I think this is just for advanced users, however, and probably just very rare instances.  But for these rare use-cases and advanced users who probably have at least some computational intuition, I think it helps a lot for the user's learning to have the server-side algorithm be "clean" and easiest to learn.  Such users would probably build up the intuition that the server adds a root_model ID column and thus understand that if they request that column again, it'd become a duplicated column.  My intuition is to NOT build a query-simplifier system on the server-side.

So **tl;dr** I think no to auto-modification of the `query` input on the server side, but yes for auto-adjusting what would get passed to this input in the Query-page's backend.

coleshaw: > So tl;dr I think no to auto-modification of the query input on the server side, but yes for auto-adjusting what would get passed to this input in the Query-page's backend.

Unfortunately I don't think this is possible without touching the server...?

Let's assume we've made the changes we discussed yesterday, and then say the user picks no additional columns, and just wants all the `rna_seq` tube names. If we do what we discussed yesterday, that would result in a query like:

```
["rna_seq","::all",[]]
```

Because we don't pass along the "tube_name" column to `/query`, as we discussed. But that's an illegal query and will throw an exception from the server. So we're stuck doing

```
["rna_seq","::all",["tube_name"]]
```

Without modifying server-side code, this results in a TSV with two identical columns.

If, instead, the user picks additional columns in the Query UI, then we have no problem -- we can do what we want!

```
["rna_seq","::all",[["gene_tpm","::slice",["ENSG00000000003","ENSG00000000005"]]]]
```

Is a valid query, and will result in a TSV with no duplicate columns, and can be presented nicely in the Timur Query UI as well as be not confusing in Vulcan.

So...the issue is really the first scenario, where someone just fetches the list of identifiers for a model.

It might be easier to explain via pairing, and happy to do so if that helps.

etna/packages/etna-js/actions/magma_actions.js
```diff
@@ -360,7 +360,7 @@ export const requestAnswer = (question, callback = null) => dispatch => {
       else return response;
   }).catch( error => {
     console.error(error);
-    throw new Error(error);
+    return Promise.reject(error);
```
corps, line 363: Minor nit, but, is there a reason we're converting these throws into Promise.rejects?
As far as I know, throwing an error in promise catch handle is equivalent to returning a Promise.reject, _except for the timing and scheduling_.

coleshaw: It does not seem to behave the same, though perhaps it's because I'm using `invoke` incorrectly. If this is not a promise and instead throws an `Error` as the original, then any error thrown causes an exception [here](https://github.com/mountetna/monoetna/blob/master/timur/lib/client/jsx/components/query/query_use_results_actions.tsx#L45). 

```
Error: [object Promise]
    requestAnswer webpack:///./node_modules/etna-js/actions/magma_actions.js?:339
    promise callback*requestAnswer/< webpack:///./node_modules/etna-js/actions/magma_actions.js?:337
    useActionInvoker webpack:///./node_modules/etna-js/hooks/useActionInvoker.js?:22
    runQuery webpack:///./lib/client/jsx/components/query/query_use_results_actions.tsx?:44
    promise callback*useResultsActions/runQuery< webpack:///./lib/client/jsx/components/query/query_use_results_actions.tsx?:42
```

Is there some other way to have the error bubble up to the `catch` block, in that Promise chain?

coleshaw: (note that the effect of the error being thrown instead of a promise reject, is a white-screen-of-death in the browser).

corps: Hmm... when you say causes an exception, do you mean the web console merely displays the error traceback?
Here's the thing -- your line that you are pointing to, when it calls invoke, it passes through the dispatch which allows that inner call to be invoked, but the catch block there cannot possibly be executing sychronously with the invoke call.  The catch block is called on the promise, providing a callback -- but promise callbacks are never invoked synchronously, they are invoked on next tick according to the spec.

That's what is confusing to me.  The only real difference should just be that Promise.reject also pushes the rejection further away in the scheduling.

regardless of wether a catch block throws a new exception or returns a Promise.reject, it should cascade to the next catch block or cause a console error if no more catch handlers exist.  Something, seems super fishy here, hmm.

corps: Actually, there is one difference -- is there a reason you have wraped `error` inside of `new Error(error)`?

You could just use `throw error` here.  In fact, the parameter of a catch callback is equivalent to the value that was originally `throw X`'d.  In this case, your line is simply wrapping the error in more error.

corps: Oh, that must be what's going on.

Try changing this line to just `throw error` with no surrounding `new Error(...)` and let me know if that works.

timur/lib/client/jsx/selectors/query_selector.ts
dtm2451: I think I understand!  So lemme re-summarize to be sure... The question is basically for the case where no extra columns are requested, Query-UI adds the identifier in order to send a valid request.  That results in a TSV with two identical columns.  We could fix this on the server-side by either allowing 1) `["rna_seq","::all",[]]` to return a single column tsv of the identifiers, or 2) by having the server-side note that tube_name of `["rna_seq","::all",["tube_name"]]` is the ID and thus to skip it because those will be added anyway.  But there's not necessarily an ideal way of achieving this 2).

Correct?

With our main goal being to square the table reported in the Query-UI with what vulcan would see **for making a visualization**, I think the id-only use case is actually not important.  A single vector of unique strings (duplicated columns or not) is not enough input to make any of our planned visualizations!

So we don't need to jump through any hoops to homogenize those outputs =).

Maybe maybe maybe we may build some path for QC'ing IDs in the future through vulcan, but we can view that as a distinct path which would have a separate workflow definition, and that workflow could have a de-duplication method built in on the python side.

etna/packages/etna-js/actions/magma_actions.js
coleshaw: Ah, nice, thanks! Yes, that behaves as expected!

timur/lib/client/jsx/selectors/query_selector.ts
coleshaw: Yes, that's a good summary. So basically I can reimplement what I did yesterday on the UI-side, but we'll assume that no one will try putting an ids-only query into Vulcan, and hopefully that's a valid assumption for the most part (certainly a good assumption for visualizations).

graft: > For example, if you wanted to get just a list of `rna_seq` tube names in the system, you can't do (though perhaps you should be able to?):
> 
> ```
> ["rna_seq","::all",[]]
> ```

You can do:
`[ 'rna_seq', '::all', '::identifier' ]` which will get you an answer of `[ [ identifier, identifier ], ... ]`

coleshaw: Yes, but isn't that the same problem, where a resulting TSV will have two, duplicate columns...?

coleshaw:
@dtm2451 , I've updated to tweak the TSV as discussed. I think it works -- can you give it another look when you have a chance?

dtm2451 (approved):
works and the column number / names match what I see in the Query page when I copy the query inputs over directly =).  Tested a few different setups.

PR #709, opened 2022-01-07
Vulcan Viz 'finalization': documentation updates and minor tweaks for current workflows

dtm2451:
Just some tweaks to metadata.json files of our Vulcan workflows.


dtm2451:
Also removing the previously exposed 'transpose' input as the format returned by this transposition is not immediately visualization-ready.  (The ID column remains, so even for the one use-case I imagine right now where transposing might be useful, == scatterplot of gene expression data, or another matrix type attribute, where x-axis would be 1 sample and y-axis would be a 2nd sample, this ID string ensures that the data will be treated as discrete rather than numeric.)

dtm2451:
@coleshaw would love a check of the vignette text and also the (?)-docs for the primary inputs!  Notes on anything else you might notice are welcome too, of course.  The text is the same across all three viz workflows everywhere except in the "Generating the plot" sections of the vignettes.

Note: If you don't have the example project in your dev environment, add "ipi" to the projects line of the <workflow>.metadata.json files in order to view via ipi.

dtm2451:
After this PR, I think we could release  =)!  Or perhaps even within this PR... we'd just need to add all the projects that we plan to target to the projects line of the <workflow>.metadata.json files.  But I want to wait til after we can discuss that together in a standup to do make that call.

vulcan/lib/server/workflows/bar-plot.cwl
```diff
@@ -9,18 +9,13 @@ inputs:
   1_Data_Source_magma_query__user_columns:
     type: string
     label: "user_columns input"
-    doc: "Optional array or comma-separated set of *quoted* strings for renaming columns of the returned data.  *Coming Soon* Suggestion: Use the Query page of Timur to build the proper dataframe, and then simply copy over the 'user_columns'-chunk presented there!"
+    doc: "Optional array or comma-separated set of *quoted* strings for renaming columns of the returned data.  Suggestion: Use the Query page of Timur to build the proper dataframe, and then simply copy over the 'user_columns'-chunk presented there!"
```
coleshaw, line 12: nit / question: what is an "array or comma-separated set"? You're saying the input will take either:

`"bar", "foo", "bim"` (comma-separated set)

or ...? What is an example of array?

dtm2451: oooh good point.  It was meant to get at that you can give either:
- `"bar", "foo", "bim"`
- `["bar", "foo", "bim"]`

Lemme update to make that more clear.

vulcan/lib/server/workflows/bar-plot.metadata.json
```diff
@@ -1,8 +1,8 @@
...
...an use Timur-Query's own documentation to learn how to set up that page to query the data you want. After setting up your query targets there: \n\n- 'query input': Copy here the 'query:' portion of the Query page's yellow box.\n- 'user_columns input': Copy here the 'user_columns:' portion of Query page's yellow box.\n- 'expand_matrices input': [Recommendation = leave as is.] This input defaults to True both here and in the Query page (where it exists as a slider button below the yellow box). Normally, you will want to leave it True as the non-expanded format is less accessible to the visualization machinery here.\n\n## Generating the plot\n\nAfter data is obtained, a new user input widget will show up in which you can select which columns to use for the x axis groupings, the data whose composition within each group should summarized in the y axis, and whether to scale the resulting bars by percentage among each group ('fraction') or absolute counts ('counts'), as well as other advanced options. NOTE: The x and y axis data for a bar plot must be a discrete (not continuous) data type.\n\n#### We plan to make the Timur-Query -> Vulcan workflow transfer process simpler in the future!",
```
coleshaw, line 5: I would probably call the query and user_columns box "green" rather than yellow...but understandably that may differ by viewer...we should probably label that section in Query, to make it more accessible.

coleshaw: `each group should summarized` -> `each group should be summarized`?

coleshaw:
Looks good!

vulcan/lib/server/workflows/bar-plot.cwl
dtm2451: I removed the array term.  I have a sense that it doesn't matter to say at all...?  Do you agree?

When the user copies from Query, they'll have the brackets, which is fine.  When a user creates the input here manually, they don't need to add the [] because it will be added in the script when needed, prior to the query call.

vulcan/lib/server/workflows/bar-plot.metadata.json
dtm2451: lol I should have checked the color with someone "color-typical"!

Made these fixes.

PR #711, opened 2022-01-10
Resource projects

coleshaw:
This PR introduces resource projects...I think it works as intended, where any project flagged as a `resource` is available to any user in the system. Within Metis, an additional check is performed so that only `viewer` buckets in a given resource project are visible, when a user does not have explicit permissions for the project. 

* UI in Janus admin to set state of a project as a resource or not.
* Updates to `Etna::User` to allow `can_view?` on any resource project. Pings Janus if user has no explicit permissions.
* Updates to `Metis::Bucket` to ping Janus to see if a user can view a bucket, for resource projects. Only allows users to view viewer access buckets for resource projects, if user does not have explicit permissions.


graft:
First, :exploding_head: 

Second - I'm confused why it is necessary to make an additional check to view a metis bucket? If they already can_view?, shouldn't that suffice?

coleshaw:
I guess it depends if you want a viewer to be able to view "administrator" buckets in a resource project (for example)?

etna/lib/etna/clients/janus/models.rb
```diff
@@ -69,6 +69,18 @@ def html
         end
       end
 
+      class JsonResponse
```
corps, line 72: Hmm, so we're not filling out some of the attributes a bit more?
One of the advantages of having the explicit attribute expansion was catching simple bugs and attribute renaming / defaults logics in the face of the api shape changing over time.  It also meant that there was a semi documentation on what sorts of things were available consistently in the response. Using a hash, we frequently have nil checks and that's about it.

coleshaw:
i.e. IPI becomes a resource project, does that mean that **all** data in Metis is viewable by everyone, or just "viewer" buckets?

etna/lib/etna/clients/janus/models.rb
coleshaw: Oh, good point! Sorry, I forgot -- I can update that.

janus/lib/client/jsx/janus-admin.jsx
```diff
@@ -45,58 +109,118 @@ const Projects = ({projects}) => (
...
-          )
-        }/>
+          />
+        </div>
+        <div className='cell submit'>
+          <Icon
+            className='approve'
+            icon='magic'
+            onClick={() =>
+              postAddProject(newproject)
```
corps, line 143: Hmm...
So firstly, it seems you can easily multi submit here right?  And in that case, which error or not error you see becomes a race condition dependent on network conditions.

This gets even weirder because of the setNewProject({}) which can actually clear project details after the button is clicked at a non deterministic time.

Again, I will always nit the issue of double or multiple submit in forms like this :P.  One alternative is to add additional state, "loading" which only allows one request to be submitted at a time.  There is incidentally already a tool for this in `useAsyncWork` which handles both cancellability and loading state.

graft:
> i.e. IPI becomes a resource project, does that mean that **all** data in Metis is viewable by everyone, or just "viewer" buckets?

I would say the behavior should be exactly the same as if Janus had put `v:ipi` on your token, meaning `administrator` and `editor` buckets would remain hidden to everyone except ipi editors and admins. This way we can have a resource project expose data to everyone, but also continue working on some parts of it in some privacy if we wish.

coleshaw:
> I would say the behavior should be exactly the same as if Janus had put v:ipi on your token, meaning administrator and editor buckets would remain hidden to everyone except ipi editors and admins. This way we can have a resource project expose data to everyone, but also continue working on some parts of it in some privacy if we wish.

Yeah, so I think that's the behavior the current PR achieves. The authorization logic regarding that is located in `Metis::Bucket`, so that's why I had to touch that class as well.

janus/lib/client/jsx/janus-admin.jsx
coleshaw: Ah, thanks, I'll try to pay more attention to this in the future...

corps (approved):

graft:
> Yeah, so I think that's the behavior the current PR achieves. The authorization logic regarding that is located in `Metis::Bucket`, so that's why I had to touch that class as well.

That is the behavior, however I am concerned that this method is brittle, and it only gives an ersatz identical behavior with the "viewer" role. I am thinking rather than having a method in Etna::User that phones Janus, the janus-phoning should happen in Etna::Auth and the resulting `v:resource_project` permission should be patched into Etna::User for the request, which, then arriving in the app, would clear can_view? and Metis::Bucket queries without having to make special checks, which seems more predictable.

coleshaw:
Ah, you mean like in [auth.rb](https://github.com/mountetna/monoetna/blob/master/etna/lib/etna/auth.rb#L96-L110), roughly in:

```
    def approve_user(request)
      token = request.cookies[application.config(:token_name)] || auth(request, :etna)

      return false unless token

      begin
        payload, header = application.sign.jwt_decode(token)

        return false unless janus_approved?(payload, token, request)
        return request.env['etna.user'] = Etna::User.new(payload.map{|k,v| [k.to_sym, v]}.to_h, token)
      rescue
        # bail out if anything goes wrong
        return false
      end
    end
```

? I started there, but I think the cleanest way to achieve what you're saying, then, is to include all the resource projects and not just the requested one. If I recall correctly, I think the issue I ran into was with the root project ERB views, like https://metis.ucsf.edu/ipi. In those cases, you can't easily detect the `project_name` value from within `auth.rb`, since it's only in the request environment as the `PATH_INFO` value (not injected into in `rack.request.params`). I suppose you could detect the project name from `PATH_INFO` as an edge case when `:project_name` isn't populated in the params but is defined in the route, but that seems brittle in its own way...

graft:
> Ah, you mean like in [auth.rb](https://github.com/mountetna/monoetna/blob/master/etna/lib/etna/auth.rb#L96-L110), roughly in:
> 
> ```
>     def approve_user(request)
>       token = request.cookies[application.config(:token_name)] || auth(request, :etna)
> 
>       return false unless token
> 
>       begin
>         payload, header = application.sign.jwt_decode(token)
> 
>         return false unless janus_approved?(payload, token, request)
>         return request.env['etna.user'] = Etna::User.new(payload.map{|k,v| [k.to_sym, v]}.to_h, token)
>       rescue
>         # bail out if anything goes wrong
>         return false
>       end
>     end
> ```
> 
> ? I started there, but I think the cleanest way to achieve what you're saying, then, is to include all the resource projects and not just the requested one. If I recall correctly, I think the issue I ran into was with the root project ERB views, like https://metis.ucsf.edu/ipi. In those cases, you can't easily detect the `project_name` value from within `auth.rb`, since it's only in the request environment as the `PATH_INFO` value (not injected into in `rack.request.params`). I suppose you could detect the project name from `PATH_INFO` as an edge case when `:project_name` isn't populated in the params but is defined in the route, but that seems brittle in its own way...

I see, that makes sense... You don't know what project is being requested until the route level at earliest.

coleshaw:
I've moved the logic into `Etna::Auth`, and add in `v` permissions for all resource projects to the environment `User` (so they don't take up space in the cookie).

coleshaw:
Also restricted the resource-project injection to just "view" routes, so that things like Metis `upload` don't overload Janus. I'm still a bit worried that Janus will see such a huge increase in requests that it may overwhelm the service, since Janus will be pinged for every "view" route by every single user across multiple applications....if that does happen, perhaps we can scale out the `janus_app` in the swarm (now that we're almost there!), or perhaps we just inject `v:project_name` into each user's token, as @corps had suggested.

PR #713, opened 2022-01-11
Some small improvements to the way airflow logs are handled.

corps:
Noticed intermitten rpc errors while fetching logs from services started via the airflow system.  Seems this just happens from time to time.

1.  Introduce a retry 'gater' class that adds exponential backoff with max retry.
2. Check for rpc errors in the streamed result and 'refetch' logs once they are hit.
3. Adds tests around this case.


coleshaw (approved):
These would be like, ETL logs that currently we view with `journalctl`?

corps:
These logs will flow throw airflow and be available via that interface.  Technically, they also are available as journctl logs on the host in which they run, but this process is consuming those logs back to the airflow scheduler that started the job so as to centralize it.

coleshaw:
:+1: Thanks for the clarification!

PR #714, opened 2022-01-11
Vulcan Viz - tweaks

dtm2451:
Goal here is mostly adding "polish":
- Allows scatter plots to be made without specifying a data column to color_by. Thus, now only X and Y are needed.
- Turns off alphabetization of the queried-DF's column names in x_by, y_by, and color_by dropdowns.  This lets the ordering of column names presented in these dropdowns match better to the order that column names would need to be given to the upstream 'user_columns input'.  (Useful for the usage path of 1) user only fills in the 'query input', perhaps manually instead of starting with Timur-Query, and performs their data query by hitting `Run`, then 2) user looks to the dropdowns of the 'Set Plot Options' widget to inform them of the columns returned, in order to then 3) go back to the 'user_columns input' to rename their columns and re-run their query.)


dtm2451:
- Adds (?)-documentation for the plot settings widgets.

dtm2451:
- adds previously missing python function documentation to the y_plotly and bar_plotly functions in archimedes.

PR #717, opened 2022-01-13
Polyphemus hook ensure containing record

coleshaw:
When investigating the DN1 `rna_seq` timepoints, I realized that for MVIR1 `clinical_lab` records, if the timepoint does not exist, it will get created, but as a disconnected timepoint (verified locally). Furthermore, the REDCap `timepoint` loader only creates timepoint records when REDCap has timepoint model data, and does not create any where there is only clinical_lab data (can be easily verified by looking at Polyphemus logs).

I'm actually not sure how all the clinical_lab timepoints that currently exist in Magma for MVIR1 were created (i.e. MVIR1-HS1-D2). Looking at my script from May 2021 from when I was asked to load clinical_lab data, I only updated clinical_lab records, but did not create any timepoints. So some other process must have created those timepoint records beforehand.

This PR proposes to introduce a hook into the loader, that lets models inject parent records that might be required to attach things to the tree.


corps:
This is kind of increasingly why I think the way recording data into magma should change.  We'll be forced to constantly hunt down and essentially "synchronize" the creation of related data due to the push nature of the system.  We'll have to consistently "hope" we got it right and there isn't missed data that didn't flow through the system.

In the magma datums approach, we would always record any data claims and claimed relationships as is, *without reading from magma itself to make a decision*.  That is to say, the state of existing magma data, or even magma's model itself, should not be an input to any etl creating new magma data.

Now, how we process and determine the integrated view is something we program as a function of the data claims after the fact.  That is the shift that would, I think, make our system more sound.  I especially think this to be true the more the external sources of data flow through the system (I suspect over time external to timur will be increasingly the norm).

coleshaw:
While I agree with you in principle, I'm not clear how datums would help here? There is apparently no "source of truth" for parent-timepoint relationships (or even "timepoints")... where would the datum pull them from, in that case? Do datums still assume that the relationship is defined somewhere?

Or, in this example, would the datum model just say "analyze all timepoints in the system, and if they match the patient regex, link them together"?

coleshaw:
To clarify my understanding of the current situation, is that there currently a missing, claimed relationship...the loader defines:

clinical_lab -> timepoint

But we're missing the relationship:

timepoint -> patient

corps:
> While I agree with you in principle, I'm not clear how datums would help here? There is apparently no "source of truth" for parent-timepoint relationships (or even "timepoints")... where would the datum pull them from, in that case? Do datums still assume that the relationship is defined somewhere?
> 
> Or, in this example, would the datum model just say "analyze all timepoints in the system, and if they match the patient regex, link them together"?

In this case, you would merely write a datum that contains everything known for any timepoint provided by redcap.
The relationship between timepoints and patients is not *recorded* here at all.  That is determined, at read time, based on what magma determines is the source of truth for names and what claimed datums exist.

IE:  Again, aggressively record what other systems provide us.  Determine the relationships and integrity after the fact.

corps:
For instance,

the 'source of truth' could be code in magma's READ side, that performs logic that matches on names.

coleshaw:
Oh interesting, thanks for clarifying! Hm, so would you even then not record the clinical_lab -> timepoint relationship? Just record all records independently (clinical_lab, timepoint, etc.), with no relationships... neat.

coleshaw:
So basically this would be like, making a bunch of tables without FK relationships, and generating the views / "FK" relationships based on some algorithm or other priority...

corps:
Absolutely!  And it goes deeper -- the read view is no longer simply a reflection of the claims.  we can have derived attributes, we can resolve conflicts between claimed attributes, we can swap attribute values by changing which source of truth's claims are used or not used.  Etc.

Imagine that the read side of magma was a reducer, whose inputs is a stream of 'claims', 'datums', 'events'.  And you can reconstruct the state of the tree at any time based on these arbitrary data pieces by executing our function, or even modifying or function.

This as opposed to a stateful file that must correctly apply deltas in the right order to achieve the same effect.

coleshaw:
Hm...interesting. Requires some deep thought to fully grok it. Would the "reducer" be kept in some sort of version control? Is that Postgres views / stored procedures, or something that would exist in Magma / Ruby code?

corps:
A performant implementation would probably use a real table and apply deltas anytime the 'stream' of inputs is merely appended new inputs.  Anytime the stream changes (the source of truth swaps, or there is updated reducer code), a new view is generated and swapped in until it becomes a proper table.

There is actually several cool psql features that can be leveraged, and I'd say we should test several options out to check performance in the end.  This kind of architecture is not as radical as it sounds -- event stream based architectures are common, and they can be built on a database (or rather, psql has tons of cool features that can help us here).

coleshaw:
Cool, thanks this helps clarify in my mind the datums idea! I will have to do more reading...

coleshaw:
@graft , thoughts on this approach?

coleshaw:
Reviewed with @graft , but I think all the changes that we discussed will be in the project-specific model definition files...

PR #719, opened 2022-01-15
Git Checkout Hook for airflow

corps:


corps:
Includes some airflow fixes,
refactoring to make mocker library available for all packages,
and Git Hook that will make it possible to sync git repositories into the airflow executor disk on a job.

coleshaw (approved):

PR #720, opened 2022-01-18
Metis_client nuke

graft:
This adds a "nuke" command to metis_client, which removes all files and empty folders in a local directory, EXCEPT those files that are not present on Metis (verified by MD5). That is, it is a "safe" removal. Unresolved data is left in place. It relies on the presence of md5sum; md5 computation is not cached.

It also adds an "exists" endpoint to Metis, which is fairly open and lets you check if an arbitrary list of md5s exists in Metis' data store. There are no checks, on the assumption that if you know a file's md5 you have it in hand, and the information is weak in any case (mere presence).


corps:
Do you think there is any value to a check that also verifies if the file with the expected md5 exists at the expected file location?  In the case that the file was, say, ingested via the CAT etl but it hasn't been moved to the correct location outside of triage?

graft:
> Do you think there is any value to a check that also verifies if the file with the expected md5 exists at the expected file location? In the case that the file was, say, ingested via the CAT etl but it hasn't been moved to the correct location outside of triage?

Yeah, there ought to be a file list endpoint covering this as well (currently I think we kludge around this by listing the parent folder), but I think that is separate from simply checking that data blocks exist.

corps (approved):

dtm2451:
Love this idea!  I think it'd see immediate usage (by Rebecca for HuMu) and would make Arjun VERY happy with how easy this should make it to remove transferred fastqs and BAMs from c4 =)

coleshaw:
> Do you think there is any value to a check that also verifies if the file with the expected md5 exists at the expected file location? In the case that the file was, say, ingested via the CAT etl but it hasn't been moved to the correct location outside of triage?

Or similarly, would it be confusing if the data block exists, but there is no active File pointing to it? Probably unlikely with more recent uploads, but I'm wondering if someone might delete a local file without knowing where it is located on Metis / able to find it, and then we have to hunt it down for them...

coleshaw (approved):

graft:
> Or similarly, would it be confusing if the data block exists, but there is no active File pointing to it? Probably unlikely with more recent uploads, but I'm wondering if someone might delete a local file without knowing where it is located on Metis / able to find it, and then we have to hunt it down for them...

I could add a third category in the response, `found, missing, trashed` indicating ones missing a reference. In any case they don't know where the file is on Metis, and finding it by md5 isn't possible at the moment, AND the file might exist but they don't have permission to see it, so I think none of this is guaranteeing that they can find the file, just that Metis knows about it.

PR #721, opened 2022-01-18
update python version

coleshaw:
I think the pypi install issue was because the default python version in the container was python 3.6. Looks like libraries like cattrs and celery have removed support for python 3.6, so this PR pins to pythonn 3.7 (which those libraries still support). We could also pin to say, 3.9 if we want to future proof a bit.

### DIFF
--- a/.github/workflows/branches.yml
+++ b/.github/workflows/branches.yml
@@ -27,6 +27,7 @@ jobs:
           - etna
           - archimedes
           - deployer
+          - airflow
       fail-fast: false
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/branches.yml
+++ b/.github/workflows/branches.yml
@@ -28,7 +28,6 @@ jobs:
           - archimedes
           - deployer
           - airflow
-          - swarm/swarm-keeper
       fail-fast: false
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/branches.yml
+++ b/.github/workflows/branches.yml
@@ -28,6 +28,7 @@ jobs:
           - archimedes
           - deployer
           - airflow
+          - swarm/swarm-keeper
       fail-fast: false
     steps:
       - uses: actions/checkout@v2
--- a/airflow/Dockerfile
+++ b/airflow/Dockerfile
@@ -1,7 +1,7 @@
 FROM known_hosts
 FROM bash_mocker
 
-FROM apache/airflow:2.2.2
+FROM apache/airflow:2.2.2-python3.7
 
 # This drops an sqlite database in the base directory which is not used in production,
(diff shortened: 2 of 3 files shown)


corps:
Excellent find!

It's weird.  Locally, this worked for me because the 3.6 python had already cached the dependencies and worked.  Did they remove support for python 3.6 and then, reupload the package?  :/

corps (approved):

dtm2451:
:raised_hands: I noticed some package downgrades in testing some new python libraries last week which are probably related to this, but didn't get around to diagnosing that side yet.  THANKS!

I can run some checks with the scRNAseq workflow for whatever python version we decide.  I think that would amount to validating all of the vulcan functionality that we currently 'guarantee'.

coleshaw:
Yeah, that's kind of weird. Not sure, it does seem like both packages explicitly did not support 3.6 at the versions pinned in requirements.txt...

corps:
> Yeah, that's kind of weird. Not sure, it does seem like both packages explicitly did not support 3.6 at the versions pinned in requirements.txt...

:shrug: 

To be clear, THIS is why I used `poetry` inside of `archimedes`.  Airflow uses `pip` which has notoriously leniency when resolving dependencies and has frequently caused me grief.  I'm not convinced its worth stuffing `poetry` into `airflow` yet, but this is a good example of why I think `poetry` does us a favor (although the lack of good dev package linking is its own hairball T_T)

.github/workflows/branches.yml
```diff
@@ -27,6 +27,8 @@ jobs:
           - etna
           - archimedes
           - deployer
+          - airflow
+          - swarm/swarm-keeper
```
coleshaw, line 31: @corps , I figured I would sneak these in here. Is this correct for swarm-keeper? It seems to work when I run locally, but we don't have any other nested projects so wanted to check. Just `swarm` doesn't work...

graft:
Why not 3.9?

.github/workflows/branches.yml
corps: yeah, it's a good open question.  Let us just run airflow right now and not swarm-keeper.
In production and staging, the swarm images get built and released, but I do worry about having them run for every branch since they honestly aren't updated often.  It's just a question of CI performance vs completeness of branch testing.

corps:
> Why not 3.9?

Archimedes is 3.8, even.

Consistency would probably be ideal.
My, only concern is that while minor release versions in general do not change behavior in most libraries, it certainly DOES change behavior of the package management in python and can be a bit of a pain.  I'm ok changing with an eye for a consistency.

The original 3.6 was the base version shipped with airflow so it wasnt't something I was particular in at the time.

PR #726, opened 2022-01-19
Updating csv modeling and readme

corps:
Followup on model migration.

1. magma development no longer requires checking out the magma repositories
2. removed project_names from development config
3. Updating readme

initialize projects and test labors setup is handled using fixture files.  There is some open work around supporting `column_name` and `dictionary` in the CSV that would further simplify, but that is left for a future PR.


etna/lib/etna/clients/magma/workflows/json_validators.rb
```diff
@@ -172,10 +172,6 @@ def validate_links
           link_attributes.each do |attribute|
             check_key("attribute #{attribute.attribute_name}", attribute.raw, 'link_model_name')
 
-            if attribute.attribute_name != attribute.link_model_name
```
corps, line 175: This _used_ to be an issue, but it is no longer.

etna/lib/etna/clients/magma/workflows/model_synchronization_workflow.rb
```diff
@@ -161,7 +161,7 @@ def ensure_model_link(model_name, link_model_name, attribute_name)
           return if target_attributes.attribute_keys.include?(target_link_model_name)
 
           add_link = AddLinkAction.new
-          add_link.links << AddLinkDefinition.new(model_name: target_model_name, attribute_name: target_link_model_name, type: source_attribute.attribute_type)
+          add_link.links << AddLinkDefinition.new(model_name: target_model_name, attribute_name: attribute_name, type: source_attribute.attribute_type)
```
corps, line 164: This bug duplicated link attributes into two attributes: the intended one, and one named after the link_model.  This is intended behavior captured correctly.

coleshaw:
nit: probably not strictly necessary, but is it also possible then to clean up [initialize-projects.sh](https://github.com/mountetna/monoetna/blob/master/magma/development/initialize-projects#L23)?

coleshaw (approved):

corps:
> nit: probably not strictly necessary, but is it also possible then to clean up [initialize-projects.sh](https://github.com/mountetna/monoetna/blob/master/magma/development/initialize-projects#L23)?

To do that, we'd still need a hook for test to setup the test project.  Yeah, that's kind of the last pin in removing all this code, hmm.

PR #729, opened 2022-01-21
use update_json to make sure dry_run flag is boolean

coleshaw:
This PR fixes a bug with the Polyphemus IPI Flow linker commands that was introduced with the `dry_run` flag (created for REDCap loading). Using the form `update` method causes `dry_run` to always be a string, which means `dry_run=false` is evaluated as true and updates would never execute with these two commands. Ideally we migrate these to ETLs at some point, or IPI flow will finally end...


corps:
I'm working on migrating these yes, but I don't have a timeline yet.

corps (approved):

PR #730, opened 2022-01-24
draggable columns for query

coleshaw:
This PR makes the columns (sans the root model identifier) draggable and re-orderable. I think it should be fine for the number of columns that folks typically use, it would probably bog down at some point with too many columns...but most likely outside of our use case.


dtm2451:
Curiously, I had to `npm add react-beautiful-dnd` to timur even after running `make update`.  I'm trying again now, as I see that it's added in the package.json, to see if that was some weird fluke.  :crossed_fingers: 

Seems to work beautifully!

dtm2451:
Probably I had the containers running while updating so it didn't take, or something like that, the first time.  Worked fine when I was careful lol.

Looks good to me!

coleshaw:
> Probably I had the containers running while updating so it didn't take, or something like that, the first time. Worked fine when I was careful lol.
> 
> Looks good to me!

Oh yeah, I think you're supposed to do `make down` before `make update`...thanks! I can wait until after you merge in your warning PR to merge this one in.

dtm2451:
:+1: merged that one!

PR #735, opened 2022-01-27
Redcap loading in chunks

graft:
This tries to paginate redcap loading into pages of roughly around 20K records, to hopefully allow the redcap client to remain within the bounds of timeouts for magma updates.

Due to the chunking it is possible that there may be validation errors in multiple segments of the upload; however, this will raise out of the first one it encounters, which might lead to some obscuring of a comprehensive error report on the entire loading operation. However, I'm not sure how best to accumulate validation errors to avoid this.

A partial update is also likely in this circumstance if --dry-run is not used. This is just a fact of life to be accepted with this chunking I think.

Somewhat a WIP - needs hand-testing first.

### DIFF
--- a/etna/lib/etna/clients/magma/models.rb
+++ b/etna/lib/etna/clients/magma/models.rb
@@ -215,8 +215,8 @@ module Etna
         end
 
         def model(model_key)
-          return nil unless raw.include?(model_key)
-          Model.new(raw[model_key])
+          return nil unless raw.include?(model_key.to_s)
+          Model.new(raw[model_key.to_s])
         end
 
         def all
@@ -434,8 +434,8 @@ module Etna
         end
 
         def attribute(attribute_key)
-          return nil unless raw.include?(attribute_key)
-          Attribute.new(raw[attribute_key])
+          return nil unless raw.include?(attribute_key.to_s)
+          Attribute.new(raw[attribute_key.to_s])
         end
 
         def build_attribute(key)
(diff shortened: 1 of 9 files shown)


coleshaw:
Hm, does this lose the hook for `ensure_records`? I don't see them appearing for `clinical_lab` when I test locally...

coleshaw:
I'm finding the results in the log kind of confusing ... for example, for `HS1`, I only see clinical labs for `D12`, but on production, I certainly see labs for other timepoints. Do the logs only show the last batch?

graft:
There is a bug merging output from magma i need to fix.

corps:
Yeah, I think long term incremental loading is the way to go.

corps:
FWIW -- if we want all or nothing like semantics for loading, we'd need something atleast _like_ magma datums or atleast a transactional buffer.

A transactional buffer could be:
1.  A separate duplicate table for each model that gets loaded, and "swapped" for the real table atomically (constant time) when committing the 'transactional load'.  This would require some sort of locking, potential an optimistic lock with versioning, that fails when other writes occur to the same models.
2.  One or more 'event buffers' that track non conflicting diff sets, and which is read from on the read side to materialize the true table's contents.  This can be swapped out again, in constant time.  In the magma datum's model, when a large event buffer is applied, we'd rely on materialized table views in postgresql -- a materialized view can render results by querying the source event buffers while at the same time caching it to disk to speed up future queries. 

So essentially, I think it's okay for now that we have more a eventually consistent model (it may happen that full updates come delayed and we have some minor inconsistency), and furthermore we have options to deal with more nuance in the future.

Either way, chunked loading is the way to go.

PR #738, opened 2022-01-28
small styling tweaks to resource projects / root view

coleshaw:
Clean up the root view styling a bit.


graft:
Couple of nits:

- if there are no resource projects, the indentation for "No resource projects" is off
- The font for "Available projects" does not match the font for "Resource projects"
- I might add a second table header for "Resource projects"
- The "role" column is too small and causes wrapping if you have privileged access
- I think I like "Your Projects"/"Resource projects" as "Available projects" is somewhat ambiguous

PR #740, opened 2022-01-28
Vulcan Viz - Start adding some Data "Transformations"

dtm2451:
At the time of PR creation:
- Pulls in some dittoSeq-inspired utils which will be useful for plotly plotters as well as the eventual static plotting system
- Adds 'x_scale' and 'y_scale' inputs to the GUI in Vulcan and to scatter_plotly and also to y_plotly (y_scale only) functions in Archimedes
- Adds a capability into archimedes `*_plotly()` functions to subset the incoming dataframe via a `rows_use` input in order to only plot targeted data points. *still needs UI* so not implemented for the workflows yet.


dtm2451:
Still a WIP!  (One confusing bug is the default zoom of yPlots when log scaling is turned on.  I'm not sure why but the y-axis minimum value is well higher than where it should be.  I'll investigate this next week.)

dtm2451:
Added a 'log10(val+1)' option which fixed the zoom issue for the log scaled yplot I was testing.  There were many zeros in the data, so those became negative infinities after log()'ing, which I believe threw off the plotly zoom determination system.

dtm2451:
Also @coleshaw in figuring out the yplot zoom issue, I found the zoom out button necessary.  I think it would actually be useful to have the zoom in/out buttons added back, as well as the reset axes button.  So I did that.  I can't remember why we originally opted to have these all trimmed away except the save button?

coleshaw:
> Also @coleshaw in figuring out the yplot zoom issue, I found the zoom out button necessary. I think it would actually be useful to have the zoom in/out buttons added back, as well as the reset axes button. So I did that. I can't remember why we originally opted to have these all trimmed away except the save button?

I think we had some thought that too many buttons might clutter up the UI, but I think it's totally fine to add back in if we need them!

dtm2451:
I think I'll stop here for this PR, as I'm envisioning that row-subsetting would operate off of a larger UI than I would want to squeeze into this otherwise relatively short PR.

coleshaw:
So, I'm running into a weird issue with input values being stale, not sure if it's related to splitting up the primary inputs, or it's just a bug that we've never noticed before...but basically I think that downstream user step inputs don't reset correctly when the primary inputs change?

I made a bar plot using one query, and it looked nice. But then I changed the query to a completely different one (from patient + demographic to rna_seq + demographic), in the same session, and the plot is still trying to plot a patient + demographic plot. It did not ask me to submit new "plot options", but instead kept the x-axis and y-axis columns as the old ones (which don't exist in my new query). I think the expected behavior is that those inputs should reset and force the user to re-select them. The resulting exception I got was

```
KeyError: "None of [Index(['demographic.value', 'patient.ipi_number'], dtype='object')] are in the [columns]"
```

It's possible the UI algorithm for clearing inputs does not work with primary inputs, only user step inputs further downstream. Can you verify that you see the same behavior?

coleshaw:
Otherwise, the log scaling looks nice! :+1:

dtm2451:
Oh, yes.  I've definitely noticed that the plot settings don't get thrown out after a query is changed.  Not sure why I never thought to bring it up or try to fix now that you point it out too, but 100% can confirm it's an issue.

PR #742, opened 2022-02-01
reduce database calls

coleshaw:
Basically this PR aggressively caches the database results for model paths (to the date-shift-root model) and parent model / record_entry model record identifiers. When testing locally, it reduced update calls to ~15s or less -- so a little bit longer than the 1 minute for all of clinical_labs that @graft  saw without date-shifting, but much better than minutes / chunk.

Thoughts?

Screenshot from 2022-02-01 11-53-29


coleshaw:
@graft , still working on refactoring into a identifier-tree utility, but `select_map` performance is about the same as just `map`, when I'm testing locally...

PR #744, opened 2022-02-03
Metis client refresh token spec

coleshaw:
Just adding a spec for metis_client to verify that it also refreshes the token when you feed in commands via stdin. Can't replicate the issue that Gracie saw, so not sure how she had set up usage of the client...will keep an eye out for the issue cropping up, so we can gather more data around it.


corps:
Should we potentially add a secondary debug log?
Like, in addition to stdout, a known file is opened in the working directory and some debug logs (for oursake) are written (appended) there?  Would help us debug user issues.

We can log things like sever interactions, version on start, etc.

coleshaw:
Added some logging...hopefully it won't explode the log file. Rest of the behavior should be the same...

corps (approved):

PR #745, opened 2022-02-04
Janus display add user error messages

coleshaw:
Displays server error messages on the Janus project page when adding a user, so user will get some better feedback if they run into an issue, like the e-mail is malformed. I did not automatically strip spaces, because we're using a standard library regex to validate the e-mail (note that this means not all strings are valid...like `foo@site` passes the regex). So instead of trying to maintain what is a valid e-mail, I figured we could leave it to the user + Ruby standard library...

Also adds favicons for  Janus, Polyphemus, and Metis.

### DIFF
--- a/janus/lib/client/jsx/janus-ui.jsx
+++ b/janus/lib/client/jsx/janus-ui.jsx
@@ -10,6 +10,8 @@ import FlagsView from './flags/flags-view';
 
 import { findRoute, setRoutes } from 'etna-js/dispatchers/router';
 
+import {Notifications} from 'etna-js/components/Notifications';
+import Messages from 'etna-js/components/messages';
 import { createEtnaTheme } from 'etna-js/style/theme';
 
 const theme = createEtnaTheme("#3684fd","#77c");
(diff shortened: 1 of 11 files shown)


graft:
Hmm, I think stripping is a good idea here actually. Since `URI::MailTo::EMAIL_REGEXP` shows that emails cannot start or end with a string, and this is pretty much an immutable fact, the intent of the user is clearly not to include spaces in the email string, so it is okay to remove them; avoiding an unnecessary error feedback is also a good.

coleshaw:
How far down the e-mail validation part do you want to go? `UrI::MailTo::EMAIL_REGEXP` also allows strings like `foo@site` without a full hostname...

graft:
>How far down the e-mail validation part do you want to go?
Just this far for the moment, because I think it would be by far the most common error case and it is easy to solve. There might be room for a better utility downstream that lets you, say, paste in a bunch of "User Name <email@ucsf.edu>" type strings, sloppily spaced, etc., if we wish to support better bulk uploading, where we might do more parsing of the input.

PR #746, opened 2022-02-05
Create a UI for subsetting a dataframe by column values

dtm2451:
Re: [Create UI for data subsetting Notion Card](https://www.notion.so/ucsfdatascience/d0c4417ff7c346cfa37379910f1a99fe?v=d1369e88cb664427be0ac53d97c3332a&p=ecbaffc244c14ada97f35198c0194496)
- Creates a 'user_input_piece' for Vulcan aimed at allowing a user to set up any number of conditions for subsetting a dataframe based on its contents. Named `subsetDataFrameInput`.
- Creates a 'user_input_piece' for establishing a numeric range, named `rangeInput`, which is used within `subsetDataFrameInput`.
- Creates a paired function in archimedes, `subsetDF_index_targets` for converting the output of `subsetDataFrameInput` into the indexes of the target dataframe which meet the user-specified conditions.  I put this function in `archimedes/functions/ui_complements`, a spot where I imagine many more such complimentary functions might go in the future.
- Implements this `subsetDataFrameInput` within the VisualizationsUI & its compliment function in the `make_[plot-type].py` scripts.
- Adjusts the `BooleanInput` full input and `dropdownInput` input piece to encourage these components to be rendered in a single line.

ToDo:
- [x] Complete `subsetDF_index_targets` by adding logic ('and'/'or') interpretation.
- [x] Make the code prettier.
- [x] Make the widget somewhat prettier
- [x] Implement this subsetting system within the scRNAseq DGE widget and script.
- [x] Add validator that can merge into other input validators.


dtm2451:
This got large, but I think it's finally ready!

vulcan/lib/client/jsx/components/workflow/user_interactions/inputs/boolean.tsx
```diff
@@ -16,10 +16,10 @@ export default function BooleanInput({onChange, label, data, ...props}: WithInpu
     />
 
   if (label) {
-    return <label className='checkbox-input-option'>
+    return <div className='checkbox-input-option' style={{display: 'inline-flex'}}>
```
dtm2451, line 19: Made this `label` to `div` change in order to be able to encourage the checkbox and label to be in one line.

coleshaw:
Hm, seems like that might break accessibility / the built-in ability to click on the label to change the checkbox state. I can test tomorrow a bit…you should be able to style a label and a checkbox on the same line…? I will fiddle.

coleshaw:
Sounds good, I'll leave you to :violin:  with it :smile: , I think there is probably some CSS trick to getting them on the same line. I found a neat library to do spreadsheet-type stuff, going to try and get that working...

dtm2451:
> Sounds good, I'll leave you to violin with it smile , I think there is probably some CSS trick to getting them on the same line. I found a neat library to do spreadsheet-type stuff, going to try and get that working...

Yea, been looking for the CSS options for label...  Is it just me not knowing where to look or is the documentation for built-in react components terrible?

coleshaw:
> Yea, been looking for the CSS options for label... Is it just me not knowing where to look or is the documentation for built-in react components terrible?

Ah so `<label>` is just a regular old HTML element, non-React, so you'd probably need to just look at CSS properties. You should be able to do the same `style` thing you did with the `div` -- does that work? 

Generally, I like the [MDN documents](https://developer.mozilla.org/en-US/docs/Web/CSS), but getting CSS layouts right is definitely frustrating and tricky...

dtm2451:
Also, NEW task I have realized I should address before pushing this in.  @corps you told me yesterday to not let perfection on this get in the way of progress, but I think I can fold in a couple-line quick-fix =).

The issue: is that the scDGEui update changes the expected structure of the input's `value` to where a value['subset'] is expected now but this would not be in anyone's saved session .json or cookies.  The widget then cannot render.

Quick fix: Work in code to detect the absence of this key, then fill in the necessary value['subset']={} with a call to onChange.

Longer term fix for similar future updates...: I'll think on the demands a bit more, then plan to bring it up in a standup next week.

coleshaw:
Just curious -- which boolean input was running onto two different lines for you? For example, when I look at "Focus on a subset of the incoming data?", it is on one line for both `label` and `div`...

dtm2451:
> Just curious -- which boolean input was running onto two different lines for you? For example, when I look at "Focus on a subset of the incoming data?", it is on one line for both `label` and `div`...

hmmm, not sure exactly.  BUT, now when I switched back to `label`, VSCode didn't squiggles-yell at my `style={{display: 'inline-flex'}}` addition && storybook renders just fine with text-clicking functional, so you are correct that it just works!  :shrug:. Putting it back to `label`!

vulcan/lib/client/jsx/components/workflow/user_interactions/inputs/scDGE.tsx
```diff
@@ -48,54 +50,60 @@ export default function DiffExpSC({
...
+      }
+      SubsetComps = 
+        <Button
+          color={"primary"}
+          onClick={(x) => {updateOldSubset()}}>
+          Step 2: Subset... UPDATE NEEDED. Click to update.
+        </Button>
+    }
+  }
+
```
dtm2451, line 98: I don't love that this ultimately brings back half the amount of code which the `subsetDataFrameInput` piece allowed me to remove, but addresses https://github.com/mountetna/monoetna/pull/746#issuecomment-1036517692

dtm2451: scDGE_update

coleshaw (approved):
Looks good to me! Did you intentionally leave out help doc (from the hover-over tooltip for the step) for the advanced options like scaling or subsetting, so the user doesn't get too confused?

PR #748, opened 2022-02-08
Vulcan UI primary input dependencies

coleshaw:
This PR makes a small change to the Vulcan UI session management and basically gives the `PrimaryInputs` a `stepName` value, to cover a small corner case in behavior.

Expected behavior was that changing a primary input would clear downstream, dependent inputs from the session, because we assume any selected values might become invalid due to the primary input change. Note that this behavior also holds for any other user input value.

However, I noticed in #740, that this behavior was not true for primary inputs. I can also replicate using a scRNAseq workflow. For example, getting a UMAP out, and adding some manual annotations, and then updating some UMAP primary input parameters would leave the manual annotations the same as the previous ones (manual annotation is downstream of the UMAP primary inputs). But, I believe the expected behavior would be to clear out the manual annotations as a downstream step, on the assumptions that clusters might change.

This PR gives the `PrimaryInputs` component a `stepName`, which then allows [filterStaleInputs](https://github.com/mountetna/monoetna/blob/master/vulcan/lib/client/jsx/reducers/vulcan_reducer.ts#L171) to do its thing. Because we also used a `null` stepName to [prevent polling from changing input state](https://github.com/mountetna/monoetna/blob/master/vulcan/lib/client/jsx/contexts/session_sync.tsx#L59-L61), `null` was overloaded for this purpose, and it seemed to make sense to assign a distinct `stepName` to the primary inputs instead of relying on `null`.

The constant string starts with a `:` to avoid collision with real workflow steps, since I think that's an invalid character for CWL (to work as YAML).

Also adds the favicon for Vulcan, which I forgot in #745.

### DIFF
--- a/vulcan/lib/client/jsx/api_types.ts
+++ b/vulcan/lib/client/jsx/api_types.ts
@@ -41,7 +41,7 @@ export const OUTPUT_COMPONENT = {
   RAW: 'raw'
 };
 
-export const PRIMARY_INPUTS = ':primary_inputs';
+export const PRIMARY_INPUTS = 'primary_inputs';
 
 export const defaultWorkflowsResponse = {
   workflows: [] as Workflow[],
--- a/vulcan/lib/client/jsx/api_types.ts
+++ b/vulcan/lib/client/jsx/api_types.ts
@@ -41,6 +41,8 @@ export const OUTPUT_COMPONENT = {
   RAW: 'raw'
 };
 
+export const PRIMARY_INPUTS = ':primary_inputs';
+
 export const defaultWorkflowsResponse = {
   workflows: [] as Workflow[],
 }
--- a/vulcan/lib/client/jsx/api_types.ts
+++ b/vulcan/lib/client/jsx/api_types.ts
@@ -41,8 +41,6 @@ export const OUTPUT_COMPONENT = {
   RAW: 'raw'
 };
 
-export const PRIMARY_INPUTS = 'primary_inputs';
-
 export const defaultWorkflowsResponse = {
   workflows: [] as Workflow[],
 }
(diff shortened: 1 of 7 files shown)


corps:
Hmm,

So to be clear, `Maybe<string>` has 3 states:

`null` => no string
`[null]` => null string
`["hello"]` => string


I suspect what may be happening is that there is confusion about the interaction between the `Maybe` step name (which could be, not a step name, as in the call to `updateFromSessionResponse`), vs `null | string` step name, which is always a step name, but could be null (thus is always either `[null]` or `["hello"]` but not `null`).

I propose there is another fix to a more subtle problem in the works, but I need to peak around.

corps:
I think I found it, would you try reverting these changes and instead making one small change?

In `input_state_management`, change this
```
    requestPoll(false, maybeOfNullable(stepName));
```

to

```
    requestPoll(false, some(stepName as string))
```

possibly in aseparate branch, just to check if it works.
If not, I can assist in helping clean this mess up.

corps:
The fundamental issue is that we need a single type that can distinguish one of each of these states:

1.  No step or primary input is associated with this action `null`
2. A primary input, which is never overlapping or confused with a step, is associated with this action `[null]`
3. A step is associated with this action, which is not a primary input `["stepname"]`

Perhaps changing the types by using a magic sentinel can work, but in either case I believe the system as designed intends for the above typing, which encourages me to consider my proposed solution.

If we think this is bad and just using a singular magic string to represent primary inputs is better, I will not oppose.

coleshaw:
> I think I found it, would you try reverting these changes and instead making one small change?
> 
> In input_state_management, change this
> 
>     requestPoll(false, maybeOfNullable(stepName));
> 
> to
> 
>     requestPoll(false, some(stepName as string))
> 
> possibly in aseparate branch, just to check if it works.

Ah, neat! So I think this works. If I understand correctly, this converts the `null` to `[null]`, which then follows through `filterStaleInputs` as expected. Cool! :tada: Thanks for pointing that out, I didn't realize `Maybe<string>` would already be able to handle the three distinct state.

dtm2451 (approved):
Seems to work as intended for me =).

PR #749, opened 2022-02-10
Vulcan figures

graft:
This is still WIP because there are several outstanding issues and I can't finish it until next week. These are:

1) The title textfield width is difficult to control.
2) Several interactions are missing on the figure list page - the three menu interactions (copy, rename, remove), some of which will require some new API endpoints.
3) When loading from the figure list page, either method (open an existing or a new figure) should check for the corresponding local storage key and remove it if the user confirms before opening the figure (i.e., discard edits).
4) Sometimes the page requires reloading to update the cached state correctly, not clear why.
5) It is difficult to detect a dirty input state compared to the saved figure; therefore, the "save" button is constantly on and the feedback for saving is poor (this also affects #3 above).

Other than these the interactions work as expected and the figure can be updated incrementally, progress shared, etc.

Not intended for this PR, but hopefully following soon:
1) Tags for figures
2) Revision history for updates to figures


vulcan/lib/server.rb
```diff
@@ -12,25 +13,27 @@ class Server < Etna::Server
...
+      get 'api/:project_name/figure/:figure_id', action: 'figure#get'
+      post 'api/:project_name/figure/create', action: 'figure#create'
+      post 'api/:project_name/figure/:figure_id/update', action: 'figure#update'
+    end
 
-      # remaining view routes are parsed by the client and must also be set there
-      get '/:project_name/workflow', as: :workflow do
-        erb_view(:client)
-      end
+    with auth: { user: { active?: true, has_flag?: 'vulcan' } } do
```
coleshaw, line 28: These can be grouped with the previous `auth` block, I think, and consolidated under the `user: can_view` instead of `user: active`?

coleshaw:
super-nit: when there are many workflows to choose from, the vertical scrolling for the "create figures" dialog feels a bit odd, since it's restricted to one workflow wide. Should that be widened or made more flexible, to show several horizontally, if there are multiple?

vulcan/lib/client/jsx/components/workflow/workflow_manager.tsx
```diff
@@ -1,37 +1,64 @@
...
+            json_get(`/api/${projectName}/figure/${figureId}`).then(
+              figure => {
+                figure.key = `${figure.project_name}/${figure.figure_id}`;
+                workflow = workflowByName(figure.workflow_name, state);
+                dispatch(setWorkflow(workflow, projectName));
+                dispatch(setSession(figure));
+              }
+            )
+          } else {
+            workflow = workflowByName(`${workflowName}.cwl`, state);
```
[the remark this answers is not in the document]
coleshaw: Note that I think `displayName` does not have to be unique, but `name` does, on the workflow object...

vulcan/lib/client/jsx/components/workflow/session/session_manager.tsx
```diff
@@ -42,7 +64,30 @@ export default function SessionManager() {
...
+
+    let params = { ...session }
+
+    if (!params.title) {
+      params.title = prompt('Set a title for this figure');
+      if (!params.title) return;
+    }
+
+    if (params.figure_id) {
+      json_post(`/api/${params.project_name}/figure/${params.figure_id}/update`, params)
```
corps, line 80: Fairly minor nit -- is there any feedback that the update was successful or in progress?
Also, is there any risk of double clicking save and having a race condition for update?
Maybe we could track the state of saving with some state, `isSaving` to help with both?
But fairly minor.

graft:
> > The title textfield width is difficult to control.
> 
> Can you elaborate what this means? It appears fine for me when I select an existing figure, in the breadcrumbs...

Try to set a long title

coleshaw:
@graft 

> When loading from the figure list page, either method (open an existing or a new figure) should check for the corresponding local storage key and remove it if the user confirms before opening the figure (i.e., discard edits).

What is your expected behavior in the following situation?

1. User creates a new figure
2. User makes edits to the workflow, runs some steps
3. User leaves the page
4. User re-creates a new figure of the same type
5. User is prompted that they have local edits. They choose to load the local edits.
6. User sees their previous work.
7. User saves the figure and adds a title.
8. User re-creates a new figure of the same type.

At this point, should the local edits be gone, and if they pick the same type, it is just a blank workflow? Basically after step 7, does the act of "saving" remove the edits from local storage for that workflow, OR, would the user expect the edits to still be there, in case they want multiple variations of that workflow?

coleshaw:
Closing in favor of #750

PR #750, opened 2022-02-10
Timur query link vulcan

coleshaw:
This PR merges into @graft 's `vulcan/figures` branch, since there remains some work there (#749). However, the linkage between Query and Vulcan is already supported, and so I think this is ready to merge in.

This PR adds a button to the Timur Query page that will ask Vulcan what `plotting` workflows are available in a given project. Once a user has constructed a query, they can open up one of the Vulcan workflows. A `figure` will be created for them, with the query params injected into the primary inputs in the appropriate places. Then can then click `Run` to move on to the next step.

Note that Vulcan tests on this branch fail, but should be the same as the Typescript checks failing on the base branch, and should also be resolved when those are.


coleshaw:
Well phoey...trying to get the diff to look right. Currently it's comparing against `master` and thus includes more changes than necessary....apologies.

coleshaw:
Thanks @graft ! diff looks cleaner now :smile:

coleshaw:
Ah yes, we can definitely open in a new tab. I had started out with that behavior, but changed it to the same tab to facilitate some debugging. Wasn't sure which behavior was preferable, so I can switch it back to new tab. :+1:

coleshaw:
Reviewed with @graft. Going to rebase on `master` and close #749 in favor of this PR.

Open tasks:

[ ] "Smarter" Timur Query link button to account for a single `plotting` workflow, as in #753.
[ ] Remove the "post" option in `workflow_manager` so not auto-run a workflow (might be heavy).
[ ] Re-wire up the ability to edit a figure title from the figure page directly.
[ ] Clear out local storage for the "create figure" entry once it is saved, so additional "create figure"s start from a blank workflow.

We may also wish to move the "confirm to discard edits" alerts to the `figure_list` component, though that would involve some duplication of logic (fetch from local storage / server). Will wait to see user feedback on if the current alerting system is too annoying if folks consistently load or refresh an edited direct-figure page.

coleshaw:
@graft, @dtm2451 , I think this is now finished / ready for review. It includes all of @graft 's figure work and the modified "smart" button in Timur Query. Slack is done right now, so can't notify you there...happy to demo this at standup later this week, too.

dtm2451:
One thing I notice is that I can't seem to use any of the actions behind the three dots of each figure on the [project page](https://vulcan.development.local/example)

coleshaw:
Huh, weird, those should work….let me check tomorrow.

dtm2451:
jk sorry.... I hadn't `git pull`ed and was very out-of-date!

dtm2451:
Unsurprisingly, the work with the up-to-date code lol!  But I think a catch is needed for when a user clicks to rename a figure, but then cancels.  It still ends up not renaming the figure, but gives a server-error that's probably not needed.

coleshaw:
Hehe!

Thanks for catching that, I didn't test that path apparently. I will check it out!

PR #752, opened 2022-02-15
Vulcan data transformation widget

coleshaw:
This PR adds a UI widget that provides an Excel-like environment in a Vulcan workflow. Users can edit a data frame using (most?) Excel-like formulas or direct text input. This means they should be able to do regex-like things (to parse MVIR1 identifiers to identify single-cell assay types, for example), or run calculations and normalizations on columns. I've currently configured it so users can add columns to make data frames a consistent size, but not add rows, so they can't "add data". If some other configurations make sense, I can try to tweak it.

The widget preserves both the rendered / calculated final values as well as the user-input values, thus preserving formulas in the session inputs (useful for invalidating the downstream cache and saving edits).

This should be useful for cross-project analysis, when we need to hand-edit data frames from projects to make them consistent across all projects being analyzed, as well as any hand-tweaks for plotting.

There is probably some maximum size that can be handled by the widget, but I haven't seen any indication of that, so we may just have to see. There is probably also some maximum size where it's not really feasible to expect people to edit it by hand, anyways...


vulcan/lib/client/jsx/components/workflow/user_interactions/inputs/data_transformation.tsx
```diff
@@ -0,0 +1,219 @@
...
+  onClose
+}: {
+  data: any[][];
+  onChange: (data: Maybe<{[key: string]: any}>) => void;
+  onClose: () => void;
+}) {
+  const classes = useStyles();
+
+  const hotTableComponent = useRef<any>(null);
+  const hyperformulaInstance = HyperFormula.buildEmpty({
```
corps, line 39: This isn't like a `useState` value?  It should be a unique instance every render?

vulcan/lib/client/jsx/components/workflow/user_interactions/inputs/data_transformation.tsx
```diff
@@ -0,0 +1,219 @@
...
+          variant='contained'
+        >
+          Save
+        </Button>
+      </DialogActions>
+    </>
+  );
+}
+
+export function toJson(input: any[][]): DataEnvelope<{[key: string]: any}> {
```
corps, line 116: Should this be exported or just local module function?  Seems fairly specific to this component and has a very generic name.

coleshaw: Exported for tests only, basically...

coleshaw: I can rename it to be more specific. I was too lazy to write a component-level test since the table opens in a dialog, so just wanted to test the core transformation logic.

vulcan/lib/client/jsx/components/workflow/user_interactions/inputs/data_transformation.tsx
coleshaw: Ah, thanks!

dtm2451:
Very powerful and seemingly simple!  :raised_hands: 

One Q is I'm wondering if your vision is that the `transform_data.cwl` workflow will be permanent?  Or is it only for immediate testing purposes with the expectation that we would fold the step into the Viz workflows and remove this workflow?

Also did some testing, and had 3 notes:
- I wasn't initially sure how to add a new column.  I think the (?)/hover-doc could hold a few notes on how to work with the widget.  E.g.: Right click for an interactions menu where you can add/remove columns. Start a cell with '=' to create functions in an Excel-like manner.
- The row number presented in your summary text includes the column header row.  I wonder if we should either subtract that 1 in order to match exactly the number of data points encoded, or maybe leave the original and just put a (12 data points) like "Your data frame has 13 rows and 6 columns, (column headers + 12 data points)".
- Noticed some lag in state updates after a save and re-open.  I think a gif will demo better than words here...
 but summary is I don't initially see my added column.  It shows up if I click to interact, but the size of the widget then doesn't update until after I clicked the scroll-bar.  (I didn't scope too far so it might be any click / any interaction to trigger these updates, idk.)
data_edit_state_lag

coleshaw:
I was thinking that current workflow in this PR would be more of an example, and you would just re-use the `data-transformation` UI widget in a real plotting or cross-project workflow.

> I wasn't initially sure how to add a new column. I think the (?)/hover-doc could hold a few notes on how to work with the widget. 
> E.g.: Right click for an interactions menu where you can add/remove columns. Start a cell with '=' to create functions in an Excel-like manner.

Ah, thanks! Yes, that is opaque unless you're expecting Excel. I can add more words to describe this.

> The row number presented in your summary text includes the column header row. I wonder if we should either subtract that 1 in order to match exactly the number of data points encoded, or maybe leave the original and just put a (12 data points) like "Your data frame has 13 rows and 6 columns, (column headers + 12 data points)".

Hm, good point. I wasn't sure how folks "counted" the size of a data frame. I'll clarify that.

> Noticed some lag in state updates after a save and re-open. I think a gif will demo better than words here...
> but summary is I don't initially see my added column. It shows up if I click to interact, but the size of the widget then doesn't update until after I clicked the scroll-bar. (I didn't scope too far so it might be any click / any interaction to trigger these updates, idk.)

Oh interesting, I haven't seen that behavior before. I'll try replicating it and seeing if I can hunt down a root cause. I had to do some weird CSS to get the dialog the rigth width when it pops open, perhaps that isn't very responsive to changes...

dtm2451:
> I was thinking that current workflow in this PR would be more of an example, and you would just re-use the data-transformation UI widget in a real plotting or cross-project workflow.

Cool!  I can fold the widget into the Viz workflows in the way I think it would go and push a commit?  Changes would be constrained to within those CWL files, so no risk of conflicts.  If so, I'll just wait til after a (?)-`doc:` update so I can carry that over.

vulcan/lib/server/workflows/transform_data.cwl
```diff
@@ -0,0 +1,48 @@
...
+  transform_data:
+    run: ui-queries/data-transformation.cwl
+    label: 'Transform your data'
+    doc: "Manipulate your data frame as needed."
+    in:
+      data_frame: get_data/data_frame
+    # data will be a Hash with keys for "data" and "source_data".
+    #   `data` is the rendered, final value after parsing formulas.
+    #   `source_data` is the original values, including formula text.
+    out: [data]
```
dtm2451, line 42: Oh man, was going to start fitting this in with the VizWorkflows, but this output structure doesn't play well with the VizUI as is.

VizUI-side: The VizUI steps (with `run: ui-queries/<bar/y/scatter>-plotly.cwl`) expect to have an input that is the json-data_frame given directly.

Here: stepUserInputs output their jsx `prop.value` at commit => as their output, so as the cwl `out: <output_name>`. Thus, if we ever want the source_data available for this stepUI, the way to accomplish that is indeed via the hash output structure currently implemented.  Thus, VizUI needs `in: data_frame: <transformation_step_name>/data**['data']**`... which I don't believe is valid CWL.

Paths forward:
1) I can build a catch into the VizUI visualization.tsx (essentially `const data_frame = (Object.keys(data['data_frame]) == ['data', 'source_data']) ? data['data_frame']['data'] : data['data_frame']`).
2) We build a secondary output method for stepUIs where perhaps when they have a `prop.multi_out=true` (false by default) the values of `prop.value` are spread into multiple outputs named by their keys? A.k.a. a way where line 42 could become `out: [data, source_data]`, and then we can pass `in: data_frame: <transformation_step_name>/data` to a VizUI step and it would be in the currently expected format.

I'm not sure if 2 is possible (can poke!) but I prefer it to 1 as the CWL is definitely the better place for this logic to go.  1 is a cumbersome hack that will likely become unmanageable in the future as we'd accumulate more and more instances of output logic from stepUI-X transferring into the jsx of stepUI-Y!

coleshaw: Hm, yeah, I'm not sure how to do #2 with the current architecture, since we made an assumption that UI inputs would have a single output. I will poke around, I like your idea to include a flag to expand a passed-in hash value. Should be doable, and yeah, that would match better with how the back-end allows multiple outputs.

But, I want to focus a bit on making progress on the figures branch first, though, since I think that might take me awhile to finalize...I've currently broken it while fixing the TypeScript issues :stuck_out_tongue_closed_eyes:

corps: #2 might not be too hard, actually.  Technically, the server already supports multi outputs, the trick is teaching the frontend UI components to support multiple outputs.  It may be possible by making a deep cut into the onChange handlers.

coleshaw:
Hm, sure, I think we can keep the viz workflows separate from this added widget for now, so feel free to merge your other stuff in. I really want to get the multiple outputs right for this widget, and won't be able to dig into that until next week. I did play around with it a bit today and think I see where to make the cuts that @corps mentioned.

dtm2451:
> Hm, sure, I think we can keep the viz workflows separate from this added widget for now, so feel free to merge your other stuff in. I really want to get the multiple outputs right for this widget, and won't be able to dig into that until next week. I did play around with it a bit today and think I see where to make the cuts that @corps mentioned.

Awesome!  That definitely makes sense to do.
And lemme also say, sorry for potentially stepping on toes if I did with the scoping comment above!  I think if we could also make it so that the UI and steps panel draw orders are more consistent, specifically by having it break ties (I think order is entirely based on the DAG right now?) based on step order of the CWL, then my [2nd apprehension](https://github.com/mountetna/monoetna/pull/752#issuecomment-1045145548) goes away too!

coleshaw:
Hm....I've never noticed that the step order is inconsistent, but I don't pay too much attention to that panel. Does that happen within the same workflow, when you refresh the page, or run?

coleshaw:
I will try to clean up my implementation and add some tests, and then we can take another look at how it integrates with the viz workflow. It wound up being a bit messier than just `onChange` updates, because I also had to handle committing from the buffer and not rendering multiple UI widgets...

dtm2451:
> Hm....I've never noticed that the step order is inconsistent, but I don't pay too much attention to that panel. Does that happen within the same workflow, when you refresh the page, or run?

perhaps that panel isn't inconsistent between renders... I've not noticed that either.  I guess I was just thinking that this CWL-based tie-breaking logic could apply to both panels (so why not, in order to be sure they stay aligned?).  Fixing for the stepUIs feels imperative.  Updating the steps side if it would ensure matching ordering between the two panels would be nice.

coleshaw:
Hm, do you have a screenshot of where the two are out of sync? I know they are slightly different, because the inputs panel only shows steps that are user-input or outputs, but it sounds like sometimes they aren't in the same order as the ones in the steps panel?

I'm not quite sure how you would apply "step order of the CWL", unless you mean the order of steps when reading it from top-to-bottom...? Because I think you can write the steps in any order you want, it's kind of arbitrary, as long as you describe the input / output relationships correctly. I think the DAG is our representation of the CWL text, though there is some algorithm that "flattens" it into a single array. Perhaps that is where you would apply this tie-breaking logic?

dtm2451:
I don't have a screenshot... but could poke for an instance with the scRNAseq workflow and see if it's a real 'possible issue' versus just a 'useless thought' for the steps side.

But yes, I mean top-to-bottom ordering from the CWL text.  Maybe this isn't an always correct idea, but I think it's a workable solution?  Although CWLs can be written in whatever order, I think expecting earliest steps at the top, latest at the bottom is a reasonable expectation of how most CWL-authors would operate. And especially if we encouraged them to do so via a documentation note about such a render-order tie-breaking mechanism for the UI?

vulcan/lib/client/jsx/components/workflow/user_interactions/inputs/input_types.ts
```diff
@@ -74,15 +75,41 @@ export function bindInputSpecification(input: InputSpecification,
...
+        }, {});
+        setInputs(inputs => ({...inputs, ...values as any}))
+      } else {
+        setInputs(inputs => ({...inputs, [input.source]: v}))
+      }
     },
     data: inputData,
-    value: input.source in buffered ? buffered[input.source] :
-      input.source in session.inputs ? some(session.inputs[input.source]) :
+    value: input.source in buffered ?
```
coleshaw, line 90: So in order to handle multiple outputs in a UI step, I wound up having to also cut into `value`, here (in order to correctly feed the set value back into the UI component), and [sourceNamesOfSteps](https://github.com/mountetna/monoetna/blob/5e2e99d40b21dcb43c7e08972547b5fbabbc1401/vulcan/lib/client/jsx/selectors/workflow_selectors.ts#L295), because it winds up being a single UI widget handling multiple output ... not sure this is the cleanest way to get the functionality, will take a closer look.

vulcan/lib/client/jsx/selectors/workflow_selectors.ts
```diff
@@ -294,14 +293,23 @@ export const sourceNameOfReference = (ref: [string, string]) => {
 };
 
 export const sourceNamesOfStep = (step: WorkflowStep) => {
-  return step.out.map((name) => sourceNameOfReference([step.name, name]));
-};
+  if (stepCollapsesOutputs(step))
+    // Assign one output here so that any downloaded input data is grabbed.
+    return [sourceNameOfReference([step.name, step.out[0]])];
```
coleshaw, line 298: Here I "hide" any multiple outputs for a single CWL step. If you don't "hide" the extra outputs, then you get duplicate UI components rendered on the screen. For example, when I added two outputs to the `transform_data` step, I got two "DataTransformationInput" components, each of which had a button and a dialog. And unfortunately one dialog would be tied to `data`, and the other tied to `source_data`, which isn't the desired behavior...

vulcan/lib/client/jsx/components/workflow/user_interactions/inputs/data_transformation.tsx
```diff
@@ -0,0 +1,236 @@
...
+            if (hotTableComponent.current) {
+              // hotTableComponent.current.hotInstance.getSourceData()
+              //   returns the raw formulas, unrendered, which we
+              //   set as input values to preserve any transformations.
+              const sourceData = hotTableComponent.current.hotInstance.getSourceData();
+              const data = hotTableComponent.current.hotInstance.getData();
+
+              onChange(
+                some({
+                  source_data: some(toJson(sourceData)),
```
coleshaw, line 112: This output structure is now tightly coupled to the CWL outputs (in other words, your CWL steps have to include these two output names, exactly). I don't really like how that works, will revisit and see if there's a way to make it more generic...perhaps we can just return an array of inputs, and one is source and one is data. But that still requires the CWL author to know which output index is which set of data, which seems better but not great.

corps: I'm not quite following: as written, this results in whichever, singular output this ui output step names, to have an object, containing two keys, source_data and data.  the 'output' is still whatever name the CWL author would like, it just happens to be a nested object.

coleshaw: Mmm, so the widget itself outputs a single Hash, yes. But then in [onChange](https://github.com/mountetna/monoetna/blob/cs/vulcan-data-transformation-widget/vulcan/lib/client/jsx/components/workflow/user_interactions/inputs/input_types.ts#L78-L84), it destructures this hash into multiple outputs.

It seemed like because this is a single UI widget controlling multiple outputs, I had to make changes to `input_types` to both `onChange` and `value`. The UI widget can pass additional information along to `onChange` (i.e. `multi=true` or some equivalent) to tell it to unpack or handle multiple output values. However, a UI widget cannot tell its parent components `UserInput` or `StepUserInput` that it requires multiple input values (`data` and `source_data` in this case)... thus, the packing and unpacking of multiple input / output values happens elsewhere, and this UI widget still only gets one input value and one output value. Thus it also unpacks the [input value](https://github.com/mountetna/monoetna/blob/cs/vulcan-data-transformation-widget/vulcan/lib/client/jsx/components/workflow/user_interactions/inputs/data_transformation.tsx#L211) and uses the part it needs.

Not sure if that makes sense. I can try to explain it better, if my thinking isn't clear...

corps: Ahh, so this brings us back to your original point about destructuring multiple outputs with either names or arrays.

What if you support both approaches: arrays OR named output matching?

IF the CWL author specifies an output by the same name as the multi output properties, those output names map to those output properties.

Otherwise,

each of the multiple output properties, in some fixed ordering (property ordering of the json object in onChange), is assigned to an output name corresponding to the array of output names in the CWL.

Any destructured output not referenced in any way (say the user only specifies 1 output but there are 3) are discarded with a warning in the system somehow.

I think the only thing that would concern me is if an output is matched to a name even when the CWL does not explicitly specify that name.  It would make analyzing the CWL for correctness much, much harder, and would make reading it tricky ("implicit output names" would make figuring out where values are coming from tricky).

coleshaw: Hm, okay, let me try putting in more flexibility like that. I was trying another approach this morning, but I think it winds up as being more brittle than this current implementation, so going to trash it...

coleshaw: And thank you for the suggestions! I've been a bit stuck trying to figure out how to make it more friendly, those ideas should definitely help.

coleshaw:
So I think this is now in a better place, and it feeds directly into the Data Viz components (you can test this out with the transform_data workflow, I've already wired it up).

dtm2451:
Testing :heavy_check_mark:.  Couple nits that I'll add, but it works well and I think I like the current two workflow setup quite well!

Perhaps there's a good way that we 
plot_menu
can make the names display something more like "Visualize directly"  and "Tweak data first" or something like that.

vulcan/lib/client/jsx/components/workflow/user_interactions/inputs/data_transformation.tsx
```diff
@@ -0,0 +1,275 @@
...
+              // hotTableComponent.current.hotInstance.getSourceData()
+              //   returns the raw formulas, unrendered, which we
+              //   set as input values to preserve any transformations.
+              const sourceData = hotTableComponent.current.hotInstance.getSourceData();
+              const data = hotTableComponent.current.hotInstance.getData();
+
+              onChange(
+                some({
+                  source_data: some(nestedArrayToDataFrameJson(sourceData)),
+                  data: some(nestedArrayToDataFrameJson(data))
```
dtm2451, line 117: Perhaps we can rename these to `formulaic_data` and `calculated_data` or something similar?  I think it'd be nice to have the contents of these both be obvious from the name as they're the outputs that workflow creators will get to work with.

coleshaw:
> can make the names display something more like "Visualize directly" and "Tweak data first" or something like that.

Hm, so currently this takes that from the `displayName` of the workflow metadata -- would you want the `Visualization` one to be different than that value?

Do you want to remove the `plotting` tag from data-manipulation (that would turn the drop-down menu into a single button)? I had originally intended that workflow to act as an example of how you could use that UI widget in other workflows. Not sure it's useful as a standalone one, or if folks are at the point where they need to edit the data frame first?

vulcan/lib/client/jsx/components/workflow/user_interactions/inputs/data_transformation.tsx
coleshaw: Sure, I can do that! `data` and `source_data` are just what the table library uses, so I mimic'd by default. I'll change those to be more descriptive.

coleshaw:
So, in lieu of the slider, I've added text about the data being different than the raw, original data, and a button to revert it. I think it may be very difficult / hacky to get a slider as originally discussed, because of what you mentioned, @dtm2451 . When the user reverts to the raw data, we reset `input.value`, and are not able to track the transformed state. Trying to store the transformed state locally in the component leads to weird behavior, and seems like introducing yet-another-input-bufer, which I don't like. I think the text and button may be a good compromise, in that the user knows they have modified the data and can revert.

Note that if they have committed their work into the session, they can then manage the three states (raw data, committed inputs, and transformed inputs) through a combination of the Reset button and the Use raw data buttons.

dtm2451:
looks great!  I really like the 'Review' oriented phrasing updates too!  :+1:

> Note that if they have committed their work into the session, they can then manage the three states (raw data, committed inputs, and transformed inputs) through a combination of the Reset button and the Use raw data buttons.

This is a great point!  It's a nice perk, although once the user does commit again, that past edited state that could be retrieved with the reset button gets replaced with the new version.  Some saving of ui histories and gui for swapping between them could help, but that's definitely not a small idea.

coleshaw:
> Some saving of ui histories and gui for swapping between them could help, but that's definitely not a small idea.

Yeah, future PR :smile:

PR #753, opened 2022-02-18
Combine all visualization workflows into one, and document in the 'Help' pages

dtm2451:
Merges the previously separate visualization workflows into a single workflow where the plot type is chosen within the plot setup widget.

(I'd piled these changes on top of my documentation update branch, so just going to close #751 and have it all over here.)

Still need to:
- [x] update the metadata file.
- [x] update io page documentation for the now single-workflow path.


dtm2451:
tldr: I think so!

I have one lingering "maybe?" but I don't think it should actually require going back to 2 separate workflows: I eventually aim to set up static plotters because those will be able to do a bunch of extra cool things that are just not possible or not easily automated in a pretty way with plotly (like density contours, showing non-target cells after a subset in gray in the back, labels or ellipses to highlight color-groups' regions within the plot).  This would create a 2nd plotter option for most visualization types, but I think I'd aim to handle that within the same plot setup interface and workflow!

dtm2451:
> This looks really cool! Should there then be the assumption on the Timur Query side, that there will only ever be one `plotting` workflow? i.e. I should get rid of the drop-down menu and just link the button to Vulcan?

We discussed in slack, but to put it all here:

I realized an additional and more likely 'maybe?' is for heatmaps... These likely have some specialized data selection, transformation, re-structuring that would necessitate a dedicated separate workflow.  One, where I could still imagine a Timur-Query as viable starting point to pull in both the target data (e.g. rna_seq#gene_tpm) and potential annotations (e.g. a patient#group, diagnostic#stage).

Your plan of making the button "smart" in #750, with a dropdown shown only if we do add more workflows matching the viz-tags criteria, seems a great way!

dtm2451:
Perhaps I should add some edits to the vulcan-help page to cover the now merged in workflows -> figures change, but outside of that, I think this PR is ready. @coleshaw @graft

coleshaw (approved):
Nice! Seems to work well for me. :+1:

PR #754, opened 2022-02-26
Inject some Material UI components into Vulcan

dtm2451:
Goal is to move as many of our input components from basic version to MaterialUI versions for visual appeal.

My plan is to update only within Vulcan for now, though we will definitely want to push to Etna and elsewhere eventually.

For keeping track...

user_interactions/inputs targets:
- [x] boolean / checkboxes
- [x] float
- [x] integer
- ~[ ] multiselect_string~ = out-of-scope
- [x] nested_select_autocomplete
  - [x] build in extra support for when options >1000
- [x] scDGE
- [x] select_autocomplete
  - [x] build in extra support for when options >1000
- [x] single_dropdown_multicheckbox
- [x] string

user_input_pieces:
- [x] subsetDataFramePiece
- [x] stringPiece
- [x] checkboxPiece
- [x] dropdownPiece
  - [x] build in extra support for when options >1000
- [x] rangePiece

*Note that many of the above rely on each other so there's less to do than the long lists make it seem!


dtm2451:
Thinking to leave the 'multiselect_string' update for a next pass because updating the Etna list input seems a rather large chunk of work on its own.

So! I've finished all my planned MaterialUI updates now!!  All that's left now is:
- [ ] fix tests
- [ ] double-check style tweaks

PR #757, opened 2022-02-28
add in missing query-input map

coleshaw:
Ran into the issue that @dtm2451  mentioned at standup, that query values weren't getting mapped to Vulcan inputs. Realized that we were missing the config values in the workflow metadata, oops...

Side-question @dtm2451 , do we have to release the visualization workflow to IPI, too, before the presentation?

### DIFF
--- a/vulcan/lib/server/workflows/visualization.metadata.json
+++ b/vulcan/lib/server/workflows/visualization.metadata.json
@@ -1,10 +1,15 @@
 {
-  "projects": ["example"],
+  "projects": ["example", "ipi"],
   "authors": ["Dan Bunis"],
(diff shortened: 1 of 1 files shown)


dtm2451:
ooooh oops!  I missed that metadata change as I think I had my PR which deleted the old workflows going when you'd added this to the separate workflows.  Sorry about that!  Lemme test first but then we should get this through!

For releasing... I think we want to release the visualization workflow even more widely than just ipi, but not sure about timing.  Maybe a question for slack as we could fold that in here if desired!

dtm2451 (approved):
nice, quick work Cole!

PR #758, opened 2022-02-28
First pass at a highly efficient tail endpoint

corps:
After analyze some performance of the ipi and mvir1 datasets under pressure,

I realized that the current path calculator has an exponential performance impact relative to folders due to the folder calculator requiring an ever increasing overhead to every batch fetch.  This start out fairly manageable but gets worse over time.

This approach creates a specialized endpoint for tailing folder and file changes that overcomes some inherent limitations of the find endpoint.

1.  Streaming, as opposed to buffered -- this means that if loading a large data set requires 15 seconds of scans, increment lines are sent over the wire immediately, rather than buffered, as postgresql performs the scan.  This is more performant for both our servers and psql itself, since it does not require nearly as much intermediate data buffering; only the temporary tables as part of the recursive.  It also means that even if the data set gets quite large, in general we can handle pulling data down in segments and handling it without large memory or json parsing issues.
2. Better use of the USE RECURSIVE mechanism.  Previous path calculation mechanisms cost postgresql considerably due to the way it would run -per- individual result look up, and furthermore the lack of a UNION resulted in duplicate parent directories being produced as part of the query for each lookup.  This use of USE RECURSIVE yields incrementally from a batch of files or folders being searched, and dedupes / prunes paths it has already taken along the way (important for deeply nested folders).
3. Ignoring the ORM intermediate.  The ORM is wonderful for constrained datasets but puts a large overhead in the way of much much bigger datasets, due to the need to instantiate ORM objects, load database rows into them, and then immediately throw all of that away and instantiate yet another hash in the form of the `to_hash` method.  The new approach takes those rows yielded directly from the postgres adapter and simply renders them as json directly to the stream, no intermediate objection construction and subsequent garbage collection.

Generalizing these adaptations to the find endpoint might be a bit harder; at the very least, it would require several steps; the orm logic to generalize the operations, selection of the two data sets as identifiers, then USE RECURSIVE handling and potentially wrapping all of this back up into the ORM objects to match that endpoint's expected outputs.  For now, I'd like to experiment with this simpler, more specialized endpoint before doing all the work of generalizing back to find. 

My intention is to deploy this and test it backing the initial data loading in airflow and measure the performance.


corps:
@coleshaw Correct, it is up to the client to put together full paths.  It guarantees that all parent folders exist in the response as type 'parent'.  This both ends up being more efficient to store for the clients (no repeating common parent directories in every single path) but significantly faster for transfer and processing.

PR #763, opened 2022-03-01
Janus token builder

graft:
This is a small two-parter that I am prepared to partition if controversy ensues.

1) A general cleanup of the Janus api, mostly moving stuff to the `/api` namespace and making most of the root namespace available to the UI. Some token-generating routes have been removed; where possible the appropriate client handlers have been changed. I may have missed something! A few ancient token routes are still in place.

2) A new token-builder and API endpoint which lets you forge new tokens from scratch and have Janus sign them. The UI appears in "settings" if you are a superuser; the API is similarly locked. Otherwise the endpoint is pretty unconstrained (it makes a few rude checks of format), which I feel is okay since only superusers can access it anyway.


etna/lib/etna/clients/janus/client.rb
```diff
@@ -10,8 +10,7 @@ class Janus < Etna::Clients::BaseClient
       def get_project(get_project_request = GetProjectRequest.new)
         json = nil
         @etna_client.get(
-          "/project/#{get_project_request.project_name}",
-          get_project_request) do |res|
+          "/api/admin/#{get_project_request.project_name}/info") do |res|
```
coleshaw, line 13: If you are editing `etna` gem, should you also bump the version in the `etna.gemspec` file?

coleshaw (approved):
Looks good!

PR #764, opened 2022-03-03
Vulcan project page remix

coleshaw:
This PR reorganizes the Vulcan figures landing page a little bit, in a more tabular format. I think it leaves room for actions like "Edit Workflow", figure tags, figure thumbnails, etc. It uses Material UI's Data Grid, so there is built-in functionality for sorting and filtering rows in each table.

vulcan-project-dashboard


graft:
Hmm, I have a bunch of thoughts which I will try to lay out here in advance of discussion on Monday.

Generally, while I think the main idea of what we want to accomplish is present in these changes, I'm not sure two tables is a good idea here, for several reasons.

1. Tables seem to obscure the particular character of rows and make the focus more on particular columns, which is not what we want people to do here.  The lack of style differentiation between columns and the use of a header means if you want to understand what is in a "row", your attention must constantly travel up to the header ("is this the description? title? etc.") as you read across the row.

2. Because the two tables are so similar in construction, I think the overall result is actually counter to our purpose - we want to disambiguate between "figure" and "workflow", but now have made their separate identities seem confused by presenting them in a similar guise.

3. The table gives undue weight to some elements. Currently less than 2% of available space is used for thumbnails, while authors + last modified get 25% of available space.

I would suggest two main changes to counter these:

1. Each row in the table could be a single element (a card, essentially); the header would be eliminated (that is, just use a list of elements, rather than a table). The meaning of the data contents can be indicated by appropriately styled elements rather than a header. E.g., you don't need to tell people that "Dan Bunis" is the "author", they will figure that out so long as "Dan Bunis" is indicated as a separate element via style, so the text "author" is unnecessary - I'd also favor the use of avatar badges instead, where several can be shown as a stack, and are more compact than a full name, which has arbitrary length.

2. The elements in each list (Workflow and Figure) should have a very different visual style to separate them out as different types of content. I'd say the figure (where we can have thumbnails showing actual results) could have more emphasis on the image than the workflow, where it is mostly decorative, and the description has more utility. The difference could thus be, long skinny cards for Workflows, emphasizing the name + description, and square cards for Figures, emphasizing the image + title.

Finally, it's unclear what the scrolling behavior should be - simply a max height for each list? Currently things look nice and succinct, with both contents visible at once, but with enough workflows, "figures" are driven off the page.

coleshaw:
Thanks for the feedback! I do think we would set a max height or rows for each list, to keep them both visible. 

One thing I like about the table is the built in filtering and sorting. I feel like once there are more than a couple dozen figures, people will want to sort or filter them in various ways. If we include just cards, then I would suggest also including some separate filtering widget.

coleshaw:
Reviewed at standup, tasks:

  - [ ] Switch to cards for workflow
  - [ ] Switch to cards for figures
  - [ ] Control button bar for workflows
  - [ ] Gray out image for figures if no thumbnail
  - [ ] Search text field for figures
  - [ ] Tags field for figures (with default `public` selected)
  - [ ] Add tags db migration for figures
  - [ ] "Mark public" tag button in specific figure
  - [ ] For now, only author of the figure can edit or change the title
  - [ ] Change section headers to different color, make bolder
  - [ ] Add "Copy" button inside of a figure (for non-editors)

coleshaw:
So I think I've made all the updates that we talked about on Monday, mostly captured in this GIF:

vulcan-updated-project-dashboard

Also, once you open a specific figure, if you are not the owner, you can only copy it:

Screenshot from 2022-03-09 10-06-14

If you are the owner, you can change its public / private status as well as save it:

Screenshot from 2022-03-09 10-06-33

PR #766, opened 2022-03-09
Production Deploy

corps:


coleshaw (approved):
Looks good! We'll unfortunately have the not-so-nice table view in Vulcan temporarily, for the projects page, until #764 gets merged in and deployed.

PR #769, opened 2022-03-10
VizualizationUI redesign

dtm2451:
Redesigns the organization of inputs into bins of settings categories to help break the many options up into better understandable chunks.

(Uses a similar <Accordian> setup to how our primary inputs are organized.)
VizUI-redesign


dtm2451:
In addition to reorganizing the inputs in the UI into category accordions:
- [x] changes the default scaling option's name from "as is" to "linear"
- [ ] Adds restriction of the options provided to users for x_by / y_by / color_by based on continuous/discrete needs of the plot type
  - Currently non-inclusive of any changes made during the data transformation step.  I need to rethink the continuous/discrete options determination method, and likely have this be performed in the react side instead of python.

PR #770, opened 2022-03-10
Timur query bug 768

### LINKED ISSUE #768: Timur Query -- bug with CodeMirror, when selection highlighted
When you have a part of the CodeMirror block highlighted in Query (i.e. `user_columns` or `query`), and then try to click elsewhere on the page to edit the query, you can get an exception that causes the page to crash:

```
Uncaught RangeError: Selection points outside of document
    checkSelection webpack:///./node_modules/@codemirror/state/dist/index.js?:876
    Transaction webpack:///./node_modules/@codemirror/state/dist/index.js?:1570
    resolveTransaction webpack:///./node_modules/@codemirror/state/dist/index.js?:1744
    update webpack:///./node_modules/@codemirror/state/dist/index.js?:1923
    CodeMirror webpack:///./node_modules/rodemirror/dist/index.js?:110
    React 6
    unstable_runWithPriority webpack:///./node_modules/scheduler/cjs/scheduler.development.js?:653
    React 3
    workLoop webpack:///./node_modules/scheduler/cjs/scheduler.development.js?:597
    flushWork webpack:///./node_modules/scheduler/cjs/scheduler.development.js?:552
    performWorkUntilDeadline webpack:///./node_modules/scheduler/cjs/scheduler.development.js?:164
    EventHandlerNonNull* webpack:///./node_modules/scheduler/cjs/scheduler.development.js?:187
    <anonymous> webpack:///./node_modules/scheduler/cjs/scheduler.development.js?:857
    js https://timur.ucsf.edu/js/timur.bundle.js:14528
```

coleshaw:
This PR closes #768 .

The root cause of the issue was when you had a selection in the CodeMirror query preview text area, and then updated the query in such as way as to reduce the text length -- i.e. delete a filter or column. If the highlighted selection included the deleted text, CodeMirror would throw an exception because the text was now shorter than the specified selection, as if you had selected outside of the document. CodeMirror throwing this exception would cause the app to crash, causing a white screen.

There was also a similar issue when you added to the query, though less drastic (no white-screen of death). For example, if you had `query` + `user_columns` selected, adding a filter would lead to losing some of the `user_columns` selection. Potentially causing user confusion, since their selection would change.

This PR turns the `selected` attribute into a controlled value via `useState` hook, and resets it whenever the text in CodeMirror changes. This should prevent both app crashes and user confusion due to the selection changing relative to the text.

### DIFF
--- a/timur/lib/client/jsx/components/query/query_results.tsx
+++ b/timur/lib/client/jsx/components/query/query_results.tsx
@@ -1,4 +1,4 @@
-import React, {useContext, useMemo} from 'react';
+import React, {useCallback, useContext, useMemo, useState} from 'react';
 import Grid from '@material-ui/core/Grid';
 import {makeStyles} from '@material-ui/core/styles';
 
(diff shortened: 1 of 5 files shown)


timur/lib/client/jsx/components/query/query_results.tsx
```diff
@@ -39,6 +40,28 @@ const useStyles = makeStyles((theme) => ({
   }
 }));
 
+// https://reactjs.org/docs/error-boundaries.html
+// For #768, because just resetting the selection doesn't
+//    happen early enough in the CodeMirror lifecycle...?
+class ErrorBoundary extends React.Component {
```
coleshaw, line 46: Yeah, also had to wrap the `<CodeMirror />` component in an `<ErrorBoundary />`, because it seemed like the `selection` updates just weren't happening quickly enough in the component lifecycle to avoid the error state. It's not a great solution, but I can't find any "real" API hooks that occur early enough in the CodeMirror lifecycle to handle this bug.

PR #772, opened 2022-03-12
Figure thumbnails

graft:
This sets up thumbnails to appear on figures in the figures list. The interface is quite explicit (the workflow author must specify a thumbnail file for every output in CWL), somewhat aided by the output helper. (NB: I'm unclear on how or where we should document such a helper.)

So far it only thumbnails plotly plots. Other types will have to be added later (say, png, pdf or svg outputs), as so far they don't appear in any workflows. Seems to work for scRNAseq & visualization workflows, though the tiny UMAP looks awful.

To put it mildly, some style issues remain, like: the thumbnails are written in the dimension of 300x200 every time, no matter what the plot output is. This means they pretty much only look good at this size. Meanwhile the card is flexible; it appears this is simply a result of ImageList, which defines columns in terms of window width. This may be untenable. In short, it does not yet look great. I'd like to tackle some of these issues elsewhere.


vulcan/lib/client/jsx/components/dashboard/figure.tsx
```diff
@@ -80,7 +86,7 @@ const Figure = ({
   const {canEdit} = useUserHooks();
 
   const workflow = workflows
-    ? workflows.find((w) => w.name == figureSession.workflow_name)
+    ? (workflows.find((w) => w.name == figureSession.workflow_name) || null)
```
coleshaw, line 89: super-nit: I actually think there is [an existing selector](https://github.com/mountetna/monoetna/blob/master/vulcan/lib/client/jsx/selectors/workflow_selectors.ts#L30) that might work instead of this?

vulcan/lib/client/jsx/components/workflow/user_interactions/outputs/plot.jsx
```diff
@@ -34,8 +34,7 @@ const PlotOutput = ({data: plots}) => {
                 zerolinewidth: 1,
                 linecolor: '#000'
               },
-              width: 800,
-              height: 800,
+              autosize: true,
```
coleshaw, line 37: Hm, I thought UMAPs had to be square. Does changing this affect the UMAP dimensions and thus utility?

vulcan/lib/server/models/figure.rb
```diff
@@ -6,15 +6,48 @@ def self.next_id
...
+      cache = {}
+
+      session.workflow.steps.map do |step|
+        case step.ui_output_name
+        when 'plotly.cwl'
+          step_target = session.orchestration.build_target_for(step.id, cache)
+
+          next unless step_target.input_files.length > 1
+
+          thumbnail = step_target.input_files[1]
```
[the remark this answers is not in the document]
graft: Sorry, I meant to write this issue up. Yes, this seems brittle, and I also wasn't sure what would be better. We haven't previously required inputs to ui-outputs to be named, but that might be less brittle than this. I think this is mostly here as a placeholder that works to demo the idea, pending a better notion.

vulcan/lib/client/jsx/components/workflow/user_interactions/outputs/plot.jsx
graft: Err, forgot about this change, I made this while fiddling with the thumbnail image - I will pull it out. While it's true the UMAP looks best square, I don't think this assumption is true for plots in general. What would be nice is if the workflow author could define a fixed aspect ratio and the user could scale the plot size responsively by scaling the window as they liked. But I don't think this quite works that way yet.

graft:
> Hm, so I'm trying to test this with the visualization workflow and the example project. But I'm not getting a thumbnail attached to the figure (the regular plot does appear in the UI widget). Do I need to do something special? I think I did `make down` and `make up`, so everything should be up to date...

Other than rebuilding the vulcan archimedes image, I don't think you should have to do anything special, other than re-run the step that generates the plot itself.

coleshaw:
Hm....so I get the thumbnail file being generated, but it's not populated in the figure model instance (and hence does not show up in the dashboard)... I'll try to narrow down where the issue is, perhaps I've changed my local code somehow.

coleshaw:
Ah, so I figured out that on my end, I hadn't saved the figure yet to the database. So I was getting the thumbnail file for my session, but since my inputs weren't saved, the server wasn't able to fetch the correspoding file. Kind of a tricky edge-case, I guess...

coleshaw:
Looks pretty neat, to see the thumbnails in the dashboard!

graft:
> Ah, so I figured out that on my end, I hadn't saved the figure yet to the database. So I was getting the thumbnail file for my session, but since my inputs weren't saved, the server wasn't able to fetch the correspoding file. Kind of a tricky edge-case, I guess...

I think this is the intended behavior since, at any given moment, there might be several completely different thumbnails built from the same saved starting point, and there is no good way to pick one of them (or even to compute their cache keys to know they exist, which the server/db lacks the information for) until the state is saved.

coleshaw:
Yeah, that makes sense. Just perhaps a little confusing from a user standpoint, though I guess we don't tell them how thumbnails are generated (or that they will be generated during the workflow), so perhaps they won't have the same assumptions that I had.

vulcan/lib/client/jsx/components/workflow/user_interactions/outputs/plot.jsx
dtm2451: Perhaps even a per-workflow **suggested** aspect-ratio, but then we let the user adjust even around that?  Thinking of the general plotting workflow as an example, a y_plot with 3 groups on the x-axis would need a different aspect ratio than one with 20, but it's the same workflow.cwl, so it'd be nice to give that 'complete' flexibility.

vulcan/lib/server/models/figure.rb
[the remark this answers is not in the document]
coleshaw: I almost wonder if this should be instead, some form of a top-level `output`, with a custom `type`?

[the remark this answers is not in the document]
coleshaw: I think the visualization workflow includes the plot as an `output`, as an example...which, if you think of the top-level `output`s as artifacts of the workflow, I think makes sense...

PR #774, opened 2022-03-14
Zc/rollup metrics

corps:
Ability to great 'accumulator' etls that process the results of other etls into summaries that can be measured and exported to prometheus.

### DIFF
--- a/airflow/opt/providers/etna/etna/__init__.py
+++ b/airflow/opt/providers/etna/etna/__init__.py
@@ -15,7 +15,7 @@ from datetime import datetime
 
 
 def __getattr__(name):
-    from .dags.decorators import system_dag
+    from .dags.decorators import system_dag, dag, rollup_dag
     from .operators import run_on_docker
     from .etls.decorators import metis_etl
     from .etls.metis import (
@@ -30,6 +30,7 @@ def __getattr__(name):
         get_project_slack_hook,
     )
     from .hooks.etna import UpdateRequest
+    from .operators.rollup_xcom_operator import rollup
 
     __all__ = list(k for k in locals().keys() if k != "name") + [
         "dags",
@@ -51,7 +52,7 @@ def __getattr__(name):
 
 # trick the static checker.
 if globals().get("notathing", False):
-    from .dags.decorators import system_dag
+    from .dags.decorators import system_dag, dag, rollup_dag
     from .operators import run_on_docker
     from .etls.decorators import metis_etl
     from .etls.metis import (
@@ -66,3 +67,4 @@ if globals().get("notathing", False):
         get_project_slack_hook,
     )
     from .hooks.etna import UpdateRequest
+    from .operators.rollup_xcom_operator import rollup
--- a/airflow/opt/providers/etna/etna/dags/decorators.py
+++ b/airflow/opt/providers/etna/etna/dags/decorators.py
@@ -49,6 +49,7 @@ def dag(
             start_date=start_date,
             end_date=end_date,
             schedule_interval=schedule_interval,
+            max_active_runs=1,
             **kwds,
         ) as dag:
             inject(fn, inject_params)
@@ -56,12 +57,9 @@ def dag(
 
     return instantiate_dag
 
-
-def system_dag(interval: timedelta) -> Callable[[Callable], DAG]:
+def rollup_dag(interval: timedelta, labels: List[str]) -> Callable[[Callable], DAG]:
     """
-    Decorator that converts the decorator function into a DAG by the same name
-    using any tasks instantiated within the function.  For use with 'system'
-    level dags whose owner should be 'administration'.
+    `system_dag` variant that sets the dag's labels so that a rollup of it's XCom values are used
     """
 
     def instantiate_dag(fn):
@@ -70,7 +68,27 @@ def system_dag(interval: timedelta) -> Callable[[Callable], DAG]:
             schedule_interval=interval,
             default_args=dict(
                 owner="administration",
-                retries=3,
+                retries=10,
+            ),
+            tags=["rollup"] + [f"label:{l}" for l in labels],
+            catchup=False,
+        )(fn)
+
+    return instantiate_dag
+
+
+def system_dag(interval: timedelta) -> Callable[[Callable], DAG]:
+    """
+    `dag` variant that sets the owner to 'administration' and a high number of retries.
+    """
+
+    def instantiate_dag(fn):
+        return dag(
+            start_date=system_epoch,
+            schedule_interval=interval,
+            default_args=dict(
+                owner="administration",
+                retries=10,
             ),
             # default_args=dict(
             # on_failure_callback=notify_slack_dag_callback("failed: "),
--- a/airflow/opt/providers/etna/etna/get_provider_info.py
+++ b/airflow/opt/providers/etna/etna/get_provider_info.py
@@ -1,5 +1,10 @@
 # https://airflow.apache.org/docs/apache-airflow-providers/
+from etna.metrics.rollup_metrics import register_rollup_metrics
+
+
 def get_provider_info():
+    register_rollup_metrics()
+
     return {
         "package-name": "etna",
         "name": "etna",
--- a/airflow/opt/providers/etna/etna/metrics/rollup_metrics.py
+++ b/airflow/opt/providers/etna/etna/metrics/rollup_metrics.py
@@ -25,7 +25,7 @@ class RollupMetricsCollector:
         bag = DagBag(read_dags_from_db=True)
         bag.collect_dags_from_db()
         for dag_name, dag in bag.dags.items():
-            if 'rollup' not in dag.tags:
+            if dag.tags is None or 'rollup' not in dag.tags:
                 continue
 
             for task_id in dag.task_ids:
--- a/airflow/opt/providers/etna/etna/metrics/rollup_metrics.py
+++ b/airflow/opt/providers/etna/etna/metrics/rollup_metrics.py
@@ -2,11 +2,12 @@ import re
 from datetime import datetime
 from typing import Generator, Iterable, Tuple, Dict
 
-from airflow.models import DagBag, XCom
+from airflow.models import DagBag, XCom, BaseOperator
 from prometheus_client import REGISTRY, Metric
 from prometheus_client.metrics_core import GaugeMetricFamily
 from prometheus_client.samples import Sample
 from pytz import utc
+
 
 rollup_label_metric_re = re.compile(r'label:(.*)')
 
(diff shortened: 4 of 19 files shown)


airflow/opt/providers/etna/etna/__init__.py
```diff
@@ -15,7 +15,7 @@
```
corps, line 16: Exporting the most useful pieces of this into the top level module to make imports easy:

```
from etna import dag, rollup_dag, etc
```

airflow/opt/providers/etna/etna/dags/decorators.py
```diff
@@ -49,19 +49,37 @@ def instantiate_dag(fn):
             start_date=start_date,
             end_date=end_date,
             schedule_interval=schedule_interval,
+            max_active_runs=1,
```
corps, line 52: Restricting the number of active runs so that the scheduler doesn't run multiple of overlapping instances of a single ETL.  Problem was that the genentech upload etl takes a really long time, so on the next interval it scheduled a second instance of the ETL which ... just duplicated the upload lol.   This is because the genentech etl upload uses polyphemus cursors still and not date ranges provided by airflow.  Low priority change, so I'd rather just prevent the concurrent execution.

It is possible to make multiple dag runs concurrently safe, but it is simply generally safer to keep concurrent runs of a single dag at a serial level.  We can revisit in the future.

airflow/opt/providers/etna/etna/dags/decorators.py
```diff
@@ -49,19 +49,37 @@ def instantiate_dag(fn):
...
+
+    def instantiate_dag(fn):
+        return dag(
+            start_date=system_epoch,
+            schedule_interval=interval,
+            default_args=dict(
+                owner="administration",
+                retries=10,
+            ),
+            tags=["rollup"] + [f"label:{l}" for l in labels],
```
corps, line 73: Really, all this dag does is add some labeling so that the prometheus exporter 'picks up' on the generated XComs of tasks inside a rollup_dag.  See below.

airflow/opt/providers/etna/etna/dags/decorators.py
```diff
@@ -70,7 +88,7 @@ def instantiate_dag(fn):
             schedule_interval=interval,
             default_args=dict(
                 owner="administration",
-                retries=3,
+                retries=10,
```
corps, line 91: There is already some back off logic built into airflow, so raising the retries just helps persist jobs that have intermittent failures (uploads, network issues)

airflow/opt/providers/etna/etna/get_provider_info.py
```diff
@@ -1,5 +1,10 @@
 # https://airflow.apache.org/docs/apache-airflow-providers/
+from etna.metrics.rollup_metrics import register_rollup_metrics
+
+
 def get_provider_info():
+    register_rollup_metrics()
```
corps, line 6: Entry point for airflow providers, registers the prometheus metric generator.  This will then "just work" with the existing prometheus-airflow tooling.

airflow/opt/providers/etna/etna/metrics/rollup_metrics.py
```diff
@@ -0,0 +1,58 @@
...
+from pytz import utc
+
+
+rollup_label_metric_re = re.compile(r'label:(.*)')
+
+# magma_record_attributes[model_name, restricted, attribute_name, attribute_type] = count
+
+class Rollup:
+    def measure(self) -> Iterable[Tuple[Dict[str, str], int]]:
+        return []
```
corps, line 18: A Roillup is really just any object that has a way of 'measuring' metrics from itself.  This class is an interface that does not directly provide behavior (an ABC in python parlance)

airflow/opt/providers/etna/etna/metrics/rollup_metrics.py
```diff
@@ -0,0 +1,58 @@
...
+
+class RollupMetricsCollector:
+    def describe(self):
+        return []
+
+    def collect(self) -> Generator[Metric, None, None]:
+        bag = DagBag(read_dags_from_db=True)
+        bag.collect_dags_from_db()
+        for dag_name, dag in bag.dags.items():
```
corps, line 27: Collects all known dags and finds any that are labeled for rollup and generate RollupXcomOperator tasks.

airflow/opt/providers/etna/etna/metrics/rollup_metrics.py
```diff
@@ -0,0 +1,58 @@
...
+                    dag.description or f"Quantity pulled from {dag_name}.{task_id}",
+                    labels=[
+                        match.group(1) for match in
+                        (rollup_label_metric_re.match(t) for t in dag.tags)
+                        if match is not None
+                    ],
+                )
+
+                now = datetime.now().replace(tzinfo=utc)
+                rollup_value = XCom.get_one(dag_id=dag_name, task_id=task_id, include_prior_dates=True, execution_date=now)
```
corps, line 47: Creates a gauge metric by taking the latest XCom result of the task and measuring it.

airflow/opt/providers/etna/etna/operators/rollup_xcom_operator.py
```diff
@@ -0,0 +1,106 @@
...
+        self.target_task_id = target_task_id
+
+    @provide_session
+    def execute(self, context: Any, session: Session = None):
+        start, stop = get_batch_range(context)
+
+        result: Optional[RollupState] = None
+
+        ti: TaskInstance = context['ti']
+        previous: Optional[TaskInstance] = ti.get_previous_ti(state=State.SUCCESS)
```
corps, line 67: Essentially, a rollup is a task using an operator `concat` to summarize a collection of XCom results of other tasks.

Idea is that eg a magma metric generator takes samples of the state of magma records which are 'rolled up' into a larger picture assessment of the current completeness of a data set.

coleshaw: How often do the rollups run? Does this mean that they are like Prometheus metrics (since above you note that it takes the last XCom result of a given task) -- sampling of other task results, not a comprehensive summation?

airflow/opt/providers/etna/etna/tests/cassettes/test_metis_files_etl/test_metis_files_etl_e2e.yaml
```diff
@@ -1,4 +1,46 @@
 interactions:
+- request:
+    body: '{"token_type": "task", "project_name": "mvir1", "ready_only": true}'
```
coleshaw, line 3: random question / nit: is `ready_only` a typo somewhere in the spec that produced this cassette, or perhaps in this YAML file? Should be `read_only`? Maybe that doesn't affect anything, since all task tokens are read_only...

Note that line 45 has the same `ready_only` text.

airflow/opt/providers/etna/etna/operators/rollup_xcom_operator.py
corps: Rollups are just standard DAG tasks, so they run as often as you configure them.

They can be a comprehensive summation -- it entirely depends on how you implement the concat operation.  The operator itself attempts to roll up all xcom results of another dag.  How you sum that is dependent on the rollup you are building.  So for completion metrics, this will be a full summation.

airflow/opt/providers/etna/etna/tests/cassettes/test_metis_files_etl/test_metis_files_etl_e2e.yaml
corps: oh snap, good catch.  I think I actually have been fetching write tokens by accident!  I'll make the change to the dictionary.

airflow/opt/providers/etna/etna/tests/cassettes/test_metis_files_etl/test_metis_files_etl_e2e.yaml
```diff
@@ -1,4 +1,46 @@
...
+        - application/json
+      User-Agent:
+        - python-requests/2.26.0
+      authorization:
+        - XXX-Auth
+    method: POST
+    uri: https://janus.ucsf.edu/api/tokens/generate
+  response:
+    body:
+      string: eyJhbGciOiJSUzI1NiJ9.eyJlbWFpbCI6InphY2hhcnkuY29sbGluc0B1Y3NmLmVkdSIsIm5hbWUiOiJaYWNoYXJ5IENvbGxpbnMiLCJwZXJtIjoiZTptdmlyMSIsImZsYWdzIjoiaW5nZXN0O3Z1bGNhbjt0aW11cmFkdmFuY2VkO3F1ZXJ5cGxvdCIsImV4cCI6MTY0Njg1MDQ5MCwidGFzayI6dHJ1ZX0.rX6ZIbgxEdTowBKyfVTty6ZSTBdh83jPvQ-9kFIEdwHbWB3nzPIxjw8Jo5FqsXbPT33-_TYTd6Am__baO0EVGwxflW5ggjiapPcc0joy7zfqMletDY9dAwLa07gwtXXDk3Ncn8X1gmPeR3CKxgwQBh7zDjJ4IRnab68LVCnvog5u9NSONtyoBeIzP_9hNsYbGA8P_xxV0_JDl-uZuf4gHVOA1RDtsArvHtJfjCpTgct6_5_J2RrQcVeow5wG3sov2IV7zuAxcL2xKHVhjXeYR5FkLEixVjyOJj_jSskazrci7DFbm62OMtJGfk2MYLkMngW_9uUaVQWPwRC57BEaa6OMWkC5hs21XvAxPg6o7HVb-TbJ9bMI7IeBcOVt_5BDKFVcQIvR6ieigFR1ZlcL9D0VGgQFD_Exp2BochnXNcrzzIgQZmzkWbpP6u-43XYHcon-RgmwBBemOaKGddlFPdL2Uh4vwGeoeJXBE4HbNJRb97xtpU-TFjm1TT9KmztGpHOe6i1q476h_tTxJ7sVvukAQNkTaOPeewfjGOfU82HAT7VrvoMwdKQK07t_PFneXZCyvTPE1uDcuiEsK19VzHpB_sud7YsTTEhvOBKJ7Fb7qulGsvgBcUr7XR_IVNQPtCfrbv6TdX5TpUWDY0pOFXQ3tsUUMppOaaBFlwm0jmI
```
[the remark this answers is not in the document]
corps: It's intentionally an expired token.  The python etna client does not validate if at oken is expired or not so it won't throw any exceptions.  This token is 'opaque' in that it's only purpose is to validate that it is sent up to the server "as is" in the cassette.

airflow/opt/providers/etna/etna/tests/cassettes/test_metis_files_etl/test_metis_files_etl_e2e.yaml
coleshaw: Huh, interesting that it is an `e` token (verified by decoding) -- it may actually be always `e` because of that supereditor thing you fixed last Friday.....it will never be `read_only`, even if you pass in the flag? Perhaps need to rework that logic in `token.rb`?

corps: Yeah, I can get a fix in and write some more tests on this branch.

corps: @coleshaw Addressed these in latest commit

coleshaw (approved):

PR #775, opened 2022-03-16
Figures page layout + style changes

graft:
This cleans up a bunch of style-related issues on the figures page, mainly:

1) introducing tags into figure cards & unifying tag format in workflow and figure cards
2) making figure cards have a fixed-width and resolve thumbnail size issues
3) adding some pointless color zazz


dtm2451:
Lookin good!

Is it possible to ensure the full figure thumbnail is shown?  Even though it would create white/wasted-space, and likely no words would be readable anyway, I think it would still be nice to see the bounds of the plot and hints of labels/titles/legends to give users the full picture of what was produced.

coleshaw:
Nice, I see you got the color-tag thing working!

super-nit, but the tags in the figure card seem oddly aligned, perhaps because the author avatar takes up so much vertical space. I kind of expect the tags to be vertically in the middle, aligned with the avatar? :shrug: 

Screenshot from 2022-03-16 09-22-18

graft:
> Is it possible to ensure the full figure thumbnail is shown?
I tried this at first, titles take up too much whitespace, making the plot area a little stamp, and it is impossible to scale them consistently to a size that looks decent.

PR #776, opened 2022-03-16
use multiline input for user key, to avoid issue with escape chars

coleshaw:
This PR is a hotfix for Janus -- it seems impossible to upload a user's RSA key via the UI, since the input is an `input` instead of `textarea`. So if you try to upload your key in raw format, the newlines are pasted as spaces (and you get `Invalid key format` error). If you try to escape yourself and add `\n` for newlines, those get escaped as `\\n` and you get the same `Invalid key format` error. This PR switches to use a `textarea` input, which accepts the newlines from pasting the raw key and does not escape them.

I'm going to push this directly to production to help solve a DS's immediate problem, but opening this PR for review.

### DIFF
--- a/janus/lib/client/jsx/__tests__/janus-settings.test.jsx
+++ b/janus/lib/client/jsx/__tests__/janus-settings.test.jsx
@@ -11,18 +11,26 @@ describe('JanusSettings', () => {
   beforeEach(() => {
     store = mockStore({
       user: {
-	email: "janus@two-faces.org",
-	name: "Janus Bifrons",
-	permissions: { }
+        email: "janus@two-faces.org",
+        name: "Janus Bifrons",
+        permissions: { }
       }
     });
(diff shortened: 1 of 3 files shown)


janus/lib/client/jsx/__tests__/janus-settings.test.jsx
```diff
@@ -11,18 +11,26 @@ describe('JanusSettings', () => {
...
-	name: "Janus Bifrons",
-	permissions: { }
+        email: "janus@two-faces.org",
+        name: "Janus Bifrons",
+        permissions: { }
       }
     });
   });
 
+  const createNodeMock = element => {
```
coleshaw, line 21: This is a workaround for a known issue with MUI multiline `TextField` for testing...

https://github.com/mui/material-ui/issues/16491

Issue #765, opened 2022-03-07
Vulcan data transformation widget -- performance constraints?


coleshaw:
Trying to use the Vulcan data transformation widget and visualization workflow with full gene expression matrices, and couple of performance issues:

* Browser complains when trying to open the table, that a script is slowing down the page. So perhaps we need to tweak the table code somehow.
* Cannot commit and run with the raw data, because the data frame causes local storage to exceed the max permissible amount (on FF at least).
* Downloading large amounts of data from Magma causes the Magma proxy to time out (i.e. all gene expression matrices for IPI), and the Archimedes cell throws a Python exception.

So, not sure what folks will expect to be able to input into Vulcan visualizations, but we will have to work around some stuff for large data sets. Or, is there some limit in which we say, you should be using a command line / full analysis pipeline, and Vulcan is just for simple exploration?

coleshaw:
Definitely some tweaking of the [table performance is possible](https://handsontable.com/docs/performance/#overview).

Perhaps we can investigate IndexDB in lieu of local storage, as well.

corps:
@coleshaw If you are looking into IndexedDB, I highly recommend https://dexie.org/

It has a typescript bindings, and the wrapper is 'minimal' -- it provides the most useful aspects of the base api in a convenient way without hiding underlying pieces.

graft:
I'm loathe to boot up a method to store gigabytes of data locally and run calculations on them in the browser, which seems like a bad pattern leading to many headaches. If possible, it might be better to move most of the calculation server-side - i.e., you could just show a head of the table with a sample of the data, allow them to compose their transformation calculations, then ship the formulas from the first row or some such. Not sure how you might effect running the table calculations on the server - either running the handsontable js or running an equivalent python lib (e.g. @vinci1it2000/formulas) - since handsontables' HyperFormulas claims to be fully compatible with Excel, this might work, though I imagine there will be some holes.

coleshaw:
Hm, that's a good point. But interpreting partial equations on the server seems complicated. How would you deal with things like `sum across an entire column`, or other formulas that span a range?  Do you make an assumption about the range?

Is there an assumption that folks will never use formulas to fill holes, but only for populating entire columns?

graft:
[...]

PR #777, opened 2022-03-16
Vulcan handsontable perf

coleshaw:
Okay, kind of a smallish PR. Went through the handsontable documentation and applied their recommendations for improving performance. Also did a small tweak for pre-loading a subset of data when given a large data frame, since it appears that [table rendering is synchronous](https://github.com/handsontable/handsontable/discussions/7708), so loading large data sets causes the UI to freeze. This implementation at least will show the user a subset of the data while the rest loads afterwards ... still kind of locks, but you have something to look at.

In reality, this addresses the first part of #765 , and is probably only impactful for medium sized data frames that people may want to hand-edit. Small data frames will load and behave as before, with no degraded performance. Large data frames will run into the other performance limitations, and will be addressed separately (in a vector-calculation PR).


vulcan/lib/client/jsx/components/workflow/user_interactions/inputs/use_handson_table.tsx
```diff
@@ -0,0 +1,56 @@
...
+  UndoRedo,
+  registerPlugin
+} from 'handsontable/plugins';
+
+// import translations
+import {registerLanguageDictionary, enUS} from 'handsontable/i18n';
+
+export default function useHandsonTable() {
+  // register individual translations
+  registerLanguageDictionary(enUS);
```
corps, line 36: Should these be in a `useEffect() => {}, [])` ?

PR #779, opened 2022-03-18
Polyphemus redcap config UI tweaks

coleshaw:
This PR makes some small changes to the REDCap config UI, to hopefully give a little more user-help.

* Run panel:
  * Save button disabled until everything filled in
  * Model names pre-populated (along with `all` option) in the `model_names` field
  * Defaults of run `Once` and `default` mode  
  * Explanatory description for the modes, so users better understand effects of each
* "Saved" message on successful save so user can see feedback to their button click (for each panel)
* Secrets clear out after save, so it's obvious something changed
* Some tooltips on the config panel to clarify what some of the options mean and when they would be used
* Sort some select options in the config panel, so more consistent and easier to find the desired option

I'd really like the config panel to have more inline help or documentation, but I'm currently out of ideas for how to provide those in a non-cluttering fashion.

### DIFF
--- a/polyphemus/lib/client/jsx/etl/etl-config.tsx
+++ b/polyphemus/lib/client/jsx/etl/etl-config.tsx
@@ -1,5 +1,5 @@
-import React, { useState, useCallback } from 'react';
-import { json_post } from 'etna-js/utils/fetch';
+import React, {useState, useCallback} from 'react';
+import {json_post} from 'etna-js/utils/fetch';
 
 import Typography from '@material-ui/core/Typography';
 import Card from '@material-ui/core/Card';
@@ -31,19 +31,19 @@ import ConfigurePane from './configure-pane';
 import RemovePane from './remove-pane';
 import LogsPane from './logs-pane';
 import SecretsPane from './secrets-pane';
-import { formatTime, runTime } from './run-state';
+import {formatTime, runTime} from './run-state';
 
-const StatusIcon = ({status}:{status:string}) => {
-  let IconComponent:any;
+const StatusIcon = ({status}: {status: string}) => {
+  let IconComponent: any;
   if (status == 'completed') IconComponent = CheckIcon;
   else if (status == 'error') IconComponent = ErrorIcon;
   else if (status == 'pending') IconComponent = ScheduleIcon;
   else return null;
-  
-  return <IconComponent size='small'/>
-}
 
-const useStyles = makeStyles( theme => ({
+  return <IconComponent size='small' />;
+};
+
+const useStyles = makeStyles((theme) => ({
   title: {
     color: 'goldenrod',
     justifyContent: 'flex-end',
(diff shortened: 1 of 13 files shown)


graft:
Hmm just getting into it, but I can't make the run `mode` and `model_names` things work, I just get `undefined, undefined, undefined` for mode and `model_names` has no autocompletion.

coleshaw:
Did you restart your Vulcan instance? I made some changes to the `redcap_model.rb` file, too, so it returns slightly more detailed information to the UI widgets.

polyphemus/lib/models/redcap_job.rb
```diff
@@ -7,8 +7,21 @@ def self.as_json
...
+            value: 'strict',
+            description: 'copy REDCap data and delete Magma records not in this run'
+          }, {
+            value: 'existing',
+            description: 'only copy REDCap records that currently exist in Magma'
+          } ],
+          model_names: {
+            type: 'options',
+            value: 'model_names'
+          },
```
graft, line 24: I might summarize: 
```
default: update existing and append new records,
strict: discard all existing records and append new records,
existing: only update existing records
```

polyphemus/lib/client/jsx/etl/run-pane.tsx
```diff
@@ -31,108 +41,245 @@ const useStyles = makeStyles( theme => ({
...
+          <MenuItem key={opt.value} value={opt.value}>
+            {`${opt.value}${opt.description ? ` - ${opt.description}` : ''}`}
+          </MenuItem>
+        ))}
+      </Select>
+    );
+  } else if (_.isObject(opts)) {
+    if (opts.type && opts.type === 'options' && opts.value === 'model_names') {
+      const modelNames = ['all'].concat(Object.keys(config).sort());
+      return (
```
graft, line 92: Hmm, might be brittle, although it's unclear to me if we'll ever use this run-pane interface with another loader/config format anyway.

coleshaw: Yeah, that's a good point. Hopefully we won't have super-complicated forms like this for other stuff....hard to generalize it.

polyphemus/lib/client/jsx/etl/etl-config.tsx
```diff
@@ -86,74 +82,152 @@ const useStyles = makeStyles( theme => ({
...
-    r => r.then(({error}:{error:string}) => setError(error))
+  const classes: any = useStyles();
+
+  const [_, postUpdate] = useAsyncWork(
+    function postUpdate(update: any) {
+      clearMessages();
+      return json_post(`/api/etl/${project_name}/update/${name}`, update)
+        .then((etl) => {
+          onUpdate(etl);
+          setMessage('Saved!');
```
graft, line 121: This is there once you save but doesn't go away if you reset or change the state, so some confusion remains. Could the buttons be grayed instead if there are no changes?

coleshaw: Oh, good catch! I only tested that it goes away if you navigate to a different panel. Thanks, will update how that works.

coleshaw: Hm...grayed from the last saved state? Probably, will have to see how to recalculate that.

graft (approved):
Nice set of improvements, looks good to me.

PR #784, opened 2022-03-24
Reconnect thumbnail output for scRNAseq workflow

dtm2451:
Maybe we push before Teusday, maybe not.


dtm2451:
Seems to work in my dev!

coleshaw:
Nice catch!

coleshaw (approved):

PR #786, opened 2022-03-28
Adding size column to datablocks

corps:
Main motivation is to make it much easier to calculate various per project size statistics.

While it is possible to calculate this with the `actual_size` method there is a fairly non trivial overhead when doing this through ruby on a per file basis.  Since datablocks are immutable anyways, adding this attribute would make calculations much more accessible and cheaper.

### DIFF
--- a/metis/lib/commands.rb
+++ b/metis/lib/commands.rb
@@ -39,6 +39,56 @@ class Metis
     end
   end
 
+  class ManualUpload < Etna::Command
+    string_flags << "--project_name"
+    string_flags << "--bucket_name"
+    string_flags << "--path"
+
+    def execute(file, project_name:, bucket_name:, path:)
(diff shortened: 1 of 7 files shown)


metis/lib/commands.rb
```diff
@@ -39,6 +39,56 @@ def execute
...
+    def execute
+      i = 0
+      Metis::DataBlock.where(size: nil).each do |db|
+        db.update(size: db.actual_size || 0)
+
+        i += 1
+        if i % 100 == 0
+          puts "Processed #{i}"
+        end
+        puts "Processed #{i}"
```
graft, line 81: Is this debugging stuff?

corps: the command was intended for backfill, I added some output while testing locally, can remove this particular line

graft (approved):

PR #793, opened 2022-03-31
Vulcan mui moves tests

coleshaw:
I think this PR fixes the broken specs to work with the new MUI components, while maintaining the intent of the original specs :crossed_fingers: . @dtm2451 , could you verify the Storybook / real behavior remains as expected? Especially this component: [vulcan/lib/client/jsx/components/workflow/user_interactions/inputs/select_autocomplete.tsx](https://github.com/mountetna/monoetna/pull/793/files#diff-76917674a42430f8a96f474ac217618564f153109c6ba622ce7b221a101ba55aR73), because I changed the `filterOptions` type to make the Typescript compiler happy ... not sure if I did the right thing?


vulcan/lib/client/jsx/components/workflow/user_interactions/inputs/select_autocomplete.tsx
```diff
@@ -70,7 +70,7 @@ export default function SelectAutocompleteInput(
       clearOnBlur={true}
       options={options.display}
       // disable filtering on client
-      filterOptions={(x: typeof options) => x}
+      filterOptions={() => options.filtered}
```
coleshaw, line 73: Should this be more like `(x: any) => x`?

vulcan/lib/client/jsx/components/workflow/user_interactions/inputs/select_autocomplete.tsx
```diff
@@ -70,7 +70,7 @@ export default function SelectAutocompleteInput(
       clearOnBlur={true}
       options={options.display}
       // disable filtering on client
-      filterOptions={(x: typeof options) => x}
+      filterOptions={(x: string[]) => x}
```
dtm2451, line 73: Ohhh this must have fallen out-of-date and is remnant from before I would have turned options from a 'string[]' itself into the container of filtered and display 'string[]'s.  :+1: seems right.  The goal right here is to turn off MUI Autocomplete's internal filtering, because we do our own.  Our filtering results get passed in via `options={options.display}` above.

dtm2451:
Seems good to me in manual testing!

dtm2451 (approved):

PR #797, opened 2022-04-02
fix redcap config tab sorting

graft:
Tiny bugfix - the redcap config seems to sort model tabs alphabetically, but then selects tabs based on their given order, resulting in selecting the wrong model config without realizing it. Edits to this work fine and persist correctly, but the confusion will lead to error.


coleshaw:
Ah, sorry! I must have introduced the bug in my previous PR 😢 Good catch -- that would be confusing to find and deal with.

coleshaw (approved):

PR #798, opened 2022-04-05
Fix for cancel helper

corps:
Adding comments to explain

### DIFF
--- a/etna/packages/etna-js/utils/cancellable_helpers.jsx
+++ b/etna/packages/etna-js/utils/cancellable_helpers.jsx
@@ -1,9 +1,9 @@
 import 'regenerator-runtime';
 import {Cancellable, cancelledAsMaybe} from "./cancellable";
-import {useState, useEffect, useCallback} from 'react';
+import {useState, useEffect, useCallback, useRef} from 'react';
 
 export function useAsyncCallback(fn, deps, cleanup = () => null) {
-  const context = useContext(new Cancellable());
+  const context = useRef(new Cancellable());
   const cancel = useCallback(() => {
     context.cancellable.cancel();
     const newCancellable = context.cancellable = new Cancellable();
--- a/etna/packages/etna-js/utils/cancellable_helpers.jsx
+++ b/etna/packages/etna-js/utils/cancellable_helpers.jsx
@@ -5,8 +5,8 @@ import {useState, useEffect, useCallback, useRef} from 'react';
 export function useAsyncCallback(fn, deps, cleanup = () => null) {
   const context = useRef(new Cancellable());
   const cancel = useCallback(() => {
-    context.cancellable.cancel();
-    const newCancellable = context.cancellable = new Cancellable();
+    context.current.cancel();
+    const newCancellable = context.current = new Cancellable();
     return newCancellable;
   }, []);
 
--- a/etna/packages/etna-js/utils/cancellable_helpers.jsx
+++ b/etna/packages/etna-js/utils/cancellable_helpers.jsx
@@ -3,14 +3,10 @@ import {Cancellable, cancelledAsMaybe} from "./cancellable";
 import {useState, useEffect, useCallback} from 'react';
 
 export function useAsyncCallback(fn, deps, cleanup = () => null) {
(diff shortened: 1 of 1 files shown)


etna/packages/etna-js/utils/cancellable_helpers.jsx
```diff
@@ -3,14 +3,10 @@ import {Cancellable, cancelledAsMaybe} from "./cancellable";
 import {useState, useEffect, useCallback} from 'react';
 
 export function useAsyncCallback(fn, deps, cleanup = () => null) {
-  const [_, setContext] = useState(() => new Cancellable());
+  const context = useContext(new Cancellable());
   const cancel = useCallback(() => {
```
corps, line 7: The idea of this callback, `cancel` is to execute `cancel()` on whatever current Cancellable context exists, returning a fresh one.  So essentially each time a context should be "reset" this function is called to start a fresh new context.

etna/packages/etna-js/utils/cancellable_helpers.jsx
```diff
@@ -3,14 +3,10 @@ import {Cancellable, cancelledAsMaybe} from "./cancellable";
 import {useState, useEffect, useCallback} from 'react';
 
 export function useAsyncCallback(fn, deps, cleanup = () => null) {
-  const [_, setContext] = useState(() => new Cancellable());
```
corps, line 6: Previously, I was keeping the state of the "current cancellable context" in a React state hook.  There was a problem I'll show below.

etna/packages/etna-js/utils/cancellable_helpers.jsx
```diff
@@ -1,16 +1,12 @@
 import 'regenerator-runtime';
 import {Cancellable, cancelledAsMaybe} from "./cancellable";
-import {useState, useEffect, useCallback} from 'react';
+import {useState, useEffect, useCallback, useRef} from 'react';
 
 export function useAsyncCallback(fn, deps, cleanup = () => null) {
-  const [_, setContext] = useState(() => new Cancellable());
+  const context = useRef(new Cancellable());
```
corps, line 6: `useRef` => very similar to useState, but crucially it is not managed in terms of render cycles and its reference is *stable*.

In some contexts, this concept is known as a "cell".  Essentially, the idea is to put a value you want to read and write into a length 1 array or an object with a single key, so that you can share that parent object's reference in order to share the mutations on the inner value.

```
a = 1
f(a)
a = 2 // f does not see the new a value, it still sees 1
```

```
a = [1]
f(a)
a[0] = 2 // f sees the updated value because it is contained within the shared reference that accepts mutations.
```

I'll explain why I use a reference instead of the state hook bellow

etna/packages/etna-js/utils/cancellable_helpers.jsx
```diff
@@ -1,16 +1,12 @@
...
   const cancel = useCallback(() => {
-    const newCancellable = new Cancellable();
-    setContext(running => {
```
corps, line 9: Here's the bug.

Normally, react state hooks provide callbacks, "setters", that allow you to update the state value while /also/ fetching the current value.  Here I used that trick to "fetch" the current cancellable while returning a new cancellable, executing the cancel on the existing value before returning the new one.

but here's the catch -- react state callbacks *do not execute synchronously when you call them!*.  When you run `setX()` react does *NOT* actually set the value immediately or execute the callback immediately.  this blew my mind.

Instead, it simply *buffers the fact that you want to execute it*, and later runs the callback *once* inside of the render loop.

In other words, if you call setX(1) setx(2) and setX(3), react simply accepts the LAST set call setX(3) and runs that during the next render loop, ignoring all the previous set calls.  This optimization normally is convenient for reducing unnecessary renders, but in our case it's problematic because if we happen to cancel twice in the same render loop, we want to cancel twice!  We need to cancel each and every context we create, even if it happens to occur quickly.

That's where `context.cancellable` comes into play.  We can mutate and acccess it synchronously like we expect, react does not deferring or batching, allowing us to consistently make the call we need.

dtm2451 (approved):

PR #800, opened 2022-04-06
Airflow impersonate user

coleshaw:
This PR adds in some Airflow functionality, per our conversation:

• Updates the project viewer permissions so that users should be able to run, edit, and inspect DAGS, but not see our connection secrets, alter users / roles, etc. Let me know if I should tweak these, I made a guess of what functionality our general DS users would need.
• Adds a top-level menu item linked to the Etna airflow docs, plus a route for serving the static help pages.
• Should run pdoc and build the help pages upon release. Not sure if I modified the ``airflow/Makefile`` correctly to achieve this -- could you please verify? It seems to work, but I'm not 100% confident it was the right way to do it :shrug:
• Allows Admin users the ability to impersonate the permissions of another system user. It is temporary and reverts as soon as you try doing something prohibited (i.e. accessing the impersonate form as a non-Admin), and it will also revert when you log out. I think it should work for our purposes? Let me know.

@corps , I'm not sure the issue with the Access denied warning upon initial login (no cookies) is possible to address. I think it is similar / related to this [known issue with OAuth login](https://github.com/apache/airflow/issues/15601), and perhaps is an issue with the underlying Flask AppBuilder. May just have to educate users until that is addressed upstream?

Note that to test this locally, you may have to clear out your user session, so it can re-sync the role permissions. Getting the new permissions in production will also require you to clear out your cookie, so the EtnaSecurityManager resyncs the project permissions.


airflow/Makefile
```diff
@@ -14,3 +14,7 @@ black:
 
 pdoc: $(shell readlink -m $(shell dirname $(firstword $(MAKEFILE_LIST))))/image-release.marker
 	docker run -u 0:0 -v $(pwd)/opt/providers:/opt/airflow/providers --rm $(baseTag) bash -c '$$@' -- pdoc -o /opt/airflow/providers/etna/docs etna
+
+release::
```
coleshaw, line 18: Is this the right way to auto-build `pdoc` on release?

airflow/opt/airflow.cfg
```diff
@@ -94,15 +94,15 @@ load_examples = False
 load_default_connections = False
 
 # Path to the folder containing Airflow plugins
-plugins_folder = /opt/airflow/plugins
+plugins_folder = /opt/airflow/providers/etna/etna/plugins
```
coleshaw, line 97: I needed to update this path in order for the plugins to be found...

airflow/opt/entrypoint-with-secret-files.sh
```diff
@@ -23,7 +23,7 @@ if [ -n "$USE_MOCKS" ]; then
 fi
 
 
-cp /opt/airflow/.airflowignore $(airflow config get-value core dags_folder)/.airflowignore
+cp /opt/airflow/.airflowignore $(airflow config get-value core dags_folder)/
```
coleshaw, line 26: When running `pdoc`, I was getting an exception on this `cp` command, where it assumed the `$(airflow config get-value core dags_folder)/.airflowignore` was a non-existent directory. Update the target to avoid the error.

airflow/opt/providers/etna/etna/tests/conftest.py
```diff
@@ -271,7 +271,6 @@ def no_op(*args, **kwargs):
             "sync_appbuilder_roles",
             "init_jinja_globals",
             "init_xframe_protection",
-            "init_permanent_session",
```
coleshaw, line 274: For some reason, when I ran tests locally using the `app` fixture, I was getting an exception that this method `init_permanent_session` didn't exist. Perhaps it was deprecated and removed recently?

airflow/opt/providers/etna/setup.py
```diff
@@ -25,6 +25,7 @@
         "etna.dags",
         "etna.etls",
         "etna.metrics",
+        "etna.plugins"
```
coleshaw, line 28: Required so that the plugins are also installed.

airflow/Makefile
corps: This, could work?  But due to the base release definition:

```
.PHONY: release
release:: release-build release-test ## Builds static docker images staged for release, runs tests against them, and pushes them to dockerhub (requires PUSH_IMAGES=1)
	@ true
```

It's possible that the new target runs after release-build and executes an additional release-build, or something.  Basically, I'm not 100% the Make behavior here when merging targets where the same target exists twice.

How about making it `release-build::` ?

airflow/opt/airflow.cfg
corps: That makes total sense.  I think I started out with a general plugins dir but deprecated it, but now we need it.  The plugins / providers difference is a bit confusing tbh.

airflow/opt/providers/etna/etna/auth/etna_cookie_auth.py
```diff
@@ -107,11 +109,59 @@ def sync_project_role(self, project_name: str) -> Role:
         cond = DagModel.tags.any(DagTag.name == project_name)
         dags_query = dags_query.filter(cond)
 
+        project_permissions = [
+            # From standard Viewer
+            (permissions.ACTION_CAN_READ, permissions.RESOURCE_AUDIT_LOG),
```
corps, line 114: Oh cool!  Dude, thank you for digging into this.  It's the sort of configuration that takes a bit of focus and experimentation to get right.

airflow/opt/providers/etna/etna/auth/etna_cookie_auth.py
```diff
@@ -107,11 +109,59 @@ def sync_project_role(self, project_name: str) -> Role:
...
                 resource_name_for_dag(dag.dag_id),
-            )
-            self.add_permission_role(role, perm_view)
+            ))
+
+        # Convert to Permissions
+        project_permissions = [self.add_permission_view_menu(*p) for p in project_permissions]
+
+        # Remove any old permissions on the role
+        for role_perm in role.permissions:
```
corps, line 157: Cool.  Originally I knew i had left out this 'removal' part of syncing, but it's wise to add it.

airflow/opt/providers/etna/etna/plugins/impersonate_user_view.py
```diff
@@ -0,0 +1,54 @@
...
+            impersonator = self.appbuilder.sm.current_user
+
+            impersonator.roles = subject_roles
+            self.appbuilder.sm.update_user(impersonator)
+
+            flash(f"Done -- you have the same Airflow permissions as {subject.first_name}. Logout to reset your permissions.", 'info')
+
+    impersonate_user_view = ImpersonateUserView()
+
+except (ImportError, ModuleNotFoundError) as e:
```
corps, line 53: I'm curious what lead to this import error?

airflow/opt/providers/etna/etna/tests/conftest.py
corps: Huh.  Possible.  I think I might have _started_ using the app fixture but backed out from using it for some reason.  I'm trying to remember the context.  

We did upgrade the airflow version at one point, so if this was added before the upgrade but not used after, that's plausible.

airflow/opt/providers/etna/setup.py
corps: Cool, that works.  I seem to recall there was also a way of specifying a plugins entrypoint specifically?  But if this works it works.  Python has a history for a rather divergent and inconsistent ecosystem for its packaging system.

airflow/opt/providers/etna/etna/plugins/impersonate_user_view.py
```diff
@@ -0,0 +1,54 @@
...
+    email = SelectField(label='User',
+        validate_choice=False, # Seems to be required to work with Airflow, since the choices are injected
+        description=('User whose permissions to impersonate'),
+        validators = [InputRequired()],
+        widget=Select2Widget())
+
+
+try:
+    from flask_appbuilder import SimpleFormView
+    from flask_appbuilder.security.decorators import has_access
```
corps, line 19: Just to clarify: basically, access to this view requires whitelisting the view in, which only the admin role has "sudo" privilege to view even without explicitly adding it to that role, right?

airflow/Makefile
coleshaw: Sorry, very much so a Make noob -- changing it to

```
release-build::
	$(MAKE) pdoc
```

but I'm not sure the order will be right? It does run `pytest` and then `pdoc`, but then doesn't seem to do anything else like build an image with the latest docs? Do I need to add some sort of `super` command to the `release-build::` to call the regular build process after `$(MAKE) pdoc`?

airflow/opt/providers/etna/etna/plugins/impersonate_user_view.py
```diff
@@ -0,0 +1,54 @@
+from flask import flash
+
+from wtforms import SelectField
+from wtforms.validators import InputRequired
+from flask_appbuilder.fieldwidgets import Select2Widget
+from flask_appbuilder.forms import DynamicForm
+
+
+class ImpersonateUserForm(DynamicForm):
```
corps, line 9: Ok, cool.  So it is possible for us to use react in airflow if we need to  (build the static assets in and add them as a plugin resource), any "simple admin-y thing" would ideally be easy enough to idea with a basic python form.  This will make administration easier over time.

airflow/opt/providers/etna/etna/plugins/impersonate_user_view.py
coleshaw: So this is what the `airflow-code-editor` does, I think because it's straddling two different versions of Airflow -- Airflow 1 doesn't use flask appbuilder, whereas Airflow 2 does, so it's perhaps unclear what the context is for the plugin. We can remove this since we'll always be using Airflow 2.

corps (approved):
Awesome due, this is a huge boon!!!  And now that we have an example of adding views / permissions more correctly, it'll be easier to customize this further in the future.  

The hardest part about airflow I'm finding is the sheer degree of options and customizability which leads to a learning curve.  But carving out these features and usages will make it easier over time to get a lot more done with it.  Thank you.

RE:  Using `release-build` is my only real ask before merge.

airflow/opt/providers/etna/etna/plugins/impersonate_user_view.py
coleshaw: Yes, only the Admin role has permissions for this Impersonate view, and it is implied (do not have to add it explicitly).

airflow/opt/providers/etna/etna/plugins/impersonate_user_view.py
corps: Yeah, I'd feel safer removing this and just getting an error during build if we accidentally use an incompatible airflow version.

airflow/opt/providers/etna/etna/plugins/impersonate_user_view.py
coleshaw: Yes, we could totally use React components, etc., if we need to at some point (i.e. the REDCap config). `airflow-code-editor` uses a Vue.js front-end and compiles the assets, so we could do the same.

airflow/Makefile
corps: AH!  Ok.  Yeah, you are correct actually.  Let me see...

corps:
> Note that to test this locally, you may have to clear out your user session, so it can re-sync the role permissions. Getting the new permissions in production will also require you to clear out your cookie, so the `EtnaSecurityManager` resyncs the project permissions.

Yes.  So I was thinking of adding this as an 'update' step similar to db migrations.  It would run as per `make update` locally and as part of deployment in production.  That way when this logic changes, we trigger it over all known projects.  The trick is we need to be able to identify all project roles to sync. We can address this separately.

corps:
> * Updates the project `viewer` permissions so that users should be able to run, edit, and inspect DAGS, but **not** see our connection secrets, alter users / roles, etc. Let me know if I should tweak these, I made a guess of what functionality our general DS users would need.

I think you picked a pretty good starting place.  There is a follow work I intend to do that allows project users to see / create secrets _for their project_ via connection scoping.  But for now, requiring admin privilege is fine, setting up a secret is easy enough for one of us to do.

airflow/opt/providers/etna/etna/tests/conftest.py
coleshaw: Looks like it was taken out in February 2022: https://github.com/apache/airflow/commit/da9d0863c7ff121c111a455708163b026943bdf1#diff-048ecd9f5ec6ccdc46b6cd7740cb9dcce877562817d3e2f997d2df39d2dcefec

airflow/Makefile
corps: Ok, alternative fix, while silly, is at the VERY top of the file, add:

```
release::
<tab>$(MAKE) pdoc
```

Before the other includes.   The multitarget `::` executes in order of definition, so we need this step to proceed the other release definition to inject it before the build image.

This works similar to the way you've written it, but I worry about Make logic invoking `release-build` twice as written.

coleshaw: Okay, I think this works as expected. The final output I get is:

```
set -e; if ! [ -z "" ]; then docker push airflow; touch /projects/ucsf/monoetna/airflow/image-release.marker; fi
make: Leaving directory '/projects/ucsf/monoetna/airflow'
```

so I think if I had included the right environment variables, it would have pushed the latest build with latest docs. :+1:

graft:
> * Updates the project `viewer` permissions so that users should be able to run, edit, and inspect DAGS, but **not** see our connection secrets, alter users / roles, etc. Let me know if I should tweak these, I made a guess of what functionality our general DS users would need.

Not quite clear to me if `viewer` here == Etna `viewer` - if so, could we escalate to `editor`? It does not seem to me that viewers (i.e., people who cannot update data for a project) would have any role inspecting or running airflow, so they should just be locked out, while editors can run, edit, inspect dags.

corps:
> > * Updates the project `viewer` permissions so that users should be able to run, edit, and inspect DAGS, but **not** see our connection secrets, alter users / roles, etc. Let me know if I should tweak these, I made a guess of what functionality our general DS users would need.
> 
> Not quite clear to me if `viewer` here == Etna `viewer` - if so, could we escalate to `editor`? It does not seem to me that viewers (i.e., people who cannot update data for a project) would have any role inspecting or running airflow, so they should just be locked out, while editors can run, edit, inspect dags.

In that case, we may need two project roles: a project edit role and a project view role.

corps:
> > > * Updates the project `viewer` permissions so that users should be able to run, edit, and inspect DAGS, but **not** see our connection secrets, alter users / roles, etc. Let me know if I should tweak these, I made a guess of what functionality our general DS users would need.
> > 
> > 
> > Not quite clear to me if `viewer` here == Etna `viewer` - if so, could we escalate to `editor`? It does not seem to me that viewers (i.e., people who cannot update data for a project) would have any role inspecting or running airflow, so they should just be locked out, while editors can run, edit, inspect dags.
> 
> In that case, we may need two project roles: a project edit role and a project view role.

Or yea you may be right, just lockout < editor.

coleshaw:
Yeah, currently Airflow `viewer` is equivalent to Etna `editor`, because we're assuming it is Data Scientists who will set up and configure the ETLs. So I can lock out all Etna `viewer` accounts, which makes sense to me.

coleshaw:
Okay, fixed the permissions a bit so that you can only see the DAGs in your editor+ projects, which means most users won't be able to see the system DAGs or any DAGs where they are only viewers.

Not sure how to limit the scope within the actual code editor though, since all folders and code are viewable by all users...?

corps:
> Okay, fixed the permissions a bit so that you can only see the DAGs in your editor+ projects, which means most users won't be able to see the system DAGs or any DAGs where they are only viewers.
> 
> Not sure how to limit the scope within the actual code editor though, since all folders and code are viewable by all users...?

This is a known issue and limitation of the current editor, _which I have a simple solution for_ but we'll address as a follow up.

PR #802, opened 2022-04-11
use parse instead of isoparse

coleshaw:
I noticed this problem on the MVIR1 data ETL, which had stopped the processing around 3-15:

```
  File "/home/airflow/.local/lib/python3.8/site-packages/etna/etls/decorators.py", line 77, in propagate_folder_updated_at
    if child.updated_at_datetime < folder.updated_at_datetime:
  File "/home/airflow/.local/lib/python3.8/site-packages/etna/hooks/etna.py", line 440, in updated_at_datetime
    return dateutil.parser.isoparse(self.updated_at)
  File "/home/airflow/.local/lib/python3.8/site-packages/dateutil/parser/isoparser.py", line 37, in func
    return f(self, str_in, *args, **kwargs)
  File "/home/airflow/.local/lib/python3.8/site-packages/dateutil/parser/isoparser.py", line 138, in isoparse
    components += self._parse_isotime(dt_str[pos + 1:])
  File "/home/airflow/.local/lib/python3.8/site-packages/dateutil/parser/isoparser.py", line 374, in _parse_isotime
    raise ValueError('Unused components in ISO string')
ValueError: Unused components in ISO string
```

It's unclear to me what might have changed around this timeframe, since it doesn't look like there were any changes to Metis's `tail` endpoint or the ETL decorators. However, testing manually, I can verify that `dateutil.parser.isoparse(self.updated_at)` throws that error on the datetimes returned from `tail` (`2020-05-01 11:07:16 +0000`), but `dateutil.parser.parse(self.updated_at)` works okay. Same kind of error as reported [in this SO question / answer](https://stackoverflow.com/a/63960104). It seems the TZ info that `tail` returns conflicts with the Python `parser` expectation for ISO format...?

Hard to test locally... so thoughts on if this will fix the issue and not break anything else? :shrug: 

Note that trying to re-run the ETL seems to not make much difference, because the `propagate_folders` task is stuck due to the above...

### DIFF
--- a/airflow/opt/providers/etna/etna/hooks/etna.py
+++ b/airflow/opt/providers/etna/etna/hooks/etna.py
@@ -326,7 +326,7 @@ class TailNode:
 
     @property
     def updated_at_datetime(self) -> datetime:
-        return dateutil.parser.isoparse(self.updated_at)
+        return dateutil.parser.parse(self.updated_at)
 
 class TailResultContainer:
     bucket_name: str
@@ -424,7 +424,7 @@ class File:
 
     @property
     def updated_at_datetime(self) -> datetime:
-        return dateutil.parser.isoparse(self.updated_at)
+        return dateutil.parser.parse(self.updated_at)
 
 
 @serialize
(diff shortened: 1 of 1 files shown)


corps:
OMG I was just looking into this, thank you!!!

PR #804, opened 2022-04-12
Polyphemus redcap separate filter logic

coleshaw:
This PR adds in a separate `filters` section to REDCap loading scripts. Users can specify one of four verbs to compare against a redcap field (`equals`, `exists`, `in`, and `match`), but this value does not get mapped to a Magma attribute (unless it is separately added to the `attribute` section). Filters can be applied to any script, and appear in the UI below the `attribute` section:

Screenshot from 2022-04-12 14-55-39

They are added via the new `Filter` icon next to "Add attribute" -- should this be moved to a more intuitive location, or icon changed?

Note that `exists` should also work in the negative sense, now ... 

This does refactor the filtering logic in `Redcap::Value` into `Redcap::Filter`, so in the future we can deprecate the usage of filtering verbs in the value mapping.


graft:
Looks great!

Some nits:
- Now that there are "filters" and "attributes" sections, it might be nice to give each their own header and move the add filter and add attribute buttons to their respective headers
- "filters" might appear above "attributes", seems consistent with what we do elsewhere (e.g. in query)
- While 'redcap_field' is required by the filter schema, the UI allows you to remove it, which seems improper

Somewhere in here are some reusable components we might bring to bear in other config editors...

Also: Bushra noticed a small UI bug where the copy script button inserts a new copy at the wrong place, can you add this change:
https://github.com/mountetna/monoetna/blob/912b6b87dc484b6c208f884480c115b28892c04a/polyphemus/lib/client/jsx/etl/redcap-form.tsx#L1119

Here `i` should be `i + (page - 1) * pageSize`

graft (approved):

PR #810, opened 2022-04-15
Vulcan select_autocomplete bug fixes

dtm2451:
Fixes bugs that shouldn't have been apparent to users, but are still worth removing:
1. calculation of options which match the user-input, which is run as an async Promise when there are more than 1000 options, wasn't cancelled properly.  -> occasional React complaint of a 'memory leak' / state update on an unmounted component.
2. On new mount of a nested_select_autocomplete with value = a nested-value, a select_autocomplete component render would be requested, before proper path was calculated, with the given value & top-level options. -> mui complaint because this nested value wouldn't actually exist among the options provided.


vulcan/lib/client/jsx/components/workflow/user_interactions/inputs/select_autocomplete.tsx
```diff
@@ -25,28 +27,26 @@ export default function SelectAutocompleteInput(
...
-  const getOptionsDelayed = useCallback(
-    debounce((text, options_in, callback) => {
-      getOptionsAsync(text, [...options_in]).then(callback);
-    }, 200),
-    [],
-  );
+  const [getOptionsDelayed] = useAsyncCallback(function*(text: string, options_in: string[], callback: Function) {
+    const options = yield getOptionsAsync(text, [...options_in]);
+    callback(options)
+  }, [])
```
dtm2451, line 40: Adds cancellation (and negates need for debounce)

vulcan/lib/client/jsx/components/workflow/user_interactions/inputs/nested_select_autocomplete.tsx
```diff
@@ -53,7 +53,7 @@ function LeafOptions({
   handleSelect: (value: string | null, depth: number) => void;
   maxOptions?: number;
 }) {
-  if (!options_in) return null;
+  if ( (!options_in) || (value!=null && !options_in.includes(value)) ) return null;
```
dtm2451, line 56: Checks if provided value exists in the given options to prevent MUI complaint bug

vulcan/lib/client/jsx/components/workflow/user_interactions/inputs/select_autocomplete.tsx
```diff
@@ -25,28 +27,26 @@ export default function SelectAutocompleteInput(
...
   
   useEffect(() => {
     // Shown from all options when user has made their selection (current text = current value)
     const query = (inputState!=value) ? inputState : ''
     
     if (options_in.length>1000) {
-      setLoadingOptions(true); setOptions([]);
```
dtm2451, line 47: This `setOptions([])` was a hangover from the example I'd used to set up our self-filtering of options.  Not needed anymore due to management of `loadingOptions` state variable given to `loading` prop below.

vulcan/lib/client/jsx/components/workflow/user_interactions/inputs/select_autocomplete.tsx
```diff
@@ -68,9 +68,8 @@ export default function SelectAutocompleteInput(
     <Autocomplete
       disableClearable={disableClearable}
       clearOnBlur={true}
-      options={options.display}
-      // disable filtering on client
-      filterOptions={(x: string[]) => x}
+      options={options_in}
+      filterOptions={(x: string[]) => options.display}
```
dtm2451, line 72: I never encountered it, but this change likely prevents similar MUI complaint of 'value not valid' in cases where the user selects a value that's not within the first 100/`maxOptions` options.

dtm2451:
Moved my changes related to the second bug (material ui complaining that nested_select_autocomplete values were invalid) to a new branch so that I can merge in the async fix that's ready now.

PR #811, opened 2022-04-17
Improvements to the airflow-code-editor user experience

corps:
Some improvements to the airflow-code-editor ahead of my talk and wider audience:

1. "Optimistic locking" in the form of tracking the state of the file before save, and ensuring it has not been concurrently modified.  this prevents two users from modifying the same file and unknowingly destroying each other's work.  An improved form may involve either introducing real time code colaboration (not that hard, libraries exist for this, but mostly we'd need the production redis to back the websocket channels) or just introducing mergeable diff on conflict.  For now, anyone who makes a concurrent write sees an error message instruction them to copy their changes and reload, essentially making the merge themselves.
2. Validates changes to DAG files before save.  Currently the only way to detect DAG errors is to reload the main page after modifying a DAG.  This is because DAG processing is asynchronous by default.  This new changes forces validating the module synchronously BEFORE writing to a file, ensuring that bad edits are obvious immediately to the user.

This also demonstrates the basic pattern for further improving the editor experience, and how we can introduce similar modules to airflow that include assets.

### DIFF
--- a/airflow-code-editor/airflow_code_editor/app_builder_view.py
+++ b/airflow-code-editor/airflow_code_editor/app_builder_view.py
@@ -1,6 +1,8 @@
 #!/usr/bin/env python
 #
 #   Copyright 2019 Andrea Bonomi <andrea.bonomi@gmail.com>
+#   Modifications included by DSCOLabs engineering team
+#   University of San Francisco, CA
 #
 #   Licensed under the Apache License, Version 2.0 (the "License");
 #   you may not use this file except in compliance with the License.
@@ -36,9 +38,6 @@ try:
         (permissions.ACTION_CAN_READ, permissions.RESOURCE_WEBSITE),
     ]
 
-    # ############################################################################
-    # AppBuilder (Airflow >= 2.0)
-
     class AppBuilderCodeEditorView(BaseView, AbstractCodeEditorView):
         route_base = ROUTE
         base_permissions = ["can_list", "can_create", "menu_acccess"]
@@ -62,6 +61,11 @@ try:
         @auth.has_access(PERMISSIONS)
         def save(self, path=None):
             return self._save(path)
+
+        @expose("/uploads/<path:path>", methods=["POST"])
+        @auth.has_access(PERMISSIONS)
+        def upload(self, path=None):
+            return self._save(path, upload=True)
 
         @expose("/files/<path:path>", methods=["GET"])
         @auth.has_access(PERMISSIONS)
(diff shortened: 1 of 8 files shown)


airflow-code-editor/airflow_code_editor/app_builder_view.py
```diff
@@ -63,6 +62,11 @@ def repo(self, path=None):
         def save(self, path=None):
             return self._save(path)
 
+        @expose("/uploads/<path:path>", methods=["POST"])
```
corps, line 65: Separate out uploads path from editor save path, versioning only affects later not former.

airflow-code-editor/airflow_code_editor/app_builder_view.py
```diff
@@ -98,68 +102,9 @@ def _render(self, template, *args, **kargs):
                 **kargs
             )
 
-
 except (ImportError, ModuleNotFoundError):
-    from airflow_code_editor.auth import has_access
```
corps, line 103: This was support for older airflow versions which we don't intent to, so removing for cleanliness.

corps:
Incidentally, our changes to airflow-code-editor are tracked here https://github.com/mountetna/airflow-code-editor/ so that we can still consume upstream improvements as well.  Hope to eventually contribute PR back.

graft:
>mostly we'd need the production redis to back the websocket channels

What's the word on a production redis anyhow?

corps:
> > mostly we'd need the production redis to back the websocket channels
> 
> What's the word on a production redis anyhow?

It is existed and runs, but it is turned off.  I want to complete the migration to the NFS storage backend and network fixes.

coleshaw:
Not to be super nit-picky, but more as an FYI, the UC Technology Licensing Office has classified the Apache 2 license (that the airflow-code-editor is licensed under) as [moderate risk when distributed externally](https://security.ucop.edu/files/documents/resources/oss-chart.pdf), I'm guessing because it is patent-granting. :shrug: Should be okay because everything we're writing is open source, anyways. But we should probably also add a "modifications" doc to our fork, to be compliant with the original license...

corps:
> Not to be super nit-picky, but more as an FYI, the UC Technology Licensing Office has classified the Apache 2 license (that the airflow-code-editor is licensed under) as [moderate risk when distributed externally](https://security.ucop.edu/files/documents/resources/oss-chart.pdf), I'm guessing because it is patent-granting. shrug Should be okay because everything we're writing is open source, anyways. But we should probably also add a "modifications" doc to our fork, to be compliant with the original license...

I started by added modification information to the individual files changed, I can also add a MODIFICATIONS DOC.  It's not explicit how modifications are called out other than that they are, so traditionally I include my modification information in individual sources next to original claims.

Yeah, my reasoning on the original license was that the stack as is is also open source.  It does not bleed into scripts run within the technology because none of the ETLs themselves distribute with the code editor's own source.

coleshaw:
I did send an e-mail to the team a month or so ago, and they pointed me to [this site](https://security.ucop.edu/resources/open-source-software-licensing.html). Not sure if they would give more specific recommendations in this case, since it is a more concrete example.

corps:
I'm down for discussing this more broadly then.  
If we don't feel comfortable with the apache license in general, there are many many tools we may need to reconsider and deploy.

An open question I suppose is how much buffer do we have between our infrastructure tooling and the pipelines themselves we run on them.  My assumption is that since our methodology and our infrastructure are distributed separately and, in theory, can be run separately, that the mix was safe.  But that's for the lawyers I suppose.

coleshaw:
Well, my take is that mostly Apache 2 is fine (note the green boxes for most of the use cases for it), especially since we're not modifying most of the libraries we're using (so all green / low-risk). Even for our use case of the code editor, I agree with your assessment that the code we're modifying (the code editor) is separate from the DAGs we're building using the editor, so seems like it should be safe. @dtm2451 , if you have a current conversation going with the licensing folks, do you want to check with them? Otherwise I can shoot them a quick e-mail to confirm our assessment.

dtm2451:
> Even for our use case of the code editor, I agree with your assessment that the code we're modifying (the code editor) is separate from the DAGs we're building using the editor, so seems like it should be safe.

I agree as well.

@cole I think you have a bit more context than me, especially since you've reached out for monoetna in the past.  I find the chart format quite underwhelming and not informative enough about why each license got the assessments they did.  Assuming we've read between the lines properly, I think we're good, but I'm in favor of double-checking if you are!  I haven't heard back on my email yet, so probably makes sense for you to just ask separately.  Thanks!

airflow-code-editor/airflow_code_editor/flask_admin_view.py
```diff
@@ -1,6 +1,8 @@
 #!/usr/bin/env python
 #
 #   Copyright 2019 Andrea Bonomi <andrea.bonomi@gmail.com>
+#   Modifications included by DSCOLabs engineering team
+#   University of San Francisco, CA
```
coleshaw, line 5: super-nit, but these modification statements should probably all be "University of California, San Francisco"?

PR #812, opened 2022-04-19
Adding archive commands and multi deploy

corps:
As part of moving the new swarm to the new NFS server (eventually we migrate the old beegfs and this NFS together on the new storage sever), I wanted to streamline our rolling logical back up process a bit to include not just databases, but any arbitrary storage target (for instance, grafana runs on a sqlite file, and prometheus produces a tbdb).

It occurred to me that since metis is and likely always will be backed up to tape and other means, the strongest form of logic backup is to use metis itself.  I had been doing this for new VM databases but in a one off way.

Inclusion of these two commands as etna commands allows me to deploy 'backup' services to any target in the swarm we like fairly easily, as well as providing an environment for recovery as well (the same service can act as both).

The backup commands here essentially upload targets to metis in a given archive bucket.  We can use scoped tokens that have unique buckets per application.  It keeps up to 14 rolling copies of any backup for now, though in the future this could be configurable (I hesitate though; backups are not free, and actually I recently had to purge much older database backups because they start to consume non trivial amounts of room on even our beegfs cluster.  I seriously ponder if a consistent policy, say 14 days, makes sense.  Remember -- this is just for the logic backups in metis, every still goes to tape from metis so it is strictly necessary to use hot storage for indefinite back ups).

Again, to emphasize -- the goal of this logical backup is to simplify recovery process.  A pull archive commands will fetch the latest version of the backup target instantly.  Everything still goes to tape in the end.


etna/lib/commands.rb
```diff
@@ -2,6 +2,7 @@
 require 'logger'
 require 'rollbar'
 require 'tempfile'
+require 'securerandom'
```
coleshaw, line 5: Are these commands meant to be in the distributed gem? If so, do you want to bump the version of the gem, so it gets published?

coleshaw: and thus I guess you can pin the `etna` gem version down, in whatever backup swarm service is running them?

corps: In this case, I'll be using the production image, so it's not imperative to bump the gem.  I think it's ok for them to be available in the gem, they're generally useful as well.  I separate this out from the metis_client however since they have a specific niche use however that I considered outside the norm for metis_client itself.

PR #815, opened 2022-04-20
Allow Vulcan UI steps to be marked to "Auto-Commit in future"

dtm2451:
Still a bit of a WIP and I do plan to add tests, but the basic concept seems to be working!

The goal is to allow for a (slightly) improved user experience in vulcan by removing certain breakpoints.  For example, in the scRNAseq workflow, a user may want to color their umap by 25 different genes.  For each new gene, the 'Set plot options' step comes after the color_by selection and it creates an additional break point where the user needs to 'Commit' this step, then hit 'Run'.  The goal here is to allow break points like this to be eliminate-able.

Biggest Requirement: Target step needs to have valid defaults.

Unfortunately, I expect that there are times when a step's defaults will appear valid but might actually be right for all use cases, or for enough use cases that the auto-pass setup would be helpful >> potentially harmful (e.g. annoying to the user because they actually needed to make a change and now they need to click `Stop` first, or maybe auto-passing hides that a step proper for only 50% of use cases is hit) and thus I think we want to give control to the people who know the full context.  *That's the workflow creator & the user.*

Workflow creator should be able to know if their primary input defaults are generally viable, and if any defaulting of later stepUIs should also be generally viable.  User should then be able to determine, if they do want to accept defaults or be asked each time that a step gets re-hit.


vulcan/lib/client/jsx/contexts/input_state_management.tsx
```diff
@@ -77,28 +78,80 @@ export function WithBufferedInputs({
...
+    } else {
+      dispatch(clearStepUIAutoPass(stepName))
+    }
+  }
+
+  function isPassable(stepName: string | null) {
+    if (stepName!=null) {
+      // ToDo: better way?
+      const this_step = state.workflow?.steps[0].filter((val) => val.name == stepName)[0]
+      if (this_step && this_step.doc!=null && this_step.doc.startsWith('SKIPPABLE')) return true
```
dtm2451, line 93: Ultimate first trigger of 'Can this step be auto-passed?'.  We could do this part differently, but what I like about it is that it operates within the bounds of CWL, falls to the workflow author to set up / maintain, and is visible to workflow users behind the (?).

corps: Seems interesting, I don't mind.  One q: does this mean that primary inputs are not "skippable"?

dtm2451: Well... not by these exact methods. Wasn't thinking towards that use case.

I imagine the downstream `triggerCommit` and `triggerRun` machinery could be utilized for primary input auto-skipping too, but we'd need a starting point that's not reliant on the user marking them for auto_complete... likely just an 'if_CWL_fetaure' then mark start out new (completely fresh) figures of that workflow with `triggerCommit: ['primary_inputs']`.  And this exact 'doc' method for marking in the CWL won't work because each individual primary input has it's own 'doc'-section... there's not a single 'doc' for all primary inputs which could be started with "SKIPPABLE".

vulcan/lib/client/jsx/components/workflow/session/primary_inputs.tsx
```diff
@@ -76,6 +77,10 @@ function PrimaryInputsInner() {
 
     if (Object.keys(withDefaults).length > 0) {
       setInputs((inputs) => ({...inputs, ...withDefaults}));
+      // Also check / trigger auto-committing after defaults are set
+      if (state.workflow && state.workflow.vignette && state.workflow.vignette.includes("Primary inputs are skippable")) {
```
dtm2451, line 81: Trigger for primary inputs = this text anywhere in the vignette.

coleshaw: Hm....that seems brittle? Is there a way to calculate this from the graph? I don't recall what the input step dependencies look like, but I imagine there should be a way to tell the UI that the primary inputs aren't needed until some future step...?

coleshaw: In fact, if the server can provide some sort of dependency graph, could you calculate for all inputs (not just primary inputs) that a given UI step is skippable, without the CWL author having to explicitly name it as such?

dtm2451: Certainly open to suggestions about these triggers, but I'm not sure I agree that a vulcan-side calculation of suitability is a good idea.  The issue with automating is that there's external context needed.  Primary defaults might appear valid (pass validations) yet be, say, tuned for only 1 of a few different use cases.  It's the workflow author who would have that context, so I think it's good to put the control into the workflow author's hands.  That's especially important in the primary input case because there's no user-toggle -- right as the user jumps in to their fresh session, if the check here passes, the primary inputs get committed, and if that works then steps are started running.

Difference between stepUI and primary inputs use cases:
- stepUI: "<target text>" at start of step 'doc' => show auto-commit toggle =user-marks-yes=> mark step to `triggerCommit` any future time that it gets marked as a bufferedStep (=validation-passes=> commited & marked to `triggerRun`)
- primary inputs: "<target text>" in workflow vignette => mark step as `triggerCommit` as soon as it's values are filled in (=validation-passes=> commited & marked to `triggerRun`)

coleshaw: Hm, okay, I guess I'm not really familiar with the use cases, so very possible I misunderstood the intent of the PR! 

Would this have any effect on modifying a "skippable" step directly? Like every time you type into a text box, it commits and runs the workflow automatically instead of waiting for you to finish typing, if that step is skippable?

dtm2451: I normally give a better description when I make a PR!  I'll add above.

>Would this have any effect on modifying a "skippable" step directly? Like every time you type into a text box, it commits and runs the workflow automatically instead of waiting for you to finish typing, if that step is skippable?

Shouldn't, no!  The triggers are essentially tied to "on-mount" of given step.  For primary inputs: at the start of the session, right after defaults are filled in.  For stepUIs: they're marked as a bufferedStep if they're being newly added to the set of shown input.  Those are the only times this system would be triggered.

dtm2451: Ooooh no, I got some of that logic wrong... a committed step can become buffered if their inputs get changed.  I'll need to rework that trigger.

dtm2451:
I think it's finally done now!

Mighta gone overboard on tests lol, but they did keep helping me find issues in the implementation as I wrote more!

dtm2451:
Testing on the Manual Annotations step, I think a more nuanced system might be needed for `MultipleInput` steps like that one, or any stepUI where default values are loaded in a piecemeal fashion rather than in a single-go.

I don't think it's worth figuring out right now... I'll just remove passability from that step (for now)

dtm2451:
Just to document: 
manual_annots_issue

PR #816, opened 2022-04-20
Airflow box etl helpers

coleshaw:
This PR adds hooks and helpers to interact with Box. The helpers should let you filter out files / folders via regexes, and then ingest them into Metis in either a default folder path or a DAG-specified path.

I'll add some annotations throughout.

Still working on the tests .... about to give up, may give it a new try tomorrow. Basically for `test_box_helpers.py`, all the tests pass when run individually, but run together `pytest` hangs after the first test. I think it's this [something like this problem](https://stackoverflow.com/questions/26350911/what-to-do-when-a-py-test-hangs-silently), but I can't figure out how to elegantly close the right connections ... any ideas, @corps ? Ready to review, otherwise.


airflow-code-editor/airflow_code_editor/code_editor_view.py
```diff
@@ -82,7 +82,6 @@ def _save(self, path=None, upload=False):
                         file.write(data.encode('utf-8'))
                         file.write("\n".encode('utf-8'))
                         file.flush()
-                        file.close()
```
coleshaw, line 85: I was running into an issue with our fork of the `airflow-code-editor` (which still isn't getting pushed to the right subtree! :angry: ), where trying to save a DAG file would result in an error like "Problem loading module: /tmp/tmpfilename.py` (from the next code block). I suspect this was closing the temp file handle before the DAG-validity check could run?

airflow/opt/providers/etna/etna/hooks/etna.py
```diff
@@ -579,7 +580,7 @@ def upload_file(
                     "size was not specified, but file is streaming.  You must specify a size hint when using a stream."
                 )
 
-        if len(dest_path) > 1:
+        if len(dest_path) > 1 and not self.folder_exists(project_name, bucket_name, os.path.dirname(dest_path)):
```
coleshaw, line 583: Tweaked this a bit, since it seemed like if someone tried uploading a file to Metis at an existing folder, the `self.create_folder` would throw an exception and derail the upload.

airflow/opt/providers/etna/etna/tests/test_box_helpers.py
```diff
@@ -0,0 +1,160 @@
+# import re
```
coleshaw, line 1: Blah... yeah, just blah about these tests.

It's tough to handle both FTP and HTTP requests, so I may wind up not bothering with these, if I can't figure out the issue with the DB connection. Would be nice to at least test some of the non-connection-related logic...

corps:
Looking through now.

airflow-code-editor/airflow_code_editor/code_editor_view.py
corps: Huh.  I suppose it's possible closing may drop the file handle of the temp file, but that surprises me because I did test this before locally and it worked.  I'll test your branch here in a bit to see.  Also I'll check the subtree issue out, I think it's likely just a permission for push issue.

airflow/opt/providers/etna/etna/etls/box.py
```diff
@@ -0,0 +1,127 @@
...
+                        self.log.info(f"Uploading {file.full_path} to {dest_path}.")
+
+                        for blob in metis.upload_file(
+                            project_name,
+                            bucket_name,
+                            dest_path,
+                            box_io_file,
+                            file.size
+                        ):
+                            self.log.info("Uploading blob...")
```
corps, line 99: Kind of silly, but I wonder if we should limit the logging by time and not just blob iterations?  Like log once every 3-5 seconds?  Haha, just that the same message logging over and over for each chunk probably is overkill?

I realize the upload_file iterator interface is odd, but I found it to be most technically correct at the time.

airflow/opt/providers/etna/etna/hooks/box.py
```diff
@@ -0,0 +1,237 @@
...
+        ```
+
+        params:
+          ftps: an open, FTP_TLS connection
+          file: an FTP file listing
+        """
+        ftps.cwd(file.folder_path)
+
+        io_obj = io.BytesIO()
+        ftps.retrbinary(f"RETR {file.name}", io_obj.write)
```
corps, line 226: I, think this approach ends up buffering the entire download into memory, due to the push nature of the write.

There are a couple of solutions, and we should consider them, since file transfer of big files is not safe in buffered mode.

1.  Use `ntransfercmd` to retrieve a low level io and expected size value.  Using this combination, you can still construct a buffer reader, but one that reads in data in chunks as needed, rather than pushing the entire contents into memory immediately.  (https://pythontic.com/ftplib/ftp/ntransfercmd)
2.  Use a different library.  It looks like paramiko supports an open interface that supports streaming buffers https://docs.paramiko.org/en/stable/api/sftp.html#paramiko.sftp_client.SFTPClient.open     ie open("blah.txt", "r", 1024)

corps: Sorry, "buffered" readers are totes ok, what I meant by buffered here is "buffering the entire contents by pushing all of it immediately into a BytesIO", which is what retrbinary does unfortunately.

Buffered readers work by fillup to a certain size (a fixed buffer) on demand from reads, improving performance.

airflow/opt/providers/etna/etna/hooks/etna.py
corps: Interesting, I actually figured create_folder gracefully handles existing folders?  But ok, it's all good.

airflow/opt/providers/etna/etna/tests/conftest.py
```diff
@@ -118,8 +118,21 @@ def _get_client(self, auth):
...
 
     # Authorization in any vcr is safely masked
-    return {"filter_headers": [("authorization", "XXX-Auth")]}
-
+    return {
+        "filter_headers": [("authorization", "XXX-Auth")],
+    }
+
+
+# When re-recording tests, you can use this callback in before_record_response
```
corps, line 126: That's fair.  I think I also ought to add additional encryption the way that our ruby tests currently do via `CI_SECRET`.  TODO

airflow/opt/providers/etna/etna/tests/test_box_helpers.py
corps: Yeah, this is gonna be hard.  I too have struggled to write good tests because this code strikes at a crucial integration point, and various mocking libraries / recording (VCR) do not handle correctly all the socket level stuff going on here.

corps: If we have to, I'm ok saying that we have a manual, smoke screen tests, a script that we run by hand that executes some tests without recordings, given some inputs from the runner, to verify integration of some parts of the system.  Again, not ideal, but better than no testing.

This level of integrations tooling gets tricky to test overall, and is a rabbit hole.  Let's be pragmatic and aim for value vs time here!

corps:
This looks very solid!

Two open areas:
1.  Finding an easier to maintain strategy for testing.  I am ok with adding manual, non CI testing scripts around some of the ingration parts.  In that, perhaps we'll need to create useful mocking classes so that a tleast unit tests are easier to write.
2.  I think it's very important we do not buffer the entire ftp download, let's try to make the switch to a chunk, pull based reader object.

corps:
I will test and play around with the code editor change tomorrow morning, I'm heading out right now.

coleshaw:
Hm, I will play around with the subtree thing tomorrow too. It looks like a subtree caching issue from the action logs, so might have to manually clear that somehow?

coleshaw:
Oh, I thought that BytesIO inherits from BufferedIOBase, so it would also be buffered? Thanks for the pointers to the other libraries, I will check them out.

corps:
> Oh, I thought that BytesIO inherits from BufferedIOBase, so it would also be buffered? Thanks for the pointers to the other libraries, I will check them out.
> […](#)
> On Apr 20, 2022, at 20:44, Zach Collins ***@***.***> wrote: ﻿ @corps commented on this pull request. In airflow/opt/providers/etna/etna/hooks/box.py: > + + eg: + ``` + with box.open_file(file) as open_file: + for line in csv.reader(open_file): + break + ``` + + params: + ftps: an open, FTP_TLS connection + file: an FTP file listing + """ + ftps.cwd(file.folder_path) + + io_obj = io.BytesIO() + ftps.retrbinary(f"RETR {file.name}", io_obj.write) I, think this approach ends up buffering the entire download into memory, due to the push nature of the write. There are a couple of solutions, and we should consider them, since file transfer of big files is not safe in buffered mode. Use ntransfercmd to retrieve a low level io and expected size value. Using this combination, you can still construct a buffer reader, but one that reads in data in chunks as needed, rather than pushing the entire contents into memory immediately. (https://pythontic.com/ftplib/ftp/ntransfercmd) Use a different library. It looks like paramiko supports an open interface that supports streaming buffers https://docs.paramiko.org/en/stable/api/sftp.html#paramiko.sftp_client.SFTPClient.open ie open("blah.txt", "r", 1024) — Reply to this email directly, view it on GitHub, or unsubscribe. You are receiving this because you authored the thread.

It is, "buffered", but that's where the words get mixed up here.  
The buffered of bufferedio means that, *during pull based reads or push based writes, it collects up to atleast X amount of data* where X is a fixed size.  However, because of the fact that retrbinary *pushes all data into the callback synchronously*, in this case the buffer is filled with the entire file contents.  IE: Buffered the entire file, not just a fixed size.

The fundamental difference here is push vs pull on the read.  retrbinary pushes all contents synchronously into the callback.  Our goal with a pull based approach is that data is read, then immediately passed up to the metis client and uploaded, before any other data is read.  Thus, the "stream" only buffers a fixed amount at any given time, since the push of writing to metis directly drives the pull of reading from FTP.

airflow-code-editor/airflow_code_editor/code_editor_view.py
corps: Huh!  Your fix is spot on, I don't know how I missed this.  I pushed a hotfix to production, but this should get to master on your branch.

coleshaw:
>  However, because of the fact that retrbinary pushes all data into the callback synchronously, in this case the buffer is filled with the entire file contents. IE: Buffered the entire file, not just a fixed size.

Hm... so you mean the issue is on the `retrbinary` side, not on the receiving `BytesIO` side? i.e. the `retrbinary` waits to push the entire blob, instead of pushing individual chunks? The [docs for ftplib](https://docs.python.org/3/library/ftplib.html) say:

```
FTP.retrbinary(cmd, callback, blocksize=8192, rest=None)

    Retrieve a file in binary transfer mode. cmd should be an appropriate RETR command: 'RETR filename'. The callback function is called for each block of data received, with a single bytes argument giving the data block. The optional blocksize argument specifies the maximum chunk size to read on the low-level socket object created to do the actual transfer (which will also be the largest size of the data blocks passed to callback). A reasonable default is chosen. rest means the same thing as in the [transfercmd()](https://docs.python.org/3/library/ftplib.html#ftplib.FTP.transfercmd) method.
```

Since it refers to individual `blocksize`s and calling the `callback` per block of data, I thought that would allow for the streaming of data directly to Metis, without waiting for the entire file ... not sure if that qualifies as a pull-based read or is still a pushed-based write?

coleshaw:
> manual, non CI testing scripts around some of the ingration parts.

What would something like this look like? Basically a Python script that resides somewhere, that we would run manually? But the script would do the entire workflow (i.e. scan Box, ingest file to Metis, verify that all works, etc.), just not run as part of CI?

coleshaw:
> Hm, I will play around with the subtree thing tomorrow too. It looks like a subtree caching issue from the action logs, so might have to manually clear that somehow?

Seems like some obscure `git subtree` thing (https://stackoverflow.com/questions/68761778/git-subtree-cache-exists) ... clearing the cache also does not seem to help, `git` just recreates the cached value...

corps:
> > However, because of the fact that retrbinary pushes all data into the callback synchronously, in this case the buffer is filled with the entire file contents. IE: Buffered the entire file, not just a fixed size.
> 
> Hm... so you mean the issue is on the `retrbinary` side, not on the receiving `BytesIO` side? i.e. the `retrbinary` waits to push the entire blob, instead of pushing individual chunks? The [docs for ftplib](https://docs.python.org/3/library/ftplib.html) say:
> 
> ```
> FTP.retrbinary(cmd, callback, blocksize=8192, rest=None)
> 
>     Retrieve a file in binary transfer mode. cmd should be an appropriate RETR command: 'RETR filename'. The callback function is called for each block of data received, with a single bytes argument giving the data block. The optional blocksize argument specifies the maximum chunk size to read on the low-level socket object created to do the actual transfer (which will also be the largest size of the data blocks passed to callback). A reasonable default is chosen. rest means the same thing as in the [transfercmd()](https://docs.python.org/3/library/ftplib.html#ftplib.FTP.transfercmd) method.
> ```
> 
> Since it refers to individual `blocksize`s and calling the `callback` per block of data, I thought that would allow for the streaming of data directly to Metis, without waiting for the entire file ... not sure if that qualifies as a pull-based read or is still a pushed-based write?

In theory, it is possible to use a callback to implement pull based read, but that's not the case here.  What's happening is that retrbinary does not return until *all of the data is consumed and callback invoked*.  In other words, control flow is push -- the moment you call retrbinary, data is being pushed to callback until all of it is done, after which retrbinary returns.  Test yourself -- retrbinary does not return until the callback has already been invoked for all data.  In that case, our callback is writing to the 'buffered' IO, but here the buffer just increases to accommodate all the data, until the buffer in memory contains all data.

Now, it is possible in theory to create a pull based stream with the callback, but to do so, we have to invert the control from push.  Think about it this way:

1.  We call retrbinary
2.  retrbinary starts invoking our callback
3.  we still cannot consume any data, because retrbinary has not returned
4.  buffer still filling up
5. we can't stream data anywhere because we're still waiting on retrbinary, and we can't tell retrbinary to STOP calling our callback.  Thus the callback has to do something -- either drop the data al together or put it somewhere.  all of it.
6. finally retrbinary returns, but the data has all arrived in the callback.  From this point on, retrbinary isn't going to give us the data again for a pull unless we put it somewhere (the entirety of it).

The fix is, we would have to introduce a thread.  This thread would have to mediate push (the data arriving) with pull (the data having a destination to leave to).  A buffer, after all, *is the place data lives after it has arrived but before it has left*.  The thread would both need to STOP the push (making it block if we are not reading to stream new data), and also WAIT the pull (making it block if we are not read with new data to consume) with a blocking q primitive and a lot of threading logic.

The other fix, would be to change the callback.  If the callback can immediately upload the data chunk without storing it anywhere, we'd have a stream process.  That is also possible, but I think it makes the interface far less generic.  Especially because while we have a stream in that case, it's still push -- no way to stop consuming the entire data, we're forced to push all the data to the callback.  Currently, the metis interface provides pull based reads which allow processors to consume partial sets of a file (say the header), which is being used for certain data.

I think the right solution may be to use the [ntransfercmd](https://pythontic.com/ftplib/ftp/ntransfercmd) command instead.  It returns a socket and an expected size.  The size is needed for the upload to metis, but the socket is implemented in the OS as pull based (the kernel is handling the mediating the incoming data and TCP for you!), and i suspect it has a file like interface already.

By the way, this problem exists all the way down the stack, in the kernel and TCP stack.  Remember, even though data is digital, *it is still physical*, so you can always imagine data as moving just like real packages.  If the package isn't on a boat coming or on a boat leaving, it has to exist somewhere, in storage.  But if storage is full, you have to turn away arriving boats and you have to inform leaving boats so they don't depart with empty hulls.  Your kernel and TCP stack mediates the traffic and retransmission of data, it stops and starts and buffers (fixed size buffers) to allow streaming knowing that asynchronous behavior means you never guarantee the OS is ready to consume data at the same time it arrives.

corps:
by the way, python io's BufferIO actually only implements a block on the read -- it makes _pulls_ wait until a useful amount of data is ready.  But it does NOT block the write -- it takes and immediately writes every set of data.

corps:
Actually, if we're using ftp entirely to upload, it's reasonable to have the callback simply perform the upload and stream that way.  In this case, however, we don't return a BufferIO object (it would contain the entire file contents), so the function and flow has to change.

BUt, I think the pull read approach, if we can get it to work, is better and more useful long term.

PR #819, opened 2022-04-21
Airflow-code-editor try reset

coleshaw:
Hm...this diff seems absurdly large. Going to try opening a PR to see if it becomes more clean?

Anyways, I tried re-adding the subtree and then cherry-picking our fixes to the code editor. The diff should be nothing...! All these files are already on `master` :stuck_out_tongue_closed_eyes:


coleshaw:
But, locally, I can push this back onto the subtree, it just dies because I don't have permissions to push directly to the repo. So I think this will work?

corps:
It looks like this diff is merge commits (which simply represent nodes that join history) and not actual fill diff commits, so I guess that makes sense???
also confused, possible weird git edge case we hit.  :crossed_fingers:

corps (approved):

coleshaw:
Bah, fails after I merged in master because of the original split commit ... will have to see if I can edit or purge that somehow...

coleshaw:
Okay...wound up pushing to the subtree manually from [8ed5164](https://github.com/mountetna/monoetna/pull/819/commits/8ed5164fe0959e179defc1a1974768973d7a787d), which seems to have at least updated the subtree, so going to close this PR. Not sure if that means future pull / push cycles will work better, but I guess let's see. :crossed_fingers:

PR #821, opened 2022-04-21
Improving the way linking parents works

corps:
Bug found while working with @dtm2451 , pushed this to production to validate.

Basically, the intention was that models that need to hang off the project should automatically get the project name filled in to save some convenience, but a typo prevented that.

Also included is the ability to override the single cell linker's model parent linking process.  Ideally, in the future, a naming service would allow airflow to pass a model name and a record name, and get back all the parent model / attributes that should be derived from it.  For now, this has to be managed project by project since there is not consistency in the lineage.


airflow/opt/providers/etna/etna/etls/metis.py
```diff
@@ -243,7 +244,7 @@ def _expand_project_parents(self, update: UpdateRequest, models: Dict[str, Model
                 continue
             model = models[model_name]
             self.log.info(f"Expanding project={project} for model {model_name}...")
-            if model.template.parent == "parent":
+            if model.template.parent == "project":
```
dtm2451, line 247: oooh is this where my xneo1 linking went wrong?

dtm2451 (approved):

airflow/opt/providers/etna/etna/etls/metis.py
corps: Yessir!

coleshaw (approved):

dtm2451 (changes requested):
Ooh, I think I just found another call that you make to `create_and_link_parents` which should be changed to just `link_parents`.  I couldn't highlight in the 'files changed' tab cuz there's nothing nearby currently, but looks like line 94, in the section with TaskGroup(f"{model_name}_processed"):

corps:
> Ooh, I think I just found another call that you make to `create_and_link_parents` which should be changed to just `link_parents`. I couldn't highlight in the 'files changed' tab cuz there's nothing nearby currently, but looks like line 94, in the section with TaskGroup(f"{model_name}_processed"):

You are correct, addressing now.

PR #825, opened 2022-04-27
Remove need for 'queryplot' flag and properly open Viz to all projects

dtm2451:
Noticed a bug with how I'd tried opening the Viz workflow up previously, and then found https://github.com/mountetna/monoetna/blob/45841a0f95285b669ba7f490285dc27169182eab/vulcan/lib/server/controllers/workflows_controller.rb#L56
Indeed, just removing the "projects" line from visualization.metadata.json works without the need for any of my additional edits to Vulcan code.  Thus, reverting the changes from my previous PR which attempted to set up the `"projects": "ALL"` method.

In the end, this all became wonderfully simple =)


coleshaw (approved):
This looks good to me, and works on local dev!

coleshaw:
Funny, the failing Timur UI test, I have no idea why it hasn't failed earlier ... I think the expected behavior has changed, so you could just delete that test, since the Query page always shows a "Plot" button now (just disabled or not).

PR #826, opened 2022-04-27
fix? attribute_report

### LINKED ISSUE #820: Timur Map attribute detail display breaks down for matrix attributes
Love the new map page details @graft!  I just noticed that it doesn't seem to render properly for matrix type attributes.... seems like the attribute list is emptied and the layout gets botched
map-matrix-bug
.

graft:
Attempt to fix #820 - it seems to work, but the `width: 0` bit is some css voodoo I don't understand, which makes me nervous.


dtm2451:
Looks good to me with the example project.

BUT I also noticed it had already rendered better for that project on master than what I put in #820 with the coprojects_template...
master-vs-fix_example-project
I wonder if the size of the validation data makes a difference?  Unfortunately I don't have a more full project in my dev that I could check with.

dtm2451:
Ahhhh I missed the scroll bar before.  I'm sure validation size is the cause of the visual difference from coprojects_template compared to example then, and adding wrapping does seem a good fix!  Curious how much space the 60k Ensembl id list still takes up when wrapped, but since it shows in the view for the single attribute, and also at the bottom below other features of the attribute, maybe it's okay for that list to take up a lot of space!

coleshaw:
60K ensembl ids seems to take up the same height, just scrolls forever. Must be a max-height set up on the validation or the container...

coleshaw (approved):

coleshaw:
No, I didn't see any sort of re-shifting of the panes like in your #820 gif. Just the text block with the ensembl ids is very long, so takes awhile to scroll to the bottom of it.

coleshaw:
Oh, but I wonder if it has to do with screen / browser width? Because I also don't get the "mobile view" like you have in your GIF above

coleshaw:
Oh, the mobile view I see is from `master`, not the PR branch....nevermind...

coleshaw:
Screenshot 2022-04-28 at 13-13-13 Dev Timur

dtm2451:
vertical scroll in the validation part itself!  Looks great!

dtm2451 (approved):

PR #828, opened 2022-04-28
Improve feedback on bad foreign key updates and projects.

corps:
Fixes taken from rollbar error messages.

1.  Admin users may have permission to view projects that don't exist, so explicitly catch and handle that case.
2. Temp id code handling is a bit... not great.  Many errors in rollbar exist in these paths, so I clarified some typing and added more explicit error messages when updating foreign keys with values that do not exist.

### DIFF
--- a/magma/lib/magma/attributes/foreign_key.rb
+++ b/magma/lib/magma/attributes/foreign_key.rb
@@ -11,7 +11,9 @@ class Magma
       return [ foreign_id, nil ] if value.nil?
 
       if value.is_a? Magma::TempId
-        [ foreign_id, value.real_id ]
+        id = value.real_id
+        raise Magma::LoadFailed(["#{model_name}.#{attribute_name} value was not a valid #{link_model.model_name} identifier."]) if id.nil?
+        [ foreign_id, id ]
       elsif link_identity && loader.identifier_exists?(link_model, value)
         [ foreign_id, loader.identifier_id(link_model, value) ]
       end
--- a/magma/lib/magma/loader/record_entry.rb
+++ b/magma/lib/magma/loader/record_entry.rb
@@ -125,13 +125,38 @@ class Magma
 
     def temp_entry
       # Replace the entry with the appropriate values for the column.
-      @record.map do |att_name,value|
+      errors = []
+
+      real_id = @loader.real_id(@model, @record_name)
+      if real_id.nil?
+        errors.push("#{@model.model_name} '#{@record_name}' has no backing identifier, could not be loaded.")
+      end
+
+      result = @record.map do |att_name,value|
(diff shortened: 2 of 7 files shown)


magma/lib/magma/attributes/foreign_key.rb
```diff
@@ -11,7 +11,9 @@ def entry(value, loader)
       return [ foreign_id, nil ] if value.nil?
 
       if value.is_a? Magma::TempId
-        [ foreign_id, value.real_id ]
+        id = value.real_id
+        raise Magma::LoadFailed(["#{model_name}.#{attribute_name} value was not a valid #{link_model.model_name} identifier."]) if id.nil?
```
corps, line 15: the `real_id` method can return nil, as which happens when the record_entry itself is nil.  https://rollbar.com/ucsf/magma/items/661/

magma/lib/magma/loader/record_entry.rb
```diff
@@ -125,13 +125,38 @@ def update_entry
 
     def temp_entry
       # Replace the entry with the appropriate values for the column.
-      @record.map do |att_name,value|
+      errors = []
+
+      real_id = @loader.real_id(@model, @record_name)
+      if real_id.nil?
+        errors.push("#{@model.model_name} '#{@record_name}' has no backing identifier, could not be loaded.")
```
corps, line 132: `real_id` can itself return nil in various cases as well, so we handle that explicitly.

magma/lib/magma/loader/record_entry.rb
```diff
@@ -125,13 +125,38 @@ def update_entry
...
+      end
+
+      result = @record.map do |att_name,value|
         attribute = @model.attributes[att_name]
         next unless attribute.is_a?(Magma::ForeignKeyAttribute)
-        [ attribute.foreign_id, @loader.identifier_id(attribute.link_model, value).real_id ]
+
+        identifier = @loader.identifier_id(attribute.link_model, value)
+
+        if identifier.is_a?(Magma::TempId)
```
corps, line 141: Sometimes the identifier value is an Integer, not a Magma::TempId, see https://rollbar.com/ucsf/magma/items/662/

magma/lib/magma/query/predicate/model.rb
```diff
@@ -67,7 +67,7 @@ def initialize(question, model, *query_args)
         child_extract(
           table.group_by do |row|
             row[identity]
-          end.first.last,
+          end.first&.last,
```
corps, line 70: https://rollbar.com/ucsf/magma/items/656/

graft (commented):
Some of these are so old it feels like saying goodbye to an old friend.

Issue #824, opened 2022-04-27
Vulcan: Re-enable the "download input params" feature


coleshaw:
In a meeting, a question was about for publication, being able to describe the workflow or input parameters (i.e. for reproducibility or otherwise saying something like "used scanpy version X, parameters Y, Z, abc"). Seems like the feature we had to download the inputs JSON file might help with that.

PR #831, opened 2022-04-30
Figure versions

graft:
Gives figures a basic history - seems to work, a final glitch remains to be sorted in that once loaded, the session does not update the outputs automatically. I think this should also handle #824 as the params hash is visible in the history interface.

I reused the RevisionHistory component from polyphemus, which might cause some typescript issues as I tried to move it into etna-js.


coleshaw:
Pretty neat! Some nits and issues:

* The update times only seem to reflect the most recent edit, or off-by-one (with the most recent one duplicated)? My original version (at the bottom of the list) should be something like March 29th:
Screenshot from 2022-05-02 09-49-43

* Comment lines don't seem like they are getting saved or displayed properly. For my edit above, the comment should have been "change title".
* Perhaps it would be nice to close the CodeMirror if you click on the revision again? Right now once you open a revision, there's no way to close it, i.e. if you want to scroll through a long list of revisions.
* Would a revision search be useful in the dialog window?
* Loading an old revision doesn't seem to change the other figure params (title) ... might be confusing to the user. Minor nit, though, since I think changing stuff like the title would make the UI think there had been an edit, which really isn't the case...perhaps in that case would need to change how the route and UI widget work.


I think it would be super cool and helpful if you could also show a thumbnail for each revision! But probably a lot more issues there, i.e. if a thumbnail was even generated, etc. Just wondering if people will find it difficult to make the connection between the inputs and what the output was for the given input.

coleshaw:
Hm...I'm also noticing (maybe?) that the increased number of figures for each revision means duplicate figure cards appear on the project's main page, with the same ID and title, etc. -- is that intended behavior? Or should there be a filter somewhere in the API that only returns the newest version of a figure?

coleshaw:
Okay, so in reference to the "updated at" bug I mentioned above, it looks like the updated_at times are getting messed up in the database itself....so that seems bad.

graft:
> Loading an old revision doesn't seem to change the other figure params (title) ... might be confusing to the user. Minor nit, though, since I think changing stuff like the title would make the UI think there had been an edit, which really isn't the case...perhaps in that case would need to change how the route and UI widget work.

Hmm, seems like changes to the title are only in memory, not in storage, so they don't persist over reloads, while the rest of changes DO persist over reloads, a strange result of the disconnect between "figure" and "session".

PR #832, opened 2022-05-03
Vulcan vectorize calculations

coleshaw:
This PR attempts to address the vector calculation idea from #765:

* Creates an `archimedes-node` container to run JavaScript workflow steps. Modifies the `archimedes` runner to account for other interpreters, extending the existing "use X image" functionality. Also updates the `Vulcan` orchestrator to find non-Python scripts specified in the CWL. Hopefully this sets up a pattern than can be used for R and other languages, as needed.
* The Vulcan UI widget for data transformation sets all data-frame cells as read-only, except for the column headers. This now lets the user only edit column headers or add new columns. In the new columns, the user can do whatever they want.
* There is a UI button that will propagate formulas defined in new columns to the entire data frame, so the user doesn't have to do the awkward scroll + drag action.
* UI widget only sends back 20 rows of data, and the server then calculates formulas for the entire data set using the same algorithm as above. This saves on data stored in the `inputs`, `localStorage`, and sent over the wire. We could cut this down if needed, but it does impact what the user sees when the re-open the dialog. The UI widget can only render what is in `inputs`, so if a user re-visits the step, it may be confusing that some previously-populated cells are blank. The Propagate button should make it convenient to repopulate, but it is something for users to be aware of.

Note that I had to do some weird things to account for how `npm` finds `node_modules` to make sure imports work ... not sure my implementation is very efficient, but I tried to bake things into the Docker images as early as possible. Hope that works as I think it works... :shrug: 

I did have to tweak some core docker stuff to get the TypeScript compiled in `etna-js`, so hoping I didn't break any of the other apps accidentally... :crossed_fingers: . Things seem to work okay locally...

Will add some annotations throughout the PR with some more details comments.

Screenshot from 2022-05-03 15-17-04


swarm/vulcan.1650532904.yml
```diff
@@ -19,10 +19,10 @@ services:
       - source: db-pw
         target: password
 
-  # Exists purely to allow updates to archimedes image on all possible nodes via service deployer.
+  # Exists purely to allow updates to archimedes and archimedes-node images on all possible nodes via service deployer.
```
coleshaw, line 22: Okay, so I'm probably doing this wrong, but wanted to make sure that Vulcan pulls in `archimedes-node` in the swarm ... not sure how to resolve the conflict or what's the right way to make sure production gets the right images, @corps ?

archimedes-node/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM archimedes-node-base
+# Perform these steps first to allow better caching behavior
+WORKDIR /app
+COPY . /app/
+
+RUN cd /app/archimedes-node && npm install && npm run build
```
coleshaw, line 6: Making sure the TypeScript files get converted to JavaScript

archimedes/archimedes/runner.py
```diff
@@ -52,6 +52,21 @@ class RunRequest:
     environment: List[str] = field(default_factory=list)
     image: str = "archimedes"
     isolator: str = "docker"
+    interpreter: str = "python"
+
+    def run_with(self, interp: str):
```
coleshaw, line 57: Some helper methods on the request object to match file extensions to the interpreter.

archimedes/archimedes/runner.py
```diff
@@ -181,6 +196,13 @@ def reserve_exec_dir(self) -> ContextManager[str]:
 
         return tempfile.TemporaryDirectory(dir=self.exec_dir)
 
+    def _cmd(self, request: RunRequest, exec_script_path: str):
+        if request.run_with("node"):
+            exec_dir = exec_script_path.replace(request.target_file, "")
+            return ["/bin/bash", "-c", f"cd {exec_dir}; cp /app/package.json package.json; npm install; node {exec_script_path}"]
```
coleshaw, line 202: Copying over the `package.json` and re-installing node modules in the same directory as the injected Vulcan script, because `node` expects to find them available locally and not globally. Hoping this isn't a burden because the dependencies should have been installed already and in the `npm` cache.

archimedes/archimedes/runner.py
```diff
@@ -245,23 +267,35 @@ def start(self, request: RunRequest, exec_script_path: str) -> LocalProcess:
...
 
+    def _cmd(self, request: RunRequest):
+        if request.run_with("python"):
+            return shutil.which('python')
+        else:
+            # If we want to support alternate language  engines for local development,
+            #   we'll have to pull the binary and archimedes-node
+            #   into the archimedes image via a multi-stage docker build.
+            # Otherwise, you just have to test via Vulcan.
+            raise ValueError(
```
coleshaw, line 283: For testing non-python scripts, I think you have to test it in the `archimedes-node` container itself, instead of via `archimedes-run` and the local isolator ... not sure if there's a way to make it work via the local isolator, but seems difficult without injecting Node, etc., all into `archimedes`, which seems like it could get messy.

docker/archimedes-node-base/Dockerfile
```diff
@@ -0,0 +1,60 @@
...
+    echo "export SCM_CHECK=false" >> ~/.bashrc
+
+# Add the docker client
+RUN curl https://download.docker.com/linux/static/stable/x86_64/docker-$DOCKER_VERSION.tgz > /tmp/docker.tgz
+RUN tar -xf /tmp/docker.tgz -C /tmp/
+RUN mv /tmp/docker/* /usr/local/bin/
+
+RUN npm install -g typescript@4.7.0-beta
+
+COPY --from=1 /etna /etna
```
coleshaw, line 57: Because some data frame utils are shared between `archimedes-node` and Vulcan UI widget, they are located in `etna-js`. So have to copy that in from the `etna-base` image to here.

etna/Dockerfile
```diff
@@ -2,3 +2,5 @@ FROM etna-base-dev
 ENV ETNA_GEM_DEVELOPMENT=1
 COPY . /etna
 RUN rm -rf /app && ln -sfT /etna /app
+
+RUN cd /etna && npm run build
```
coleshaw, line 6: My attempt to compile the TypeScript files in `etna-js` into the image, so any imports in the other apps can work.

etna/packages/etna-js/package.json
```diff
@@ -3,14 +3,19 @@
...
   },
   "author": "",
   "license": "GPL-2.0-only",
   "dependencies": {
     "animate.css": "^4.1.0",
+    "hyperformula": "^2.0.0",
     "react-notifications-component": "^2.4.0"
   },
+  "bundledDependencies": [
+    "hyperformula"
```
coleshaw, line 17: This seems to make the other apps not break, even though I've introduced a new `etna-js` dependency ?? :shrug: Without this, non-affected apps like Metis would break upon webpack build because they have to include all `etna-js` dependencies, even if they don't use them directly, so we had to add the new dependencies to every application. However, I think there is some [side effect](https://stackoverflow.com/a/25044361) when used with local packages, where these dependencies are then included or found? They did automatically propagate to the `package-lock.json` files in each app, without me having to add them to `package.json` manually, which seems like a plus...

coleshaw: Either this line or explicitly calling it out in line 13 above...

etna/packages/etna-js/selectors/maybe.ts
```diff
@@ -0,0 +1,66 @@
+export type Maybe<T> = null | [T];
```
coleshaw, line 1: Because some of these methods are used in the data frame utils, moving them to `etna-js`

etna/packages/etna-js/utils/dataframe.ts
```diff
@@ -0,0 +1,171 @@
+import {HyperFormula} from 'hyperformula';
```
coleshaw, line 1: Shared between `archimedes-node` and `Vulcan`, so consolidating code in `etna-js`

vulcan/config.yml.template
```diff
@@ -28,6 +28,9 @@ default: &default
   :token_name: JANUS_DEV_TOKEN
   :log_level: info
   :archimedes_run_image: etnaagent/archimedes
+  :archimedes_interpreters:
```
coleshaw, line 31: Will this get translated to production `config.yml`?

vulcan/lib/server/models/workflow.rb
```diff
@@ -161,8 +162,15 @@ def input_as_json(input)
       def lookup_operation_script
         return nil if script_name.nil?
 
-        script_path = File.join(Vulcan.instance.config(:workflows_folder), "scripts/#{script_name}.py")
-        if ::File.exists?(script_path)
+        script_path = nil
+
+        Find.find(File.join(Vulcan.instance.config(:workflows_folder), "scripts/")) do |path|
```
coleshaw, line 167: This change allows non-Python scripts to be found. It also means we can organize the scripts into project-specific ones, if that is the long-term goal to support editing.

vulcan/lib/server/workflows/visualization.cwl
```diff
@@ -38,18 +38,25 @@ steps:
     in:
       data_frame: get_data/data_frame
     out: [formulaic_data, calculated_data]
+  extend_user_formulas:
```
coleshaw, line 41: Adding this new step to the viz workflow seems to work ...

swarm/vulcan.1650532904.yml
corps: Yes, that is correct.  You can test this before merge by doing a `PUSH_IMAGES=1 IMAGES_PREFIX=etnaagent/ IMAGES_POSTFIX=:production make -C archimedes-node release` before hand (pushing the image to dockerhub) and then making this change inside of portainer under the vulcan stack.

etna/Dockerfile
corps: :+1: 
Ultimately, a "single package.json / webpack.js" setup will look like this, too -- build into etna-base-dev, and share that out to other images.

etna/packages/etna-js/package.json
corps: Yes, but I need to warn you -- my experience is that each npm release __changes its mind about how this works__ so we may see issues in the future.  The only real solution which is beyond our scope is to move towards that single npm package approach.

vulcan/config.yml.template
corps: Nope, we'll want to add this into production.

```
./bin/portainer-login
./bin/portainer-edit-config vulcan.production.yml
```

corps:
The docker changes look good!  This approach can work I think.   Makes me definitely think that moving the scheduling into airflow has some benefit here, a future project.  But this approach is sound.

coleshaw:
Thanks! Will review your other comments in detail tomorrow. 

I also think airflow will help with figure versions and pinning workflows to a specific SHA … one of the things I hope to discuss with you later this week when we can find time to pair. 

Thanks!

etna/packages/etna-js/package.json
coleshaw: Well phooey... okay, so I guess we'll get bitten by this at some point and revisit ... :cry:

vulcan/config.yml.template
coleshaw: Those are some nice / handy scripts for updating the config and deploying! :clap: :tada:

PR #835, opened 2022-05-12
Cs/airflow box fix binary transfer

coleshaw:
This PR fixes a couple of issues with the Box ETL helpers:

* Switched from `ntransfercmd` to just `transfercmd`, since the size returned was always `None` ... so not really useful.
* Spaces in the folder path did not play well with the `cwd` and then `transfercmd` commands, so I've gotten rid of the `cwd`. Since we have the full paths to each file, it isn't necessary to call `cwd` anyways.
* Large binary files weren't being transferred correctly (the last blob wasn't uploading right), and the file would get corrupted at the destination. Turns out with the lower-level `ntransfercmd` or `transfercmd` commands, you have to explicitly force the connection to binary via `TYPE I` command before creating the socket, to force binary transfer mode. Default mode is ASCII, which I didn't catch with my initial test (text) files.
* Small quality of life feature to allow for a `split_folder_name` for non-flattened ingestion, to better mimic the Box folder structure in Metis.
* Trying to keep a single FTPS connection for transferring all files was not working well, because `ftplib` would not consistently flush the status string of the previous commands, thus causing a second `transfercmd` call to fail. (It would report `226 Transfer Complete` on the second call instead of returning a socket, which was the final status of the first call). I've been unable to force a clearing or flush of these situations, and the cleanest way I've found is to just create a new FTP connection per transfer.


airflow/opt/providers/etna/etna/hooks/box.py
```diff
@@ -211,16 +211,17 @@ def retrieve_file(self, ftps: FTP_TLS, file: FtpEntry) -> Tuple[BinaryIO, int]:
...
         """
-        ftps.cwd(file.folder_path)
-        sock, size = ftps.ntransfercmd(f"RETR {file.name}")
+        # force binary file transfers via the socket
+        ftps.voidcmd('TYPE I')
 
-        return sock.makefile(mode='rb'), size
+        # Required because...otherwise the next ftps call may return the status
+        #   message from the previous command. :shrug:
+        ftps.voidcmd('NOOP')
```
corps, line 219: :facepalm: Nice catch

corps (approved):

corps:
Part of the problem we're experiencing, too, is that it seems FTP server implementations can vary significantly due to ambiguity in the spec as part of connection state handling.  So, yeah.  Excellent catches!

PR #836, opened 2022-05-12
+ model_names and attribute_names verbs

graft:
Adding model_names and attribute_names verbs to the query API, just to make this information easier to retrieve. Again the return types are a bit squirrelly, just bare json arrays.

### DIFF
--- a/magma/lib/magma/server/query.rb
+++ b/magma/lib/magma/server/query.rb
@@ -7,9 +7,13 @@ class QueryController < Magma::Controller
     return failure(401, errors: ["You are unauthorized"]) unless @user && @user.can_view?(@project_name)
 
     begin
-      if @params[:query] == "::predicates"
-        return success(Magma::Predicate.to_json, "application/json")
+      case @params[:query]
+      when "::predicates"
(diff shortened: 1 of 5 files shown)


magma/lib/magma/server/query.rb
```diff
@@ -7,9 +7,13 @@ def action
     return failure(401, errors: ["You are unauthorized"]) unless @user && @user.can_view?(@project_name)
 
     begin
-      if @params[:query] == "::predicates"
-        return success(Magma::Predicate.to_json, "application/json")
+      case @params[:query]
+      when "::predicates"
+        return success_json(Magma::Predicate)
+      when "::model_names"
+        return success_json(answer: Magma.instance.get_project(@project_name).models.keys, format: [ 'String' ])
```
coleshaw, line 14: super-nit: do you want to also provide a `"type": "Array"` here, for consistency with the other endpoints, i.e. `::attribute_names`?

coleshaw (approved):
Nice, works well!

corps:
Cool, I think this could be quite useful.

PR #838, opened 2022-05-13
tweak config for scheduler health check timeout

coleshaw:
Trying to run the Box ingest ETLs, and I was running into [this problem](https://github.com/apache/airflow/issues/19277), I think, where the Job would get killed with a Sigterm, like the reported issue (text is from the GH issue, I can't figure out how to get old logs from our Airflow instance?):

```
worker>     [2021-10-28 02:51:29,343] {local_task_job.py:208} WARNING - State of this instance has been externally set to None. Terminating instance.
worker>     [2021-10-28 02:51:29,347] {process_utils.py:100} INFO - Sending Signals.SIGTERM to GPID 25492
worker>     [2021-10-28 02:51:29,347] {taskinstance.py:1236} ERROR - Received SIGTERM. Terminating subprocesses.
```

This happened multiple times today, and each time Airflow restarts the ingest DAG task, which at this point basically re-ingests the same files before it gets killed again :cry: 

Following the threads, seems like it might be an issue with the task job not being able to reach out to the scheduler via a health check within the specified timeout, so the scheduler kills the job. I think this happens during large file uploads (5GB), so seems possible? The threads indicate this should be fixed in 2.2.2, but we're running 2.2.4 already ... so I've already pushed a production image with an increased health check timeout to see if it helps our case. Will only merge this PR if the issue seems to be fixed.

### DIFF
--- a/airflow/opt/airflow.cfg
+++ b/airflow/opt/airflow.cfg
@@ -833,7 +833,7 @@ pool_metrics_interval = 500.0
 # If the last scheduler heartbeat happened more than scheduler_health_check_threshold
 # ago (in seconds), scheduler is considered unhealthy.
 # This is used by the health check in the "/health" endpoint
-scheduler_health_check_threshold = 60
+scheduler_health_check_threshold = 120
 
 # How often (in seconds) should the scheduler check for orphaned tasks and SchedulerJobs
 orphaned_tasks_check_interval = 300.0
--- a/airflow/opt/providers/etna/docs/etna/hooks/box.html
+++ b/airflow/opt/providers/etna/docs/etna/hooks/box.html
@@ -155,7 +155,7 @@
 <span class="kn">import</span> <span class="nn">os</span>
 <span class="kn">import</span> <span class="nn">re</span>
 <span class="kn">from</span> <span class="nn">typing</span> <span class="kn">import</span> <span class="n">List</span><span class="p">,</span> <span class="n">Dict</span><span class="p">,</span> <span class="n">Optional</span><span class="p">,</span> <span class="n">ContextManager</span><span class="p">,</span> <span class="n">Tuple</span><span class="p">,</span> <span class="n">BinaryIO</span>
-<span class="kn">from</span> <span class="nn">ftplib</span> <span class="kn">import</span> <span class="n">FTP_TLS</span><span class="p">,</span> <span class="n">error_reply</span>
+<span class="kn">from</span> <span class="nn">ftplib</span> <span class="kn">import</span> <span class="n">FTP_TLS</span>
 <span class="kn">from</span> <span class="nn">socket</span> <span class="kn">import</span> <span class="n">socket</span>
 
 <span class="kn">import</span> <span class="nn">cached_property</span>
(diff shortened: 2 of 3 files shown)


coleshaw:
Just some additional logging notes from the scheduler:

```
default, priority_weight=1, operator=_PythonDecoratedOperator

[2022-05-13 18:48:23,793] {scheduler_job.py:300} INFO - 2 tasks up for execution:

	<TaskInstance: mvir1_data_metis_etl15.bulk_rna_seq_attributes.process_attributes_file_ scheduled__2022-05-12T18:45:35.048215+00:00 [scheduled]>

	<TaskInstance: mvir1_data_metis_etl15.propagate_folder_updated_at scheduled__2022-05-12T18:45:35.048215+00:00 [scheduled]>

[2022-05-13 18:48:23,797] {scheduler_job.py:329} INFO - Figuring out tasks to run in Pool(name=default_pool) with 122 open slots and 2 task instances ready to be queued

[2022-05-13 18:48:23,797] {scheduler_job.py:357} INFO - DAG mvir1_data_metis_etl15 has 5/5 running and queued tasks

[2022-05-13 18:48:23,798] {scheduler_job.py:364} INFO - Not executing <TaskInstance: mvir1_data_metis_etl15.bulk_rna_seq_attributes.process_attributes_file_ scheduled__2022-05-12T18:45:35.048215+00:00 [scheduled]> since the number of tasks running or queued from DAG mvir1_data_metis_etl15 is >= to the DAG's max_active_tasks limit of 5

[2022-05-13 18:48:23,799] {scheduler_job.py:357} INFO - DAG mvir1_data_metis_etl15 has 5/5 running and queued tasks

[2022-05-13 18:48:23,799] {scheduler_job.py:364} INFO - Not executing <TaskInstance: mvir1_data_metis_etl15.propagate_folder_updated_at scheduled__2022-05-12T18:45:35.048215+00:00 [scheduled]> since the number of tasks running or queued from DAG mvir1_data_metis_etl15 is >= to the DAG's max_active_tasks limit of 5

[2022-05-13 18:48:23,799] {scheduler_job.py:433} INFO - Setting the following tasks to queued state:

	

[2022-05-13 18:50:19,988] {scheduler_job.py:1137} INFO - Resetting orphaned tasks for active dag runs

[2022-05-13 18:50:19,993] {scheduler_job.py:1160} INFO - Marked 1 SchedulerJob instances as failed

[2022-05-13 18:50:20,003] {scheduler_job.py:1201} INFO - Reset the following 1 orphaned TaskInstances:

	<TaskInstance: xhlt2_ingest_from_d2b8.cytof_files.ingest scheduled__2022-05-12T18:24:41.710381+00:00 [running]>

[2022-05-13 18:55:20,476] {scheduler_job.py:1137} INFO - Resetting orphaned tasks for active dag runs

```

That text "Marked 1 SchedulerJob instances as failed" is flagged in the GH issue above as an indicator of the problem :shrug:  Note that the ingest DAG task is flagged as orphaned, so sounds bad...this is post-my config change above. Note that there also appear to be issues where more than 5 DAGs are running right now, so some are getting delayed. Not sure if we should increase that limit?

coleshaw:
Note that the logging tweak hasn't been pushed to production. I just noticed that multiple chunks are uploaded each second...so multiple lines are getting logged on 5-second intervals. This change should make just a single line gets logged every 5 seconds.

PR #839, opened 2022-05-14
Map diff

graft:
More map improvements:
1. Adds 'group' column to the attribute table for a model. There is no tag behavior yet as discussed today.
2. Adds sorting to the attribute table.
3. Adds column-specific filtering
4. Adds map diff:
map-diff

Assumes #836 will close.

### DIFF
--- a/timur/lib/client/jsx/components/model_map/model_report.jsx
+++ b/timur/lib/client/jsx/components/model_map/model_report.jsx
@@ -1,15 +1,15 @@
 import { useDispatch } from 'react-redux';
 import React, { useState, useCallback } from 'react';
-import { sortAttributes } from '../../utils/attributes';
-import {useReduxState} from 'etna-js/hooks/useReduxState';
+import { sortAttributeList } from '../../utils/attributes';
+import SelectProjectModelDialog from '../select_project_model';
 import {requestAnswer} from 'etna-js/actions/magma_actions';
+import {getDocuments} from 'etna-js/api/magma_api';
 import {useActionInvoker} from 'etna-js/hooks/useActionInvoker';
-import {selectUserPermissions} from 'etna-js/selectors/user-selector';
+import {isEqual} from 'lodash';
 
 import Grid from '@material-ui/core/Grid';
 import Typography from '@material-ui/core/Typography';
 import TextField from '@material-ui/core/TextField';
-import InputAdornment from '@material-ui/core/InputAdornment';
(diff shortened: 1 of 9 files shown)


coleshaw:
Looks neat! I'm a bit confused about the color-scheme ... which color means "in the original model", "in the other model", and "in both models" (maybe I'm just too much of a diff n00b)? Might be nice to have a legend or additional icons on the page (like `<<` or `>>`), especially since I'm not sure red / green are color-blind friendly and might appear exactly the same to some users.

Also, is there a reason you have to click the checkmark next to the model name in the dialog, to fetch the attribute names? Seems a bit counterintuitive UX behavior ... like, I would expect that the attribute names would appear right after selecting the model?

graft:
> I'm a bit confused about the color-scheme .

Yeah I worried about this, i think adding some +/- indicators is right.

The checkmark on the input is so you can type in a project name, the dropdown only shows what is on your token.

timur/lib/client/jsx/components/model_map/model_report.jsx
```diff
@@ -137,38 +179,106 @@ const ModelReport = ({ model_name, updateCounts, counts, template, setAttribute
...
+        model_name,
+        record_names: [],
+        attribute_names: 'all'
+      },
+      fetch
+    ).then(({models}) => setDiffTemplate(models[model_name].template));
+  }
+  const attributes = Object.values({
+    ...template.attributes,
+    ...diffTemplate && diffTemplate.attributes
```
corps, line 241: hmm, does this expression work out?

assume `diffTemplate = null` => `...null && (null.attributes)` => `...null`.
I forget, but doesn't expanding a null result in an exception?

corps: Also, in javascript, `{}` is truthy, not falsy

graft: Seems to work as intended, ...falsy expands to nothing, i.e., `x = { ...null } => { }`.

coleshaw: That's fascinating! Because I also thought `{...null}` would throw an exception, but it does seem to work in an assignment...

```
> {...null}
Uncaught SyntaxError: expected expression, got '...'
debugger eval code:1:1
> x = {...null}
Object {  }
```

PR #842, opened 2022-05-17
switch box etl to use cursor instead of file timestamps

coleshaw:
Per discussion in Slack, seems like using a cursor-based method for scanning new Box files may be more reliable than the FTP timestamps (and perhaps even the API timestamps). This PR implements the change by storing the cursor in an Airflow Variable, and using the file `size-modify-create` as a hash to detect changed files.


corps (approved):

coleshaw:
Also added in a convenience method for alerting to slack, from Box helpers, and removed references to `batch_start` and `batch_end` to avoid confusion about how it operates.

PR #843, opened 2022-05-18
Remove attribute

corps:


magma/lib/magma/migration.rb
```diff
@@ -152,7 +172,13 @@ def column_entry name, type
     end
 
     def remove_column_entry name
-      "drop_column :#{name}"
+      now = DateTime.now.to_time.to_i
+      new_name = "#{name}_#{now}_backup"
+
+      [
+        "drop_foreign_constraint_if_exists :#{name}",
```
coleshaw, line 179: Does dropping the foreign constraint affect the existing data in that column at all?

Separately, is there some other entry that has to get modified in the `attributes` table, in this case? I think there might be a reference from the foreign table attribute ... i.e. if

`sc_seq` --> `sc_seq_pool`

And you drop the `sc_seq_pool` attribute on `sc_seq`, not only do you have to change the `sc_seq_pool` attribute where `model_name=sc_seq`, but will you also have to remove the `sc_seq` collection attribute where `model_name=sc_seq_pool`? Hopefully that makes sense...

magma/lib/magma/actions/remove_attribute.rb
```diff
@@ -0,0 +1,75 @@
...
+    def validate_attribute_exists
+      return if attribute
+
+      @errors << Magma::ActionError.new(
+        message: 'Attribute does not exist',
+        source: @action_params.slice(:attribute_name, :model_name)
+      )
+    end
+
+    def validate_restricted_attribute
```
graft, line 41: Superfluous?

magma/lib/magma/migration.rb
graft: Yeah I'm concerned we'll have to do this in two parts, the easy part is removing data attributes, the hard part is unlinking models, which has all sorts of issues if we are touching parent attributes (like, how to maintain a connected graph while changing fundamental relations) - I'd say `remove_attribute` should only deal with the former.

corps: > Does dropping the foreign constraint affect the existing data in that column at all?

Nope, in fact, removing the constraint is what allows us to preserve the existing data better.  Basically, the data can remain always preserved and not subject the model to update failures from an invisible attribute.  If we don't remove the constraint, we are forced to  add data to the 'removed' relation on every new record (because nil or 0 won't work if it's not a real relationship).

corps: > Yeah I'm concerned we'll have to do this in two parts, the easy part is removing data attributes, the hard part is unlinking models, which has all sorts of issues if we are touching parent attributes (like, how to maintain a connected graph while changing fundamental relations) - I'd say remove_attribute should only deal with the former.

What is the goal of remove attribute?  I'd say, the focus of remove attribute is primarily to clean and refactor models when we need to make major changes.  In that case, I really feel handing it clear, useful power is the correct approach here.  I think we could spend a lot of time over engineering tools that prevent 'bad cases' but also don't solve 'necessary cases'.  

Here's my thought: removing a parent attribute and unlinking a model has no catostrophic effect other than to temporarily unlink the model until a new parent attribute is applied.  No data loss, no system downtime, just a temporarily inconvenient state that can be repaired.  BUT, it might be the obvious natural path in a multi step process to making a refactor of some other sort.

I think we should allow that step, because the alternative feels like additional work that will result in an unintuitive tool.  Eg: autocomplete trying to fight me about what I'm trying to type.  Sure, it might prevent some mistakes, that were never a big deal anyways, but it also definitely causes me "flight blindness" when I expected typing A -> "A", but I get A -> "A....<bs><bs><bs>Furniture"

magma/lib/magma/actions/remove_attribute.rb
corps: Yup, from copy paste, will remove.

graft:
A few thoughts stemming from recent conversations:

1) In discussing the single-cell model implementation, there is a desire to implement a tree within a single model (that is, some sc_seq records are parents of other sc_seq records). Although we might be able to technically reason what the reciprocal pair of links are in this specific case, I think it now behooves us to make the pairing of links a formal magma requirement so that this kind of link-tracing can be done more correctly in general.
In fact this has been an ad-hoc requirement for a while, although Magma does not explicitly test it, and many of the querying paths will not work correctly if they do not exist reciprocally. In addition, the link-pair formalizes the relationship between two models, which are reciprocal in fact (i.e., if I am your parent, you must be something to me, whether or not the relationship has been formally described). Thus, imposing this requirement is only forcing us to be explicit about what must already be the case. This can be done by adding the property `link_attribute_name` to the attribute, which will allow the two halves of the link relationship to point to each other (i.e., each will have link_model_name and link_attribute_name).
Needless to say if we make this change it would have implications for the way remove_attribute works with respect to links; this suggests to me leaving links out of it for the moment and only dealing with data attributes.

2) Previously the update_model API did not introduce breaking changes to the graph; that is, every change made produced a valid graph that Magma could continue to query. This is easily done when adding new models or links between models, but removal is more complex in the case where parenting is involved. This means avoiding an operation which cracks a valid link in half and makes it invalid, or unparents a model from the graph. I think this can be done with two link-related removal operations:
- `remove_link` - removes a `link` attribute (i.e., a non-parenting foreign key), and automatically removes its reciprocal also. The graph remains intact.
- `reparent_model` - Since the graph requires a model to always be descended from the project model, we cannot simply remove a parent attribute, we must instead remove it and set a new one at the same time. Like add_model, this requires `model_name`, `parent_model_name` and `parent_link_type`. We remove the old `parent` foreign key attribute and its reciprocal, and add a new parent attribute pointing to the new parent model, along with its reciprocal attribute in the new parent model (that is, four attributes are involved in the operation, two created and two removed). The graph remains intact, although necessarily the data below the reparented model will become detached and new parent identifiers will have to be set again via `update`.

PR #848, opened 2022-05-20
metis validate

graft:
Adds a `validate` command to metis, to match the `nuke` command - this one merely checks MD5s and reports discrepancies, but does not actually remove files, in case people want to ensure their backup to Metis is correct but continue working with local files. Perhaps it should be a flag to `nuke` instead, but this seems less accident-prone.


coleshaw:
Hm, but this doesn't verify that they are in the local directory structure, like the opposite of `nuke`? I would check perhaps with Emily, but I feel like this might be misleading or not exactly the expected behavior ... will it be confusing if the md5 exists, but in a different location, and then they can't find it (but `validate` said it's there!)? 

From the ask, I thought that they would want something more like, "check this directory structure and make sure the files exist in Metis at the same locations, plus have matching md5sums"?

graft:
`nuke` also doesn't care whether there is a matching directory structure on Metis with similar file names, etc., so I thought this seemed the most direct counterpart. It seems like 99% of the time the sequence will go like `put <dir>`, `validate <dir>`, and possibly later `nuke <dir>`, so even if the file exists somewhere else, the dir will be mirrored. Perhaps a separate `compare <local_dir> <remote_dir>`?

coleshaw:
Ooohhh, didn't realize `nuke` behaved that way. From the phrasing in the help doc, I thought it literally checked directory structure ... may not need a `compare` command, perhaps it's just my misunderstanding.

I think you're right, if the workflow is typically `put`, `validate`, `nuke` (maybe), with no expectation that files are (manually?) deleted on Metis during some intervening time period between running those commands, the output behaves as expected.

:+1:

PR #850, opened 2022-05-23
Vulcan figures store hashes

coleshaw:
This PR is a proposal / attempt at implementing a simple form of computational-reproducibility into Vulcan, where `archimedes` and `archimedes-node` versions are specified by the DockerHub image SHA.

Would like to present this and show next week at standup once everyone is available, to get live feedback. Opening the PR to collect feedback or thoughts, as I feel there is a high chance I have introduced some bugs or weird behavior.

Assumed workflow / outcome that this PR implements:

* Each Figure is pinned to a specific SHA of vulcan / archimedes / archimedes-node, as well as a specific snapshot of its workflow. Currently the vulcan SHA is unused, but perhaps if we move the scheduling to Airflow, it may be of use :shrug: 
* Whenever you submit to a given Figure, the orchestrator runs the pinned workflow-snapshot version as well as the specific archimedes / archimedes-node versions. This does mean that if an image is not available locally on Vulcan machine, docker will download it, so computation may take more time.
* If a workflow snapshot or image SHA does not exist or is not specified, then the current version is used.
* The UI will update the workflow steps to reflect the snapshot of a Figure. 
* Selecting a past Revision in the UI means the workflow steps may change visually (see previous bullet point). I don't have a clear way to indicate this right now...so perhaps will be confusing.
* You can only run past Revisions and tweak them, but you cannot save them -- trying to avoid any issues with forking or branching of Figures, but perhaps with better UI we might want to let folks fork / save revisions. 
* Included a convenience command to populate the workflows and dependency SHAs for current figures.

Note: not touching any sort of workflow-editing tools, though I think they could integrate with this workflow-snapshot idea. But I think something like this PR would be required before we get to workflow-editing.


vulcan/db/migrations/006_add_workflow_snapshots.rb
```diff
@@ -0,0 +1,29 @@
...
+      String :name, null: false
+      json :projects
+      String :vignette, text: true
+      String :display_name
+      String :description
+      String :cwl_version, null: false, default: "v1.1"
+      json :inputs
+      json :outputs
+      json :steps, null: false
+      json :other_metadata
```
coleshaw, line 17: This currently stores non-CWL-type stuff, like the params used by the Query page to render the button. Seems like a catchall, could model it differently.

vulcan/lib/client/jsx/components/workflow/session/advanced_session_controls.tsx
```diff
@@ -0,0 +1,69 @@
...
+    <>
+      <IconButton
+        onClick={(e) => {
+          e.stopPropagation();
+          setMenuAnchor(e.currentTarget);
+        }}
+      >
+        <MoreVertIcon />
+      </IconButton>
+      <Menu
```
coleshaw, line 59: Currently updating a figure's dependencies (SHA and workflow-snapshot) is a distinct action. I could also imagine this as a check-box on the "Save" button, so you could update at the same time you save the figure. Not sure which is preferable, since I tend to view updating dependencies as something you would rarely do, so have left it as separate for now.

vulcan/db/migrations/006_add_workflow_snapshots.rb
corps: So, to be clear -- are we planning to use this field immediately for the query page render button?  Should we rename this field to reflect that?  I feel like we'd want to avoid any 'catch all' objects and simply explicitly extend the model to meet new needs so that it is more clear what the data actually is.

vulcan/db/migrations/006_add_workflow_snapshots.rb
```diff
@@ -0,0 +1,29 @@
+Sequel.migration do
+  up do
+    create_table(:workflow_snapshots) do
+      primary_key :id
+
+      foreign_key :figure_id, :figures, null: false, unique: true
+      json :authors, null: false
```
corps, line 7: Does it possibly make sense to model snapshots as just, a raw yml payload with none of the parsing being normalized?

`cwl_yaml` + `image`?
Also, does image itself need to be stored here?  What if the image url is simply deterministic based on the snapshot id itself?

corps: Maybe I'm wrong though, I'm just curious how you felt: what were your thoughts into the table being broken out this way?

vulcan/db/migrations/006_add_workflow_snapshots.rb
coleshaw: Okay, sure I can rename it. It is currently used for the query page render metadata.

vulcan/db/migrations/006_add_workflow_snapshots.rb
coleshaw: I wasn't sure if we wanted to keep it as CWL YAML, since at one point we had talked about perhaps moving workflows to a database, etc., and then generating the YAML on the fly. Since the UI-side uses this JSON representation, I thought it would be safe to mirror that, though you're right, it introduces complexity in the conversion. 

I also wasn't sure if this would more closely reflect the format of workflow editing, though since that isn't really well defined, perhaps it shouldn't / doesn't carry much weight at this point, as a consideration.

corps: Which metadata again?  Sorry.

coleshaw:
Per feedback at standup:

* [ ] Change workflow snapshots to YAML
* [ ] Show the SHAs in the revision diff
* [ ] Also store monoetna / vulcan git SHA
* [ ] Let the orchestrator checkout the git files from the above SHA and load the Python scripts from there (solves Vulcan script-changing issue)
* [ ] Better handle the case when the Archimedes SHAs do not exist in DockerHub -- fallback to `:production`
* [ ] UI should not let you "update dependencies" unless in the newest revision.

coleshaw:
> What if the image url is simply deterministic based on the snapshot id itself? 

Unsure what this means in practice? Like, have some lookup table that just says snapshot id -> image (as blob or something)?

In general, I'm not sure if we need to be worried too much about the image, or perhaps even most of the non-YAML metadata. The only things currently used by the UI should be in the YAML, I believe...

corps:
> > What if the image url is simply deterministic based on the snapshot id itself?
> 
> Unsure what this means in practice? Like, have some lookup table that just says snapshot id -> image (as blob or something)?

No state, no lookup tables.

If id = 1, url = https://vulcan.ucsf.edu/snapshot-images/1

Deterministic from identifier.

coleshaw:
Oh, interesting! :thinking:

corps:
Also, so long as snapshots are immutable, we can use that fact to hash snapshot images deterministically just based on the id.

Makefile
```diff
@@ -1,5 +1,6 @@
 export COMPOSE_PROJECT_NAME=monoetna
 export NODE_ENV=development
+export MONOETNA_SHA=$(shell git rev-parse HEAD)
```
coleshaw, line 3: This is basically used in development to allow testing of the git-sha being stored in a figure's `dependencies`

docker/archimedes-base/Dockerfile
```diff
@@ -25,6 +25,7 @@ RUN apt-get update \
       gnupg \
       bash-completion \
       git \
+      tk \
```
coleshaw, line 28: Ran into an issue where poetry install was throwing this as missing for one of the dependencies...not sure if the underlying image changed at some point?

docker/build_support/build_image
```diff
@@ -100,6 +101,7 @@ function runBuild() {
     --cache-from $tag \
     --cache-from $baseTag \
     --build-arg APP_NAME=$baseTag \
+    --build-arg MONOETNA_SHA=$monoetnaSha \
```
coleshaw, line 104: For the production / staging images, this bakes the git sha into the image directly.

docker/build_support/compose
```diff
@@ -23,8 +23,8 @@ export GROUP_ID=$(id -g)
 
 cp $DIR/docker-compose.shared.yml /tmp/composed.yml
 for compose_yml in $@; do
-    docker-compose -f $compose_yml config > /tmp/next.yml
-    docker-compose -f /tmp/next.yml -f /tmp/composed.yml config > /tmp/composed.yml.next
+    docker-compose -f $compose_yml config --no-interpolate > /tmp/next.yml
```
coleshaw, line 26: Because the Vulcan image requires `MONOETNA_SHA` to be an environment variable, when that gets composed into the overall `docker/docker-compose.yml`, if it is interpolated then it becomes static and won't update unless the `docker/docker-compose.yml` file is rebuilt. I think this file only gets rebuilt if any of the app `docker-compose.yml` files changes, so code changes or new commits wouldn't necessarily update this value. But, if we don't interpolate at this step, in `docker/docker-compose.yml` it remains as a reference to the environment variable, and subsequent `make up`, `make restart`, etc., calls will include an updated SHA.

swarm/vulcan.1650566317.yml
```diff
@@ -63,7 +63,7 @@ services:
       - |
         set -e
         export METRICS_PW="$$(cat /run/secrets/metrics-pw)"
-        exec /bin/deploy.sh migrate 'docker pull etnaagent/archimedes:production' 'docker pull etnaagent/archimedes-node:production' -- /app/bin/puma.sh
+        exec /bin/deploy.sh migrate 'docker pull etnaagent/archimedes:production' 'docker pull etnaagent/archimedes-node:production' 'docker pull etnaagent/vulcan:production' -- /app/bin/puma.sh
```
coleshaw, line 66: Also need to make sure that the `etnaagent/vulcan:production` image with tag are up-to-date on the production system, so we can pin the dependency.

vulcan/docker-compose.yml
```diff
@@ -41,6 +41,7 @@ x-base-etna-services:
     environment:
       <<: *base-etna-env
       WAIT_FOR_DB: 1
+      MONOETNA_SHA: ${MONOETNA_SHA}
```
coleshaw, line 44: Pass this into the container

vulcan/lib/orchestration.rb
```diff
@@ -282,73 +269,77 @@ def build_target_for(step_name, cache = {})
...
+          elsif (step = workflow.find_step(step_name))
             if step.ui_behavior?
-              ref = session.material_reference_for([step_name, step_out.id])
-              source = material_source(ref)
-              input_files << source.take_as_input(step_out.id)
-              script[step_out.id] = input_files.last
+              script = {}
+            elsif step.script_name
+              # Check snapshot first for the stored script
+              script = session.snapshot_script(step.id)
```
coleshaw, line 304: In the midst of the formatting change, this is the change that first checks for the script being defined in the snapshot. If it's not there, then take the current definition on disk. This seemed better than pinging GH for a raw script file, since we'd have to store the path to the file anyways in order to fetch it.

coleshaw:
So I think this has addressed the feedback from standup a couple weeks ago.

I haven't done anything with regards to the suggestions for serving up images from the snapshots, because those images aren't currently displayed anywhere. I think that might be more relevant to workflow editing, which I tried to leave out of this PR, so have left it for the future when we get around to working on editing.

PR #858, opened 2022-05-25
use retry logic for all requests

coleshaw:
I think this wraps all the non-upload metis_client requests with the same retry logic, and manual testing seems to work for the various Metis commands and Janus token refresh ...


corps (approved):
:+1:

PR #861, opened 2022-05-26
Janus community projects model

coleshaw:
This PR sets up the Janus models for community projects as well as the code of conduct agreement log.


corps:
Kyewl.  But one note -- should we be validating pi emails?  Possibly stripping whitespace as well?

coleshaw:
:+1: for stripping whitespace. I think we declined to validate emails in other contexts -- this code should check that it's a ucsf.edu e-mail, are there other validations you would want to run?

corps:
That's the major one, I think the others are fine.

corps (approved):

janus/lib/models/cc_agreement.rb
```diff
@@ -0,0 +1,11 @@
+class CcAgreement < Sequel::Model
+
+  def to_hash
+    {
+      user_email: user_email,
+      cc_text: cc_text,
+      agreed: agreed,
+      project_name: project_name
+    }
```
dtm2451, line 9: Per https://www.notion.so/ucsfdatascience/Community-Projects-Spec-92aea94eb8944810b375390eac07411d#1dc62853e2394e1ba0ab6dda01ec89b7, could we add an updated_at or created_at (either... just one needed) to this table so that we can determine the most recent entries when setting permissions?

PR #862, opened 2022-05-26
Polyphemus redcap loading identifier field repeating instrument

coleshaw:
This PR fixes three issues with the REDCap loader that affected loading comorbidities for AutoIPI:

* Boolean fields were previously assumed to have "Yes" / "No" text as the values in REDCap. This is unfortunately not true in all cases, so I've changed the code to be a bit more flexible...basically assuming that "No" translates to `false`, and all else is `true` at the moment. We may probably need to continue tweaking that, depending on projects...
* The combination of `identifier_fields` and repeat instruments did not work at all. This is because the main `identifier_fields` are never populated on a repeat instrument, so the field was always an empty string `""` in the flat records. I have changed this to merge in any `identifier_fields` from the non-repeating instrument into the repeating instrument records, so that it is available when constructing the table records.
* This last issue was never a problem, except for the above when we're trying to munge in data from `flat` records into the `redcap_record` (built from the `eav` record). It seems that REDCap returns numeric values for `flat` records for `redcap_repeat_instance`, but for `eav` records it is numeric for the first instance and a string for all subsequent instances. The result of this was that when calling `flat_record`, you couldn't actually match the `eav` and `flat` records to use them together, because we grouped things by `[ record_id, [ repeating_instance, repeating_instance_id ] ]`, so something like `[ foo, [ form, 2 ] ]` does not match `[ foo, [ form, "2" ] ]`.


coleshaw:
Okay, changed the boolean loader to `Yes / True` and `No / False`, so per our zoom call going to merge this in. Will deploy tomorrow @bsamad , and then we can test it out.

PR #863, opened 2022-05-26
Metis checksum process should not get behind, keep processing

corps:
Metis checksum command only processes up to 10 before quitting.  If this value is too low, the delay of running this process results in it getting behind due to fix processing speed (10 per minute).

With more projects uploading, it becomes more likely to get behind with a fixed processing speed, so the change simply keeps processing in a loop until there are no unprocessed checksums.

### DIFF
--- a/metis/lib/commands.rb
+++ b/metis/lib/commands.rb
@@ -111,9 +111,13 @@ class Metis
     usage "Checksum files."
 
     def execute
-      needs_hash = Metis::DataBlock.where(md5_hash: Metis::DataBlock::TEMP_MATCH, removed: false).order(:updated_at).all[0..10]
-      puts "Found #{needs_hash.count} data blocks to be checksummed."
-      needs_hash.each(&:compute_hash!)
+      while True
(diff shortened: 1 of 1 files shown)


coleshaw:
Hm, I can't check the CI logs for that job, but I think this makes sense, especially since these jobs run in the swarm now, so it's a separate process and we shouldn't be concerned about it affecting the production Metis containers.

coleshaw (approved):

metis/lib/commands.rb
```diff
@@ -111,9 +111,13 @@ class ChecksumFiles < Etna::Command
     usage "Checksum files."
 
     def execute
-      needs_hash = Metis::DataBlock.where(md5_hash: Metis::DataBlock::TEMP_MATCH, removed: false).order(:updated_at).all[0..10]
-      puts "Found #{needs_hash.count} data blocks to be checksummed."
-      needs_hash.each(&:compute_hash!)
+      while True
```
coleshaw, line 114: Oh haha, Python 😛 Should be `true`?

PR #866, opened 2022-06-01
Janus guest auth (community projects)

coleshaw:
This PR updates various auth-related classes to handle a new `guest` role ... specifically:

* Airflow, `etna_user` should be able to parse and understand `guest`. However, they shouldn't be able to edit or view any DAGs in Airflow, since `guests` will have `can_edit = False` as `viewer` does.
* `etna-js` should recognize and unpack the `g` role. It shouldn't impact anything at the moment.
* `Etna::Auth`, `Etna::User`, `Etna::Permission`, and `Janus::Token` should all recognize the `g` role, and be able to parse / handle it.
* `Etna::Auth` should redirect the user to a Janus page `/:project_name/cc` if a token does not have permissions for a project, and it is a community project. The assumption is that if a user has already agreed to the code of conduct, the token should have some permission for that project.

No UI work was done with this (outside of the `etna-js` updates), so would still need a Janus page for handling the code of conduct, as well as handling the list of `guest` users in the admin pages.

### DIFF
--- a/etna/packages/etna-js/components/RootView.jsx
+++ b/etna/packages/etna-js/components/RootView.jsx
@@ -62,7 +62,7 @@ const RootView = () => {
         n1 < n2 ? -1 : n1 > n2 ? 1 : 0
       );
 
-    const resourceProjects = projects.filter((proj) => proj.resource);
+    const resourceProjects = projects.filter((proj) => proj.resource && !proj.requires_agreement);
 
     return {my_projects, resourceProjects};
   });


coleshaw:
Also totally possible I missed some areas where `g` handling is required ... hopefully those will come up as we make more progress, but I fixed where current tests failed, so hopefully we get good coverage through those.

etna/lib/etna/auth.rb
```diff
@@ -141,10 +201,19 @@ def approve_user(request)
       begin
         payload, header = application.sign.jwt_decode(token)
 
+        payload = symbolize_payload_keys(payload)
+
         return false unless janus_approved?(payload, token, request)
+
+        # If the user has not signed the code of conduct for a community
+        #   project, redirect them.
+        cc_required?(payload, token, request)
```
corps, line 210: I'm going to be honest -- I don't like how this logic works.  We're introducing an exception whose job is merely to overload the return value (true, false, do other unrelated thing), but also complicates control flow (is user authorized?  hold on, do some other thing by excepting up to a rescue block way above).

Part of it is that I think our etna auth library is getting bulky and confusing under its own weight and it's worth taking a look at some small refactors / break down to peel parts away more concertedly.

first off, I think that the cc redirect logic should entirely fit within fail or redirect.  Basically, in the original redirect, it should simply be able to determine if there is the possibility of a cc signage helping the user to login.  That should be the nature result of not being authorized by some other means.

corps: ah, taking that back, I think I see why you had to take this approach.

`approve_user` isn't actually responsible for authorizing, ONLY authenticating.

Maybe this logic ought to exist somewhere else (authorization layer, not authentication layer)

corps: Ok -- here's my ask.  See if you can move the cc redirect logic into `route.rb` around the `user_authorized` call.

You can not only check if authorization is failing before executing the redirect, you can capture the project_name sensibly and issue the redirect based on that.

etna/lib/etna/auth.rb
```diff
@@ -64,6 +72,20 @@ def fail_or_redirect(request, msg = 'You are unauthorized')
       return [ 302, { 'Location' => uri.to_s }, [] ]
     end
 
+      # If the application asks for a code of conduct redirect
+      def cc_redirect(request, msg = 'You are unauthorized')
+        return [ 401, { 'Content-Type' => 'text/html' }, [msg] ] unless application.config(:auth_redirect)
```
graft, line 77: Seems like most of this could be merged with fail_or_redirect, possibly using request.env. I suspect you can also avoid using an exception this way (i.e., set request.env['etna.redirect'] and defer to fail_or_redirect as usual).

coleshaw:
Okay, moved the redirect logic into `route.rb`.

PR #870, opened 2022-06-02
Janus guest permission setter function

dtm2451:
Creates a function which adds/removes guest permissions for a user based on their most recent CoC agreements entries per project.

(Bases off #866 so can either merge in there or rebase to master after that goes in)

### DIFF
--- a/janus/lib/models/user.rb
+++ b/janus/lib/models/user.rb
@@ -116,12 +116,8 @@ class User < Sequel::Model
     end
   end
   
-  def fix_guest_permissions!
-    # Get all coc agreements for user
-    # For each project
-    # Pick most recent
-    # if agreed, ensure have a permission, add guest if not
-    # if not agreed, remove permission if role==guest
+  def set_guest_permissions!
+    # Per project, if agreed, add guest access when no access already; remove guest access if !agreed
     CcAgreement.where(user_email: email).all.group_by(&:project_name).each do |project_name, cc_agreements|
       most_recent = cc_agreements.sort_by(&:updated_at).last
       project = Project[project_name: project_name]
--- a/janus/lib/models/user.rb
+++ b/janus/lib/models/user.rb
@@ -115,4 +115,27 @@ class User < Sequel::Model
         permission.project&.project_name == project
     end
   end
+  
+  def fix_guest_permissions!
+    # Get all coc agreements for user
+    # For each project
+    # Pick most recent
+    # if agreed, ensure have a permission, add guest if not
+    # if not agreed, remove permission if role==guest
+    CcAgreement.where(user_email: email).all.group_by(&:project_name).each do |project_name, cc_agreements|
+      most_recent = cc_agreements.sort_by(&:updated_at).last
+      project = Project[project_name: project_name]
+      next unless project
+      perm = Permission.where(user: self, project: project).first
+      if most_recent.agreed
+        if !perm
+          Permission.create(project: project, user: self, role: 'guest')
+        end
+      else
+        if perm && perm.role == 'guest'
+          perm.delete
+        end
+      end
+    end
+  end
 end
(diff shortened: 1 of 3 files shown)


janus/lib/models/user.rb
```diff
@@ -115,4 +115,23 @@ def has_role?(project,*roles)
         permission.project&.project_name == project
     end
   end
+  
+  def set_guest_permissions!
+    # Per project, if agreed, add guest access when no access already; remove guest access if !agreed
+    CcAgreement.where(user_email: email).all.group_by(&:project_name).each do |project_name, cc_agreements|
+      most_recent = cc_agreements.sort_by(&:updated_at).last
```
dtm2451, line 122: One question: Is explicitly checking the 'update_at' necessary?

If we will always create a new entry whenever a user interacts with the CoC-ui, then 1) we could simplify to cc_agreements.last as this would (I think) always be the most recent, and 2) we also wouldn't need an 'updated_at' field as it would be redundant with 'created_at'.

janus/spec/user_spec.rb
```diff
@@ -75,6 +75,79 @@
...
+    user.set_guest_permissions!
+    expect(Permission.count).to eq(1)
+
+    perm = Permission.first
+    expect(perm.project_name).to eq('gateway')
+    expect(perm.user).to eq(user)
+    expect(perm.role).to eq('viewer')
+  end
+  
+  it 'removes guest permission if user most recently did not agree to coc' do
```
coleshaw, line 126: nit: might be nice to also include a test that some higher-level permission is **not** removed when not agreed to the cc. For example, if the user later is granted `viewer` access, after they've already declined the cc?

janus/lib/models/user.rb
corps: Good questions.

In general, I do believe that `created_at` should be redundant with `updated_at`.  This table absolutely should be append only immutable.  That said, there is one plausible way in which `updated_at` might be meaningful: if we ever extend the table and need to provide backfill.  It's, unlikely but not impossible.  In that case, `updated_at` is more of a 'debug' column that helps us potentially understand backfill state.

Ordering by id does benefit from immutability in ways that relying on `updated_at` or `created_at` does not, but to be clear, **there is not a guarantee of monotonic increasing values** from id.  99% of the time the id will increment, but it's important to remember how auto increments are actually, implemented.  They are state that can be manipulated (you can set the autoincrement with SQL commands directly) by bugs, and furthermore, they are by default backed by 32 bit integers which can overflow (in which case the ids will actually return to the lowest unused value in the table, in case of gaps, which could occur due to failed transactions or deletes).

The most likely case of id overflowing is a bug: something accidentally inserts or attempts to insert a huge amount of bad data and forces the id to overflow.  It, does happen.  But, not often.

My intuition is that `created_at`, `id` is the best ordering here, leaving `updated_at` as merely debug information in the case of necessary future backfill.

janus/spec/user_spec.rb
corps: :+1: ^^

dtm2451: sure, will add!

janus/lib/models/user.rb
dtm2451: Swapped to sorting by `created_at`.  Assuming it would resort to `id` next if there ends up being a `created_at` tie, seems like that gets the intended behavior

janus/spec/user_spec.rb
```diff
@@ -76,7 +76,7 @@
     expect(payload['flags']).to eq(user.flags.join(';'))
   end
 
-  it 'adds guest permission if user has agreed to coc, not on project already' do
+  it 'guest-setter adds guest permission if user has agreed to coc, not on project already' do
     user = create(
```
graft, line 80: the usual way to do this is like:
```
context "guest-setter" do
  it "does whatever a guest-setter does" do
  end
end
```

PR #871, opened 2022-06-02
Community projects all apps verification

coleshaw:
This PR attempts to verify that all the applications should treat `guest` users the same as `viewer`:

* Updates tests for Janus, Magma, Metis, Timur, and Vulcan, so that things `viewer` can do, `guest` should be able to. Also checks that if `viewer` is not allowed to do something, `guest` is similarly denied.
* Updates Vulcan so that `guest` users cannot create or update Figures with a `public` tag. 
* Updates the Vulcan UI to disable the "Make public" button in the session manager. Users can try to manually type in "public" into the "Edit tags" text box, but that will throw a server error.


corps (approved):

vulcan/lib/server/models/figure.rb
```diff
@@ -76,5 +82,13 @@ def to_hash(storage: nil)
         updated_at: updated_at.iso8601,
       }
     end
+
+    private
+
+    def self.validate(payload, user, project_name)
+      project_permissions = user.permissions[project_name]
+
+      raise ArgumentError.new('Guests cannot save public figures.') if project_permissions[:role] == :guest && payload[:tags]&.include?('public')
```
dtm2451, line 91: I wonder if guest users would understand what to do when they see this.  Should we explain/recommend copying the figure and removing the public tag in order to successfully save?  (And prioritize soon making a `save as copy` route)

dtm2451 (changes requested):
Vulcan side seems to have most intended behaviors, but I can't seem to use the `copy` button from a guest-perm'd account.  It gives the "Guests cannot save public figures." error message when I try to do so.  Probably need to remove the public tag in order to let that route work.

coleshaw:
:+1: Good catch!! Yes, I'll clear out the tags when you "copy" from within the session, too. Before I think that only happened when you copied from the dashboard, with all the figure cards.

dtm2451:
Hmmmm.  Can copy now :+1:.

New, separate?, bug in testing... if I manually add a public tag from a guest account and then hit `save`, the error catches :heavy_check_mark:, BUT that error in saving seems to lead to the figure to be removed from the database :x:.  If I try to refresh after I get:
image and the copied figure nolonger shows up on the project page.

coleshaw:
Ah, you know, I think this is because of how the figures get saved / archived on the back end ... :thinking: basically, the "source" gets archived first and then the new one gets created, so if there is a problem creating the new one, the source is already archived. Let me change that order around...

dtm2451:
nice!  works fine now =).  And I think that order change sounds right... makes sense to only archive the 'old' version once we know the 'new' version could successfully replace it!

dtm2451 (approved):

coleshaw:
Ah, good catch! Yes, perhaps the "copy figure" button should strip all tags (just like "copy" does in the drop-down menu from the landing page).
 
 
-----Original Message-----

PR #873, opened 2022-06-03
Zc/cc landing page

corps:
Final integration piece of community project first pass.


etna/lib/etna/route.rb
```diff
@@ -124,7 +124,13 @@ def process_call(app, request)
       update_params(request)
 
       unless authorized?(request)
-        return cc_redirect(request) if cc_available?(request)
+        if cc_available?(request)
+          if request.content_type == 'application/json'
+            return [ 403, { 'Content-Type' => 'application/json' }, [ { error: 'You are forbidden from performing this action, but you can visit the project home page and request access.' }.to_json ] ]
```
corps, line 129: api calls should still just 403 here, since there is not a browser context to request guest access with.

janus/lib/client/jsx/cc-view.jsx
```diff
@@ -0,0 +1,116 @@
...
+  useEffect(() => {
+    json_get(`/api/admin/${project_name}/info`)
+      .then(
+        ({project}) => setProject(project),
+      )
+  }, []);
+
+  const classes = useStyles();
+
+  const requiresAgreement = true ? true : project ? project.requires_agreement : false;
```
corps, line 49: These two lines are changed once we have the janus admin ui piece in place, this just allows simpler testing with inline copy of the ipi cc.  Will remove before merging.

etna/lib/etna/route.rb
coleshaw: Ah, that makes sense, since it'll be from an XHR request.

Does this mean that the client libraries in `etna-js` will have to do the right messaging or redirects to the code of conduct page? For example, if someone opens Timur and from the root view clicks into IPI, I think the first time they'll hit an endpoint that triggers this code is when the client app does a fetch from Magma `/retrieve` ... ? So would that have to catch the 403 and then redirect the app? (same problem whether it's the 302 or the 403)

janus/lib/server/controllers/admin_controller.rb
```diff
@@ -157,6 +165,19 @@ def update_cc_agreement
       agreed: !!@params[:agreed]
     )
 
+    @user.set_guest_permissions!
+
+    if project.contact_email and @params[:agreed]
```
coleshaw, line 170: With #874 , the `contact_email` can be `""`, which I think would evaluate to true in this case, but then try to send an e-mail to an empty string address? Should this do a check for `&.empty?` here, to handle that case?

coleshaw (approved):
Looks good! Just the one question about handling the 403, which doesn't affect this PR, but would affect additional integration work that's needed.

etna/lib/etna/route.rb
corps: Not quite -- check timur/server.rb:

```
    with auth: { user: { can_view?: :project_name } } do

...
      # remaining view routes are parsed by the client and must also be set there
      get ':project_name', as: :project do
        erb_view(:client)
      end

      get ':project_name/*view_path', as: :project_view do
        erb_view(:client)
      en
```

So in other words, even the document views will trigger this from the browser.

janus/lib/server/controllers/admin_controller.rb
corps: right right -- I'll check for empty here.

etna/lib/etna/route.rb
coleshaw: Ah, nice, I missed the top of the block when I was scanning :facepalm: 

:+1:

graft: Hmm, Janus server.rb currently has: `get '/*views' do erb_view(:client) end`; the project-specific page is unguarded and wouldn't be caught this way, they would just show up and get a broken/empty page.

corps: Right, so I need to swap out the 'get project' endpoint for 'get projects' on the cc view, that way they can still pull open projects without permission and see the page correctly, making the change.

corps:
Note to self: do not merge until after janus editing lands, so I can remove the fake cc text and test e2e.

PR #874, opened 2022-06-03
code of conduct editing

graft:
Allows editing code of conduct, etc., via the project page. Some issues:

- I am consolidating the two project flags into a single dropdown state
- Setting resource projects has been moved off the "projects" page to the individual project page
- The update_project API didn't allow blanking a contact email, perhaps this was an oversight
- I included a warning if you set your project type to something else


coleshaw:
> The update_project API didn't allow blanking a contact email, perhaps this was an oversight

Good catch!

coleshaw (approved):

janus/lib/client/jsx/project-view.jsx
```diff
@@ -147,11 +192,36 @@ const ProjectView = ({project_name}) => {
   let retrieveProject = useCallback(
     () => {
       json_get(`/api/admin/${project_name}/info`)
-        .then(({project}) => setProject(project))
+      .then(({project}) => {
+        setProject(project);
+        setProjectDetails(project);
```
corps, line 197: Hmm, what's the difference here between these two?

corps: Ah, I see, one of them is the project as it is being edited, and the other is the project since the last load, right?
Nit: Name clarification?

PR #875, opened 2022-06-03
fix bug with alternate ids

coleshaw:
So the issue with the duplicate IDs during REDCap loading was because the alternate ID field wasn't getting wrapped in an array correctly, as the other loader fields do ... I apparently broke the behavior with #862 by changing how `Redcap::Record` constructs these alternate identifier fields from the EAV.

Magma was getting updates like:

```
{"model": {
  "record_1": {
    "name": "r"
  },
  "record_2": {
    "name": "r"
  }
}
```

instead of

```
{"model": {
  "record_1": {
    "name": "record_1"
  },
  "record_2": {
    "name": "record_2"
  }
}
```

This PR fixes that wrapping, and also updates the specs to accurately catch the scenario.


polyphemus/lib/etls/redcap/lib/script.rb
```diff
@@ -212,10 +212,20 @@ def alternate_identifier_fields(record_id)
       non_repeating_key = non_repeating_record_key(record_id)
       
       return {} if non_repeating_key.nil?
+      return {} unless @flat_records[non_repeating_key]
 
       @flat_records[non_repeating_key].first.slice(*identifier_fields)
     end
 
+    def some(data)
```
graft, line 220: curious why this is called `some`

coleshaw: To be honest ... I would love to know if I'm using the term correctly, but basically trying to mirror the terms [used on the client-side](https://github.com/mountetna/monoetna/blob/master/etna/packages/etna-js/selectors/maybe.ts#L12), since that seems to be how the REDCap data structure is also organized. Otherwise I would go with something wordier like, `wrap_values_in_arrays(data)`. @corps , can you provide the background / technical explanation for the client-side naming?

graft (approved):

polyphemus/lib/etls/redcap/lib/script.rb
corps: `some` comes from the Algebraic Constructor `Some` as part of the product type representation of `Optional` used in various languages.  

`Optional<A> = Some A | Nothing`

Since typescript doesn't have algebraic types by default, I use single element arrays instead, like `Optional<A> = [A] | null` and keep the name `some` from its consistent usage in other languages.

In this case I'd just call this method `wrap_X` or even `lift` which is a common function name for wrapping singleton values into a context.

graft: This could also go inside Redcap::Record itself, I think we should put more logic in there in general.

coleshaw: Ah, interesting, okay, thanks for the insights and suggestions! I will see about renaming and if there is a better place to put the logic.

PR #881, opened 2022-06-06
Update guest permissions setter

dtm2451:
Update to:
- scope checks/edits only to current project
- only remove guest access if explicitly disagreed to coc.  (vs before: If a project-admin explicitly gave guest access, so the user received this without agreeing to the coc, let's that guest-perm remain)


corps (approved):

corps:
Thank you for the fix!

PR #884, opened 2022-06-08
Durable hash computation

corps:
In the original implementation, if an error in process occurs after a file is moved but before the record is updated with its new hash, dataloss occurs.  this happened a handful of times since 2020, but fortunately it was possible to manually recover that dead.  This prevents the original issue altogether by always ensuring that there is a data file present at either location (thew new md5 or the temp id location) before making change to the data.


janus/lib/server/controllers/admin_controller.rb
```diff
@@ -173,7 +173,7 @@ def update_cc_agreement
       send_email(
         "Project Lead of #{project_name}",
         project.contact_email,
-        "A new guest user has agreed to the #{project_name} Code of Conduct",
+        "#{project_name} Code of Conduct was signed by #{@user.email}",
```
graft, line 176: Intentional? 😋

metis/lib/models/data_block.rb
```diff
@@ -88,22 +86,38 @@ def compute_hash!
...
-          file_location(md5_hash)
-        )
+      if ::File.symlink?(location) && !temp_hash?
+        original_file = ::File.readlink(location)
+        ::File.rename(original_file, location)
+      end
 
-        # Now update the record
-        update(md5_hash: md5_hash)
+      unless has_data?
```
coleshaw, line 118: Is the effect of this that any 0-Byte data blocks will be automatically removed when the Metis command tries to calculate the md5 hash?

metis/lib/models/data_block.rb
```diff
@@ -74,10 +73,9 @@ def temp_hash?
       md5_hash =~ TEMP_MATCH
     end
 
-    def compute_hash!
-      if has_data? && temp_hash?
+    def compute_hash!(&test_hook)
+      if temp_hash? and has_data?
```
graft, line 77: Any particular reason for `and` here?

metis/lib/models/data_block.rb
graft: has_data? just checks file existence so 0 byte files should be ok.

janus/lib/server/controllers/admin_controller.rb
corps: Yes, because it was the small change we talked over in standup and probably isn't worth it's separate PR.

metis/lib/models/data_block.rb
corps: Nope, programming language fuzziness is all.

PR #886, opened 2022-06-08
Vulcan UI tweaks [close #878]

### LINKED ISSUE #878: Vulcan: swap the logic / button order for "discard local changes"
Seems like having to click "Cancel" to keep your local version of a figure is counter-intuitive, so let's swap OK and Cancel actions, update the wording, so when you click "OK" on the alert, you keep the local changes. Cancel should revert to the server's version of a figure.

coleshaw:
This is a small PR based on feedback from recent standups:

* Changes the text and logic around the dialog window for local edits. Now "OK" keeps your local edits, and "Cancel" reverts to the saved figure. This should be more intuitive for users (#878).
* Changes the formula propagation for the data frame to only require 1 cell below the header cell, instead of 2.


vulcan/lib/client/jsx/components/workflow/workflow_manager.tsx
```diff
@@ -59,20 +59,20 @@ export default function WorkflowManager({
...
+          let useLocal = true;
 
           if (
             localSession &&
             !_.isEqual(localSession.inputs, figureResponse.inputs)
           ) {
-            fromDatabase = confirm(
-              'You have an edited, local version of that figure. Discard it?'
+            useLocal = confirm(
+              'You have an edited, local version of that figure. Click OK to use your local version, Cancel to discard it.'
```
dtm2451, line 69: Could we beef this up a bit?  I think it'd be nice to be explicit that unsaved changes will be discarded if the user cancels.  No getting them back without recreating the work, which reminds me of how video games always try to be explicit when a (re)load might lose unsaved progress.

Maybe:
"Your local version of this figure diverges from the last saved version. Do you want to keep your changes?  Click 'OK' to use your local version, or 'Cancel' to discard all unsaved changes and load the last saved state."

dtm2451 (changes requested):
Tested that formula propagation works and I like the swapped ok/cancel roles on workflow-page refresh!  I'm just concerned that we may want to be even more explicit about what cancel will do.

vulcan/lib/client/jsx/components/workflow/workflow_manager.tsx
graft: Edited for verbosity: "You have unsaved changes. Click 'OK', etc."

dtm2451: So "You have unsaved changes. Click 'OK',  to use your local version, or 'Cancel' to discard all unsaved changes and load the last saved state.", just trimming the beginning of my edit?  :+1:  I like that.

coleshaw: Heh, slightly contradictory with being more explicit? @dtm2451 , does cutting down the first half but leaving the explicit wording in the OK / Cancel statements provide enough detail?

dtm2451: I think so!

PR #887, opened 2022-06-08
Vulcan Viz - allow reordering of discrete groupings with new widget

dtm2451:
Overall goal was allowing reordering of x-axis / y-axis groups of bar_plots.  Unfortunately, I couldn't do that exactly because of the plotly function behind our bar_plots.  So this just implements for y_plot x-axis groups.

Vulcan:
- Adds new ui piece `reorderPiece` which makes use of the same tooling for customized, drag-and-drop, reordering as is used in timur-query column reordering.
- Preps that widget for use use in reordering x-axis, y-axis, and color groupings.
- Implements that piece for the 'y_plot' plot type, for x-axis reordering (with associated updates to its archimedes function)
- Implements that piece for the 'scatter_plot' plot type, for color ordering
- Does NOT implement the piece for 'bar_plot' plot type because the underlying plotly.express.bar function does not support the required 'category_orders' functionality. =(
- Adds 'disabled' control to `SelectAutocompleteInput` (no longer actually touched here, but probably still useful for something in the future :shrug:)

Archimedes:
- Adds and implements `x_order` input in `y_plot` & retools `color_order` in `scatter_plot`
- Adds some extra input typing into the plotting & list manipulation functions
- new list `reverse` convenience function
- adds `decreasing` input to list `order` function.


dtm2451:
Re-tooled so that we can use this ordering system for scatter_plot coloring too!

- Widget nolonger disables when data_target is continuous.  Instead it just removes custom as an option.
- Also added in a safegaurd against page-refresh or data updates; resets to 'increasing' when needed.

coleshaw (approved):

archimedes/archimedes/functions/list/manipulations.py
```diff
@@ -19,6 +19,9 @@ def _flatten(l):
                 yield el
     return list(_flatten(ls))
 
+def reverse(ls: list):
```
coleshaw, line 22: nit: is this type hint wrong? Should it be something like `List[str]`?

dtm2451: For this function, the list can be of anything so I don't think it makes sense to use any content-typing restriction like that. (`list` and `List[<type>]` are both real options for python typing, just the latter lets you additionally hint at typing of the contents.)

coleshaw: Ah, got it! I thought `list` was only the invocation for a new list. Thanks!

corps:
Adding the typing hints is awesome, thank you!

PR #897, opened 2022-06-13
Timur issue 868 [close #868]

### LINKED ISSUE #868: Timur Query UI should allow selecting file md5sums
Magma /query supports this, we should make it visible.

coleshaw:
This PR adds in the ability to get file MD5 hashes in the Query UI and downloaded TSV. It introduces a dropdown for the available predicates, with `url` as the default.

Screenshot 2022-06-13 at 11-59-33 Dev Timur

When selecting `md5`, the data frame preview will be the hash as a string.


dtm2451 (approved):
Works for me!

PR #899, opened 2022-06-14
Vulcan bug fix - stepui drawer unwanted closing

dtm2451:
Simplifies the control of when 'StepUserInputWrapper' are open/closed so that we won't close the drawer unnecessarily.

The fix: Instead of constantly trying to determine whether the drawer should be open/closed, we can just set 'open'ness on mount. The drawer will still auto-close upon step completion because once `Run` is hit, the system will unmount the drawer while the step runs and the remount it afterwards.
-> exact change = determine 'open'ness within the `useState` call and remove the `useEffect` that controlled 'open'ness later on.


dtm2451:
Lovely when fixing a bug just involves removing code =)

coleshaw (approved):
:+1:

PR #903, opened 2022-06-15
single-var bar_plots

dtm2451:
Updates the plotly bar_plotter to ... when x_by==y_by:
- Be able to run the summarization calcs 
- Skip assigning colors per category, and just use one (orange)
- Ignore the user's 'scale_by' request and always use 'counts' (because 'fraction' would always be 1, so doesn't make sense to use!)
image


coleshaw:
I think this works, but small nit:

> Ignore the user's 'scale_by' request and always use 'counts' (because 'fraction' would always be 1, so doesn't make sense to use!)

Does this reflect in the UI, or is there a way to make it reflect in the UI, that this choice doesn't matter? i.e. by removing the box, removing the `fraction` option, etc.? Wondering if it'll be confusing if they accidentally pick `fraction` but nothing changes, or they wonder what the heck that means...

dtm2451:
> I think this works, but small nit:
> 
> > Ignore the user's 'scale_by' request and always use 'counts' (because 'fraction' would always be 1, so doesn't make sense to use!)
> 
> Does this reflect in the UI, or is there a way to make it reflect in the UI, that this choice doesn't matter? i.e. by removing the box, removing the `fraction` option, etc.? Wondering if it'll be confusing if they accidentally pick `fraction` but nothing changes, or they wonder what the heck that means...

Thanks for the push!  Wasn't gonna do anything except maybe a note in the docs, but I can disable the input when x_by==y_by!

dtm2451:
bar_plot-disable

dtm2451:
Title also doesn't need the x_by info added.
... gonna sit with this for a day to make sure I don't notice another!

PR #904, opened 2022-06-15
Zc/upgrade ruby

corps:
1.  Consolidates the gemfile into etna, all projects share single gemfile.  Updating it just involves a `make -C etna bash` + `bundle install`.
2.  Upgrades everythong to ruby 2.7
3.  Includes the streaming change to thumbnails

Still testing around with this, this PR is getting close but in need of manual validation.


timur/lib/commands.rb
```diff
@@ -52,7 +52,7 @@ def execute(version: nil)
 
     def setup(config)
       super
-      Timur.instance.setup_db
+      Timur.instance.setup_db(false)
```
coleshaw, line 55: Does this throw an exception? I thought the [`setup_db` method doesn't take arguments](https://github.com/mountetna/monoetna/blob/master/timur/lib/timur.rb#L7)?

vulcan/lib/commands.rb
```diff
@@ -236,7 +236,7 @@ def execute(version: nil)
 
     def setup(config)
       super
-      Vulcan.instance.setup_db
+      Vulcan.instance.setup_db(false)
```
coleshaw, line 239: I guess same question as the above, for Timur.

metis/bin/metis_client
```diff
@@ -354,8 +350,12 @@ class MetisClient < Client
     raise MetisClient::Error, "Could not upload file #{upload.file_path}\n#{error(response)}" unless response.code == "200"
   end
 
+  def encode_path(path)
+    path.split('/').map { |v| URI.encode_www_form_component(v) }.join('/')
```
graft, line 354: err wups, wrote this same method the other day... Probably should prefer URI.encode_www_form_component to CGI.escape.

timur/lib/commands.rb
graft: Hmm yeah I don't get this either - won't this not load the models?

corps: @coleshaw I added the argument to make this work.

Without it, tests fail to migrate.  It seems with the ruby upgrade the import ordering changes?  But on the migrate command the models should not be loaded, since we're likely creating tables for things before they exist.

Anyways, it is required now or tests fails in CI when importing models that aren't migrated yet.

coleshaw: :+1: I didn't see the `timur.rb` change when I made my comment, makes sense now.

PR #905, opened 2022-06-15
fix metis_client put

graft:
Correctly puts to a different folder (not `.`). I also fixed (?) a deprecation warning about `URI.escape`


coleshaw (approved):
Works well for me :+1:

PR #909, opened 2022-06-16
metis_client update

graft:
Adding an 'update' command and some stale-version-checking which should make it easier for people to stay up-to-date. Not as great as @corps' bundled Ruby version, but hopefully helpful until that is in play.


metis/bin/metis_client
```diff
@@ -1472,13 +1502,25 @@ EOT
     end
   end
 
+  def report_stale_version
+    script = github_contents
+    return unless script
+    old_script = File.read(__FILE__)
+
+    if script != old_script
+      puts "Your version of metis_client is stale. Run `metis_client metis:// update` to update."
```
coleshaw, line 1511: Is this correct?

```
$ ./bin/metis_client update
Your version of metis_client is stale. Run `metis_client metis:// update` to update.
> update
Updating to latest version...
> exit
$ ./bin/metis_client metis:// update
Invalid command. Try:
help	mv	ls	project	cd	cp	rm	mkdir	pwd	lcd	get	put	set	exit	nuke	validate	peek
```

graft: It doesn't work exactly in dev, since you're ahead of master, not behind. But it does successfully replace your dev version with whatever is on master, thus there is no "update" command after updating.

I also wasn't sure how to test this because of this strangeness...

coleshaw: Hm....but even if you're ahead, the file doesn't match and I would expect the command to try updating in the first example

coleshaw: Ooohhh, I see what you mean. Interesting...because I chained them together. Not necessarily because it's ahead (since it's different). Cool!

coleshaw (approved):

PR #910, opened 2022-06-17
Add static plotnine plotters to vulcan

dtm2451:
Exposing this work now that it's getting closer.

This PR creates plotnine-based plotters for the Vulcan Viz pipeline, as alternatives to the current plotly plotters.  It integrates those plotters into the VizUI system.  And it updates the current PlotOutput step_output to navigate between outputing plotly.json via the current framework or a plotnine.png as an .

Still ToDo:
- [ ] See if svg/pdf plot outputs can be used instead of png
- [ ] Add categories to the plot type dropdown
- [ ] bar_plot_static (maybe for a next PR?)
- [ ] Documentation


archimedes/archimedes/functions/plotting/scatter_plot.py
```diff
@@ -88,3 +95,251 @@ def scatter_plotly(
 
     return fig
 
+def scatter_plotnine(
+    data_frame: pd.DataFrame,
```
corps, line 99: This, function looks, kinda awful lol.  So many arguments, this implies really that there is "some structure" `class ScatterPlotNineOptions` which lets you build the figure based on several options.  It'd allow you to break out this big function into smaller utilities and probably do so code sharing, too.

Right now, if I'm honest, this feels like a mess :P

PR #915, opened 2022-06-22
Bug Fix: Vulcan Session Load for the Viz workflow

dtm2451:
Adds:
- projects=null case in 'okay to set this?' check of the vulcan reducer action setWorkflow
- `string[] | null` as typing of `projects?` in Workflow interface to help remind us of that case when writing code.

### DIFF
--- a/vulcan/lib/client/jsx/api_types.ts
+++ b/vulcan/lib/client/jsx/api_types.ts
@@ -102,7 +102,7 @@ export interface Workflow {
   steps: [WorkflowStep[]];
   vignette?: string;
   image?: string;
-  projects?: string[];
+  projects?: string[] | null;
   authors?: string[];
   displayName?: string;
   description?: string;
--- a/vulcan/lib/client/jsx/reducers/vulcan_reducer.ts
+++ b/vulcan/lib/client/jsx/reducers/vulcan_reducer.ts
@@ -82,7 +82,7 @@ export default function VulcanReducer(
       };
     case 'SET_WORKFLOW':
       const workflowProjects = action.workflow.projects;
-      if (!workflowProjects || !workflowProjects.includes(action.projectName))
+      if ( !workflowProjects || (workflowProjects != null && !workflowProjects.includes(action.projectName)) )
         return state;
 
       return {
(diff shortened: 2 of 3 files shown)


vulcan/lib/client/jsx/api_types.ts
```diff
@@ -102,7 +102,7 @@ export interface Workflow {
   steps: [WorkflowStep[]];
   vignette?: string;
   image?: string;
-  projects?: string[];
+  projects?: string[] | null;
```
corps, line 105: :+1: 
Interesting aside, `projects?` implies `projects | undefined`, specifically also that the `projects` key is possibly non existent.

It's a bit of a mind bender, but recall that in javascript there are 3 "empty" states for an object property.
1.  The property exists and is the `undefined` value.  (Generally meant to signal the explicit desire to 'remove' a key)
2.  The property exists and is the `null` value.  (Generally meant to signal the explicit desire to 'keep' a key as an empty value)
3.  The property does not exist, and is `undefined` when accessed.  (Generally meant to signal the explicit lack of knowledge of the key).

:exploding_head:

dtm2451: yes... I stopped short, but I also think we could make this a non-optional property.  Just probably would involve updating a bunch of tests and maybe other real code too.  I do think we'll want workflow maintainers to always give an explicit projects scope.

dtm2451: Thing is, all the other required bits of the Wrokflow type come from the CWL (except `class`. idk what that is.) so I'm not sure if I should remove the `?` or not.

vulcan/lib/client/jsx/reducers/vulcan_reducer.ts
```diff
@@ -82,7 +82,7 @@ export default function VulcanReducer(
       };
     case 'SET_WORKFLOW':
       const workflowProjects = action.workflow.projects;
-      if (!workflowProjects || !workflowProjects.includes(action.projectName))
+      if ( workflowProjects===undefined || (workflowProjects !== null && !workflowProjects.includes(action.projectName)) )
```
dtm2451, line 85: @corps updated this since you reviewed last

corps: :+1:

corps (approved):

PR #917, opened 2022-06-23
Graft/etna package

graft:
Finally sorting this etna-package out. Still a few things todo, but I'd like to get some feedback on where things are at the moment.  The pypi package is called `mountetna` and installs okay (I pushed an initial version). Some issues:

1) I am not pleased with the way I attached things into airflow (using the etna-py Dockerfile), because it is extremely slow to test updates to etna-py in the airflow context (it takes ~30 seconds to rebuild every time you make a change to etna-py). Not sure how to do it better...

2) There are a few remaining issues involving things that airflow likes, for example the `token` variable in TokenAuth is of type `bytes` and must be decoded, which is probably not normally the case.

3) I patched in the zero-byte updates to the Metis upload classes, could use some review @coleshaw .

4) There are some python tests for the etna-package which are not hooked up to CI yet, nor is there a build/deploy action setup yet.


etna/packages/etna-py/src/mountetna/magma.py
```diff
@@ -0,0 +1,320 @@
...
+@dataclasses.dataclass
+class QueryRequest:
+    query: List[typing.Any] = dataclasses.field(default_factory=list)
+    project_name: str = ""
+
+@serialize
+@deserialize
+@dataclasses.dataclass
+class QueryResponse:
+    answer: List[typing.Any] = dataclasses.field(default_factory=list)
```
graft, line 204: I'm not sure how to correctly type-hint a nested list of lists or strings...

dtm2451: perhaps `List[List[str]]`

graft:
Also cutting out magby from archimedes, could use some eyes @dtm2451

etna/packages/etna-py/src/mountetna/magma.py
dtm2451: ah, yes.  Perhaps `List[Union[str, List[str]]` then?  (`from typing import Union`)

airflow/Dockerfile
```diff
@@ -30,8 +31,10 @@ RUN pip3 install -r /opt/airflow/requirements.txt
 
 COPY --chown=airflow:0 ./opt/ /opt/airflow/
 COPY --from=1 /mocker /usr/lib/mocker
+COPY --chown=airflow:0 --from=etna-py /etna-py /etna-py
```
corps, line 34: So this is the trick as far as performance goes, and it's an unfortunate consequence of how python packaging works.

In ruby and node, we work around this by performing the 'install' on just the requirements files first -- so, the `Gemfile` or the `package.json`, THEN copying the actual source over.  This means that as long as the requirements files themselves do not change, docker rebuilds see the install steps as cached and do not re-execute them.

Unfortunately, python installs of packages actually perform a --hard copy-- on install, rather than linking to their source path, meaning that if the entire source code isn't installed at the time of `pip install`, then the source code is not copied into the final virtualenv location.

If we wanted to improve the performance here, essentially what we'd want to do is

```
COPY --chown=airflow:0 --from=etna-py /etna-py/{setup.py,... project skeleton files only} /etna-py
RUN pip install /etna-py
COPY --chown=airflow:0 --from=etna-py /etna-py (actual installed destination of the etna-py library)
```

In that case, the first two steps involved with 'installing' are executed only when the actual skeleton files (usually for new dependencies) change, while the source code is a mere copy on delivery situation.  But that involves some magic similar to the `link-etnapy` and `link-magby` development steps to determine the correct locations.

etna/packages/etna-py/src/mountetna/magma.py
graft: Hmm, apparently something like: `Query = List[Union[ str, 'Query' ] ]` then `query : Query` using https://peps.python.org/pep-0484/#forward-references

graft:
More build badness - if you update etna-py dependencies, its poetry.lock changes - but this also results in the poetry.lock for archimedes becoming out of date, so it must be updated also, which I found required doing `poetry add mountetna` and then restoring the path link in the poetry.lock by hand. I'm somewhat suspicious this also might be pulling from pypi in any case, which is not really great. Long story short, adding dependencies to etna-py is a pain.

corps:
> More build badness - if you update etna-py dependencies, its poetry.lock changes - but this also results in the poetry.lock for archimedes becoming out of date, so it must be updated also, which I found required doing `poetry add mountetna` and then restoring the path link in the poetry.lock by hand. I'm somewhat suspicious this also might be pulling from pypi in any case, which is not really great. Long story short, adding dependencies to etna-py is a pain.

This is part of the reason why I ended up joining all the ruby images under a single parent image that builds the ruby etna gem with a unified lock file.

Yeah, this might be a little rabbit hole, and is part of why I tried to avoid it initially.  There is a solution here that is "nice", but requires some build magickry.  I may have to spend some time on this in the near future to help get this over the finish line.

coleshaw:
> I patched in the zero-byte updates to the Metis upload classes, could use some review @coleshaw .

Yes, I think you captured the required change for zero-byte updates.

PR #918, opened 2022-06-28
Support comma separate attribute group entries

corps:
Support comma separate attribute group entries for @dtm2451 effort to update coprojects template.

work to use attribute groups in the frontend to follow up.

### DIFF
--- a/magma/lib/magma/attribute.rb
+++ b/magma/lib/magma/attribute.rb
@@ -221,7 +221,9 @@ class Magma
     end
 
     def validate_attribute_group_format
-      return if !attribute_group || attribute_group =~ SNAKE_CASE_WORD
+      return unless attribute_group
+      parts = attribute_group.split(",", -1)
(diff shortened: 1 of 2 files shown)


graft (approved):

dtm2451:
Investigating (if proper versions were used, if maybe a format within the csv isn't right, ...), but I still seemed to get the old snake-case error just now when trying to build the template sc_seq model in dev.

dtm2451:
Ahhhhh the error isn't even getting thrown for an attribute with comma-separated-values lol!  'vendor' attribute_group is "SHARED".  I'll try again after making these entries lowercase...

dtm2451 (approved):
Worked once all attribute_groups were lowercase/snake-compliant!

1 note:
- Another missed task comes from the 'parent_modality' and 'associated_modalities' link-collection pair of attributes that we wanted to add.  Requires spec'ing and its own PR, so I think the plan should be to push the current updates without these attributes, so we can unblock current loading work, and add them later

PR #920, opened 2022-06-29
Vulcan image sha cache

coleshaw:
This PR addresses a couple of issues with Vulcan:

* Fixes broken thumbnails by adding in `reference_figure_id` to the `Figure#to_hash` method. This then gets translated into `Session#from_figure`, which will correctly set the reference figure and thus dependencies, used to calculate thumbnail storage locations.
* Adds in a simple cache for checking image existence in Docker Hub, to hopefully improve run performance. Ideally this would expire keys and have size limits, but I think it will be okay to test and see if this helps, since I think archimedes does not change very often, so we shouldn't have many entries in the cache for now.


vulcan/lib/dependency_manager.rb
```diff
@@ -101,9 +105,13 @@ def latest_image(image_name)
     end
 
     def image_exists?(image_name)
+      key_check_result = @existing_images_cache[image_name]
+      return key_check_result unless key_check_result.nil?
+
       # Run a docker command to check that the sha or image exists.
       exists = `docker manifest inspect #{image_name} > /dev/null`
```
corps, line 112: Part of the performance likely also is the required spawning and process management overhead for the operation.  I believe docker hub's api is a simple HTTP interface; can we figure out what URL is required to ping the existance of an image?

corps: Maybe just followed up PR work

coleshaw: Oh, I see, that makes sense. Yeah, I can add that as a follow on task to change from using the docker CLI to pinging the API directly. The nice thing about the CLI would be that it would stay updated (hopefully) regarding any underlying API changes...

corps: That's true, although it's also _sensitive_ to changes in the CLI, including flags and how authentication works, too.  It's also more tricky to consistently test as a result, compared to an HTTP mock.

coleshaw: :+1:

dtm2451 (approved):
Can confirm that thumbnails are reattached.

Vulcan has always run fast in my dev environment, so can't test whether slowness is affected, but I could check that steps still run, :heavy_check_mark:.

vulcan/lib/dependency_manager.rb
corps: That's probably the right call, and it might be contributing to why production is slow.

__not all nodes in production are authenticated__ with dockerhub.  leader nodes do the image pulling and propagation in many cases.  I believe in fact that unauthenticated access IS available but behind rate limiting.

coleshaw: Oooh, interesting. I'll have to try some additional endpoints, because so far everything I do manually returns 401 asking me to authenticate...

corps: that might have actually changed recently now that I recall so I may be wrong lol

coleshaw: The v1 API seems open to unauthenticated access, but I can't quite tell if the manifest information is available through it (image manifest seems like a v2 thing?)...

coleshaw: Might be some way to try pulling a given SHA directly, and seeing what happens... :shrug:

Issue #919, opened 2022-06-29
Etna gem: Using Ruby 2.7 to create projects throws exception


coleshaw:
Using the command `$ etna administrate project create` works fine when using  Ruby 2.5.7 and `etna` version `0.1.44`. However, if you upgrade your local environment to Ruby 2.7.6 to match our containers, the same command throws an exception:

```
$ ruby --version
ruby 2.7.6p219 (2022-04-12 revision c9c2245c0a) [x86_64-linux]
$ etna administrate project create test_proj2 "Test Project2" --environment development
Creating Janus project.
Done! Adding you as an administrator on the project.
Done with setting up the project in Janus!
Creating the project in Magma.
Done! Adding your new project record.
Traceback (most recent call last):
	12: from /home/e/.rbenv/versions/2.7.6/bin/etna:23:in `<main>'
	11: from /home/e/.rbenv/versions/2.7.6/bin/etna:23:in `load'
	10: from /home/e/.rbenv/versions/2.7.6/lib/ruby/gems/2.7.0/gems/etna-0.1.44/bin/etna:18:in `<top (required)>'
	 9: from /home/e/.rbenv/versions/2.7.6/lib/ruby/gems/2.7.0/gems/etna-0.1.44/lib/etna/application.rb:211:in `run_command'
	 8: from /home/e/.rbenv/versions/2.7.6/lib/ruby/gems/2.7.0/gems/etna-0.1.44/lib/commands.rb:140:in `execute'
	 7: from /home/e/.rbenv/versions/2.7.6/lib/ruby/gems/2.7.0/gems/etna-0.1.44/lib/etna/clients/magma/workflows/create_project_workflow.rb:89:in `create!'
	 6: from /home/e/.rbenv/versions/2.7.6/lib/ruby/gems/2.7.0/gems/etna-0.1.44/lib/etna/clients/magma/workflows/create_project_workflow.rb:84:in `setup_magma_project!'
	 5: from /home/e/.rbenv/versions/2.7.6/lib/ruby/gems/2.7.0/gems/etna-0.1.44/lib/etna/clients/magma/workflows/create_project_workflow.rb:62:in `create_magma_project_record!'
	 4: from /home/e/.rbenv/versions/2.7.6/lib/ruby/gems/2.7.0/gems/etna-0.1.44/lib/etna/clients/magma/client.rb:80:in `update'
[...]

PR #921, opened 2022-06-29
use update_json when creating Magma project record

coleshaw:
This PR updates the `etna` gem to `0.1.45`. It changes the method called when creating a Magma project record in the `create_project_workflow`, which seems to address the issues seen in #919.

To test it, you'll need to build the `etna` gem locally and install it.

```
$ make -C etna bash
-> gem build etna
-> exit
$ cd etna/
$ gem uninstall etna (type y to confirm, when prompted)
$ gem install etna
$ gem list | grep etna (should show etna  0.1.45)
$ etna administrate project create <test_project> "Test project" --environment development
```

For me this works on 2.7.6 now, even if I run it multiple times against the same arguments:

```
$ ruby --version
ruby 2.7.6p219 (2022-04-12 revision c9c2245c0a) [x86_64-linux]
$ etna administrate project create test_proj6 "Test Project6" --environment development
Creating Janus project.
Done! Adding you as an administrator on the project.
Done with setting up the project in Janus!
Creating the project in Magma.
Done! Adding your new project record.
Add more users? Y/n
n
All complete!
You need to visit Janus to refresh your token.
You can now log into any app to manage your data.
$ etna administrate project create test_proj6 "Test Project6" --environment development
Creating Janus project.
Done! Adding you as an administrator on the project.
Done with setting up the project in Janus!
Creating the project in Magma.
Done! Adding your new project record.
Add more users? Y/n
n
All complete!
You need to visit Janus to refresh your token.
You can now log into any app to manage your data.
$ gem list | grep etna
etna (0.1.45)
```

The `magma_client.update` method parses everything into a multipart/form-data structure, which causes the nested `revisions` hash to flatten into a set of array-inputs. It looks like the `multipart-post` library [changed six months ago](https://github.com/socketry/multipart-post/blame/e52ef1e11b1f1edfc3ae16eff24a71a53745c967/lib/multipart/post/multipartable.rb) to use the `transform_keys` method, which assumes a hash is passed in. Previously it used `map` on the params, which would have worked for our Array-ish inputs, I think.

`magma_client.update_json` sends the request as just a JSON structure, which is all we need in this case. We only require `magma_client.update` when sending large updates, I think...


coleshaw:
And I think the reason it worked for me on 2.5.7 is because I have an older version of `multipart-post` installed:

```
$ gem list | grep multipart
multipart-post (2.1.1)
```

If you freshly installed things in your local 2.5.7, you would have gotten the latest `multipart-post` (2.2.3), which would have just led to the same bug / issue.

coleshaw:
I've pinned the dependency to the last version of `multipart-post` that works with our current code. Scanning the code, I'm not sure that the latest versions would be compatible with our version in `multipart_serializable_nested_hash.rb`, which I think we would want to verify is a solid replacement before using newer versions.

corps:
Excellent catch!

corps (approved):

dtm2451 (approved):
Works from my system that had had a too-recent multipart-post beforehand.
```
dan@pop-os:~/GitHub/monoetna$ gem list | grep etna
etna (0.1.45)
dan@pop-os:~/GitHub/monoetna$ etna administrate project create test_project "Test project" --environment development
Creating Janus project.
Done! Adding you as an administrator on the project.
Done with setting up the project in Janus!
Creating the project in Magma.
Done! Adding your new project record.
Add more users? Y/n
n
All complete!
You need to visit Janus to refresh your token.
You can now log into any app to manage your data.
dan@pop-os:~/GitHub/monoetna$
```

PR #925, opened 2022-06-29
Map page includes autocompletion, handling of multiple attribute groups

corps:


magma/lib/magma.rb
```diff
@@ -150,7 +150,9 @@ def server_pid
     else
       # Oh boy.  Not ideal, but best effort here.  This could end up just restarting the wrong puma process if
       # it was hosted together.  Ideally we're running these processes in separate containers so it should be ok.
-      `pidof puma | tail -n 1`.chomp.to_i
```
corps, line 153: So, firstly, the pidfile approach isn't reliable -- I'm not actually sure that the pidfile is working in containerization altogether, so this branch is the one that gets hit.
Previously, `pidof puma` fails because the actual process is `ruby`, running the `puma` script.
Fortunately `pgrep` is much better and finds the process.  Better yet, we now return nil (rather than 0) when we can't find the process, which means we'll catch these problems in the future due to the kill command raising an exception in that case.

The result of the previous command was `"".to_i` => `0` => kills the entire process group, forcing swarm or docker compose to restart the entire container as opposed to merely, reboot the puma process, which is orders of magnitude faster.

magma/lib/magma/actions/model_update_actions.rb
```diff
@@ -128,7 +128,7 @@ def update_model_tables
     end
 
     def restart_server
-      return if Magma.instance.test?
+      return if Magma.instance.test? || Magma.instance.server_pid.nil?
```
graft, line 131: Is it safe to simply return here if the server_pid is nil? We are still responsible for ensuring a restart even if we were unable to determine the pid...

corps: The use case I was targeting is the magma local client.  Perhaps I could signal by some other means here?

timur/lib/client/jsx/components/model_map/model_report.jsx
```diff
@@ -281,30 +313,24 @@ const ModelReport = ({ model_name, updateCounts, counts, template, setAttribute
...
-      InputProps={{
-        startAdornment: (
-          <InputAdornment position='start'>
-            <SearchIcon />
-          </InputAdornment>
-        )
-      }}
-      onChange={(e) => setFilterString(e.target.value)}
-    />
+    <AutoComplete options={acOptions}
```
graft, line 316: Hmm, not so sure about using an autocomplete here. I am seeing only dropdown options for `type:<types>`, but that is already pretty verbose and doesn't yet include like `attribute:<each of the attributes>`, which maybe leaves you wanting a way to filter the list of autocomplete options...

dtm2451: I kinda really like the idea of giving those options in the dropdown because, tbh, I would not have know they were options otherwise.  (I'd never read that placeholder text until now.)  I like the set that was picked to show up too, as the attribute and description columns work directly with text typed into the box.

Perhaps some different placeholder text could be used to signal that you can 'type here to filter attributes by name or description'. The user then finds out there are other options to when they have need to search.  Could possibly give a version of the current placeholder details via 'helperText' below.

Or second alternate, give up on the dropdown, but instead make searching across type and group work without adding 'type:' and 'group:'?  Roll them right in with how attribute_name and description fields are searched.

dtm2451: Ohhh!  Hmmmm, yet another idea, but I think it's the best so far...
Current behavior of the Autocomplete is that clicking a dropdown item adds to the search bar yet leaves the typing cursor in place for the user to keep typing. So we could 1) give `attribute:` `type:` `group:` and `description:` as the dropdown options 2) turn off filtering of options completely so that all the options remain up no matter what the user types, and 3) maybe use the categories ability of mui Autocomplete to label the dropdown options as "optional syntax" or something like that.  And that's it.  The dropdown would highlight the optional helpful syntax for to corroborate the current placeholder text, yet the whole thing would remain slim visually, unified (all columns still in one widget), quick to use, and quick to code.

graft: Do you mean like, the dropdown would show "type", "group", etc., but clicking "type" in the dropdown would just add "type:" or " type:" to the end of the filter? I like that...

dtm2451: Yes!  (I was thinking it would show the `:`s in the dropdown too, but otherwise, yes.)

PR #926, opened 2022-06-30
handle healthcheck requests properly when redirecting

coleshaw:
Adds a check against a "localhost" request when handling redirects.


coleshaw:
Going to handle this a bit differently and update the healthchecks accordingly...

PR #927, opened 2022-06-30
Etna server noauth options route

coleshaw:
This PR provides a `noauth` endpoint (OPTIONS `/`) for each of the application servers.

It seems like the healthchecks might be noisy in Docker Swarm, because the `/` route has a bit of a conflict with the healthchecks, due to the deployment of `redirect.rb`.

We constantly get this message:

```
2022-06-30 13:10:30 +0000 Rack app ("GET /" - (127.0.0.1)): #<NoMethodError: undefined method `join' for nil:NilClass>
```

While the 500 seems to report as "okay" health (portainer still shows containers as up and healthy), seems like we should provide a working endpoint to get more accurate behavior.

We traced the above exception back to the healthcheck [request against `localhost`](https://github.com/mountetna/monoetna/blob/master/swarm/janus.1655506506.yml#L55) throwing [an exception in the redirect domain checking](https://github.com/mountetna/monoetna/blob/master/etna/lib/etna/redirect.rb#L32-L34). Since `request.hostname == "localhost"`, and `"localhost"` has no `.` in it, the code tries to join `nil` and fails.

It doesn't really make sense to allow a localhost redirect in `redirect.rb`, so here we're providing an alternate `noauth` endpoint that the healthcheck can run against. 

Once this is deployed, I'll edit the Portainer stack configs to point the healthchecks to the new routes, by changing:

```
bash -c 'curl http://localhost:3000 || exit 1'
```

to 

```
bash -c 'curl -X OPTIONS http://localhost:3000 || exit 1'
```

for each of the stacks.

### DIFF
--- a/etna/lib/etna/server.rb
+++ b/etna/lib/etna/server.rb
@@ -3,7 +3,19 @@ module Etna
   class Server
     class << self
       def route(method, path, options={}, &block)
-        @routes ||= []
+        # For healthchecks, set up servers
+        #   with an OPTIONS route on /, with noauth
+        @routes ||= [
+          Etna::Route.new(
+            'OPTIONS',
+            '/',
+            {
+              auth: {
+                noauth: true
+              }
+            }
+          )
+        ]
 
         @routes << Etna::Route.new(
           method,
--- a/etna/spec/server_spec.rb
+++ b/etna/spec/server_spec.rb
@@ -167,13 +167,13 @@ describe Etna::Server do
   it 'sets route names' do
     Arachne::Server.get('/silk/:name', as: :silk_road)
 
-    expect(Arachne::Server.routes.first.name).to eq(:silk_road)
+    expect(Arachne::Server.routes.last.name).to eq(:silk_road)
   end
 
   it 'guesses route names' do
     Arachne::Server.get('/web/:silk', action: 'web#silk')
 
-    expect(Arachne::Server.routes.first.name).to eq(:web_silk)
+    expect(Arachne::Server.routes.last.name).to eq(:web_silk)
   end
 
   it 'looks up route names' do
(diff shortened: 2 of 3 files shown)


etna/lib/etna/server.rb
```diff
@@ -3,7 +3,19 @@ module Etna
...
+          Etna::Route.new(
+            'OPTIONS',
+            '/',
+            {
+              auth: {
+                noauth: true
+              }
+            }
+          )
+        ]
```
graft, line 18: Hmm, won't this conflict with the route introduced by Etna::DescribeRoutes? That layer should trap the request even before Etna::Server sees it. Also we can't disable Etna::DescribeRoutes since it is still used in a bunch of clients unfortunately.

etna/spec/server_spec.rb
```diff
@@ -20,6 +20,22 @@ def silk
...
+    end
+
+    expect(Arachne::Server.routes.length).to eq(2)
+
+    @app = setup_app(Arachne::Server, [Etna::DescribeRoutes])
+
+    options '/'
+
+    expect(last_response.status).to eq(200)
+    expect(last_response.body =~ /OPTIONS/).not_to eq(nil)
```
graft, line 35: you can do like `expect(last_response.body).to match(/OPTIONS/)`

etna/lib/etna/server.rb
coleshaw: It actually does not conflict, and the routes_json is returned. I think you need it as an actual `Etna::Route` on the server, in order for `Etna::Auth` to recognize the `noauth` option. Note that `Etna::DescribeRoutes` does not introduce an `Etna::Route` to the server, it just hijacks the request, so I don't think you can circumvent `Etna::Auth` via `Etna::DescribeRoutes`...

coleshaw: So the `Etna::Route` as defined here does not capture / return any data, as you point out, it just allows `Etna::Auth` to do its thing...

coleshaw: Perhaps "less testable" isn't quite the right way to word it, but then it's just spread out across all the apps...

PR #928, opened 2022-06-30
update timur create_routes

graft:
This repairs an issue in Timur's CreateRoutes command, which creates a JS file containing functions that the Timur client uses for mapping routes. Following the 2.7 upgrade, this (untested) command began to fail, causing the main page for Timur to make a request for magma for the entire `project` record. This, for reasons yet unknown, causes Magma to OOM and kill its container.

This solves a part of the problem by fixing the Timur command, so there is a proper routes.js, so that a view can correctly be retrieved, so that Timur won't ask magma for too much data by default.

It is possible there are other underlying issues having to do with the change to using `URI.encode_www_form_component` and `URI.decode_www_form_component` instead of `URI.encode` and `URI.decode` (for example, outstanding Metis and Vulcan slowness); these are undiscovered so not dealt with here.

### DIFF
--- a/etna/lib/etna/route.rb
+++ b/etna/lib/etna/route.rb
@@ -38,11 +38,11 @@ module Etna
 
     UNSAFE=/[^\-_.!~*'()a-zA-Z\d;\/?:@&=+$,]/
 
-    def self.path(route, params=nil)
+    def self.path(route, params=nil, &block)
       if params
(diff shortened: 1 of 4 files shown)


coleshaw (approved):
The plots and views routes seem to work correctly for me!

Issue #907, opened 2022-06-16
Airflow ipi integral_data raw fastq's


coleshaw:
Sometimes it seems like the ETL only links 1 file to a file_collection attribute, when there are two files in the folder. I've seen this happen with multiple samples, it seems inconsistent though... perhaps something to do with paging the updates, or the updates are not grouped correctly?

![Screenshot from 2022-06-16 11-40-10](https://user-images.githubusercontent.com/4930129/174109706-0c58cee9-f247-4dca-847b-c46d433a58e5.png)
![Screenshot from 2022-06-16 11-40-13](https://user-images.githubusercontent.com/4930129/174109715-46313976-4896-4d5d-9e91-63e44143988e.png)
![Screenshot from 2022-06-16 11-40-29](https://user-images.githubusercontent.com/4930129/174109721-6b953d8b-0ff8-4ea5-8b32-ec20db652242.png)

coleshaw:
Hm, so regardless of the root cause of this issue (not the current ETL code, which as @corps pointed out, searches by folder and batches all files related to a single folder in a single update, so won't have the truncation / pagination issue that showed a similar symptom), I do think that the IPI integral_data ETL is a bit weird, process-wise, compared to the other ETLs.

All the other ETLs that process data from Metis assume that folders + files are uploaded and then "immutable", in the sense that when a folder is updated, you can take the contents of that folder and link them or process them. This works well for single-cell data, which is always uploaded from a pipeline.

However, IPI integral_data is a different use case, since the folder structure is sometimes made much earlier. I messed up the timestamps yesterday, but many of the folders were created back in Dec 2020, when Beth and I first set up integral_data. Sample folders don't get touched or re-created when new data comes in -- just the relevant subfolders or assay folders. Thus, when the ETL goes through the structure looking hierarchically for experiments / patients / samples / etc., that have been updated, I think it doesn't necessarily catch updated assay files or folders -- because those are filtered out at the top level as not meeting the "updated since the last ETL run" filter, and thus don't trickle down? Perhaps this is why some records still showed this old 1-file problem, instead of being updated to the latest data. Also might explain why nuking all ETL runs and re-running from scratch fixed the linking. :shrug:

coleshaw:
I feel like you'd almost want to just track new files in integral_data, and then back up to the folder and figure out what level you're at (patient, sample, assay, etc.) and then grab all files in that folder to relink.

coleshaw:
But that seems very convoluted...

[...]

PR #948, opened 2022-07-13
Airflow metis linking files

coleshaw:
This PR makes a small change to how files within a folder are grouped for Airflow ETLs, to better handle `file_collection` attributes. It uses a manual "reduce" instead of `itertools.groupby`, since Metis `/tail` does not guarantee that the result is sorted by `folder_id`. I did not just call `sort` on the tail results to avoid performance issues with sorting potentially large payloads.

[`itertools.groupby` assumes that the input is already ordered by the desired key](https://docs.python.org/3/library/itertools.html#itertools.groupby). If it is not, then you will miss elements due to the internal behavior of the method. Minimal code example:

```
Python 3.9.7 (default, Sep 10 2021, 14:59:43) 
[GCC 11.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import itertools
>>> results = [{"folder_id": 1, "name": "file_01"}, {"folder_id": 4, "name": "file_02"}, {"folder_id": 1, "name": "file_03"}]
>>> key_func = lambda f: f['folder_id']
>>> {k: list(v) for k, v in itertools.groupby(results, key=lambda f: f["folder_id"])}
{1: [{'folder_id': 1, 'name': 'file_03'}], 4: [{'folder_id': 4, 'name': 'file_02'}]}
>>> results.sort(key=key_func)
>>> {k: list(v) for k, v in itertools.groupby(results, key=lambda f: f["folder_id"])}
{1: [{'folder_id': 1, 'name': 'file_01'}, {'folder_id': 1, 'name': 'file_03'}], 4: [{'folder_id': 4, 'name': 'file_02'}]}
```

Note that the unsorted input results in one element missing for `folder_id = 1`.

Metis `/tail` sends back results ordered by `updated_at` (I think), so does not guarantee any sort of grouping by `folder_id`. I think this behavior was not observed for IPI and MVIR1, due to the way the source buckets were organized. IPI was reading bulk RNA Seq data out of `integral_data`, which always had the fastq's for a tube_name linked at the same time, so they would basically have the same `updated_at`. Similarly, with MVIR1, our waiver-bucket ETL constantly re-organizes the Metis data by patient / tube-name, so fastq files are naturally grouped together in time.

For AutoIPI, though, the updated_at timestamps are scattered and most likely due to when the files finished uploading and were subsequently hashed. There is no "filtering" process that would naturally group related files together, like for our other projects, and thus we observed some records linking 1 fastq while others would correctly link 2.

Related to the original / first issue identified in #907, but does nothing to address the question of "how do you handle when files are updated but the folder is not" (which I think @corps  is handling separately).


airflow/opt/providers/etna/etna/tests/cassettes/test_metis_files_etl/test_metis_files_etl_e2e.yaml
```diff
@@ -15504,24 +15504,24 @@ interactions:
...
 
-        {"type":"file","id":58307,"parent_id":51335,"node_name":"MVIR1-HS15-D0PBMC1-RSQ1_S3_L001_R2_001.fastq.gz","updated_at":"2020-09-10
-        15:09:50 +0000","created_at":"2020-09-10 15:09:50 +0000","file_hash":"a98fa07eb6246b1f4c31579a7d32bf5c","archive_id":"im_uCsoNSNplfWadC-Ulyka_OGlGIH451ZJM32dqqC1b_44h3VLHYVScz1_nGkqhuT3m9-cYiqnv16RON0Y1Nevr0C8zPh2MXeSf80KH7Tgoqi2vWjKECr54_224EuNwYwxjthRPsA"}
-
         {"type":"file","id":58308,"parent_id":51336,"node_name":"MVIR1-HS15-D0PBMC1-RSQ3_S37_L002_R1_001.fastq.gz","updated_at":"2020-09-10
         15:18:15 +0000","created_at":"2020-09-10 15:18:15 +0000","file_hash":"51fe26a68df4047bea5b8c01e7f319f1","archive_id":"2IRLXTFo727MpqYAEK8SVVYREWs7oo-ok66cTG0VIEy5KbgsIbqkkVfgws-BOk9Qj5qULB9k6KPvXTJXCFxO44bCbBsa0zOcTuKzNDCohhjW6gPB6PlW_iw-rzjSlXgoNWP9Td3MCw"}
 
+        {"type":"file","id":58307,"parent_id":51335,"node_name":"MVIR1-HS15-D0PBMC1-RSQ1_S3_L001_R2_001.fastq.gz","updated_at":"2020-09-10
```
coleshaw, line 15510: Sorry, I got lazy and didn't want to re-record the entire cassette....so just manually edited a bit to simulate the test condition we see that causes the issue...

corps: That's fair.

airflow/opt/providers/etna/etna/etls/metis.py
```diff
@@ -580,7 +575,14 @@ def list_match_folders(matches):
             project_name = matches[0].project_name
             bucket_name = matches[0].bucket_name
             with self.hook.metis() as metis:
-                files_by_parent_id = { k: list(v) for k, v in itertools.groupby(metis.tail(project_name, bucket_name, 'files', folder_id=[m.folder_id for m in matches])[0], key=lambda file: file.folder_id) }
+                files_by_parent_id = {}
```
coleshaw, line 578: Did not just add a `metis.tail().sort()` in case the payload is large, since sorting could be expensive and not strictly necessary.

corps: To be honest, in place sorting is not that expensive, it's not really clear that it would be more expensive than what you ended up doing.  (sort, other than the comparison part, is implemented natively in C, so it runs faster than the python loop).  That said, it's really not a huge deal either way.  This may just be safer implementation.

coleshaw: Ah, okay, thanks, that's good to know! I wasn't sure if that would be an issue, so used the explicit / wordy way :smile:

PR #951, opened 2022-07-14
Zc/docker operator options improvement

corps:
3 Changes here

1.  Swarm Operator now, by default, adds constraint to not run on manager nodes.  It also includes a 0.5 cpu maximum that is configurable.  Incidentally, this also means that the base service for any job does not copy its own resource reservations.

2.  The materialize metis data workflow operates abstractly on file systems, but when copying from metis -> metis, the copy files bulk command is far more efficient than downloading and re-uploading.  I believe this may have contributed to some of the metis overhead we were seeing (the materialize mvir1 job for genentech)

3.  Include parent directories in the tail logic for metis etls -- if a file is modified, each of its parent directories are also given as updated directories, triggering all parent record folder finds in the process.


corps:
I need to re-record some cassettes to get this passing, hold on.

coleshaw (approved):
Nice improvements! I agree, I think these will improve swarm resource management. Good catch on copy versus download / upload!

PR #953, opened 2022-07-15
refactor and extract test

coleshaw:
Small refactor to improve testing around the metis file grouping logic, so it's not reliant on the cassette and thus won't get clobbered on re-recording ... :man_facepalming:


corps:
that makes total sense.

TBH the cassette tests are very useful early in the project because so much was changing that higher level tests helped ensure that integration wasn't breaking, but over time I think moving towards effective mock / unit tests makes more sense.

coleshaw:
Mostly I see this as karma for my hand-editing a cassette... :nose:

Issue #939, opened 2022-07-07
Path to always use latest images in vulcan


dtm2451:
For development purposes, it would be incredibly helpful to be able to simply refresh the page and re-`Run`, after editing scripts and rebuilding vulcan/archimedes.  Especially considering that the 'Update dependencies' option does not work in dev currently because it targets `:production` images.  Currently, I need to `create figure` anew each time.

coleshaw:
Hm, update-dependencies should work if you change your config.yml to just run your local archimedes, instead of production? i.e.

```
:archimedes_run_image: archimedes
```

But, I think this is a great idea! Should definitely be doable -- do you imagine this would be a different button that shows up in development, i.e. run-on-latest?

dtm2451:
I'd need to update the lines below too?

From
```
  :archimedes_run_image: etnaagent/archimedes
  :archimedes_interpreters:
    :r: etnaagent/archimedes-r:latest
    :node: etnaagent/archimedes-node:latest
    :python: etnaagent/archimedes:latest
```

To
```
  :archimedes_run_image: archimedes
  :archimedes_interpreters:
    :r: archimedes-r:latest
    :node: archimedes-node:latest
    :python: archimedes:latest
```

or just the one line?

coleshaw:
Yes, I think all lines like that would let the "Update dependencies" button to capture the local dev versions accurately. Might try `archimedes:latest` on the `archimedes_run_image` one too, to make sure you capture the latest one you build.

dtm2451:
> do you imagine this would be a different button that shows up in development, i.e. run-on-latest?
Yea, as a button could be great!  DEf some control mechanism needed... I tend to want this behavior as the default, but it would need to be able to be turned off for testing of the image-pinning system.

PR #956, opened 2022-07-19
Vulcan dev dependency bypass

coleshaw:
@dtm2451 , I think this works for #939 ? This PR implements an `ignore_dependencies` flag on the backend and adds a checkbox for it in the Vulcan UI. If you check the box and then click `run`, Vulcan will use the latest image for `archimedes` and all the interpreters. Default is not checked. The flag only appears if you have the `vulcandev` flag set on your token, so you'll want to make sure to add that. The idea is this would only be a local dev thing ...


coleshaw:
I think one way to test this locally would be:

* Give yourself the `vulcandev` flag in Janus.
* Create some obvious changes in `archmedes`. I've been doing this by editing `archimedes/archimedes/functions/dataflow/__init__.py` into some breaking changes, like

```
from .path import input_path, output_path, output_tsv
import json
```

* Build a local version of this `archimedes` with `IMAGES_POSTFIX=:production make -C archimedes/ release` (this has to match whatever you use in `vulcan/config.yml`
* Run a saved figure in Vulcan. It should work with the saved `archimedes` SHA.
* Check the "Ignore dependencies" box.
* Run again. It should fail or show the behavior of your edited `archimedes` (whatever you did to it).
* Uncheck the "Ignore dependencies" box.
* Run again. The behavior should revert to the saved SHA behavior.

dtm2451 (changes requested):
Testing with this in vulcan/config.yml
```
:archimedes_run_image: archimedes
  :archimedes_interpreters:
    :node: archimedes-node
    :python: archimedes
```
and just `make -C vulcan bash` for building updated images...

This works to run off the latest images instead of the ones tagged to the figure.  :heavy_check_mark: 

But it's not achieving the fully desired effect of letting me say: test changes to just one step at the end of a workflow. Upon clicking the ignore deps checkbox for the first time, all past steps have their status reset to pending and the session starts from the beginning.

coleshaw:
Oh yes, because right now the flag `ignore_dependencies` is used in the cell hash calculation.

> test changes to just one step at the end of a workflow

Hm...this seems trickier. Like, if you want to ignore dependencies to just one step at the end, how do you know that the previous steps wouldn't change the downstream results? Wouldn't that make the cache unreliable? Like, if you can have two different archimedes writing to the same output file (cell hash) ... how do you know how that data was calculated?

coleshaw:
Would it make sense to run the entire workflow with the box checked, instead of toggling it only at the desired step?

dtm2451:
> how do you know that the previous steps wouldn't change the downstream results? Wouldn't that make the cache unreliable?

Yes it would, but compared to having the ease of testing new code more immediately, instead of needing to wait 20+min for the scRNAseq pre-processing to re-run, I don't particularly care if my dev cache is imperfectly reliable.

dtm2451:
I suspect that'll be the case for outside editors as well if/when we build to a 'productionized' vulcan editting environment.  Dev-cache unreliability seems a caveat to me more than a problem.

coleshaw:
> I suspect that'll be the case for outside editors as well if/when we build to a 'productionized' vulcan editting environment. Dev-cache unreliability seems a caveat to me more than a problem.

Hm....but if the productionized Vulcan editing environment touches the production cache, wouldn't then the production cache be unreliable, too, which seems bad? (understanding that we haven't really defined what productionized Vulcan editing means...)

dtm2451:
Perhaps it could be read-only access to production cache, with ability to only write to its own cache

coleshaw:
Notes:

- [ ] Make it so you can only ignore dependencies in `development` (like the Vulcan archimedes token config)
- [ ] Remove `ignore_dependencies` from the cache SHA calculation

coleshaw:
- [ ] Also need to run latest script instead of stored script when the box is checked.

coleshaw:
So in general, we'll build a system-wide dev config where you can disable dependency management (gets back old behavior), but leave the system in place in production.

coleshaw:
I think the changes we talked about work now, where the flag in `config.yml` can disabled the dependencies for dev environments.

dtm2451:
How would I set up my config.yml to achieve this?  I can't tell what the flag is that I'd want to set up.

coleshaw:
Sorry, forgot there isn't a nice UI now...

In `config.yml`, under `development` you would add `ignore_dependencies: true`, like:

```
:development:
  :ignore_dependencies: true
```

Leaving it out or making it `false` would keep the same production behavior.

dtm2451:
It doesn't seem to work?  I commented out the dataflow input/output functions in that __init__.py, cuz that's a wonderful suggestion of a catch-nearly-all to test!, rebuilt archimedes with `make -B -C archimedes bash`, then restarted everything with `make restart`.  I have this at the end of the :development: section in vulcan/congfig.yml:

```
  :db:
    <<: *default
    :database: vulcan_development
  :advisory_lock_file: /app/storage.lock
  :ignore_dependencies: true
```

And with all that set up and rebuilt properly, I think, my scRNAseq.cwl run is still plugging along and inputting/outputting data as normal.

dtm2451:
I'd forgotten to change the :production tags for the way I run... restarting vulcan after doing that and it worked.  :+1:

So many moving parts!

coleshaw:
Interesting! I was trying to replicate the issue and couldn't quite figure out how the commands you use work together to make the images, etc., (I don't use the `make -B` flag, so unfamiliar with the details of how it behaves and builds images). 

Glad you figured out the puzzle, hehe, it's definitely confusing.

dtm2451:
:heavy_check_mark: also catches / refreshes step statuses if I update the script of an early step and then commit an input change.  And if I hit Run, it hits the error message that I put into that script, as expected:
image

dtm2451:
The -B option came from Zach and his explanation was that `-B` tells the system to rebuild the involved docker images.  I don't see it in the docker compose docs anywhere so I can't really defend or explain that other than to say that it has seemed to work that way when I've used it lol

dtm2451 (approved):
Seems to work and hits the major goals of creating a dev build setup to allow rapid testing of both 1- archimedes function & 2- workflow script updates!

PR #958, opened 2022-07-20
Production Deploy

dtm2451:


dtm2451:
@coleshaw requested review just to get your eyes on the one airflow commit by Zach that I think you might have the context for?

coleshaw (approved):
:+1: I think the swarm config changes are fine, I assume those are most likely in production already, anyways... (the swarm workflow is a bit different).

dtm2451:
gotcha!  Prefect, suspected that it was already through

PR #959, opened 2022-07-21
fix regex for single cell linker

coleshaw:
Reviewing the AIP1 single-cell data with Emily today, and we realized that some of the linker regexes need to be fixed. This PR fixes a couple, and then we'll have to re-run the current linkers.


coleshaw:
I've made the regex changes manually to the IPI linkers, since the regexes there are defined in the ETL, but this would affect others that use the single-cell helpers (like aip1 and mvir1)

dtm2451 (approved):
Both these changes seem safe!

- \d+_umap == the standard syntax from the pipeline
- possorted... == what this attribute is actually meant to get.

good catches =)

coleshaw:
Thanks! Just a comment about:

>\d+_umap == the standard syntax from the pipeline

Emily pointed out that there are multiple UMAPs generated, but the one we should probably show is the one with name like `IDENTIFIER-CODE1_umap`, instead of the others, which are like `IDENTIFIER-CODE1_some_description_umap`...

dtm2451:
> Emily pointed out that there are multiple UMAPs generated, but the one we should probably show is the one with name like `IDENTIFIER-CODE1_umap`, instead of the others, which are like `IDENTIFIER-CODE1_some_description_umap`...

Looking at the umap outputs, that is the one I'd expect when coming to that attribute, so I agree with Emily.

PR #960, opened 2022-07-21
Airflow cat connection refactor

coleshaw:
This PR tweaks the Airflow CAT ingestion to use different mechanisms to stream the files. It turns out that `paramiko` prefetches the entire file at once (and seems to block while doing so), even if you use `getfo` and `putfo`, so it basically gets hung up on > 1GB files, of which we have a lot. This is documented in various places [a](https://stackoverflow.com/questions/12486623/paramiko-fails-to-download-large-files-1gb), [b](https://github.com/paramiko/paramiko/issues/926), etc. I tried the various configuration suggestions, but none seemed to help.

I wound up going back to the same approach used on the Polyphemus / Ruby-side of things. Use `lget` directly on C4 (very, very fast) and `curl` for Metis. I'm not very confident the `curl` implementation is correct -- I tried to mimic the multi-threaded approach in `etna/filesystem.rb`, but would certainly appreciate another set of eyes on it. It seems slow right now when I'm testing locally (like 300-500kB/s upload), but unsure if that is because Metis is overloaded with uploads, downloads, and ETLs, if my local upload speed just stinks, or if I need to somehow tweak the `curl` / `Popen` implementation? I keep getting 500 errors when testing (connection pool timeout to postgres) and have temporarily disabled an IPI ETL (re-ran for single-cell linking) to reduce the load.

Will let this percolate until next week when Zach is back.


airflow/opt/providers/etna/etna/hooks/c4.py
```diff
@@ -95,48 +98,47 @@ def __init__(self, hook: C4Hook):
...
+                continue
+
+            path_to_folder = os.path.join("/", "/".join(traversed), folder)
+            try:
+                sftp.chdir(path_to_folder)
+            except IOError:
+                sftp.mkdir(path_to_folder)
+            traversed.append(folder)
+
+        cmd = f"lftp sftp://{cat_hook.connection.login}:{cat_hook.connection.password}@{cat_hook.connection.host}  -e \"get {source_file.full_path} -o {os.path.join(full_path, file_name)}; bye\""
```
corps, line 131: i'm, a little worried that this can allow for escaping issues.  A malicious user could inject arbitrary shell code?  into the login value here.  The fix would be to see if we could `exec_command` with an array of arguments rather than a singular string (in that case, no matter what, the interpolated parts are only considered part of one string argument to the exec, rather than potentially being invoked by the shell itself).

coleshaw: Sure, I can make that change :+1:

airflow/opt/providers/etna/etna/hooks/c4.py
```diff
@@ -95,48 +98,47 @@ def __init__(self, hook: C4Hook):
...
+
+        with self.ssh() as ssh:
+            stdin, stdout, stderr = ssh.exec_command(cmd)
+
+            # exec_command is non-blocking, so we'll
+            #   add a blocking call here.
+            # We shouldn't deadlock since the LFTP output is small
+            # https://stackoverflow.com/a/28486051
+            exit_status = stdout.channel.recv_exit_status()
```
corps, line 140: I'd consider reading the solution here https://stackoverflow.com/a/31650750 just in case, it should be a small change.

So what's happening is that, when stdout is attached to a buffered output, it actually blocks when that buffer is full, under the assumption that some underlying reader needs to catch up.  stdout is designed to be loss-less; if there is more output than the consumer could keep up with, you only have two options: drop the extra output or block.  That's a principle of streaming.

This 'works' whenever the default buffer is enough to receive data from the stdout, but in general we should consume all the stdout if possible, even if we're throwing it away (I wouldn't actually print it, for instance, just read it).

airflow/opt/providers/etna/etna/hooks/c4.py
coleshaw: Hm, okay, so per the doc, [`exec_command` only take a string](https://docs.paramiko.org/en/stable/api/client.html#paramiko.client.SSHClient.exec_command).

Thoughts on using the same escaping utility as `subprocess`, (`list2cmdline`) a la https://stackoverflow.com/questions/3163236/escape-arguments-for-paramiko-sshclient-exec-command ?

corps: That makes sense.  Over ssh, you are relying on the configured shell of the remote user, so you don't know how the `exec` or `shell` processes will work, thus it is not possible to send the arguments as an array, only as a literal shell command.

Ok, can we atleast sanity check the interpolation parts against a regex?  Keeping it purely to alphanumeric?

corps:
Taking closer look at the upload mechanism.

corps:
The `retrieve` logic seems correct using the os pipe and opening that into the context.  It all looks correct.

As for performance, it's hard to say -- it looks correct and should be reasonably performant.  What is the transfer (download) performance of the `sftp` call directly?  SFTP is definitely not a fast transfer protocol -- it's mostly just easy to setup and secure (built on top of ssh).

coleshaw:
Hm, will test SFTP speed and post here.

Looking at the Ruby `curl` results, I think the common average is like 4.5MB/s, so I hope the Python side will be similar...

corps:
When you say "ruby curl results" what do you mean?  You mean the speed of ruby consuming curl?  What's the speed of the raw curl commands minus any environment?

corps:
And specifically when you fetch from `sftp` like in this code.

corps:
I guess basically without more instrumentation it's not clear.  It could either just be that curl from `sftp` is not very fast, or that the upload process is slow (which is basically the main bottleneck in the flow) due to some other server interaction.

Right now I'm looking into the pooling issue.

coleshaw:
> ruby curl results

I meant when I look at the ETL running in Polyphemus, using the Ruby version of the code, those are the average reported speeds for `curl` to download a file...

corps:
> > ruby curl results
> 
> I meant when I look at the ETL running in Polyphemus, using the Ruby version of the code, those are the average reported speeds for `curl` to download a file...

True, but those are also for downloading directly from metis, and not these sftp targets.
I don't think your code is introducing any major overhead as implemented.  The python is not getting in the way here - you are using OS primitives.

corps:
Err- you mean the SFTP etls?  Not the metis ones?  Sorry my context is getting twisted.  Hmm.

corps:
Either way, I'd try pushing this manually to production (after pulling latest) and possibly testing it there for diect comparison.

coleshaw:
Yeah, the Polyphemus ETLs that pulled from the CAT and uploaded to Metis (equivalent of this one)

Okay, that sounds good. I'll try it in production and see, hopefully the nodes can handle it better than my overheating laptop...

PR #962, opened 2022-07-26
Magma link collection same model; [close #947]

### LINKED ISSUE #947: Magma: one-to-many link within the same model
To support sc_seq, Magma needs to be able to construct one-to-many (collection) relationships within a single model. This would mimic the `reference_patient` (one-to-one, child) attribute in IPI, and would be used for constructing a "bundle" of related `sc_seq` records.

coleshaw:
This PR addresses #947 .

* It adds a `link_attribute_name` to the `attributes` table, with a migration to update values for all existing links. Should work in production :crossed_fingers: .
* Updates Magma `/query` and `/retrieve` to accomodate intra-model links / collections / child.
* Updates Magma `/update_model` to tweak how the `add_link` action behaves.
* Updates the `etna` gem to correctly parse and generate a new column in the model CSVs, `link_attribute_name`
* Updates the `etna` gem to handle and find reciprocal links using `link_attribute_name` instead of assuming it is the parent model name.

@graft , I'm happy to pair and review this if you'd like? Otherwise I think manually testing the gem, it seems to work locally... I can get link / collection attributes to appear in Timur and edit them via the UI.


etna/lib/etna/clients/magma/formatting/models_csv.rb
```diff
@@ -255,7 +261,11 @@ def process_parent_config(models, row_processor)
             end
 
             row_processor.process(:parent_link_type, :model_name, :parent_model_name) do |parent_link_type, template|
-              reciprocal = models.find_reciprocal(model_name: template.name, link_attribute_name: template.parent)
+              reciprocal = models.find_reciprocal(
+                model_name: template.name,
+                link_attribute_name: template.name,
```
graft, line 266: Should this come from template.name? That means that we expect the attribute to be `parent <model_name>` - if we wanted the parent attribute to be called something else this would not work. This is another "soft" kind of rule about the data graphs that we should clarify.

coleshaw:
Bug:

- [ ] Creating link via gem does not set `link_attribute_name` in the attributes table ...

coleshaw:
Okay, I think this fixes the bug listed above with the gem, plus fixes some bugs with the `/update` endpoint and creating the collection records from the parent (`link`) record.

PR #963, opened 2022-07-26
better error handling for janus token generation; close #957

### LINKED ISSUE #957: Janus: expired token results in 500 when calling generate_task_token
It seems like when I use an expired token to call the Janus "generate task token" API, I get a 500 server error. It might be nice to have a more descriptive error message or response sent, and to have this handled more gracefully, instead.

coleshaw:
This PR more gracefully handles a situation where a user uses an expired token to generate a new task token (#957). Had to modify `TestAuth.rb` in order to mimic actual, production behavior.

### DIFF
--- a/etna/lib/etna/test_auth.rb
+++ b/etna/lib/etna/test_auth.rb
@@ -54,6 +54,9 @@ module Etna
 
       return false unless token
 
+      # Useful for testing certain behavior
+      params = request.env["rack.request.params"]
+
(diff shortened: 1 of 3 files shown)


coleshaw:
Okay, I know this is super-hacky, but trying to make `test_auth.rb` behave more like `auth.rb` (and return `false` on any token-expired errors) breaks tons of tests, and I'm not sure it's worth making that kind of breaking change...

PR #966, opened 2022-07-27
Graft/gnomon

graft:
A work in progress for Gnomon. I am still sketching out the front end. Much work remains. A todo list:
- [x] - project view
- [x] - decompose-identifier view
- [x] - compose-identifier view
- [x] - rule-editor view
- [ ] - decompose API
- [ ] - compose API
- [ ] - rule-update API

Screenshot_20220727_152450


gnomon/Makefile
```diff
@@ -0,0 +1,20 @@
+include ../make-base/stubs.mk
+include ../make-base/docker-compose.mk
+include ../make-base/etna-ruby.mk
+include ../make-base/node.mk
+
+.projects-mark:
+	git clone git@github.com:mountetna/redcap-projects.git lib/etls/redcap/projects || REDCAP_CLONE_FAILED=1
```
coleshaw, line 7: Do you still need these repo's for gnomon?

gnomon/db/migrations/001_polyphemus_db_init.rb
```diff
@@ -0,0 +1,10 @@
+Sequel.migration do
```
coleshaw, line 1: hm, do these migrations still apply?

gnomon/lib/client/jsx/gnomon-nav.jsx
```diff
@@ -0,0 +1,17 @@
+import React, { useEffect } from 'react';
+import Nav from 'etna-js/components/Nav';
+import {selectUser} from 'etna-js/selectors/user-selector';
+import {useReduxState} from 'etna-js/hooks/useReduxState';
+
+const Logo = () => <div id='logo'/>;
+
+const NavBar = ({user}) => <div id='nav'></div>
+
+const PolyphemusNav = () => {
```
coleshaw, line 10: nit: `GnomonNav`?

gnomon/lib/client/scss/logo.scss
```diff
@@ -0,0 +1,8 @@
+#logo {
+  width: 50px;
+  height: 50px;
+  background-size: 100%;
+  border: 1px solid $gnomon-color;
+  border-radius: 50px;
+  background-image: url('~etna-js/images/gnomon.svg');
```
coleshaw, line 7: Do you need to add this file to the PR?

dtm2451:
I love this identifier breakdown figure concept for the project homepage!  It's actually kinda beautiful :heart_eyes:.

My one nit would be the element numbering along the bottom... if it will even stay, cuz it doesn't seem pointed to by any other text at the moment:
- Numbers should probably just get a single character count as their may be 1+,10+,and 100+ occupying that same niche.  Or perhaps even the entire structural units (project, species, day, biospecimen type, etc.) could just get a single number too? If I remember correctly, HuMu breaks some of what we'd probably consider rules nowadays as it has some models where, for example, Lymph Node is 'L' in the identifier and others where it is 'LN'.

gnomon/db/migrations/001_polyphemus_db_init.rb
graft: Err, just got lazy cleaning up stuff and started making the front-end. I'll clean this up when I get to the backend stuff.

graft:
> My one nit would be the element numbering along the bottom... if it will even stay, cuz it doesn't seem pointed to by any other text at the moment:

Yeah those numbers are mostly there because it helped me make my JSON fixtures more easily; it isn't required but I thought it might be useful/neat. It can go as it doesn't really seem necessary and it just adds clutter.

gnomon/Makefile
```diff
@@ -0,0 +1,20 @@
+include ../make-base/stubs.mk
+include ../make-base/docker-compose.mk
+include ../make-base/etna-ruby.mk
+include ../make-base/node.mk
+
+.projects-mark:
```
corps, line 6: Is this build target necessary?  Looks like maybe just copy paste.

coleshaw:
Have you thought about putting the UI into Timur, as a widget somewhere? I wonder if another app will be confusing to users ...

graft:
>Have you thought about putting the UI into Timur, as a widget somewhere? I wonder if another app will be confusing to users ...

Yeah, debating this one a lot. I agree that having yet another app is not great, but I don't think this really belongs in Timur, which is an interface to magma/the Data Library. Gnomon is meant to allow name creation to propagate entities into the library, but it isn't exactly the same thing. I think this will be more confusing if it is going on alongside the normal Timur/DL interactions. Plus I think it is nice if it is its own separate thing so that users of those functions can just understand them in isolation.

coleshaw:
Would it make sense to have the backend API inside of Magma, with a Gnomon UI separately (like Timur)? I'm worried if we get to the point where we want Magma to only create records for Gnomon-verified IDs, having them in two separate systems / databases might cause complexity / latency in having to make network calls for every single update action.

coleshaw:
Did you have some other Gnomon technical spec document? The one I found on Notion (titled Gnomon (Naming service)) seems like the original concept doc, not the technical doc you shared later...

graft:
> Did you have some other Gnomon technical spec document? The one I found on Notion (titled Gnomon (Naming service)) seems like the original concept doc, not the technical doc you shared later...

I added the planning doc to the bottom of that page. Re: putting the API in magma, I think that makes sense, I'm not sure what benefit there is in having it in a separate database, and allowing the Gnomon identifier list to be joined directly to Magma queries seems much more useful.

graft:
Ok, added the gnomon models to Magma. I am figuring we can make future gnomon PRs against this branch until the backend is wired up, but we could also just merge.

coleshaw:
Do you need to add the migration files to the PR? I don't see them ...

coleshaw:
And shall we pair today or tomorrow to divide up the remaining tasks?

magma/db/migrations/011_add_gnomon_tables.rb
```diff
@@ -0,0 +1,27 @@
...
+      column :config, :json, null: false
+      DateTime :created_at, null: false
+
+      index [:project_name, :version_number], unique: true
+    end
+
+    create_table(:identifiers) do
+      primary_key :id
+
+      String :project_name, null: false
```
coleshaw, line 17: Would it be helpful in any way to also record which `model` an identifier belongs to? Thinking it may save some sort of computation when searching for identifiers perhaps, being able to construct the rule directly instead of having to iterate through all rules to find the matching one?

magma/db/migrations/011_add_gnomon_tables.rb
```diff
@@ -0,0 +1,27 @@
+Sequel.migration do
+  change do
+    create_table(:rules) do
+      primary_key :id
+
+      String :project_name, null: false
+      Integer :version_number, null: false
+      column :config, :json, null: false
+      DateTime :created_at, null: false
```
coleshaw, line 9: I think for the other apps we used the timestamp plugin for Sequel, which also requires a column for `:updated_at`. While we may not use that column, the plugin is quite useful -- should we add the `:updated_at` column anyways?

magma/lib/magma/identifier.rb
```diff
@@ -0,0 +1,7 @@
+class Magma
+  class Identifier < Sequel::Model
+     many_to_one :rule
+     many_to_one :renamed_to, class: self, key: :renamed_to_id
```
coleshaw, line 4: Hm....is `:renamed_to` many to one, or one-to-one? Like ... you might have many identifiers renamed to the same one? That seems ...  hm, kind of mind boggling. Like, PAT1 -> PAT3, PAT2 -> PAT3, etc.?

magma/db/migrations/011_add_gnomon_tables.rb
```diff
@@ -0,0 +1,28 @@
...
+    create_table(:identifiers) do
+      primary_key :id
+
+      String :project_name, null: false
+      String :identifier, null: false
+      String :author, null: false
+      String :rule, null: false
+      foreign_key :grammar_id, :grammars, null: false
+      foreign_key :renamed_to_id, :identifiers
+      DateTime :created_at, null: false
```
coleshaw, line 23: Do we need to then add an "outside collaborator identifiers" list here, to track any third-party mappings?

coleshaw:
nit: On the identifier creation page's dendrogram type graphic, it might also be cool to have the on-hover also change the boldness of the horizontal brackets, not just bolding the line linking the section to the identifier area (hopefully that makes sense):

181383084-ca31574f-4375-456e-b10b-42483ba29efe

Right now it seems like only the linking lines turn bold, but then you have to visually follow those to the right identifier section ...

coleshaw:
@graft , you'll want to make sure to add `gnomon` to the list of apps to test, in `.github/workflows/branches.yml`

PR #968, opened 2022-07-28
Airflow cat metis ingest dies

coleshaw:
Testing this in production ... set the sub-process buffersize to default, instead of unbuffered. Seems to prevent stuck streams (?). Will keep monitoring.


corps:
Yup!  That makes sense, I did see the buffsize=0 but didn't mention anything before.

Terminology is confusing, but streaming data using a *fixed size* buffer is good practice for performance and preventing deadlocks.  that is different from *buffering* the full contents of a response (which is antithetical to streaming).

corps:
Also, *unbuffered* generally refers to *blocking* consumer / producer pipes.

IE:  If I write X data, my program stops until the receiver reads that X data, so that the memory of X data is never 'sitting around'.  It's either in the receiver or the sender end.

This can lead to the deadlock case if the reader and writer cannot independently block.  I imagine that could be a problem in Python.

coleshaw:
Oooohhh, I thought unbuffered would be like a pure stream — wonderful! Ha, didn’t realize it would be blocking.

corps:
It is a pure stream -- but in that there is _no intermediate buffer_ between reads and writes.  If you can't buffer a write, for instance, then only an equal read can unblock the write (otherwise the written data disappears, right?).

I always think of it in terms of shipping.  Two ports connected via freight boats ship things back and forth.  "Unbuffered" would mean that there is no shipyard -- trucks load and unload directly from the boats.  This is fine so long as there is perfect equilibrum on the rate of incoming goods to be shipped, and arrival of ships to do the shipping.  But once there are trucks arrive with goods and no ships to take them, the trucks either turn around and fail to ship, or they block awaiting a new ship.  In the buffered world, the trucks can drop their goods off in a shipyard (up to some fixed size), allowing a batch of boats to arrive on their own time.

Amoratized over time, the rate of transfer of boats is relatively fixed (which is why the fixed buffer is ok), but given a storm or an acute set back, there will be temporary imbalances.  Buffers, fix that, by providing my shipyard to store goods awaiting transfer.

This also works on the OTHER side, the read side.  If ships arrive with goods but there are no trucks, the ships must either wait for a truck (blocking), give up their goods (loss-y), or store it temporarily in the shipyard (buffer) if possible.

corps:
In other words, although "data" is ephemeral to us as humans, it is actually very real, very physical, and must always "exist somewhere".  Buffers are, well, a place for "data to exist" between the hand off of a write and a read.

coleshaw:
:thinking:

coleshaw:
Hm...can you give a concrete example of when you might be able to achieve data equilibrium without buffering (i.e. no shipyard)? Would you have to control both sides of the pipe?

PR #975, opened 2022-07-29
Mountetna python retry metis

coleshaw:
This PR adds a "retry" ability to the `upload_blob` functionality in Metis, for the Python mountetna package. Hope this might make the ETL uploads more stable?

It also catches and retries on [`RequestException`](https://github.com/psf/requests/blob/2a6f290bc09324406708a4d404a88a45d848ddf9/requests/exceptions.py#L12), which is a bit more generic than `HTTPError` and should catch more scenarios, like `metis_client` does (i.e. ProxyErrors, Timeouts, etc.).

### DIFF
--- a/etna/packages/etna-py/src/mountetna/metis.py
+++ b/etna/packages/etna-py/src/mountetna/metis.py
@@ -13,7 +13,8 @@ from serde.json import from_json, to_json
 from .etna_base import EtnaClientBase
 from .utils.multipart import encode_as_multipart
 from .utils.streaming import iterable_to_stream
-from requests import HTTPError
+from requests import RequestException
+from time import sleep
 
 @serialize
 @deserialize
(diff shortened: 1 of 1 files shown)


etna/packages/etna-py/src/mountetna/metis.py
```diff
@@ -503,8 +504,9 @@ def upload_parts(
                 )
 
                 unsent_zero_byte_file = False
-            except HTTPError as e:
+            except RequestException as e:
                 if remaining_attempts > 1:
+                    sleep(30)
```
corps, line 509: checkout `gater.py`.  I believe I already have exponential backoff implemented there?  In any case, I think we want exponential back off starting with a lower threshold than 30 seconds.  But this is totally fine as a start.

corps (approved):
In general, too, I think we should aim to see if we can get resuming of uploads on job retry working reasonably.  That requires care since we never want to resume uploading a file if the underlying file's hash has changed.

coleshaw:
Ooh yeah! That would be much better, I wasn't aware of `gater.py` -- will keep that in mind, thanks!

PR #977, opened 2022-08-01
only reraise the exception if not going to retry it

coleshaw:
Testing this out in production, but basically the code was re-raising the exception instead of actually re-trying the `upload_blob` request ... so no retry behavior. :disappointed: Trying again.


corps (approved):

etna/packages/etna-py/src/mountetna/metis.py
```diff
@@ -504,19 +519,7 @@ def upload_parts(
                 )
 
                 unsent_zero_byte_file = False
-            except RequestException as e:
```
coleshaw, line 507: Seems like when an exception is raised inside a generator (`yield` / `yield from`), it doesn't get handled inside the generator itself, but the calling method has to catch and handle it. You can see this in the stacktrace below, where the exception bubbles up from line 517 in `metis.py`, which is the generator call inside of this `except` block .. and which isn't subsequently caught.

```
[2022-08-01, 14:32:15 PDT] {remote_helpers_base.py:133} INFO - Uploading blob... (9.24%)
[2022-08-01, 14:32:17 PDT] {logging_mixin.py:109} INFO - Error 502, <!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
<html><head>
<title>502 Bad Gateway</title>
</head><body>
<h1>Bad Gateway</h1>
<p>The proxy server received an invalid
response from an upstream server.<br />
</p>
</body></html>
. Retrying...
[2022-08-01, 14:32:47 PDT] {taskinstance.py:1718} ERROR - Task failed with exception
  File "/home/airflow/.local/lib/python3.8/site-packages/mountetna/metis.py", line 495, in upload_parts
    self.upload_blob(
  File "/home/airflow/.local/lib/python3.8/site-packages/mountetna/metis.py", line 439, in upload_blob
    response = self.session.post(
  File "/home/airflow/.local/lib/python3.8/site-packages/requests/sessions.py", line 577, in post
    return self.request('POST', url, data=data, json=json, **kwargs)
  File "/home/airflow/.local/lib/python3.8/site-packages/requests/sessions.py", line 529, in request
    resp = self.send(prep, **send_kwargs)
  File "/home/airflow/.local/lib/python3.8/site-packages/requests/sessions.py", line 652, in send
    r = dispatch_hook('response', hooks, r, **kwargs)
  File "/home/airflow/.local/lib/python3.8/site-packages/requests/hooks.py", line 31, in dispatch_hook
    _hook_data = hook(hook_data, **kwargs)
  File "/home/airflow/.local/lib/python3.8/site-packages/mountetna/etna_base.py", line 38, in assert_status
    raise HTTPError(error_message, response=response)
requests.exceptions.HTTPError: Bad Gateway

During handling of the above exception, another exception occurred:

  File "/home/airflow/.local/lib/python3.8/site-packages/airflow/models/taskinstance.py", line 1334, in _run_raw_task
    self._execute_task_with_callbacks(context)
  File "/home/airflow/.local/lib/python3.8/site-packages/airflow/models/taskinstance.py", line 1460, in _execute_task_with_callbacks
    result = self._execute_task(context, self.task)
  File "/home/airflow/.local/lib/python3.8/site-packages/airflow/models/taskinstance.py", line 1516, in _execute_task
    result = execute_callable(context=context)
  File "/home/airflow/.local/lib/python3.8/site-packages/airflow/decorators/base.py", line 134, in execute
    return_value = super().execute(context)
  File "/home/airflow/.local/lib/python3.8/site-packages/airflow/operators/python.py", line 174, in execute
    return_value = self.execute_callable()
  File "/home/airflow/.local/lib/python3.8/site-packages/airflow/operators/python.py", line 188, in execute_callable
    return self.python_callable(*self.op_args, **self.op_kwargs)
  File "/home/airflow/.local/lib/python3.8/site-packages/etna/etls/cat.py", line 162, in ingest
    self.handle_metis_ingest(
  File "/home/airflow/.local/lib/python3.8/site-packages/etna/etls/remote_helpers_base.py", line 123, in handle_metis_ingest
    for blob in metis.upload_file(
  File "/home/airflow/.local/lib/python3.8/site-packages/mountetna/metis.py", line 517, in upload_parts
    yield from self.upload_parts(
  File "/home/airflow/.local/lib/python3.8/site-packages/mountetna/metis.py", line 491, in upload_parts
    blob_data = upload.read_next_blob()
  File "/home/airflow/.local/lib/python3.8/site-packages/mountetna/metis.py", line 288, in read_next_blob
    raise ValueError(
ValueError: Upload source stream requires restart, upload has failed.
```

So, moving the retry logic to `upload_file`, to see if the upload will retry more than once.

corps: Hold up, this seems wrong, let me check.

coleshaw: :+1: Was hard to test without triggering an exception somehow ... and now that production is stable, haven't seen a retry yet ...

coleshaw: Oh oh, is it because `ValueError` is not caught by the retry block, so it just gets re-raised and bubbles up? Hm....

corps: IE: Make it possible for us to provide alternative functions rather than upload_start and upload_blob to the retry logic, making it much simpler to test

coleshaw: :+1:

coleshaw:
```
[2022-08-02, 10:59:10 PDT] {remote_helpers_base.py:133} INFO - Uploading blob... (77.69%)
[2022-08-02, 10:59:13 PDT] {taskinstance.py:1718} ERROR - Task failed with exception
  File "/home/airflow/.local/lib/python3.8/site-packages/airflow/models/taskinstance.py", line 1334, in _run_raw_task
    self._execute_task_with_callbacks(context)
  File "/home/airflow/.local/lib/python3.8/site-packages/airflow/models/taskinstance.py", line 1460, in _execute_task_with_callbacks
    result = self._execute_task(context, self.task)
  File "/home/airflow/.local/lib/python3.8/site-packages/airflow/models/taskinstance.py", line 1516, in _execute_task
    result = execute_callable(context=context)
  File "/home/airflow/.local/lib/python3.8/site-packages/airflow/decorators/base.py", line 134, in execute
    return_value = super().execute(context)
  File "/home/airflow/.local/lib/python3.8/site-packages/airflow/operators/python.py", line 174, in execute
    return_value = self.execute_callable()
  File "/home/airflow/.local/lib/python3.8/site-packages/airflow/operators/python.py", line 188, in execute_callable
    return self.python_callable(*self.op_args, **self.op_kwargs)
  File "/home/airflow/.local/lib/python3.8/site-packages/etna/etls/cat.py", line 162, in ingest
    self.handle_metis_ingest(
  File "/home/airflow/.local/lib/python3.8/site-packages/etna/etls/remote_helpers_base.py", line 123, in handle_metis_ingest
    for blob in metis.upload_file(
  File "/home/airflow/.local/lib/python3.8/site-packages/mountetna/metis.py", line 523, in upload_parts
    raise e
  File "/home/airflow/.local/lib/python3.8/site-packages/mountetna/metis.py", line 510, in upload_parts
    self.upload_blob(
  File "/home/airflow/.local/lib/python3.8/site-packages/mountetna/metis.py", line 439, in upload_blob
    response = self.session.post(
  File "/home/airflow/.local/lib/python3.8/site-packages/requests/sessions.py", line 577, in post
    return self.request('POST', url, data=data, json=json, **kwargs)
  File "/home/airflow/.local/lib/python3.8/site-packages/requests/sessions.py", line 529, in request
    resp = self.send(prep, **send_kwargs)
  File "/home/airflow/.local/lib/python3.8/site-packages/requests/sessions.py", line 652, in send
    r = dispatch_hook('response', hooks, r, **kwargs)
  File "/home/airflow/.local/lib/python3.8/site-packages/requests/hooks.py", line 31, in dispatch_hook
    _hook_data = hook(hook_data, **kwargs)
  File "/home/airflow/.local/lib/python3.8/site-packages/mountetna/etna_base.py", line 38, in assert_status
    raise HTTPError(error_message, response=response)
requests.exceptions.HTTPError: Service Unavailable
```

Yeah, I'm thinking the retry logic definitely isn't working as I thought. Will have to better understand generators and how they interact with exceptions, because seems like I'm not wrapping it correctly.

etna/packages/etna-py/src/mountetna/metis.py
```diff
@@ -466,10 +466,25 @@ def upload_file(
...
         authorization = self.authorize_upload(project_name, bucket_name, dest_path)
         upload = Upload(file=file, file_size=size, upload_path=authorization.upload_path)
-        return self.upload_parts(upload, uuid.uuid4().hex, max_retries)
+        num_tries = max_retries
+        last_status_code = None
+        metis_uid = uuid.uuid4().hex
+        while num_tries > 0:
+            try:
+                return self.upload_parts(upload, metis_uid, last_status_code == 422)
+            except (RequestException, HTTPError) as e:
```
corps, line 475: Here's the problem, I think --- this `except` clause won't catch the `upload_parts` exceptions because this function is returning the generator, not running it.  You need the except inside `upload_parts`.

corps: In other words, when you call `upload_parts`, you don't run any part of upload parts at that point.
When upload_parts is running is during the `for _ in upload_parts()` or any `yield from` chained on it.

But, if your exception handling is inside upload_parts, then it DOES catch, since upload_parts itself is run in the generator context.

corps: try this on:

```
def m():
  print("hi!")
  yield 1

gen = m()
next(gen)
```

corps:
The original code, I think, was closer to being correct, but may have been missing something.

coleshaw:
I agree, I wonder if it's just a logic thing where I need to catch more of the exceptions thrown in the rest of the upload code, and retry on those...

coleshaw:
Hm... so I've swapped in the `RetryGater`, and I think the correct behavior is actually to retry all the requests, not just `upload_parts` (since with some of the previous uploads, we've run into issues with `metis_client` where a "list folders" might fail as well, just depends on timing). So something like the current code? I think this is thus more testable and will see if I can write up some unit tests around the retry method.

corps:
> Hm... so I've swapped in the `RetryGater`, and I think the correct behavior is actually to retry all the requests, not just `upload_parts` (since with some of the previous uploads, we've run into issues with `metis_client` where a "list folders" might fail as well, just depends on timing). So something like the current code? I think this is thus more testable and will see if I can write up some unit tests around the retry method.

That is probably true.  For background, the reason that specifically upload has _custom_ retry logic is because in some cases it is necessary to restart the entire session and not simply retry the request specifically.  

Essentially, idempotent (GET) requests are naively retryable.  The failure has no chance of also modifying some 'state' associated with a 'sessoin'.

Non idempotent, side-effect causing requests (POSTS) are not always retryable without the client also handling some specialized logic.  As is the case for upload.

etna/packages/etna-py/src/mountetna/metis.py
```diff
@@ -436,7 +445,8 @@ def start_upload(self, request: "UploadStartRequest") -> "UploadResponse":
         return response_obj
 
     def upload_blob(self, request: "UploadBlobRequest"):
-        response = self.session.post(
+        response = self._retry_request(
+            self.session.post,
```
corps, line 449: This isn't retryable -- a failure does not mean that the previous blob was not saved.  It's possible, depending on the error, that the previous upload DID succeed on saving to the file system but FAILED due to say the connection or some cleanup tasks.  Which is why the previous logic attempted to infer if restarting the upload was necessary.

etna/packages/etna-py/src/mountetna/metis.py
```diff
@@ -504,27 +521,22 @@ def upload_parts(
...
+                else:
+                    raise e
 
             yield upload
 
     def create_folder(
         self, project_name: str, bucket_name: str, folder_path: str
     ) -> FoldersAndFilesResponse:
-        response = self.session.post(
+        response = self._retry_request(
```
corps, line 538: Note:  Retry here is tricky, because it's not idempotent.  For instance, it's possible the first request DOES create the folder, but the connection breaks or there is a 500 due to some other subsequent logic bug (maybe just apache has an issue).  So then the client retries -- but this time it returns a invalid request exception because the folder exists from the previous request.  In this case, a lot of `create_folder` calls may throw false negative exceptions when all that is happening is retry logic is kicking in.

Ideally, we would make most metis requests idempotent (for instance, not throwing an exception if two identical requests are made).  But that may be a bit of work.

corps:
So, I'm ok with this hack for now, but I thought I'd let you know -- if we do want to have more custom, shareable retry logic, another place to consider adding it is in the EtnaSession class.  Incidentally, _there is already some retry logic there_.  But it can't retry everything naively (specifically uploads).  It'd be worth checking if that base retry logic isn't already catching certain classes of issues, or if it can be extended.

coleshaw:
Oooh, I see! Nice! Yes, looks like we could configure that to retry the upload requests, though like you mention, I guess the risk would be that not all the metis endpoints are idempotent...? My guess is it does not catch the folder create or upload POST calls right now ...

coleshaw:
k, so I think this works, but sometimes the retries aren't long enough to bridge a service update (still saw a bad gateway get through). But much better, so going with this for now! Will keep monitoring.

corps:
@coleshaw I'm not convinced that it's service updates causing the gateway issue:

```
root@vulcan-node:~# docker service ps metis_app
ID             NAME              IMAGE                        NODE          DESIRED STATE   CURRENT STATE           ERROR                         PORTS
ic69qkbex6ul   metis_app.1       etnaagent/metis:production   dsco3         Running         Running 19 hours ago                                  
9aqx3xxlwt6u    \_ metis_app.1   etnaagent/metis:production   dsco3         Shutdown        Shutdown 19 hours ago                                 
hupm3zb3vpwx    \_ metis_app.1   etnaagent/metis:production   vulcan-node   Shutdown        Shutdown 47 hours ago                                 
246u4e1n21dz    \_ metis_app.1   etnaagent/metis:production   vulcan-node   Shutdown        Shutdown 3 days ago                                   
ye3ewfpl9fj7    \_ metis_app.1   etnaagent/metis:production   vulcan-node   Shutdown        Failed 3 days ago       "task: non-zero exit (137)"   
4uemnqmuihqv   metis_app.2       etnaagent/metis:production   dsco2         Running         Running 19 hours ago                                  
qyvqxxsdv1aa    \_ metis_app.2   etnaagent/metis:production   dsco2         Shutdown        Shutdown 19 hours ago                                 
xhuj4i6zj1lv    \_ metis_app.2   etnaagent/metis:production   dsco2         Shutdown        Shutdown 47 hours ago                                 
g6bqsbx1t2li    \_ metis_app.2   etnaagent/metis:production   dsco2         Shutdown        Shutdown 3 days ago                                   
i184gvkam2h1    \_ metis_app.2   etnaagent/metis:production   dsco2         Shutdown        Shutdown 6 days ago                                   
ux02kh67i09u   metis_app.3       etnaagent/metis:production   vulcan-node   Running         Running 19 hours ago                                  
grknwybtnyk4    \_ metis_app.3   etnaagent/metis:production   vulcan-node   Shutdown        Shutdown 19 hours ago                                 
root@vulcan-node:~# docker service ps metis_app_fe
ID             NAME                 IMAGE                               NODE          DESIRED STATE   CURRENT STATE          ERROR                              PORTS
zjm916wlks5m   metis_app_fe.1       etnaagent/metis_app_fe:production   dsco2         Running         Running 7 days ago                                        
igq9fdvwavgs    \_ metis_app_fe.1   etnaagent/metis_app_fe:production   vulcan-node   Shutdown        Failed 7 days ago      "No such container: metis_app_…"   
bgux2nairoy4    \_ metis_app_fe.1   etnaagent/metis_app_fe:production   dsco2         Shutdown        Shutdown 7 days ago                                       
0300ldcppa6x    \_ metis_app_fe.1   etnaagent/metis_app_fe:production   vulcan-node   Shutdown        Shutdown 4 weeks ago                                      
lp65ldk1whg3    \_ metis_app_fe.1   etnaagent/metis_app_fe:production   vulcan-node   Shutdown        Failed 5 weeks ago     "No such container: metis_app_…"   
ou11njvged5c  
```

coleshaw:
hmmm, interesting. The last bad gateway error was this morning, 7:32am pacific time:

```
[2022-08-04, 07:32:02 PDT] {taskinstance.py:1718} ERROR - Task failed with exception
  File "/home/airflow/.local/lib/python3.8/site-packages/mountetna/metis.py", line 495, in upload_parts
    self.upload_blob(
  File "/home/airflow/.local/lib/python3.8/site-packages/mountetna/metis.py", line 439, in upload_blob
    response = self.session.post(
  File "/home/airflow/.local/lib/python3.8/site-packages/requests/sessions.py", line 577, in post
    return self.request('POST', url, data=data, json=json, **kwargs)
  File "/home/airflow/.local/lib/python3.8/site-packages/requests/sessions.py", line 529, in request
    resp = self.send(prep, **send_kwargs)
  File "/home/airflow/.local/lib/python3.8/site-packages/requests/sessions.py", line 652, in send
    r = dispatch_hook('response', hooks, r, **kwargs)
  File "/home/airflow/.local/lib/python3.8/site-packages/requests/hooks.py", line 31, in dispatch_hook
    _hook_data = hook(hook_data, **kwargs)
  File "/home/airflow/.local/lib/python3.8/site-packages/mountetna/etna_base.py", line 38, in assert_status
    raise HTTPError(error_message, response=
```

Might be something else triggering a 502 on the specific request, then?

There was one last night, too, but the timing corresponded exactly with the production deploy around 5:55pm pacific.

PR #987, opened 2022-08-12
Performance improvements

corps:
Will discuss the changes in the morning


coleshaw:
Ah, is this to get internal network routing, i.e. `metis_app` instead of hitting the public URL?

corps:
> Ah, is this to get internal network routing, i.e. `metis_app` instead of hitting the public URL?

Yup.  I'm seeing a 20% performance increase for the materialize gne jobs using the internal routing.

coleshaw:
Nice, that's awesome!

coleshaw (approved):

airflow/opt/providers/etna/etna/metrics/rollup_metrics.py
```diff
@@ -55,4 +55,5 @@ def collect(self) -> Generator[Metric, None, None]:
                 yield metric
 
 def register_rollup_metrics():
-    REGISTRY.register(RollupMetricsCollector())
\ No newline at end of file
+    # REGISTRY.register(RollupMetricsCollector())
+    pass
```
coleshaw, line 59: Do we lose any metrics by taking this out?

corps: Nope, those metrics in the end aren't being used.  I think it's safe to remove these now.

Issue #990, opened 2022-08-12
Vulcan: bug with escaping JSON strings?


coleshaw:
I think the flip side of #989 is that Vulcan is actually escaping the JSON array inputs now ... so it sends them back to Magma escaped. We may actually not need quoted strings now?

```
Cleaning up docker container

b'Traceback (most recent call last):\n  File "/archimedes-exec/tmpzsg69tn9/script.py", line 16, in <module>\n    tsv_result = query_tsv(magma, project_name,\n  File "/app/archimedes/functions/magma/magma.py", line 24, in query_tsv\n    query_result = magma.query({\n  File "/root/.cache/pypoetry/virtualenvs/archimedes-9TtSrW0h-py3.8/lib/python3.8/site-packages/mountetna/magma.py", line 301, in query\n    return from_json(QueryResponse, response.content)\n  File "/root/.cache/pypoetry/virtualenvs/archimedes-9TtSrW0h-py3.8/lib/python3.8/site-packages/serde/json.py", line 75, in from_json\n    return from_dict(c, de.deserialize(s, **opts), reuse_instances=False)\n  File "/root/.cache/pypoetry/virtualenvs/archimedes-9TtSrW0h-py3.8/lib/python3.8/site-packages/serde/json.py", line 49, in deserialize\n    return json_loads(s, **opts)\n  File "/root/.cache/pypoetry/virtualenvs/archimedes-9TtSrW0h-py3.8/lib/python3.8/site-packages/serde/json.py", line 34, in json_loads\n    return json.loads(s, **opts)\n  File "/usr/local/lib/python3.8/json/__init__.py", line 357, in loads\n    return _default_decoder.decode(s)\n  File "/usr/local/lib/python3.8/json/decoder.py", line 337, in decode\n    obj, end = self.raw_decode(s, idx=_w(s, 0).end())\n  File "/usr/local/lib/python3.8/json/decoder.py", line 355, in raw_decode\n    raise JSONDecodeError("Expecting value", s, err.value) from None\njson.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)\n'

[...]

PR #991, opened 2022-08-12
Vulcan timur fix query plot

### LINKED ISSUE #989: Timur: bug with query plot params?
Just noticed this, when using the Plot button from Timur query, it seems like the "user columns" param is converted incorrectly now. Not sure what changed, but it seems like the Timur side sends the query as a JSON string, but the user columns as an actual array:

Screenshot from 2022-08-12 15-33-46

This means it breaks in Vulcan, since they aren't quoted strings:

Screenshot from 2022-08-12 15-34-30

My guess is this worked before we converted the inputs to Material UI, probably due to some type-casting on our end, but the current behavior doesn't make the same assumptions and thus breaks.

I will fix the query side to always send a JSON string.

coleshaw:
This PR attempts to close #989 and #990. I noticed on production that if you try to go from Query -> Vulcan to do any new visualiztion, the query part breaks when you click "Run" in Vulcan, now...

Probably works for existing figures that have already fetched data and the data exists in the cache. I suspect those will break if you try to modify the queries and re-run, unfortunately...

So far the PR only addresses #989 on the Timur-side of things, to make sure JSON strings are sent to Vulcan.

 - [x] Fix Vulcan query to Magma (currently throws a JSON error, I suspect because something is wrong with the token or query formulation?)

@dtm2451 , I will try to figure out a fix for #990 before Monday or on the plane, but it may (likely) be a hack that you may decide to improve later on ...

### DIFF
--- a/etna/packages/etna-py/pyproject.toml
+++ b/etna/packages/etna-py/pyproject.toml
@@ -1,13 +1,14 @@
 [tool.poetry]
 name = "mountetna"
-version = "0.1.4"
+version = "0.1.5"
 description = "Client package for Mount Etna data library"
 
 license = "GPL-2.0-only"
 
 authors = [
     "Zachary Collins <zachary.collins@ucsf.edu>",
-    "Saurabh Asthana <saurabh.asthana@ucsf.edu>"
+    "Saurabh Asthana <saurabh.asthana@ucsf.edu>",
+    "Cole Shaw <cole.shaw@ucsf.edu>"
 ]
 
 readme = 'README.md'  # Markdown files are supported
--- a/etna/packages/etna-py/src/mountetna/magma.py
+++ b/etna/packages/etna-py/src/mountetna/magma.py
@@ -1,5 +1,6 @@
 import dataclasses
 import typing
+from io import StringIO
 from inspect import isgenerator
 from typing import Dict, Optional, List
 from serde import serialize, deserialize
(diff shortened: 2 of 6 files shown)


dtm2451:
Shoot!  Glad you caught these before Tuesday.... I'd not noticed these issues =/.  I'll take a look at this on Monday, though likely won't get a chance until you'll already be in the air.

coleshaw:
Okay, I think the problem is in `etna-py`, where it [doesn't handle a TSV correctly](https://github.com/mountetna/monoetna/blob/master/etna/packages/etna-py/src/mountetna/magma.py#L301), like [Magby did](https://github.com/mountetna/monoetna/blob/master/etna/packages/magby/magby/Magma.py#L50-L69), it assumes a JSON response. Will try to push a fix.

coleshaw:
This seems to work for me locally (sorry @graft , my hack wound up being in the mountetna Python library code, not Vulcan UI code). Maybe someone can review / test on Monday, and we can merge / push it to production same day?

etna/packages/etna-py/pyproject.toml
```diff
@@ -1,13 +1,14 @@
...
+version = "0.1.5"
 description = "Client package for Mount Etna data library"
 
 license = "GPL-2.0-only"
 
 authors = [
     "Zachary Collins <zachary.collins@ucsf.edu>",
-    "Saurabh Asthana <saurabh.asthana@ucsf.edu>"
+    "Saurabh Asthana <saurabh.asthana@ucsf.edu>",
+    "Cole Shaw <cole.shaw@ucsf.edu>"
```
graft, line 11: :smiley:

PR #998, opened 2022-08-17
Cs/build remove sync subtrees

coleshaw:
As we talked about yesterday, this PR removes the subtree syncing to improve build / CI performance. This means that the individual app repos will not be updated, but no one uses those for their workflow anyways ... so basically they will be deprecated.

### DIFF
--- a/.github/workflows/gem-etna-push.yml
+++ b/.github/workflows/gem-etna-push.yml
@@ -1,0 +1,31 @@
+name: Ruby Gem
+
+on:
+  push:
+    branches: [master]
+
+jobs:
+  build:
+    name: Build + Publish
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


.github/workflows/gem-etna-push.yml
```diff
@@ -0,0 +1,31 @@
...
+        with:
+          ruby-version: 2.5.x
+
+      - name: Publish to RubyGems
+        run: |
+          mkdir -p $HOME/.gem
+          touch $HOME/.gem/credentials
+          chmod 0600 $HOME/.gem/credentials
+          printf -- "---\n:rubygems_api_key: ${GEM_HOST_API_KEY}\n" > $HOME/.gem/credentials
+          gem build *.gemspec
```
coleshaw, line 25: Here still need to do a `cd etna/` since this is from `monoetna` now...

PR #1005, opened 2022-08-23
Airflow debug metis 403 on file

coleshaw:
For some reason the Box ingest for xneo1 repeatedly failed on a specific file (Metis returned 403 on upload_start), so trying to add some debugging and batch handling to figure out why.

```
[2022-08-22, 16:02:02 PDT] {remote_helpers_base.py:120} INFO - Uploading /Root/DSCOLAB_FCS/XNEO1/XNEO-POOL-CYM2/20220107_XNEO1_Pool_CYM2_01(2 calibration channels missing).FCS to cytof/XNEO-POOL-CYM2/20220107_XNEO1_Pool_CYM2_01(2 calibration channels missing).FCS.
[2022-08-22, 16:02:03 PDT] {taskinstance.py:1718} ERROR - Task failed with exception
  File "/home/airflow/.local/lib/python3.8/site-packages/airflow/models/taskinstance.py", line 1334, in _run_raw_task
    self._execute_task_with_callbacks(context)
  File "/home/airflow/.local/lib/python3.8/site-packages/airflow/models/taskinstance.py", line 1460, in _execute_task_with_callbacks
    result = self._execute_task(context, self.task)
  File "/home/airflow/.local/lib/python3.8/site-packages/airflow/models/taskinstance.py", line 1516, in _execute_task
    result = execute_callable(context=context)
  File "/home/airflow/.local/lib/python3.8/site-packages/airflow/decorators/base.py", line 134, in execute
    return_value = super().execute(context)
  File "/home/airflow/.local/lib/python3.8/site-packages/airflow/operators/python.py", line 174, in execute
    return_value = self.execute_callable()
  File "/home/airflow/.local/lib/python3.8/site-packages/airflow/operators/python.py", line 188, in execute_callable
    return self.python_callable(*self.op_args, **self.op_kwargs)
  File "/home/airflow/.local/lib/python3.8/site-packages/etna/etls/box.py", line 91, in ingest
    self.handle_metis_ingest(
  File "/home/airflow/.local/lib/python3.8/site-packages/etna/etls/remote_helpers_base.py", line 123, in handle_metis_ingest
    for blob in metis.upload_file(
  File "/home/airflow/.local/lib/python3.8/site-packages/mountetna/metis.py", line 489, in upload_parts
    self.start_upload(
  File "/home/airflow/.local/lib/python3.8/site-packages/mountetna/metis.py", line 436, in start_upload
    response = self.session.post(
  File "/home/airflow/.local/lib/python3.8/site-packages/requests/sessions.py", line 577, in post
    return self.request('POST', url, data=data, json=json, **kwargs)
  File "/home/airflow/.local/lib/python3.8/site-packages/requests/sessions.py", line 529, in request
    resp = self.send(prep, **send_kwargs)
  File "/home/airflow/.local/lib/python3.8/site-packages/requests/sessions.py", line 652, in send
    r = dispatch_hook('response', hooks, r, **kwargs)
  File "/home/airflow/.local/lib/python3.8/site-packages/requests/hooks.py", line 31, in dispatch_hook
    _hook_data = hook(hook_data, **kwargs)
  File "/home/airflow/.local/lib/python3.8/site-packages/mountetna/etna_base.py", line 46, in assert_status
    raise HTTPError(error_message, response=response)
requests.exceptions.HTTPError: Request failed with 403: You are forbidden from performing this action.
```

Maybe because the file name has `()` in it? But those characters should be valid and pass the Metis regex, so trying to collect more information. Failed 4 times on the exact same file ...


etna/lib/etna/auth.rb
```diff
@@ -148,7 +148,7 @@ def approve_hmac(request)
       hmac_params = {
         method: request.request_method,
         host: request.host,
-        path: request.path,
+        path: encode_path(request.path),
```
coleshaw, line 151: This is probably the critical part for you to take a look at, @graft  ... explained more fully in #1006.

coleshaw:
Should close #1006 :crossed_fingers:

coleshaw:
To reproduce the problem, locally or on Metis:

* Have a simple file with `(` or `)` in the file name, i.e. `upload (testing).txt`
* Try uploading the file to Metis. Can use either metis_client or the Metis UI.
* You should get a 403 / "You are not authorized" message.

On this branch, try uploading the same file. It should work! :tada:

coleshaw:
Similary, if you change an existing file's name (already in Metis) to include `(` and `)`, and try downloading it, will not work on `master` or production. But should work on this branch.

etna/lib/etna/auth.rb
```diff
@@ -172,5 +172,15 @@ def approve_hmac(request)
...
       return true
     end
+
+    def encode_path(path)
+      # request.path is pulled from the webserver. In our case, Apache via
+      #   ProxyPass.
+      # However, Apache mod_proxy encodes this differently than URI.encode_www_form_component ... so we normalize it here according to how Etna::Route.path() behaves, so that Metis HMAC signing for
+      # upload / download URLs is compatible.
+      path.split("/").map do |c|
+        URI.encode_www_form_component(URI.decode_www_form_component(c))
```
graft, line 182: Hmm, I would really worry about how this behaves in cases we aren't testing explicitly here... For example if there are encoded-looking things in the original path this would not work. E.g., if there are two files in a dir, "hey/what+up" and "hey/what up", both of which should be able to exist as independent files in Metis, these both get encoded to the same path by `encode_path`.

coleshaw: Ah, so that's an interesting point! Hm, I will have to test and play around with this. I actually expect this branch would correctly protect all the special characters (like `+`).

```
> URI.encode_www_form_component("file+example.txt")
=> "file%2Bexample.txt"
```

So the `+` should get encoded and protected, and when you call `URI.decode_www_form_component()`, it converts back to a `+` instead of a space? I guess it does depend on if `mod_proxy` passes the `+` through as `%2B` or `+`, in which case we might have to resort to using the `nocanon` flag on Apache, because we have no way to tell in the application if it's supposed to be a `+` or a space.

Note that in production, you currently cannot upload a file like `hey/what+up` because of the same problem as the `(` and `)` -- the `+` gets percent-encoded multiple times and results in a 403 from Metis upon calling `upload#start` due to an HMAC mis-match. A file with spaces in the file name works fine. So while we may want such files to independently exist, they currently cannot be uploaded as such -- can only exist if you edit the files post-uploading.

coleshaw:
Okay, I think this is now better and correctly handles the use case of `+` in file names. Basically the RFC spec for query params is different than URLs. `%20` should be used in URLs to escape spaces, but `%20` or `+` can be used in query params. The Ruby method `URI.encode_www_form_component` is used for query params, so we shouldn't be using it for URL path components. I switched over to `Addressable::URI`, since there doesn't seem to be a standard library method that correctly encodes URL path components, with the deprecation of `URI.encode`...

All of these files were uploadable and downloadable on this branch:

Screenshot from 2022-08-25 09-55-18

PR #1008, opened 2022-08-24
Update ci secret

coleshaw:
So ... this updates the VCR cassettes to use a new `CI_SECRET` environment variable. It is going to fail several of the CI test suites ...

I would propose that early some morning, Eastern time, I update the `CI_SECRET` in the repo, and then re-run the test suite. If all the tests then pass, I would merge in this PR and we can use the new secret. During that window, it would mean that no other PRs could pass tests and be merged in. Probably after I get back, or I can find some non-critical time when you two don't have work you want to merge in?


dtm2451:
I won't be looking to merge anything myself this week :shrug:

PR #1015, opened 2022-08-26
Airflow specify backup mount option docker swarm

coleshaw:
In order for Airflow backup ETLs to write directly to the /data3 mount, I've added a config flag to the docker swarm operator to make sure that exists in the backup tasks. Testing this to see if it works ...


coleshaw:
Decided not to go this route -- instead will mimic the other `_backup` service configs for the swarm. That way can get a consistent username:group to set permissions correctly, and easier to update via config rather than having to deploy new code (i.e. if mount points change in the future).

PR #1016, opened 2022-09-08
Polyphemus mvir1 waiver update json

coleshaw:
This PR updates the MVIR1 cascade waiver job to call `update_json` instead of `update` -- not sure if this helps, but related to #729 and #921, might affect behavior.

### DIFF
--- a/docker/archimedes-base/Dockerfile
+++ b/docker/archimedes-base/Dockerfile
@@ -7,7 +7,7 @@ WORKDIR /app
 ENV DOCKER_VERSION 19.03.12
 ENV PYTHON_VERSION 3.8
 ENV DOCKERIZE_VERSION 0.5.0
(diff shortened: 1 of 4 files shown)


docker/archimedes-base/Dockerfile
```diff
@@ -7,7 +7,7 @@ WORKDIR /app
 ENV DOCKER_VERSION 19.03.12
 ENV PYTHON_VERSION 3.8
 ENV DOCKERIZE_VERSION 0.5.0
-ENV POETRY_VERSION 1.1.4
+ENV POETRY_VERSION 1.1.15
```
coleshaw, line 10: Looks like we hit this problem: https://github.com/python-poetry/poetry/issues/6300

So bumping the poetry version.

graft (approved):

PR #1019, opened 2022-09-08
Vulcan reduce UI bundle size

coleshaw:
Just opening this as a PR to make sure it doesn't get lost ...

Goal of the PR is to combine all the `package.json` and npm information into a single `package.json` and `package-lock.json` in `etna`, much like was done for Ruby's `Gemfile` and `Gemfile.lock`. The benefits would be an quicker build process, with `npm install` only having to run once in the `etna` image, and an easier path to maintain the UI code -- greater consistency in dependency versions, upgrade paths, etc., across all the apps.

This also minifies the production webpack builds, which makes the final bundles smaller, which leads to faster load times for users.

For development, with each of the apps, you may have to manually delete any `node_modules` folders and equivalent files that have been moved to `etna`, i.e. `package-lock.json`, `webpack.config.js`, etc. Those are all linked now to the version in `etna`, but that change has to be manually implemented.


coleshaw:
@graft @dtm2451 , I think this is ready for review, but it would probably be good for one of you to test things locally -- especially the Metis "download folder" feature. That has been tripping me up the last couple of days, because the forked library we were using stopped working during this consolidation process :angry: . I've pushed a version that seems to work to our [fork](https://github.com/mountetna/downzip), but the behavior was inconsistent locally, and I'm not very confident it actually works :crossed_fingers: 

Note that my force push of our  `downzip` fork probably broke any Metis deployments that will happen before this branch gets merged in... :cry: But I couldn't figure out how our fork actually worked (:confused: ), so felt like I had to reset it on top of the `upstream` repo for it to make sense to me.

This PR basically consolidates all the `package.json` files and NPM dependencies into `etna`, and should help speed up build processes across all applications, since only a single `npm install` is now required, instead of one per app. Should also help us do version upgrades more easily in the future.

I have tried testing all the apps manually, and things seem to work for me. There is high risk of breakage, though, so we should be prepared that some UI features may need to be fixed as this gets rolled out to production.

To test locally:

* Check out this branch
* Manually remove any `<app>/node_modules/` folders that you may have from over a year ago (when NPM dependencies were put into a Docker volume)
* Manually remove any `.eslintrc.js`, `package-lock.json`, `babel.config.js`, `webpack.config.js`, `jest.config.js`, `tsconfig.json`, or `.eslintignore` file that still resides within each app directory.
* Run `make down` to stop all apps
* Run `make -C etna update` to install the NPM dependencies
* Run `make up` to bring up all apps
* Stuff should run as before! :crossed_fingers:

graft:
Looks good to me so far... `janus/package-lock.json` seems to be missing.

docker/etna-apache/httpd.conf
```diff
@@ -62,7 +62,7 @@ LoadModule filter_module modules/mod_filter.so
 #LoadModule substitute_module modules/mod_substitute.so
 #LoadModule sed_module modules/mod_sed.so
 #LoadModule charset_lite_module modules/mod_charset_lite.so
-#LoadModule deflate_module modules/mod_deflate.so
+LoadModule deflate_module modules/mod_deflate.so
```
graft, line 65: why these deflate-related changes?

coleshaw: This would only affect dev, should help make things load faster in dev, too.

coleshaw:
> Looks good to me so far... `janus/package-lock.json` seems to be missing.

Oh, thanks! Perhaps I forgot to delete and get the symlink on my end, will check and verify next week.

PR #1022, opened 2022-09-09
consolidate to single action

coleshaw:
Apparently cannot downgrade to v1 of the checkout action for a workflow to trigger another workflow, and per [the docs](https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#triggering-a-workflow-from-a-workflow), you basically can't do this without a PAT. Since the recommendation is for PATs to expire within a year (you can generate forever PATs, but not recommended), this PR just consolidates the staging image build and push to happen on merge into master, since that mimics the desired behavior...

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -11,6 +11,11 @@ concurrency:
 jobs:
   push-staging:
     runs-on: ubuntu-latest
+    env:
+      IS_CI: 1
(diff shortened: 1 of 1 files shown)


coleshaw:
One potential side-benefit of doing it this way (I think), is that the staging image build and push will be much faster, since the images should already exist in the Action runner from building the master images...? :crossed_fingers:

PR #1024, opened 2022-09-09
try pruning to save disk space

coleshaw:
Previous PR ran out of disk space ... seeing if pruning helps.


coleshaw:
If this doesn't work, I think I'll just have to use a PAT....

PR #1028, opened 2022-09-09
Metis copy from magma buckets [close #997]

### LINKED ISSUE #997: Metis: Make Magma bucket copyable ?
This would greatly help things like the GNE ETL, or possibly also support a Magma view inside of Metis.

One thing that we can't handle well is that Magma does not clean up its bucket. Generally should be okay, but there might be cruft (either someone deleted a file link, or reduced the number of items in a file_collection), so you would have to always fetch records and file / file_collection attributes from Magma first, and then run any sort of copy command. Do not copy blindly.

Perhaps related or duplicate of #894 ?

coleshaw:
This PR updates Metis's `file_controller#bulk_copy` endpoint, to allow the following behavior:

* Copy files out of the `magma` bucket into somewhere else in the same project.
* Copy files out of the `magma` bucket into some other project, where you have permissions.
* Use of the Timur UI to upload new files.
* Use of the Timur UI to link files via the Metis path.

This should **not** allow the following behavior:

* Copy files **into** the `magma` bucket for the same project.
* Copy files **out of** a different project's `magma` bucket, if you do not have permissions for that project.

I have tested this manually and it seems to behave as expected. Also includes some new specs to verify.

CI seems to be dying due to some issue between GH and the Docker registry ... so will try kicking that again later / next week?

### DIFF
--- a/metis/lib/models/bucket.rb
+++ b/metis/lib/models/bucket.rb
@@ -15,6 +15,8 @@ class Metis
       guest: 3
     }
 
+    READ_ONLY_BUCKETS = ['magma']
+
     def self.valid_bucket_name?(bucket_name)
       !!(bucket_name =~ /\A\w+\z/)
     end
--- a/etna/lib/etna/clients/metis/workflows/sync_metis_data_workflow.rb
+++ b/etna/lib/etna/clients/metis/workflows/sync_metis_data_workflow.rb
@@ -22,26 +22,27 @@ module Etna
         end
 
         def copy_file(dest:, url:, stub: false)
-          # This does not work due to the magma bucket's restrictions, but if it did work, it'd be super sweet.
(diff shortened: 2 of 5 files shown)


metis/lib/models/bucket.rb
```diff
@@ -15,6 +15,8 @@ class Bucket < Sequel::Model
       guest: 3
     }
 
+    READ_ONLY_BUCKETS = ['magma']
```
coleshaw, line 18: I will refactor this a bit next week to put the "is this a read-only bucket" into the `Metis::Bucket` class, so it's cleaner and should then automatically extend to any other type of "app" bucket, too. i.e. `vulcan` owned `vulcan` bucket, if we want to manage the cache in Metis.

PR #1029, opened 2022-09-12
do not allow user to remove all where clauses [close #1004]

### LINKED ISSUE #1004: Timur Query: app crashes when closing out the one remaining "and" clause in a filter
As seen by Bushra today, it seems like if you "close" the one and only "and" clause in a filter, the app crashes and you have to reload the page.

coleshaw:
This PR removes the "remove clause" icon from the Where filters in Timur Query, when there is one and only one clause in a given filter. As noted in #1004, previously, if a user clicked the "x" and tried removing the last clause, the app would crash with a white-screen.

### DIFF
--- a/timur/lib/client/jsx/components/query/query_slice_control.tsx
+++ b/timur/lib/client/jsx/components/query/query_slice_control.tsx
@@ -39,7 +39,7 @@ const QuerySliceControl = ({
           modelNames={modelNames}
           isColumnFilter={true}
           patchClause={handlePatchClause}
-          removeClause={removeSlice}
+          removeClause={() => {}}
           showRemoveIcon={false}
         />
       </Grid>
--- a/timur/lib/client/jsx/components/query/query_slice_control.tsx
+++ b/timur/lib/client/jsx/components/query/query_slice_control.tsx
@@ -39,12 +39,12 @@ const QuerySliceControl = ({
           modelNames={modelNames}
           isColumnFilter={true}
           patchClause={handlePatchClause}
-          removeClause={() => {}}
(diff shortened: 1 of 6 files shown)


timur/lib/client/jsx/components/query/query_slice_control.tsx
```diff
@@ -40,6 +40,7 @@ const QuerySliceControl = ({
           isColumnFilter={true}
           patchClause={handlePatchClause}
           removeClause={() => {}}
+          canRemove={false}
```
dtm2451, line 43: Seems I don't have my dev environment set up properly to be able to test directly, but the idea makes sense to me and the code seems right, aside from maybe here... should this be *true instead?

dtm2451 (changes requested):

timur/lib/client/jsx/components/query/query_slice_control.tsx
coleshaw: Note that this funky extra `RemoveIcon` in the wrapping `QuerySliceControl` component is for UI alignment of headings ...

coleshaw: This is what the first clause in a filter will look like:

Screenshot from 2022-09-12 14-31-13


And when you have more than one:

Screenshot from 2022-09-12 14-31-19

dtm2451: Gotcha!

dtm2451 (approved):
Still can't test directly, but in reading through the code changes, all seems good to me!

PR #1035, opened 2022-09-15
Timur query add subclauses; [close #1009]

### LINKED ISSUE #1009: Timur Query: Allow additional "AND" conditionals inside of each Where Filter
In working through example queries with Bushra, I've realized that the UI is still limiting and needs some tweaking to accomodate all queries.

For example, she wants to query `rna_seq` compartments for samples that have `live`, `myeloid`, and `tcell` compartments, but where all have EHK > 7.

For now, you can simulate this by adding a filter like:

```
Sample -> any rna_seq where compartment == live
                  any rna_seq where compartment == myeloid
                  any rna_seq where compartment == tcell
                  every rna_seq where EHK > 7
```

And it will return results that kind of look accurate.

But, this will fail if we ever have a sample with additional compartments with EHK <= 7, for example, it would fail with:

```
Sample -> rna_seq.live, EHK = 8
                  rna_seq.myeloid, EHK = 9
                  rna_seq.tcell, EHK = 10
                  rna_seq.treg, EHK = 7
```

Would result in the entire sample + three rna_seq compartments not being returned.

Ideally, the UI would let you construct a filter like:

```
Sample -> any rna_seq where compartment == live AND EHK > 7
                  any rna_seq where compartment == myeloid AND EHK > 7
                  any rna_seq where compartment == tcell AND EHK > 7
```

I believe the Query API would already allow this, it's just a matter of updating the UI widget and query builder.

coleshaw:
This PR adds in the ability to Timur Query UI to add "subclauses" to each where filter clause (refer to #1009). It looks like:

Screenshot from 2022-09-15 09-44-51

Functionally, it means there can be additional `::and` statements inside of a subquery condition (an `::any` or `::every` filter). And means you can do queries like:

"Give me all the samples that have (any rna_seq record with EHK > 7 AND compartment == live) AND (any rna_seq record with EHK > 7 AND compartment == myeloid)" to ensure that samples have good quality **live and myeloid** compartments, but disregard the other compartments.

Note that the backend API already supported this functionality, it just wasn't exposed in the UI, so this PR only contains UI changes.

Also, note that this PR changes the internal filter JSON structure that is sync'd to the URL, so the PR includes a migration path for old, saved query URLs. You can test this by opening up an old query on dev, which should still populate the form correctly.

You can test the subclause query part locally by setting up something like the following:

Screenshot from 2022-09-15 10-02-10

Then, if we want to see what rna_seq records come from samples with good quality **live and myeloid** compartments (disregarding other compartments), we can try constructing a query.

The old query UI would have let you construct something like this, and return 0 results (due to the `every EHK score >= 8` filter hitting the `tcell` compartment and saying the sample `IPIADR001.T1` doesn't meet the requirements):

Screenshot 2022-09-15 at 10-03-24 Dev Timur

With subclauses, we would modify the query and get our expected result:

Screenshot 2022-09-15 at 10-03-52 Dev Timur


coleshaw:
Hm, so I wound up upgrading to `CodeMirror6` instead of 5 (I think some underlying dependencies changed), and thus having to also update the View Editor and Manifest Editor.

I'm pretty sure the View Editor behaves as it did before, perhaps with some improvements, like code folding and better "find" within the editor.

I'm pretty sure the Manifest Editor does not do syntax highlighting like it did before, since it's a custom language, though I think it works as an editor ... since this is rarely used, perhaps that is acceptable?

graft (approved):
There are a few other uses of CodeMirror e.g. in polyphemus and the RevisionHistory pane, possibly on different versions, using a Json schema validator. Dunno if those will break... The manifests seem ok to ignore

coleshaw:
Ah, that's good to know. That means probably in Vulcan, too. For now, they are different `node_modules` folders, so probably okay ... if anyone can ever finish #1019, then they would have to be migrated to 6, too. But I think if Revision History uses the standard JSON / Javascript editing features, it would be fairly straightforward to take what's in the View Editor and use that code.

coleshaw:
Just some context for why I upgraded to CodeMirror6:

https://github.com/sachinraja/rodemirror/issues/139

Basically running into the same issue. The solution was to pin to lower versions of the plugins (which didn't seem to work for me), or just upgrading to 6 ...

PR #1043, opened 2022-09-19
Metis issue 1021; [close #1021]

### LINKED ISSUE #1021: Metis: Add a "Copy from" widget to copy a file to a folder given a metis path
Would be nice to have some sort of "copy file from this location to that location" in the UI, feature

coleshaw:
This PR adds a new "paste from Metis path" icon to the Metis UI, that lets a user put in a Metis path from the same project, to copy into the given folder. Just a convenience method, that would be useful for these little IPI tweaks and renames. I've overlayed the `paste` icon on top of `cloud` (to be similar to the Timur UI File icon that we used, which was just `cloud`), but happy to change that if something else will make more sense.

Screenshot from 2022-09-19 16-15-15
Screenshot from 2022-09-19 16-15-46

Wound up updating `eslint` in Metis, which necessitated a bunch of other file updates, too, so most of this PR is linting.


metis/lib/client/jsx/components/folder-view.jsx
```diff
@@ -1,137 +1,197 @@
...
-              ref={ (input) => this.uploadDirInput = input }
-              onChange={ this.dirSelected.bind(this) }
-            />
-          </ControlBar>
-        </div>
-        <div className='listing-group'>
-          <ListHead columns={ COLUMNS }/>
-          <ListBody widths={ COLUMN_WIDTHS } folder_name={folder_name} bucket_name={bucket_name}/>
-        </div>
+    if (metis_path) {
```
coleshaw, line 113: This is the meat of the change

metis/lib/client/jsx/actions/file_actions.jsx
```diff
@@ -123,4 +124,14 @@ export const touchFile = ({bucket_name, file}) => (dispatch) => {
     .catch(
       errorMessage(dispatch, 'warning', 'File touching failed', error => error)
     );
-}
+};
+
+export const copyFile = ({metis_path, dest_metis_path}) => (dispatch) => {
```
coleshaw, line 129: Along with this new action.

PR #1046, opened 2022-09-20
Polyphemus issue 1044

coleshaw:
This removes the hard-coded model / attributes and filters from the Ruby ETL code, and instead accepts those data structures from files, which we can inject via Airflow.

Sets the default dataset to basically nothing, except a listing of `comet_plus` patient records with only the `name` and `comet_plus` attributes listed.

To test this locally, without the injected files, you can do the standard Ruby ETL command in a Polyphemus bash environment:

```
-> ./bin/polyphemus etl materialize_gne_magma_records_etl run
```

Assuming your MVIR1 project is populated somewhat.

You can also create the two files:

* /model_attributes
* /model_filters

to see the Airflow behavior, or in Airflow, you can create an ETL and inject the files, like:

```
    run_on_docker(
        'polyphemus_etl_run_materialize_gne',
        'polyphemus_app',
        ["/app/bin/polyphemus", "etl", "materialize_gne_magma_records_etl", "run"],
        env=dict(
            ETNA__PRODUCTION__MAGMA__ENDPOINT_FILE="/magma_host",
            ETNA__PRODUCTION__METIS__ENDPOINT_FILE="/metis_host",
        ),
        include_external_networks=True,
    ).accepts("/magma_host", "/metis_host", "/model_attributes", "/model_filters")(
        prepare_magma_config(), prepare_metis_config(), prepare_model_attributes(), prepare_model_filters()
    )
```


coleshaw:
Also fixes a bug with the copy code, where the parent folder for the destination may not exist and needs to be created first.

graft:
How will we inject the files?

coleshaw:
This set of Airflow statements (`accepts`, and the subsequent function call) injects files into the containers, as is done with other ETLs:

```
    run_on_docker(
        'polyphemus_etl_run_materialize_gne',
        'polyphemus_app',
        ["/app/bin/polyphemus", "etl", "materialize_gne_magma_records_etl", "run"],
        env=dict(
            ETNA__PRODUCTION__MAGMA__ENDPOINT_FILE="/magma_host",
            ETNA__PRODUCTION__METIS__ENDPOINT_FILE="/metis_host",
        ),
        include_external_networks=True,
    ).accepts("/magma_host", "/metis_host", "/model_attributes", "/model_filters")(
        prepare_magma_config(), prepare_metis_config(), prepare_model_attributes(), prepare_model_filters()
    )
```

coleshaw:
So you can have the method `prepare_model_attribute()` return a dict, for example, with the relevant information, and that will appear in the container's filesystem. You can see what this looks like in the `polyphemus_jobs.py` ETL code (I've already put it in, for when we enable the materialize ETL).

coleshaw:
Probably better is to set the file names to be configurable, too, and inject those via the environment...hm...

graft (approved):

coleshaw:
So to clarify, while testing I figured out the Airflow injection is actually the contents as an XCom value, not the name of a file. So tweaked the code to accomodate that. Basically you inject values into the application's `config`.

PR #1052, opened 2022-09-21
Magma return nil for unknown matrix values; [close #1048]

### LINKED ISSUE #1048: Magma: matrix slicing does not work through multiple models
It seems like Matrix slicing in Magma for query does not work when traversing down multiple sets of models.

If you execute a query with say, `rna_seq` as the root model, and add a slice like:

```
["gene_tpm", "::slice", ["gene_id']]
```

This works. The matrix `::slice` verb returns a table + identity that match up:

```
{:ilsbyxqrxr_tube_name=>"MVIR1-HS1-D0PBMC1-RSQ1"}
ilsbyxqrxr_tube_name
```

However, if you try traversing through multiple models, this does not work.

For example, using `patient` as a root model, you would have to specify (for MVIR1) that your slice includes `timepoint`, i.e. `timepoint where day==0`, and then a slice. This seems to lead to `patient` being the Matrix table used, with joins incorrectly done agains timepoint and rna_seq, such that the `::slice` verb no longer gets the right information. Here are the table + identity returned in this case:

```
{:irremsgxqm_name=>"MVIR1-HS1", :shvmywpehi_name=>nil, :kaeublonpm_tube_name=>nil}
kaeublonpm_tube_name
```

Because the starting model is `patient`, no data is joined from the sub-tables, as expected, and a `nil` identifier is provided. Seems like a bug somewhere.

coleshaw:
Did more digging for issue #1048. The problem is basically that you can construct queries such that you require filters on columns while also doing matrix slices, i.e., if you are using `patient` as a root model, and it has a collection of `timepoint` records, and each `timepoint` has a collection of `rna_seq`, to get gene expression data into a flat data frame, you would have to add timepoint filters in the predicates in addition to rna_seq slices:

```
["timepoint",["day","::=",0],"::first","rna_seq",["tube_name","::matches","PBMC"],"::first","gene_tpm","::slice",["ENSG00000159403","ENSG00000173372"]]
```

Currently, for any record that does not match the filter requirements, `nil` is returned to the [matrix.rb predicate](https://github.com/mountetna/monoetna/blob/master/magma/lib/magma/query/predicate/matrix.rb#L21), and you wind up with the situation described in the issue, where the final model identifier is `nil` because no intermediate model meets the filter criteria:

```
(table) {:irremsgxqm_name=>"MVIR1-HS1", :shvmywpehi_name=>nil, :kaeublonpm_tube_name=>nil}
(identity) kaeublonpm_tube_name
```

This becomes unfetchable when serializing the answer, resulting in a `MatrixJsonError` being thrown.

This PR changes the behavior to populate the matrix attribute values with `nil` instead, since it seems plausible to expect a query with column filters + slices to have no qualifying data ...

I will also note, that I think we want to steer folks away from constructing queries in this fashion. Ideally they should construct queries that **do not** require slicing + filtering in the columns, and instead construct the queries "bottom up" so that only slicing is required. I would also be open to the idea of not allowing this type of query in the Timur UI, though unclear if we would want to remove the functionality from the API itself...thoughts?


graft:
Could you exemplify "bottom up" queries?

coleshaw:
Sure! A top-down query where you would need to filter + slice in the columns would look like:

```
["patient",["covid_status","::matches","Confirmed"],"::all",[["admission_level"],["oxygen_homesupport"],["consent"],["covid_status"],["deceased"],["hosp_los"],["vent_duration"],["timepoint",["day","::=",0],"::first","day"],["timepoint",["day","::=",0],"::first","rna_seq",["tube_name","::matches","PBMC"],"::first","gene_counts","::slice",["ENSG00000159403","ENSG00000173372"]]]]
```

I believe a mostly equivalent (but better) bottom-up query would look like:

```
["rna_seq",["::and",["timepoint",["patient",["covid_status","::matches","Confirmed"],"::any"],"::any"],["timepoint",["day","::=",0],"::any"],["::has","gene_counts"]],"::all",[["timepoint","patient","name"],["timepoint","patient","admission_level"],["timepoint","patient","oxygen_homesupport"],["timepoint","patient","consent"],["timepoint","patient","covid_status"],["timepoint","patient","deceased"],["timepoint","patient","hosp_los"],["timepoint","patient","vent_duration"],["timepoint","day"],["gene_counts","::slice",["ENSG00000159403","ENSG00000173372"]]]]
```

I say "mostly" because the bottom query will return all `rna_seq` records, whereas the first one has to do slicing (`["tube_name","::matches","PBMC"]`) and doesn't guarantee that you get all `rna_seq` records anyways (because there are patients with multiple PBMC `rna_seq` records on D0) -- so it is super difficult to guarantee you get all the data while also getting a flat data frame in the top-down case.

graft:
Aha - I think those "top-down" style queries are avoidable in some circumstances, but I can imagine them turning up if you wanted to do something like, compare expression for some genes for multiple samples from the same patient, for example, or compare expression and some other patient-level variable. Perhaps all of these can be rotated into a "bottom-up" query? Not clear...

coleshaw:
I see, that makes sense. In which case, I think it's important that a fix like this gets applied, otherwise the top-down queries will error out when the filter returns no results for a given root model record...

PR #1057, opened 2022-09-27
fix matrix retrieval and query

graft:
Paginates queries, and retrievals for matrix attributes.


coleshaw (approved):
Great job digging into this! Looks good to me.

PR #1060, opened 2022-09-29
add gnomon_db service

coleshaw:
This is a small PR to add the definition for `gnomon_db` into the `docker-compose.yml`. Without it, you cannot run `make -C gnomon update` to install NPM dependencies. This is due to how the Makefile build process works, where on `update` it always tries to bring up the `_db` service, so one much exist. Since they are double-colon targets, I'm not sure there is a way around that.

Post-merge of #1019, it **may** be possible to rethink the build process and remove this requirement, since then all Ruby and NPM packages would be installed exclusively in the `etna` image. However, the other apps will (perhaps) need to run DB migrations on `update`... so may still need a `gnomon_db` service.

### DIFF
--- a/gnomon/docker-compose.yml
+++ b/gnomon/docker-compose.yml
@@ -13,7 +13,7 @@ volumes:
     driver: local
   gnomon-node-modules:
     driver: local
-      
+
 x-base-etna-service:
   &base-etna-service # Base configuration used by most etna services.
   image: etna-base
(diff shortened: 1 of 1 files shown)


graft:
Hmm, but didn't we axe gnomon_db because we are putting tables into magma?

coleshaw:
> Hmm, but didn't we axe gnomon_db because we are putting tables into magma?

Yes, that's true, and I think we can follow that in making our production swarm stack. However, in dev environments, I think it's required due to our build process -- if you try something like `make update`, it will fail because it tries to run `update` -> `update-ready` -> `$ docker-compose up -d gnomon_db` which will fail unless that service (`gnomon-db`) is defined in `gnomon/docker-compose.yml`.

So we would need to do some adjustments of the build process, in order to be able to run `make update` without defining a `gnomon_db` service ...

PR #1064, opened 2022-10-04
Gnomon enforce grammar

coleshaw:
:shrug: 

This includes some enforcement of various grammar rules, and some specs / examples of how you could use the rule regexes.


coleshaw:
K, this should be a good diff now.

Will add some of the enforcements on top of this.

PR #1070, opened 2022-10-09
Gnomon decompose API

graft:
A start on the decompose API - this does the basic thing, but still lacks the reporting on identifier creation, in the Gnomon identifier tables and in the library tables. Also the front end is not wired up yet to use this.


magma/lib/magma/gnomon.rb
```diff
@@ -0,0 +1,6 @@
+require_relative './gnomon/rule_errors'
+require_relative './gnomon/rule_definition'
```
coleshaw, line 2: Nice, thanks for separating these out more cleanly!

coleshaw (approved):
Nice! I started a "counter" API that builds off of this, but haven't been able to make much progress...

graft:
Ok, bunch of fixes, I think it should be ready to go, barring a few tests.

This is maybe a bit over the top:
create

dtm2451:
> This is maybe a bit over the top

bahahaha, but I love it!

coleshaw:
That is psychadelic!

coleshaw:
Can you make it generate fireworks and sparks / smoke on click?

PR #1089, opened 2022-10-13
try building etna app first

coleshaw:
Will only see an effect after merging into master ...


coleshaw:
Thanks for the suggestion to try this, @graft , @dtm2451 ! Seems much better than without pre-building `etna` (~42 minutes vs 1hr), though not as fast as I had hoped for ...

PR #1095, opened 2022-10-14
Metis nuke from file

graft:
Computes checksums to a file before nuking; also rescues some crashes.


metis/bin/metis_client
```diff
@@ -1222,45 +1222,135 @@ EOT
...
 
-  class Nuke < Command
-    usage "nuke <local_directory>"
+  class Checksum < Command
+    usage "checksum <local_directory>"
+
+    def parse(args)
+      options[:file] = "md5checksum.txt"
+      parse_opts(args) do |opts|
+        opts.on("-f", "--file FILE", "Recursive folder removal") do |file|
```
coleshaw, line 1231: Is this option description accurate? Or from a copy / paste of `rm`?

metis/bin/metis_client
```diff
@@ -1222,45 +1222,135 @@ EOT
...
     def run(local_directory)
       # collect all files and folders
       folders, files = Dir[::File.join(local_directory, "**", "*")].partition { |f| ::File.directory?(f) }
 
       puts "Found #{files.count} files in #{folders.count} folders, total size #{byte_format(files.map{|f| ::File.size(f)}.sum)}."
 
-      deleted_files = deleted_folders = 0
-      remaining_size = 0
+      md5files = files.select do |file|
+        ::File.basename(file) == options[:file]
```
coleshaw, line 1260: Will this work if the user does not specify a `-f` option when invoking the command?

metis/bin/metis_client
```diff
@@ -1222,45 +1222,135 @@ EOT
...
+        load_md5_cache(md5file)
+      end
+
+      # compute md5sums
+      files.each.with_index do |file,i|
+        file = Pathname.new(file).cleanpath.to_s
+
+        print "Computing md5s - #{progress_bar(i.to_f / files.length)} - #{i} of #{files.length}      \r"
+
+        next if ::File.basename(file) == options[:file] || md5_cache[file]
```
coleshaw, line 1273: What if the file `md5` has changed since it was last cached in `md5checksum.txt` -- should that recalculate the value? Probably defeats the purpose of caching...

metis/bin/metis_client
coleshaw: Oh, nevermind, I see what you're doing! ... assumes you have the same "md5 checksum filename" in some subdirectories. What's not quite clear from the option description...

metis/bin/metis_client
graft: Yeah we have to assume the cache is accurate for this to work...

PR #1099, opened 2022-10-19
fix behavior of column labels and model / attribute changes to be mor…

### LINKED ISSUE #1096: Timur: Update query columns labels when attribute changes
If same as the default original one, should update so users aren't confused, or only keep if it's not blank.

coleshaw:
…e intuitive

Closes #1096.

If the attribute in a column changes, and the label is the previous default, then it updates to reflect the new attribute name. If the label has been changed manually, then the label will not change.

i.e.

"patient.notes" => "patient.processor"
"custom column name" => "custom column name"

If the model changes, the label resets to empty string (previously it stayed the same ... which doesn't quite make sense, either).


coleshaw:
And thanks @dtm for the suggestion on Monday for an approach to improving the UX of this interface!

PR #1104, opened 2022-10-21
Timur add attribute model

coleshaw:
This PR adds a "Add Attribute" button to Timur map, so administrators can start editing models in the UI (non-admins do not see the button). For now, user can just add the `name`, `type`, `description`, and `group`, since I think those may be the most commonly used parameters?

Screenshot from 2022-10-21 10-14-05

Could also add all (or more) options, if we think they might be useful. Validation might be one that is also useful. I figured an "edit attribute" widget would present all the attribute params, so folks could hide, mark read-only, mark restricted, etc.


dtm2451 (approved):
Fantastic idea!!  And I think I'm also behind the idea of giving additional options in the (future) edit feature while keeping this one cleaner.

Agree that validation is a primary other piece that would get used in attribute editing, but also that having it 'hidden' for UX in the initial add path seems right, though we might want to expose it here in the future?  Since the `add attribute` button is only visible to admins, I think that would be mostly people who understand the regex utilization of validations?  But I can also imagine other features which would be nice to have to the side of a gui-based validation editor: percentage of current values that pass the validation, + perhaps a way to see values that fail!

Tested an add + that trying to add an attribute with same name as an existing attribute failed with an understandable error!

graft:
This is awesome and will be loved by all! However I have a serious concern that adding "add" without "remove" or even "hide" will cause junk attributes to pile up (or removal requests)

coleshaw:
Hm, well I was trying to break things into manageable chunks. I can certainly build off of this branch to add in a hide or remove feature before merging, though remove will take more work (since it requires backend updates).

graft:
Def agree with adding things in manageable chunks - I think the "safe" order would be:
1 - Edit/hide non-link attributes
2 - Remove non-link attributes
3 - Add attributes
4 - Add/remove links
5 - Add/remove models

coleshaw:
Hm, what about `reparent_model`, @graft ?

coleshaw:
Hm, well I think it would behave like the other "remove" actions, where the column gets renamed / backed up (i.e. `patient_id` => `patient_id_<timestamp>_backup`, the foreign key constraint gets removed, and you would have a new column i.e. `biospecimen_id`. So you could theoretically re-construct the relationship if you ever needed to, for records that pointed to a patient before you reparented.

Hm, I guess re-attaching to the new model would be an entirely manual exercise, all the records would become detached from the graph. There would be no technical limitation though, I think.

coleshaw:
@graft , I would also refer you to [this thread and your previous comments](https://github.com/mountetna/monoetna/pull/843#issuecomment-1132297459) on the feature.

coleshaw:
Because if you remove a model ... you have no way to every find it again, without mucking around in the database? With reparent at least you could say, insert a new `timepoint` or `biospecimen` model and keep the graph intact. Though you would have to manually reparent records.

coleshaw:
Per comment at standup, let's allow reparent_model if the model has no data (good for designing a project)

coleshaw:
New task: on new project creation, should copy over the coproject_template structure

dtm2451:
Re-read all this to get current on this discussion.  On thing I'd add is for the new task...

> New task: on new project creation, should copy over the coproject_template structure. #1116

I'd perhaps have a ui-point for selecting what project to base the models off of.  It should default to coprojects_template, but I think there may be times where basing off of a different project, perhaps like XNEO1, might reduce the tweaking needed to get to the final state.

PR #1107, opened 2022-10-21
show ids matching current identifier selections + add create button

dtm2451:
Goal is providing feedback about names / records that match a user's current selections on the compose-identifier page.


dtm2451:
Working, but very much WIP:
- Styling leaves a lot to be desired
- I have it running off of magma records via a /retireve identifier call, but it should work off the gnomon db instead.

dtm2451:
- [x] Runs off gnomon
- [x] restyled

I've also started adding display of upstream ids once the compose selections are all filled in!  WIP, ToDos:
- [ ] Disable create button when name already exists
- [ ] Error reporting for create button responses
- [ ] Create-all button?

Current look:  (Note: project, subject, and biospecimen names were all 'created' from this page!)
image

dtm2451:
Updated plan:
- [x] single 'create' button.  (With work of figuring out upstream rule's ids that should also be created falling to the ruby-side, with that already having been worked into the 'graft/gnomon-decompose branch'!)

magma/lib/magma/gnomon/grammar.rb
```diff
@@ -109,6 +109,7 @@ def decompose(identifier)
 
         {
           tokens: tokens,
+          rule_name: rule.name,
```
dtm2451, line 112: Just noting this line for @graft.  There was a new merge conflict here this morning when I went to merge and I had to add the `rule.` portion to this line to make it work with the updated methodology.  I'm pretty that I got it correct because the target rule is still prioritized in <idTreeTable>s, but still wanted to get your eyes on it!

PR #1111, opened 2022-10-25
Timur edit attribute

coleshaw:
This PR adds an "Edit" button to the attribute report, and gives you a fuller set of options to change, and also removes the stuff you can't change yet (like the `type`):

* attribute_name
* description
* group
* validation
* display_name
* hidden
* read_only
* restricted

If you change the name, Timur will fire a `rename_attribute` action along side the `update_attribute`. 

Only admins can see the button.

Merging this into the "add attribute" branch, so they can all go into the main branch together.

Note that at some point after "remove attribute" is implemented, it may also be nice to put in a way to view hidden attributes in the model report, so that there is a way to see / edit hidden attributes.


dtm2451 (changes requested):
Lookin pretty good!  Tested all the options and all but one seem to work.
1) The validation field seems un-editable, at least in this one (likely common) use-case.  Perhaps the implementation doesn't work for creating a new validation?  I'm trying to edit a freshly created attribute, so with whatever the likely blank default value for the field would be, and I can't seem to change the value from `null`.
2) Stylistically, perhaps the "Edit Attribute" text at the top can be styled differently somehow to look more like a title?  It blends in with the rest of the widget to me.

coleshaw:
> The validation field seems un-editable, at least in this one (likely common) use-case. Perhaps the implementation doesn't work for creating a new validation? I'm trying to edit a freshly created attribute with nothing originally in the field and I can't seem to change the value from null.

Hm, thanks for finding this! I must have broken something in the JSON translation, because that worked at one point for me, when I treated the input as just a raw string.

coleshaw:
> Stylistically, perhaps the "Edit Attribute" text at the top can be styled differently somehow to look more like a title? It blends in with the rest of the widget to me.

Not sure what you mean, to look like a title -- it should look like a button? Or is the CSS not getting through correctly, so it's just text on your end?

Screenshot from 2022-10-26 11-59-57

dtm2451:
> > Stylistically, perhaps the "Edit Attribute" text at the top can be styled differently somehow to look more like a title? It blends in with the rest of the widget to me.
> 
> Not sure what you mean, to look like a title -- it should look like a button? Or is the CSS not getting through correctly, so it's just text on your end?
> 
> Screenshot from 2022-10-26 11-59-57

Not that button; that's fine. The widget title after I click the button:
image

coleshaw:
Ah, I see. Okay thanks for clarifying!

graft:
A huge win for us, I hope, this will make updating the models much more accessible and make a lot of the model features more usable (like attribute groups, validations and hiding).

A few UI notes:
- I might consider adding a mask to the inputs for the name in the Edit Attribute and the Add Attribute dialogs, so that any non-snake-case characters are excluded; the validation catches it anyway but it might be nicer to make the UI prevent these kind of simple mistakes. Similarly for the comma-separated list of attribute_groups, and eventually validations (where a "type" dropdown and matching component might be best).
- Just a weirdness, once "hidden" is set you have one chance to undo what you have done before clicking away from the attribute or reloading the map, subsequently the UI won't let you get back to the Edit button. Hopefully resolved in the future by a "hidden attribute" toggle somewhere.

coleshaw:
> Just a weirdness, once "hidden" is set you have one chance to undo what you have done before clicking away from the attribute or reloading the map, subsequently the UI won't let you get back to the Edit button. Hopefully resolved in the future by a "hidden attribute" toggle somewhere.

Yes! That's what I was referring to in the original PR comment:

> Note that at some point after "remove attribute" is implemented, it may also be nice to put in a way to view hidden attributes in the model report, so that there is a way to see / edit hidden attributes.

coleshaw:
So I've updated the following things:

* Changed `validation` to a dropdown with types (`Array`, and `Regexp`), which reveals a second string box to type in a comma-separated list (for Array) or regex string.
* Changed dialog header to be bolder and have larger font, so hopefully stand out more as a header.
* Added error-checking for `name` (to be `snake_case`) and `group` / `validation` (Array) (to be `comma,separated`).

coleshaw:
So I've updated the following things:

* Changed `validation` to a dropdown with types (`Array`, and `Regexp`), which reveals a second string box to type in a comma-separated list (for Array) or regex string.
* Changed dialog header to be bolder and have larger font, so hopefully stand out more as a header.
* Added error-checking for `name` (to be `snake_case`) and `group` / `validation` (Array) (to be `comma,separated`).

Screenshot from 2022-10-27 14-50-34

PR #1112, opened 2022-10-26
with migration file

coleshaw:
Oops, forgot to commit the `db` migration files for #1110  :facepalm:


coleshaw:
Note that the default value will set the timestamps for all existing projects and users to whenever the migration is run, but then the `timestamps` plugin overrides that for any new entries.

PR #1118, opened 2022-11-09
Magma timur add remove model

coleshaw:
This PR implements `add_model`, `remove_model`, and `reparent_model`.

One niggling thing about the UX I don't quite like, is that you don't have feedback on the screen that a model (or its children) has records and you cannot reparent it ... the server will reject if you try, but it would be nice if the UI would just hide the button. The reason that the user themselves cannot do the check (via a `::count` query API call), is that the current `reparent_model` logic does not let you reparent if there are restricted records in the model tree, either. These are records that the user may not be allowed to see, but that reside in the model or a child model -- in this case, I think the right behavior is to not allow `reparent_model`, but the user may not be able to see why. Perhaps sligtly confusing, though if we just use it to construct project trees before data is populated, much less of an issue.

It should be merged before #1115.

Please thoroughly test :sweat_smile:


dtm2451 (changes requested):
Thanks for implementing this Cole!

I do have a few questions/suggestions/gripes:
- Reparet Model:
  - I was wondering why the Reparent Model button remained for models that I knew had data, so I'm glad to have your explanation there.  Perhaps there could be a different server api interaction here that pokes deeper but runs no risk of revealing the possibly restricted data to the user?  Something where, regardless of the user's permissions, a proper determination of 'is there any data in the model' can be made?  I'm not sure of the long-term plan though... if we plan to keep the button working even when some data does exist, the knowledge of whether any records are hidden specifically to this user could also cause the button to be disabled.
    - I think this ping could be run either 1) upon coming to the page -- returned as an array of the viable model_names, or 2) upon clicking on a model -- returned as a logical and the button only shows in the proper case.
- Add Model ui:
  - I think the "child model name" piece should come first as it serves as the title for the new model, and also perhaps having "child" will be confusing because this window feels like it refers to that model?
  - Perhaps the "Identifier" label could say "Identifier attribute name" to be more clear about what it represents?
  - The "Link Type" dropdown... I think the options are confusing and I'm not sure how best to fix that here because I'm not sure what the purpose of ever choosing "child" would be?  Is there a benefit to restricting linkage to be 1:1?  But the real problem to me is just that I wasn't expecting to need to make this type of decision.  Could it instead become a "Model-type" dropdown (or switch) with options "standard model" vs "table".
  - (Least committed to this comment) there could also be an info line showing (or dropdown for changing) the parent model.

coleshaw:
Thanks for the feedback!

> I was wondering why the Reparent Model button remained for models that I knew had data, so I'm glad to have your explanation there. Perhaps there could be a different server api interaction here that pokes deeper but runs no risk of revealing the possibly restricted data to the user? Something where, regardless of the user's permissions, a proper determination of 'is there any data in the model' can be made? I'm not sure of the long-term plan though... if we plan to keep the button working even when some data does exist, the knowledge of whether any records are hidden specifically to this user could also cause the button to be disabled. 

Hm, I think long-term, we decided not to allow reparent for models with data, and it would only be used (basically) at the start of a project when constructing the models? Maybe I can investigate some query param, like `::total_count` instead of just `::count`, that could be used in a ping.

Good suggestions for the Add Model UI! For the Link Type, I think @graft  and I had discussions about `child` at one point ... I forget the conclusions, but I think it was for a super obscure use-case, that maybe we never have hit in a real project? I like the other wording -- I can take out `child` and see if we can remember why it's there, haha.

> there could also be an info line showing (or dropdown for changing) the parent model.

I can definitely add an info line. If we wanted to give the ability to change the parent model, is that confusing from a UX standpoint, if they already selected a model from the tree, but then are given a choice to change it? Makes me feel like the "Add model" button should then be somewhere on the top-nav (a general set of actions), not in a specific model's set of buttons, though it does seem more convenient so that you don't have to pick a model in the graphic, first.

graft:
This is, simply, huge. The UI is nice and simple, and the instant visual feedback is pretty cool as well.

Couple of things I noticed:
- the Remove Model butten just says "Model" rather than "Remove Model"
- When you "Add Model", if you specify an identifier but choose "table" type, it silently ignores the "identifier" argument. It might be better if this were disabled appropriately when you chose the link type (and the UI order reflected this). I agree we can hide "child" type for now.
- Ibid that Reparent Model is a bit confusing, since the error turns up late in the game. It might be simple to check if there is at least one connected record for the model with a query, but I think the restricted data question makes this hard (i.e., the user can't legitimately query for this data right now). We could:
  - Disable the button based on a query anyway (taking care of the case where there is visible data)
  - Let them know in the error that the data is invisible due to restriction if the reparent fails for this reason. This way at least they won't be going nuts trying to understand why they can't reparent.

coleshaw:
K, so I've updated the following:

* Moved the "Add model" button to be a top-level thing, and it includes a parent model selector.
* Reworded various add-model dialog input labels, per feedback above.
* Hides "reparent_model" button, based on `::count`, and updates the error message if there are restricted records.

One side-effect due to how I implemented the reparent_model with `::count` is that clicking on a model now shows the # of records on the top-right, which you used to only get when you click "Count" in the drop-down menu.

Screenshot from 2022-11-10 14-13-39

I did this to cache the count value so we don't make a lot of queries to the API, if someone clicks around the model tree a lot. However, if we don't want the chip to show up by default, I can make a separate reducer for the model counts ... a little redundant in terms of data storage, but would be a cleaner UI. I can also perhaps find a way to keep the same data store but add a flag on the "Count" button to reveal the chip. Opinions on preferred behavior?

graft:
Since we are now creating a requirement to have this query it seems great to reuse the data and present it automatically. However, it causes the menu item to be disabled. If we're going to automatically count records, can we amend the menu item to simply count attributes and be conditioned on that?

graft:
I think I prefer the "Add Model" button being in the ModelReport rather than at the top level. First all the other modeling buttons are already there, and it also saves entering the parent model name (which can just be clicked on the map instead).

I like the new toggle - I might be a little more verbose about the two options, like, make the toggle `collection <-> table`, and when one or the other is selected an adjacent label reads either:
"The model will be a collection of named entities"
"The model will contain tabular data associated with the parent record"
Hopefully this helps people understand the presence/absence of the identifier field given these two options. I think it's useful to say "collection" and "table" because these are the attribute types they will see in the interface, so it is good to connect and explain those here.

dtm2451:
Hmmm I liked my 'standard vs table' because it makes it immediately clear that in most cases, what they'll want is the standard/collection option.  But I am also a fan of forced learning when it doesn't get in the way of immediate usability... and I think with the extra explanations, this can enter that category!  Just might amend the collection description to end in "named records" instead of entities.

graft:
> Hmmm I liked my 'standard vs table' because it makes it immediately clear that in most cases, what they'll want is the standard/collection option. But I am also a fan of forced learning when it doesn't get in the way of immediate usability... and I think with the extra explanations, this can enter that category! Just might amend the collection description to end in "named records" instead of entities.

I'm not sure I agree that one is "standard" - they have different usages, and people will have to use both in different contexts. Tables get used a lot - the immunoassay and flow models both depend on them, as do all the clinical tables. In this context they are "standard" certainly. I think it's worthwhile instead to teach people the important distinction, viz., one is "named records" (records is way clearer than "entities"), the other is unnamed, tabular data.

coleshaw:
> However, it causes the menu item to be disabled. If we're going to automatically count records, can we amend the menu item to simply count attributes and be conditioned on that?

Oof, I didn't test the menu item, didn't realized it wouldn't work now. Yes, good point, thank you, I can amend the item to just count attributes.

Other thoughts on the placement of the "Add Model" button? If I move it back to the ModelReport, would you still let them change the parent model? I kind of liked being able to just add model from the button without having to click on the map, first...

coleshaw:
We should demo this at a standup perhaps and get @bsamad feedback, too?

coleshaw:
K, made changes based on the latest feedback.

dtm2451:
Hmmm the one thing I think is missing now is an indicator of what the parent model is.  Currently, the user needs to intuit that the model will be a child of the model they'd clicked from, and also remember which model that was.  Could we add back a line displaying the parent model?
(My other thought would be to just add "child" to either the button text or its tooltip, but perhaps that would get confusing if the user does understand the link child/collection difference?  Thus thinking just an info line within the pop-out widget instead, but it may be a silly worry and thus a viable alternative!)

coleshaw:
Thanks for the suggestion! Here is what it looks like with an info-type line of the selected parent model.

Screenshot from 2022-11-15 12-01-16

coleshaw:
One more request from Bushra:

- [ ] Re-confirm for Remove Model, if there is data in it (i.e. bring up a dialog, make them type something).

PR #1120, opened 2022-11-10
fix formatting of author for metis files

coleshaw:
This PR (I think?) fixes a bug I'm hitting in Metis, where the use of [`authorFormat()`](https://github.com/mountetna/monoetna/blob/master/etna/packages/etna-js/utils/format.jsx#L5-L9) returns an object, not a string, and Metis UI doesn't know how to render that. I get the following error:

```
Uncaught Error: Objects are not valid as a React child (found: object with keys {name, email}). If you meant to render a collection of children, use an array instead.
    in div (created by ListEntryUpdatedColumn)
    in div (created by ListEntryColumn)
    in ListEntryColumn (created by ListEntryUpdatedColumn)
    in ListEntryUpdatedColumn (created by ListFile)
```

And the app crashes with a white screen if you try to load files or folders in a bucket (main Metis root view works fine).


graft:
Ergh, my bad. I thought I got all these...

dtm2451 (approved):

PR #1125, opened 2022-11-16
Janus create project template; [close #1116]

### LINKED ISSUE #1116: Janus: add option for applying coproject_template when creating a new project

coleshaw:
This PR adds a widget to the Janus screen for picking a template to copy, when creating a new project. Default should be `coprojects_template`. 

Screenshot from 2022-11-16 08-58-25

The copying happens in a separate thread because it takes so long, but the user receives a notice to check Timur in 5-10 minutes. Unfortunately because it is in another thread, I'm not sure how we'd notify the user if there are errors ... Keeping it in the same thread means the client times out and doesn't refresh the page to add the new project to the list, which seems like worse UX.

Screenshot from 2022-11-15 12-03-10


coleshaw:
Because right now it just runs through the same workflow as using the gem, where it makes changes one action at a time.

I suppose that I could try to consolidate the logic into a single set of actions, but I'm not very confident that Magma would order those correctly and worry that it might crash with an incompletely drawn project graph...

graft:
I think a single set of actions is maybe risky, but I think it could be at least bundling things by model - there's no reason you couldn't add all of the data attributes for a single model at the same time without conflict, especially for a blank model. Not sure if this would suffice to make the process run in a reasonable request cycle, though...

coleshaw:
But for things like links, you'd have to make sure you got the other model created first...

coleshaw:
I think also, even for a single model, you'd have to wait for Magma to restart...so at a minimum, 9-10 roundtrip HTTP requests from Janus -> Magma, which I think naturally won't fall into a user's reasonable request expectation, since they're operating in a single HTTP request time's frame-of-reference?

PR #1128, opened 2022-11-18
Metis mv folder bug 1124; [close #1124]

### LINKED ISSUE #1124: Metis: move folder command (client or UI) does not handle trailing /
> I want to move folders from one bucket to another bucket within the same project in Metis. Is this possible? When I either use the mv command in metis_client in my terminal or click the ‘Move folder’ dropdown menu option in the website and enter a path that’s in another bucket, the transfer does not work. metis_client tells me “Could not rename file” and the website tells me “Server Error.” 

> The metis_client mv seems to be a result of a server-side bug which the client is unable to avoid. The UI "Move folder" function appears to work okay, you must simply get the pathname correct. There appears to be a bug that causes it to crash if there is a trailing / in the path in your move command. If you leave this out (i.e. enter "foldername" instead of "foldername/", similarly for the bucket name), the UI ought to work. Please let me know if you have any success trying this, or if not what exact paths you are entering into the UI that cause this to fail.

coleshaw:
I think this PR addresses some of the issues identified in #1124, with trailing `/` and the `mv` folder command in the UI and in the client.

You should now be able to:

* Use the UI to move a folder, even if you accidentally include a trailing slash.
* Use metis_client to mv a folder to another bucket, by including a metis path as the target. Previously this behavior only worked for `cp`. Still have to teach people to use a metis path in this case, instead of a relative path (this is fairly unintuitive).


metis/lib/path.rb
```diff
@@ -24,8 +24,14 @@ def self.filepath_match
...
 
+      def self.remove_double_slashes(path)
+        return ::File.join(path.split("/")) unless path.start_with?("metis://")
+
+        ::File.join("metis://", path.split("metis://").last.split("/"))
+      end
+
       def self.path_from_parts(project_name, bucket_name, file_path)
-        "metis://#{project_name}/#{bucket_name}/#{file_path}"
+        ::File.join("metis://", project_name, bucket_name, file_path.split("/"))
```
graft, line 34: Won't this cause `metis:///`?

coleshaw:
```
irb(main):001:0> ::File.join("metis://", "example_project", "bucket", "path/to/file.txt".split("/"))
=> "metis://example_project/bucket/path/to/file.txt"
```

coleshaw: :smile: Somehow `::File` figures it out.

Issue #1123, opened 2022-11-15
tsv download of a project's / rule's names from gnomon


dtm2451:
another idea that came up in the gnomon demo today.  Jess brought it up for merging gnomon in with sample tracking systems for current projects. Probably the intent is a rule focused download (ex: all scRNAseq samples) so could be a button that pulls from the table that's already in the compose page.  I can reach out to Jess to scope more if we'd like!

Issue #1122, opened 2022-11-15
sortable tables in gnomon


dtm2451:
an idea that came up in the gnomon demo today

PR #1130, opened 2022-11-19
table sorting + download

graft:
Adds sorting to tables in Gnomon along with a TSV download button, closing #1122 and #1123. The TSV-ing is just inline janky stuff, I didn't figure on adding a dependency like `csv`.


coleshaw (approved):
Looks good to me

dtm2451 (changes requested):
Looks good to me, and tsv download worked!  I like how you used a background color highlight to deal with the sorting of tables now overriding our previous order-at-top method way of highlighting the directly targeted name!

Only thing I'd change if you can is making the file format fill in with `.tsv` or `.txt` instead of it's current `.csv`.  Cuz that's not correct.

Issue #1126, opened 2022-11-16
make option that automatically 'y's config updates in dev builds


dtm2451:
Just a thought of something which could speed up the process of testing out each others PRs.

A lot of the time I spend waiting for a project rebuild in order to test one of our PRs comes from simply not noticing that I'm being asked "vulcan/config.yml.template has been updated recently, copy it into config.yml? y/n" while focusing elsewhere.  And it often happens for multiple of our apps.

Something like `make up --update-configs` if that would work?

Perhaps not a high prioirity task because it only affects dev builds and even then only a subset of those builds, but could be nice!

coleshaw:
Oh yes, I have that happen to me, too! I never hit `y` though ... I like moving it out into its own task, or separating the two things. Perhaps now the `config.yml` is more stable, so less of an issue that an app breaks without you knowing about changes in config...

dtm2451:
This may be poor practice on my part, but I've so rarely needed to make changes to any configs for things to 'just work', (exceptions being just targeted vulcan dev work and times that something was called out directly in a PR), so I do just accept the default templates as fine and accept updates without thinking about it.

Hasn't been a source of issues for me, so...
>Perhaps now the `config.yml` is more stable

perhaps indeed!

PR #1133, opened 2022-11-21
separates out update of config files to own make task

coleshaw:
I think this addresses #1126 ?

Adds a new `make update-configs` command to iterate through the `config.yml.template` files and prompts you if you want to update.

Removes this check / prompt from the `make up` or other `make` commands.

Also adds `gnomon` to the `make clean` task.

To test locally, touch or edit a `config.yml.template` in an app directory. You should be able to run the following without being prompted now if you want to update configs:

* `make up`

But if you then run `make update-configs`, it should catch your updated `config.yml.template` and prompt if you want to update your `config.yml`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 export COMPOSE_PROJECT_NAME=monoetna
 export NODE_ENV=development
-export MONOETNA_SHA=$(shell git rev-parse HEAD) 
+export MONOETNA_SHA=$(shell git rev-parse HEAD)
 
 .PHONY: help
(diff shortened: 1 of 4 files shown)


dtm2451:
Oof, this PR has been waiting for me for a while!

I think this is nice that the `make up` process wouldn't get hung up, but it could also lead to configs becoming problematically out-of-date if I don't regularly run the separate `make update-configs` command...  I don't know whether I'm in favor of it or not?  @graft would you want this change to the build system?

PR #1146, opened 2022-12-01
metis_client updates

graft:
This adds a couple of bugfixes to metis_client.

### DIFF
--- a/metis/bin/metis_client
+++ b/metis/bin/metis_client
@@ -19,6 +19,10 @@ require 'singleton'
 CONFIG_FILE='~/.metis.json'
 LOG_FILE='~/.metis.log'
 
+def clean_path(path)
+  Pathname.new(path).cleanpath.to_s.force_encoding('UTF-8')
+end
+
 def github_contents
   uri = URI('https://raw.githubusercontent.com/mountetna/monoetna/master/metis/bin/metis_client')
   response = Net::HTTP.get_response(uri)
@@ -822,7 +826,7 @@ EOT
   class Project < Command
     usage "project [<project_name>]"
     def run(project_name=nil)
-      raise ShellError, "No such project#{project_name && " '#{project_name}'" }. Try:\n#{@shell.project_list.join("\t")}" unless @shell.project_list.include?(project_name)
+      raise ShellError, "No such project#{project_name && " '#{project_name}'" }. Try:\n#{@shell.project_list.join("\t")}" if (@shell.project_list & [ project_name, 'administration' ]).empty?
 
       @shell.project_name = project_name
       @cwd = nil
@@ -982,8 +986,7 @@ EOT
       raise ShellError, 'No remote folder selected!' unless @shell.relative_path(remote_path)
 
       path = local_path =~ %r!^/! ?
-        local_path :
-        Pathname.new(::File.join(*[Dir.pwd, local_path].compact)).cleanpath.to_s
+        local_path : clean_path(::File.join(*[Dir.pwd, local_path].compact))
 
       begin
         # Download a folder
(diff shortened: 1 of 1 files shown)


metis/bin/metis_client
```diff
@@ -822,7 +826,7 @@ EOT
   class Project < Command
     usage "project [<project_name>]"
     def run(project_name=nil)
-      raise ShellError, "No such project#{project_name && " '#{project_name}'" }. Try:\n#{@shell.project_list.join("\t")}" unless @shell.project_list.include?(project_name)
+      raise ShellError, "No such project#{project_name && " '#{project_name}'" }. Try:\n#{@shell.project_list.join("\t")}" if (@shell.project_list & [ project_name, 'administration' ]).empty?
```
graft, line 829: Let admin be admin

metis/bin/metis_client
```diff
@@ -1076,6 +1079,15 @@ EOT
 
   class Put < Command
     usage "put <input_file1> ... [<input_fileN>] <remote_folder_path>"
+
```
graft, line 1082: Normally metis and the client will barf on badly-named files; this attempts to resolve by replacing illegal characters with `_`.

metis/bin/metis_client
```diff
@@ -1125,10 +1137,19 @@ EOT
       end
     end
 
+    NON_FILE_MATCH=/[\]\[<>:;,?"*\|\/\x00-\x1f]+/
```
graft, line 1140: This also includes `[]` although Metis allows those, because they seemed to give me trouble.

metis/bin/metis_client
```diff
@@ -1278,21 +1299,21 @@ EOT
...
       files.each.with_index do |file,i|
-        file = Pathname.new(file).cleanpath.to_s
+        file = clean_path(file)
 
         print "Computing md5s - #{progress_bar(i.to_f / files.length)} - #{i} of #{files.length}      \r"
 
         next if ::File.basename(file) == options[:file] || md5_cache[file]
 
-        md5_cache[file] = %x{ md5sum "#{file}" }.split(' ')[0]
+        md5_cache[file] = %x{ md5sum #{Shellwords.escape(file)} }.split(' ')[0]
```
graft, line 1308: Some filenames may contain a quote character.

metis/bin/metis_client
```diff
@@ -1278,21 +1299,21 @@ EOT
...
       end
       puts "Computing md5s - #{progress_bar(1)} - #{files.length} of #{files.length}      "
 
-      File.open("md5checksum.txt", "w") do |f|
-        output =  md5_cache.map do |file_name, checksum|
-          "#{checksum}  #{file_name}"
-        end.join("\n")
+      output =  md5_cache.map do |file_name, checksum|
+        "#{checksum}  #{file_name}"
+      end.join("\n")
```
graft, line 1314: If output computation fails/errors for some reason, this could previously blank the md5checksum.txt file, which represents many precious hours of computation.

metis/bin/metis_client
```diff
@@ -1358,9 +1379,9 @@ EOT
       folders.sort.each do |folder|
         next unless ::Dir.exists?(folder)
         # remove checksum file if necessary
-        files = Dir[folder]
+        remaining_files = Dir[folder]
 
-        if files.size == 1 && is_subdir_checksum_file?(files.first)
+        if remaining_files.size == 1 && is_subdir_checksum_file?(remaining_files.first)
```
graft, line 1384: prevents miscount of total number of files below

coleshaw:
Nice job hunting all these edge cases down!

Should the same `corrected_file_name` logic be applied to `Mv` and `Cp`, where someone has the opportunity to rename a file, and thus inject unicode / unwanted characters?

graft:
> Should the same `corrected_file_name` logic be applied to `Mv` and `Cp`, where someone has the opportunity to rename a file, and thus inject unicode / unwanted characters?

Hmm, since this behavior is behind a flag, I'm not sure it makes sense? Metis will actively refuse to make badly-named files anyway, the idea of the `-F` flag is to allow this to happen automatically when we are putting many files. If the user is explicitly writing out the filename, they can see the error and fix it themselves hopefully.

coleshaw (approved):

PR #1147, opened 2022-12-02
Airflow retries task behavior

coleshaw:
This PR adds some passthrough configuration parameters to the Box and CAT ingest airflow tasks, so that the ETL author can set number of retries and a strict dependency on past runs succeeding. These are defaulted to 10 and `False`. I think this works for these types of ETLs because each run is actually batch loading data and checking against a cursor (an Airflow variable) of past data, instead of relying on the given batch date range. This is due to the way we're polling both Box and the CAT, given limitations of the available interfaces.


coleshaw:
This should help prevent the problem where a single task fails after X retries, for some reason (i.e. Metis is down for an extended period of time), but then the entire ETL run is stuck because the given tasks require past runs to have succeeded. Hopefully the next run of the ETL would just re-run the failed task and pick up the missed files.

graft:
I was considering this for a job the other day - however, the situation I am worried about is if the task merely fails silently and does not block, will it simply be ignored? And as a result some data is not pushed?

coleshaw:
Hm, but we have the exact same problem now -- if some task is blocked and no one complains, data does not get pushed. Without some sort of notification system that a task fails, or someone occasionally monitoring the Airflow dashboard, tasks fail silently all the time ...

coleshaw:
Merging this in, because this only affects those specific types of ETLs, and we can override the behavior in the ETL config.

PR #1163, opened 2022-12-16
does not refetch count when equals to 0; [close #1162]

### LINKED ISSUE #1162: Timur: thrashes and constantly makes requests for records on Map page, when 0 records for a model
I thought I had fixed this, but noticed it is still an issue today. We should fix this bug...

Basically go to the Map page and select a model with 0 records. You'll see that Timur is constantly making the `::count` request to Magma, but it shouldn't re-fetch.

coleshaw:
This PR sets the logic check to account for 0 record count, so it won't continuously refetch the model count data.

### DIFF
--- a/timur/lib/client/jsx/components/model_map/model_report.jsx
+++ b/timur/lib/client/jsx/components/model_map/model_report.jsx
@@ -498,7 +498,7 @@ const ModelReport = ({
   }, [model_name, models]);
 
   useEffect(() => {
-    if (counts[model_name]?.count) {
+    if (counts[model_name]?.count >= 0) {
       setCanReparent(counts[model_name].count <= 0);
     } else {
       getAnswer([model_name, '::count'], (count) => {


timur/lib/client/jsx/components/model_map/model_report.jsx
```diff
@@ -498,7 +498,7 @@ const ModelReport = ({
   }, [model_name, models]);
 
   useEffect(() => {
-    if (counts[model_name]?.count) {
+    if (counts[model_name]?.count >= 0) {
```
dtm2451, line 501: Hopefully I'm just being dense, but this doesn't seem like it would fix the issue to me?

dtm2451 (approved):

timur/lib/client/jsx/components/model_map/model_report.jsx
coleshaw: Should be working in production now, if you check the same cytof model you were looking at before? :crossed_fingers:

dtm2451: confirmed!

PR #1168, opened 2022-12-22
use Variable.update instead of Variable.set

coleshaw:
Getting some errors on the CAT ingest ETL (wonder why we didn't see these before):

```
  File "/home/airflow/.local/lib/python3.8/site-packages/sqlalchemy/engine/base.py", line 1276, in _execute_context
    self.dialect.do_execute(
  File "/home/airflow/.local/lib/python3.8/site-packages/sqlalchemy/engine/default.py", line 608, in do_execute
    cursor.execute(statement, parameters)
psycopg2.errors.UniqueViolation: duplicate key value violates unique constraint "variable_key_key"
DETAIL:  Key (key)=(cat_ingest_cursor-c4-cat_ingestion_etl3) already exists.
```

Changing to use [`Variable.update`](https://airflow.apache.org/docs/apache-airflow/2.2.4/_api/airflow/models/variable/index.html#airflow.models.variable.Variable.update) instead of `Variable.set` to try and avoid this problem. Testing in production ...

### DIFF
--- a/airflow/opt/providers/etna/etna/hooks/box.py
+++ b/airflow/opt/providers/etna/etna/hooks/box.py
@@ -244,7 +244,7 @@ class Box(object):
         """
         Save the cursor to the database.
         """
-        Variable.set(self.variable_key, self.cursor, serialize_json=True)
+        Variable.update(self.variable_key, self.cursor, serialize_json=True)
 
     def remove_file(self, ftps: FTP_TLS, file: FtpEntry):
         """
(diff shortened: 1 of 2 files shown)


coleshaw:
hm, just noting, but I don't think this helps, and the same exception is thrown. Might be some sort of race condition, where multiple tasks are trying to update the same key at the same time, because when the task re-runs, no exception is thrown...

PR #1172, opened 2023-01-05
WIP

coleshaw:
Just opening this to pin down the compare, don't actually use / merge this, since we figured out that the group_by in SQL is not performant.


coleshaw:
Current working branch is much better...ideas have been taken from this branch, but I think this isn't needed anymore.

PR #1175, opened 2023-01-11
Metis linker

graft:
This is a WIP of the universal linker, which basically creates the UI part of the linker. Mostly this cribs off the existing polyphemus stuff, which allows us to define and hold etl configs of various kinds based on a JSON schema. The backend changes are mostly teaching Polyphemus about the new metis job and not trying to run it via the `RunEtlJob` command. The front end mostly just adds a new custom form for editing the Metis loader config (by default you just get a JSON editor). I did a little bit of refactoring out common components between this and the Redcap form, but maybe more could be done (or less?).

Remaining work is basically making the ETL job that runs this thing. This is more or less the existing linker code + a polyphemus client requesting current ETL configs. It would be nice if this is mostly in monoetna I think and not in the airflow code editor, so it can be tracked and have tests around it, especially since this won't have any project-specific information in it.


dtm2451:
Been a bit here (apologies!) so correct this summary if it's wrong: I remember you'd had additional ideas for this linker, but all can be scoped for a next PR?

I think this is great!  All my notes below are pretty nitty.

Some notes on the current version:
- The `type` options dropdown could be made more clear.  Incorrect or not, for the 'table' option, my newer-to-polyphemus brain went to "table model" and not the true meaning "load from a table". Perhaps the display names for these options could be more specific?  Maybe --> "Link file", "Link file collection", "import from table" or "parse table file"?
- The placeholder text for the file_match field would be more clear saying "Regular expression matching file \*name\*"
- type = table: In the column_map 'ADD mapping' modal, the placeholder "Mapped Column Name" could be more clear as "Mapped Attribute Name".
- type = table: I'm not sure how extracted_columns is used.  Are these columns where the column name already matches the attribute name?  Maybe that can be clarified in the placeholder or with a tooltip over either the input field or the "extracted_columns" text

graft:
Okay, this seems to be working. A few notes:

1) I did not get this in a test harness. I tried to use airflow's DebugExecutor to test this at first, but could not get it to work, so I gave up. Newer versions of airflow implement the much simpler dag.test() and don't use an executor at all, but this requires an upgrade, which I did not have the patience for here. Thus, much remains untested. Another difficulty in testing is that the dev environment doesn't actually seem to support airflow talking to the apps. Adding them to the airflow network helps, but I couldn't figure out how to get airflow to support dev SSL, so in order to test I had to spike all of the client requests with `verify=False`, rather painful.

2) There are a number of changes around how Polyphemus implements versioning for the `etl_config` model, mostly around using `config_id` to determine uniqueness instead of `name`. There might be some residual issues with this change. There are also several unsatisfying features of the model, as it conflates the requirement of maintaining a config history with that of remaining the state of the job running this config, which is now maintained in the "current" record for each etl_config (e.g., making changes to a config for a given config_id will create records with varying version_number, but the latest version_number will hold `status, run_interval, output`, etc.

3) I added an `add_output` endpoint to Polyphemus to smooth over some of the logging issues. This allows you to append log lines into a db field, but I would like to soon transition to a better logging system, since I'm not sure how much text one can append to a db string column before trouble starts.

dtm2451:
Tested with both `file` and `file_collection` types and both worked great!  This is awesome.

Logs seem pretty bare.  I know that's a spot you have thoughts and potential future plans, so no worries there, just one suggestion:
- maybe there could be a log line added like "No new files" when it's the case.  That'd be nice feedback for a user trying to dial in their file regex.  Maybe they got the regex wrong in the first run.  Afterwards, they'd really need to delete and re-make the loader or touch all target files in order to get the ETL to care about them.

A few UI/UX notes:
- **The only thing here that I think NEEDS to be done before release**. Remove the 'data_frame' option from the loader type dropdown.  Lots of great work in the ui for that option, but if the backend isn't there, better to hide than let users build a perception that the route is unintentionally broken!
- I'd recommend moving the `+` add model button to the left side, it'd be more visibile.  Took me a moment to notice this button was there at all my first time.
- The fact, regardless of config changes, that a particular loader will only pay attention to new files may not be the expectation of all users.  Some sort of 'Re-check previous files' button for the ui would be nice.  Perhaps it would essentially just handle the delete (-> 'previous') and re-create (-> 'new') loader path which is the easy way a user would achieve this currently. And perhaps so that the log history is not lost, the logs from the 'previous' loader could be copied into the 'new' loader within this action.

graft:
>The fact, regardless of config changes, that a particular loader will only pay attention to new files may not be the expectation of all users. Some sort of 'Re-check previous files' button for the ui would be nice. Perhaps it would essentially just handle the delete (-> 'previous') and re-create (-> 'new') loader path which is the easy way a user would achieve this currently. And perhaps so that the log history is not lost, the logs from the 'previous' loader could be copied into the 'new' loader within this action.

Adding a "touch folder" button which touches all of the files in the indicated folder path might do this.

PR #1179, opened 2023-01-19
project-specific existence check

graft:
Adds a --project-name flag to nuke and validate, and the corresponding "exists" endpoint on Metis, to allow those commands to only check (or remove) files that are linked into some project. Doesn't know anything about folder structure or buckets, just that the file is somewhere in the project's metis collection. A file that exists, but in a separate project (e.g. triage) will be treated as if it does not exist in Metis.

### DIFF
--- a/metis/spec/client_spec.rb
+++ b/metis/spec/client_spec.rb
@@ -751,6 +751,10 @@ describe MetisShell do
   end
 
   describe MetisShell::Nuke do
+    after(:each) do
+      FileUtils.rm('md5checksum.txt', force: true)
+    end
+
     it 'nukes a local directory except for missing files' do
       bucket = create( :bucket, project_name: 'athena', name: 'armor', access: 'editor', owner: 'metis')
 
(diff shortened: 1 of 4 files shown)


metis/spec/client_spec.rb
```diff
@@ -809,5 +834,22 @@ def token_with_exp(exp)
 
       FileUtils.rm_r('athena-copy')
     end
+
+    it 'only looks within the specified project' do
+      bucket = create( :bucket, project_name: 'athena', name: 'armor', access: 'editor', owner: 'metis')
+
+      helmet_file = create_file('athena', 'helmet.jpg', HELMET, bucket: bucket)
+      stubs.create_file('athena', 'armor', 'helmet.jpg', HELMET)
+      folly_file = create_file('ate', 'folly.txt', WISDOM.reverse, bucket: bucket)
```
coleshaw, line 843: Hm, does this do what you intend (create another file in a project called `ate`)? I'm not sure `create_file` will create the project and generate a `Metis::File` record for it? Plus, wouldn't you need to have the `bucket` in that case, belong to the specific project?

Perhaps a good counter-test would be to have the exact same setup code and queried without the `--project-name athena` flag, and see if the `folly.txt` file is returned...

I believe this idea would apply to all the new tests.

graft: Hmm, the bucket is definitely wrong here. I think it will work anyway because the actual check only relies on the `project_name` string on file, otherwise there is no project entry. I added some more fixtures to the default cross-project search for good measure.

Issue #1171, opened 2023-01-04
Magma: Query -- JSON agg and JSON build array


coleshaw:
```
magma_development=# SELECT p.name as "project_name", COALESCE((SELECT json_agg( json_build_array(scp.tube_name, COALESCE((SELECT json_agg(json_build_array(sc.tube_name, sc.tube_name)) FROM ipi.sc_rna_seqs as sc where sc.sc_rna_seq_pool_id = scp.id), '[]'::json) ) ) FROM ipi.sc_rna_seq_pools as scp WHERE scp.project_id=p.id), '[]'::json) as "sc_rna_seq_tube_names" FROM "ipi"."projects" as p;
```

```
 project_name |                                                                                                           
                                                                                                              sc_rna_seq_t
ube_names                                                                                                                 
                                                                                                        
--------------+-----------------------------------------------------------------------------------------------------------
--------------------------------------------------------------------------------------------------------------------------
--------------------------------------------------------------------------------------------------------------------------
--------------------------------------------------------------------------------------------------------
 ipi          | [["IPIPOOL001.P1.scrna.xvipbmc", []], ["IPIPOOL001.P1.scrna.xvipbmc2", []], ["IPIPOOL001.P1.scrna.xvipbmc3
", []], ["IPIPOOL001.P1.scrna.xvipbmc4", []], ["IPIPOOL002.P1.scrna.xvipbmc", []], ["IPIPOOL002.P1.scrna.xvipbmc2", []], [
"IPIPOOL002.P1.scrna.xvipbmc3", []], ["IPIPOOL002.P1.scrna.xvipbmc4", []], ["IPIPOOL003.P1.scrna.xvipbmc", []], ["IPIPOOL0
03.P1.scrna.xvipbmc2", []], ["IPIPOOL003.P1.scrna.xvipbmc3", []], ["IPIPOOL003.P1.scrna.xvipbmc4", []]]
```

graft:
Seems like we can also handle `::first` at the same time this way (like just LIMIT 1 instead of doing a json_agg).

coleshaw:
Hm, that might work! Let's get this refactor working first, and then we can change `::first` as a separate issue / PR.

Issue #937, opened 2022-07-05
Magma: OOM issue with specific queries


coleshaw:
I've been trying to root-cause the OOM issue with specific Magma queries, and I think @corps is correct in what is happening. A good fix is currently beyond my brainstorming abilities -- I've played around with the GC, but I can't seem to make a dent in the memory explosion or trigger a clean up. I have a very minor fix that addresses part of this, but I think a real, significant fix would require re-architecting everything query-related.

I'm actually able to reproduce the issue locally, where a query will consume >12GB of memory and it is not released upon completion of the request (works against `/retrieve` and `/query`). Profiling the issue shows that basically there are huge numbers of allocated strings, i.e.

```
  1 Total·allocated:·4.57·GB·(19773266·objects)¬
  2 Total·retained:··53.68·kB·(488·objects)¬
  3 ¬
  4 allocated·memory·by·gem¬
  5 -----------------------------------¬
  6 ···4.16·GB··sequel-5.57.0¬
  7 ·410.21·MB··app/lib¬
  8 ···3.48·kB··puma-5.6.4¬
  9 ···3.40·kB··etna/lib¬
 10 ¬
 11 allocated·memory·by·file¬
 12 -----------------------------------¬
 13 ···4.15·GB··/bundle/ruby/2.7.0/gems/sequel-5.57.0/lib/sequel/adapters/postgres.rb¬
 14 ·410.01·MB··/app/lib/magma/query/predicate/model.rb¬
 15 ···7.76·MB··/bundle/ruby/2.7.0/gems/sequel-5.57.0/lib/sequel/dataset/actions.rb¬


451 Allocated·String·Report¬
452 -----------------------------------¬
453 ···1520643··"UCSF·Immunoprofiler"¬
454 ···1520640··/bundle/ruby/2.7.0/gems/sequel-5.57.0/lib/sequel/adapters/postgres.rb:610¬
455 ·········2··/bundle/ruby/2.7.0/gems/sequel-5.57.0/lib/sequel/adapters/postgres.rb:780¬
456 ·········1··/app/lib/magma/query/predicate/model.rb:82¬
457 ¬
458 ····760401··"ritpwlgosh"¬
459 ····760354··/app/lib/magma/query/predicate/model.rb:262¬
460 ········34··/app/lib/magma/query/predicate/column.rb:41¬
461 ·········7··/bundle/ruby/2.7.0/gems/sequel-5.57.0/lib/sequel/dataset/sql.rb:686¬
[...]

PR #1180, opened 2023-01-19
Magma subselect instead of join

### LINKED ISSUE #1165: Magma adding models (or attributes?) still restarts the container
The PID function still doesn't seem quite right in production. Updating the models today for xrep1, and it has died a couple times when the Magma container dies and restarts, instead of just the Puma process. So probably something still needs tweaking, there.

coleshaw:
This PR implements subselects instead of using joins, to handle collecting data from collection / table attributes. This is a significant performance increase, and should close #1165, #1171, #937.

This is an entire re-architecture of the retrieve / query / question internals, and so unfortunately is a very large PR. Please review and provide any feedback that you wish. My goals were:

- Make sure current tests pass, so we are fairly confident that we won't be breaking any obvious use cases. :sweat_smile: 
- Factor out subselect classes, to hopefully future changes can be built on top of those and there is a bit clearer understanding of what is going on under the hood. There should thus be minimal impact to the current predicate classes.
- Factored out and created a separate `Answer` class to clarify the expected answer tuple or direct value cases. I think this makes some of the downstream logic for TSV writing a bit clearer...

Note that due to the way that identifiers from intervening models are not available to the subselects (since no joins), I have removed the Matrix caching feature. I wasn't very confident that it would behave the way we would want, given we may not have an identifier for the matrix attribute's model, and that we risk providing incorrect values back (i.e. cache on `monster_name` would have multiple `monster` -> `victim` -> `sidekick` -> `weapon_proficiencies` matrices. Since the subselect is fairly performant, this may not be an issue.

I'm sure we will find corner cases that were not exercised by the current specs, but hopefully we will still support all known use cases. If we get new bug reports, we will have to add specs and fix them as we go.


coleshaw:
:pray: @graft

magma/lib/magma/query/predicate/with_date_time_predicate_methods.rb
```diff
@@ -0,0 +1,7 @@
+class Magma
```
graft, line 1: Not sure I understand how these modules are used - why can't these just be in the DateTimePredicate itself?

coleshaw: They could be left there, but then they would have to be duplicated in the `date_time_subselect.rb` class, too. Otherwise I think the verb actions that call those methods would not work?

coleshaw: Generally, the new subselect predicates copy the verbs from the regular predicates, but are subclasses of a new `ColumnSubselectPredicate`, and they share different `#extract` code to handle the subselect nested data. So I was treating these modules as mixins, where you need the same methods and verbs for regular predicates and subselect predicates. 

I supposed I could also organize it the other way, subclass the subselect predicates like `DateTimeSubselectPredicate < DateTimePredicate`, and include a module for `WithSubselectExtract` that just overrides `#extract`. Would that be better?

graft: I have the vague impression that the SubselectPredicate classes could all be collapsed into their counterpart, and instead have the various `select` and `extract` bits for on the basis of something like `#is_subselect?`, which might be a little more compact. Could that work?

graft:
This is great stuff, a powerful revision to Magma that has been due for many years. Due to the large size it's difficult to review all of the new logic introduced for function, so I think we'll just have to trust the specs. Seems like the important auxiliary bits (disconnected data, restriction) in addition to the bread-and-butter querying work, so :crossed_fingers: 

For the rest probably a big legibility factor is due because of all the accumulating changes we've made, but probably not necessary at the moment.

Regarding the cached matrix stuff, it seems fine to axe that, as it wasn't working properly anyway due to multiple puma instances not sharing a cache.

magma/lib/magma/query/predicate/with_date_time_predicate_methods.rb
coleshaw: That's the way I originally worked it out when exploring this idea, but I think that is less clear and maintainable, since you wind up having a ton of `is_subselect` `if` statements laying around, and you have to pass that flag around quite a bit. It's less extensible (if we wind up having a third "subtype" of predicate, for example), and would seem to violate the single-responsibility principle (the same class / method would be doing two different things).

By organizing all the subselect-related code into their own set of classes, I think that will help isolate any bugs and changes that would be related to the feature.

PR #1184, opened 2023-01-24
etna materialize

graft:
Surprisingly easy delivery of a "materialize" command, which just involves connecting the Materialize workflow to a local Etna::Filesystem. `etna materialize --project-name my_project` results in a local folder `my_project` containing the project data.

If the Filesystem can report size information (i.e. responds to `stat`) sync_metis_data_workflow will skip a file if it exists with the same file size. Does not check hashes. Hopefully this makes resuming/retrying less painful. Records (.json) still get re-downloaded every time.

### DIFF
--- a/etna/lib/commands.rb
+++ b/etna/lib/commands.rb
@@ -248,13 +248,14 @@ class EtnaApp
 
       string_flags << "--project-name"
       string_flags << "--log-file"
+      string_flags << "--log-level"
+      string_flags << "--concurrency"
 
-      def logger
-        @logger ||= @logger = Etna::Logger.new(@log_file, 0, 1048576)
-      end
 
-      def execute(project_name:, log_file:'/dev/stdout')
-        @log_file = log_file
+      def execute(project_name:, log_file:'/dev/stdout', log_level: ::Logger::INFO, concurrency: 1)
+        logger = Etna::Logger.new(log_file, 0, 1048576)
+
+        logger.level = log_level
 
         workflow = Etna::Clients::Magma::MaterializeDataWorkflow.new(
             model_attributes_mask: model_attribute_pairs(project_name),
@@ -265,7 +266,7 @@ class EtnaApp
             logger: logger,
             project_name: project_name,
             model_name: 'project', filesystem: filesystem,
-            concurrency: 1)
+            concurrency: concurrency.to_i)
 
         workflow.materialize_all(project_name)
         logger.info("Done")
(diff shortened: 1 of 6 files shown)


etna/lib/etna/clients/metis/workflows/sync_metis_data_workflow.rb
```diff
@@ -82,20 +94,7 @@ def copy_file(dest:, url:, stub: false)
                 if start_time == end_time
                   next
                 end
-
```
coleshaw, line 85: Just a thought, but this has been useful in debugging ETLs....might be nice to put verbose logging like this behind some sort of `--debug` or `--verbose` flag in the future, if we find a need for it.

etna/lib/commands.rb
```diff
@@ -238,6 +238,58 @@ def execute(email:, task: false, project_name: nil)
...
+      include WithEtnaClients
+
+      string_flags << "--project-name"
+      string_flags << "--log-file"
+
+      def logger
+        @logger ||= @logger = Etna::Logger.new(@log_file, 0, 1048576)
+      end
+
+      def execute(project_name:, log_file:'/dev/stdout')
```
coleshaw, line 256: This is awesome! I can see where this might be expanded in the future, since folks might only want `sc_seq` data for example. Not for this PR, thought.

coleshaw (changes requested):
This looks great! Should we bump the gem version in `etna.gemspec` so it gets re-published? Then we can pass that info along to users.

etna/lib/etna/clients/metis/workflows/sync_metis_data_workflow.rb
coleshaw: Also not necessary for this PR...

graft: It seemed hard to follow in any case, especially with concurrency turned up, because it doesn't print file names, just rates. We could just put it behind logger&.debug and let them turn up the log level.

What I wish for is a nice progress bar here if using STDOUT, but I'm not sure how to deal with concurrency in this case.

coleshaw: Oh good point. Perhaps would be nice to include file names along with the rate on the log line. Yeah, progress bar with concurrency seems difficult ...

coleshaw (approved):

PR #1186, opened 2023-01-25
tweak uniq call until after all predicates have been built

coleshaw:
I think moving this `.uniq` call helps remove the flakiness seen in the Magma tests. Unclear why those tests would sometimes fail, since the bugs and failed tests we observe in testing are not reproducible, even with the same seed. So there may still be an underlying issue...

### DIFF
--- a/magma/lib/magma/query/question.rb
+++ b/magma/lib/magma/query/question.rb
@@ -155,11 +155,11 @@ class Magma
     end
 
     def built_predicate_select_statements
-      (predicate_collect(:select)).uniq.map do |predicate|
+      (predicate_collect(:select)).map do |predicate|
         predicate.is_a?(Magma::SubselectBase) ?
           predicate.build_with_alias :
           predicate
-      end
+      end.uniq
     end
 
     def order_by_attributes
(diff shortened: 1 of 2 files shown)


graft:
Hmm, unclear what tests you mean, could you specify what was flaky? But is it fair to use uniq here? At least it seems it would prevent me from mapping the same column twice, which ostensibly the api should let me do?

coleshaw:
You can see an example in [CI, here](https://github.com/mountetna/monoetna/actions/runs/4008024033/attempts/1). If you try running the tests locally with the same seed, you will most likely get either 1) all tests pass, or 2) some other random test fails. 

The [previous behavior also used `uniq`](https://github.com/mountetna/monoetna/blob/ec2d4fb16e235ca864c7da34554ee83f765b50c1/magma/lib/magma/query/question.rb#L131) to prevent mapping the same column multiple times. It works as intended, because if the user requests duplicate columns via the API, each predicate has a different `alias_name` value, so those are not considered duplicate columns even though they might extract identical information from the database.

graft (approved):

PR #1194, opened 2023-01-30
also render the "sample" button for identifier attributes

### LINKED ISSUE #1176: map 'sample' button for identifiers
'Sample' seems only possible for string types, but identifiers are all strings, so should be easy (?? :crossed_fingers: )

coleshaw:
This PR closes #1176.

@dtm2451 , is this what you had in mind? It looks like the other "sample" results, just listed at the bottom of the pane:

Screenshot from 2023-01-30 16-35-42

I think there was also a comment about making that box larger so you could see more data, but I don't recall -- might have just been a verbal comment at the meeting?

### DIFF
--- a/timur/lib/client/jsx/components/model_map/attribute_report.jsx
+++ b/timur/lib/client/jsx/components/model_map/attribute_report.jsx
@@ -243,6 +243,13 @@ const AttributeReport = ({attribute, model_name, isAdminUser}) => {
     );
   }, [executeAction, model_name, removeLink, attribute]);
 
+  const showSampleButton = useMemo(() => {
+    if (!attribute?.attribute_type) return false;
+
+    const allowedTypes = ['string', 'identifier'];
+    return allowedTypes.includes(attribute.attribute_type);
+  }, [attribute]);
+
   if (!attribute) return null;
 
   return (
@@ -257,7 +264,7 @@ const AttributeReport = ({attribute, model_name, isAdminUser}) => {
               handleRemoveLink={handleRemoveLink}
             />
           )}
-          {attribute.attribute_type == 'string' && (
+          {showSampleButton && (
             <Tooltip title='Show data sample'>
               <Button
                 onClick={showSample}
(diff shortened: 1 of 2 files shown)


graft:
>I think there was also a comment about making that box larger so you could see more data, but I don't recall -- might have just been a verbal comment at the meeting?

The value box has a `maxHeight: 90px` and `overflowY: auto` so it should be possible to see all of the data by scrolling for long lists of values in a sample; not sure if this is sufficient. Perhaps it should also be expandable?

coleshaw:
Or remove the `maxHeight` to let it fill the bottom of the pane?

I can make a couple variations if you two want to look at them.

coleshaw:
If you make it so the `sample` data fills the space, it would scroll / look like this:

Screenshot from 2023-01-31 10-17-56

dtm2451:
Thanks Cole, this is great!

Re making the box larger: Giving more space for the display could be nice.  With this `sample`ing being a function that the user actively clicks to request, using all the available space by removing `maxHeight` seems like it could improve the functionality.  (Tbh, this return starts to feel a bit unwieldy to me once it gets too long, but maybe that's just me?)

My other thought focusing on this right now is that we don't explain that the displayed are just the unique / duplicate-removed values.  Maybe we update the tooltip and/or the row name to include the word "distinct"?

coleshaw:
>  (Tbh, this return starts to feel a bit unwieldy to me once it gets too long, but maybe that's just me?)

That's true ... perhaps we can present columns (like in Browse), or re-direct users to Search or Query. I'm not sure we originally planned for this field to be "useful", rather to provide a sample of the data, and if the user wanted to do something more with it, they would use one of the other tools. :shrug: We could also make a "copy out" button if that would make the data more immediately useful.

> Maybe we update the tooltip and/or the row name to include the word "distinct"?

That makes sense. I can update the row label to be something like `sample (distinct values)`

graft:
Make sure you can still copy the string as json, also important here. I used maxHeight because other value fields may also be long, and two overflowing boxes in one pane is no good.

coleshaw:
Ooo, so copy out as JSON is useful for other use cases, is what I'm understanding? That could be separate from the visual presentation though, if we had a distinct "copy samples" button?

Hm... regarding the competing maxHeights, do the other fields also have maxHeight set, so this would be the only one expanding (without maxHeight)? Or other ones might not have maxHeight set, and so we'd have multiple expanding boxes?

coleshaw:
:+1: Let me revert to the original change, in that case.

PR #1197, opened 2023-01-31
remove deprecated url encoding library

coleshaw:
Small change, since the Magma logs are super noisy. We should remove usage of this obsolete library anyways, as we did in #1005 and #928.


magma/lib/magma/storage.rb
```diff
@@ -51,14 +53,18 @@ def hmac_url(method, host, path, expiration=0, headers={})
...
             expiration: (Time.now + expiration).iso8601,
             nonce: Magma.instance.sign.uid,
             id: :magma,
             headers: headers
           ).url_params
         )
       end
+
+      def encode_path(path)
+        path.split('/').map { |c| Addressable::URI.normalized_encode(c) }.join('/')
```
graft, line 66: Somewhere we use URI.encode_www_form_component here, what is the difference?

graft (approved):

coleshaw:
> Err, just recalled this issue, which may cause trouble for files with a ? in them:
> 
> ```
> pry(#<FolderController>)> Addressable::URI.normalized_encode("blah?blah")
> => "blah?blah"
> ```

Hm...I believe you are right, that may cause problems, but is that not the same behavior as `URI.encode`?

```
irb(main):001:0> URI.encode("blah?blah")
(irb):1: warning: URI.escape is obsolete
=> "blah?blah"
```

coleshaw:
Perhaps that scenario also has implied consequences for other places in the code, too, not just here...

graft:
However:
```
irb(main):003:0> URI.encode_www_form_component("blah?blah")
=> "blah%3Fblah"
```

coleshaw:
Yes, that's right. So the problem is ... for some characters, `encode_www_form_component` does what we want (`?`), because it doesn't assume the `?` is the query character like the other methods, and thus encodes it. But on the other hand, ` ` is encoded incorrectly as if for a form and not a URL path, which breaks for us because the file name becomes part of the URL path ... :thinking:

coleshaw:
I think trying to account for that difference will require changes to other parts of the code, too, not just removing this one deprecated method call.

coleshaw:
I wonder if we just need our own encoding methods to account for the ` ` as a separate use case on top of `encode_www_form_component`. As far as I know, that might be the one character that behaves differently, depending on if it's in the query params or URL path?

coleshaw:
Okay, tweaked this a bit to have the encoding parts use a custom encoder, based on `URI.encode_www_form_component`, but then making sure any stray `+` are converted into the URL-friendly `%20`. Hopefully that is the only special character that we'll have to deal with ...

Issue #1196, opened 2023-01-31
Etna gem: build retry into the client


coleshaw:
With a backoff time, like metis_client currently does. This way materialize workflow will better handle situations when the service might be down.

coleshaw:
Also refresh the Janus token if almost at expiration

PR #1198, opened 2023-02-02
Etna gem client robustness

coleshaw:
This PR attempts to address #1196 .

It adds:

* Automatic retries to the core Etna::Client, which means our workflows should retry commands to Magma / Metis if the service is down (i.e. during a deploy)
* Token refresh when it appears the user's token is about to expire. This does not request a task token, since those are project-scoped, and I think that would be difficult to check per request. So just request a regular user token. This means the configured user token expiration duration, but with automatic refresh should hopefully just keep going until the job is done.

I added some specs and also tested manually:

* Against Metis production, when downloading a large file, I shut off my WiFi connection. The download paused, and then resumed when I re-enabled WiFi.
* Against Metis local, when I stopped the container, downloads paused and showed a "retrying" message (with `--log-level debug`).
* With an almost-expired production token, initializing a materialize command requested new tokens and the command kept executing even after the token's original expiration time.


etna/lib/etna/client.rb
```diff
@@ -3,11 +3,39 @@
...
+
+    def initialize(host, token, routes_available: true, ignore_ssl: false, max_retries: 10, backoff_time: 15, logger: nil)
       @host = host.sub(%r!/$!, '')
       @token = token
       @ignore_ssl = ignore_ssl
+      @max_retries = max_retries
+      @backoff_time = backoff_time
+
+      default_logger = ::Etna::Logger.new('/dev/stdout', 0, 1048576)
+      default_logger.level = ::Logger::DEBUG
```
graft, line 37: Probably should be on Logger::WARN by default?

etna/lib/etna/client.rb
```diff
@@ -149,46 +183,220 @@ def request_headers
...
+            @logger.debug("Received #{e.class.name}, retrying")
+            return retry_request(uri, req, retries: retries)
+          end
+
+          begin
+            retry_codes = ['503', '502', '504', '408']
+            if retry_codes.include?(response.code)
+              @logger.debug("Received response with code #{response.code}, retrying")
+              return retry_request(uri, req, retries: retries)
+            elsif response.code == '500' && response.body.start_with?("Puma caught")
```
graft, line 312: Hmm, what happens to other 500s here?

graft:
Looks great! I really like the encapsulation of the Retrier, makes it easy to plunk in around any request we want to make.

etna/lib/etna/client.rb
coleshaw: They should bubble up and get caught by the subsequent call to `status_check!`, which means they would behave the same as before. i.e. 500 "Server Errors" will just feed back to the client, etc. I believe this was put in place for metis_client, because the `Puma caught` is the raw message sent back when Puma is waiting for the app to come back up.

PR #1204, opened 2023-02-06
Timur query magma api bugs

### LINKED ISSUE #1202: Magma: Query API
It seems like ~~`/retrieve`~~ `/query` now does not return "Page # not found" when there are no results for a `::count` query, instead returns a "Server error". So the subselect stuff probably introduced a bug somewhere...

### LINKED ISSUE #1203: Timur Query: page crashes when switching attribute?
Hm...seems to be a bug in the Query page. I was constructing a filter on `rna_seq`, and had the attribute selected as `compartment`. When I tried switching the attribute to `eisenberg_score`, the app crashed with:

```
TypeError: e.replace is not a function
    r https://timur.ucsf.edu/js/timur.bundle.js:95
    value https://timur.ucsf.edu/js/timur.bundle.js:95
```

Looks like some sort of switching of the operator or escaping, perhaps:

```
var r=JE(function(e){return e.replace(/</g,"&lt;").replace(/>/g,"&gt;")}
```

coleshaw:
This PR fixes a couple of small issues:

* Fixes issue with `::count` and zero-results with Magma queries
* Prepopulates identifier fields in Timur Query (small quality of life improvement)
* Fixes crashing Timur Query app when switch a filter attribute type.

close #1202 , close #1203

### DIFF
--- a/magma/lib/magma/query/predicate/model.rb
+++ b/magma/lib/magma/query/predicate/model.rb
@@ -141,7 +141,11 @@ class Magma
     verb '::count' do
       child Numeric
       extract do |table,return_identity|
-        Magma::Answer.new(table.first[count_column_name])
+        if table.empty?
+          Magma::Answer.new(0)
+        else
+          Magma::Answer.new(table.first[count_column_name])
+        end
       end
       format { 'Numeric' }
 
--- a/magma/spec/query_spec.rb
+++ b/magma/spec/query_spec.rb
@@ -1235,6 +1235,13 @@ describe QueryController do
       expect(json_body[:format]).to eq(['labors::labor#name', 'Numeric'])
     end
 
+    it 'reports zero ::count' do
+      query(['labor', ['name', '::equals', 'Eating'], '::count' ])
+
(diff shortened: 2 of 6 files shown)


graft:
Hmm looks like a client-side fix to the Timur Query crashing? I can't get search page to produce a 500, it still says "Page not found", can you clarify how you repaired this issue?

coleshaw:
I may have flagged the root cause incorrectly in the issue write-up, but I think what fixed the issue I saw it [this change to the model predicate `::count` verb on the Magma server-side](https://github.com/mountetna/monoetna/pull/1204/files#diff-305d7f5ed18d0c804aca2a5912bfa3465118a970b1385bf2ba4b6310917dbe44R144-R148). Basically the `::count` query to update the page values (in both `/retrieve` and `/query`) would throw a 500. 

With this branch, no extraneous 500 error is thrown, and instead you get the standard "Page 1 not found" message.

coleshaw:
Also, did you test on the latest branch of `magma`? In production, you should be able to produce the 500 response pretty reliably...

Issue #1205, opened 2023-02-06
Etna: retry should throw exception regardless of block_given? state


coleshaw:
Upon code inspection, it seems like `Etna::Retrier#retry_request` should always throw an exception when the number of retries exceeds the configured max, ignoring the `block_given?` status.

https://github.com/mountetna/monoetna/blob/master/etna/lib/etna/client.rb#L298

PR #1209, opened 2023-02-08
bump etna gem version; always throw exception if exceed number of ret…

coleshaw:
…ries

Should address #1205 ?


graft:
Can we also put in a fix for the null :size issue here? I don't understand why this fails, but at least preventing a crash here, if not forcing another retry, seems necessary.

PR #1213, opened 2023-02-10
ADR 0001 timur query

coleshaw:
This is my first attempt at writing an [ADR](https://github.com/joelparkerhenderson/architecture-decision-record), per @graft 's suggestion in Slack. This ADR is for the Timur Query page.

I'd welcome feedback on:

* Is this enough detail? Or not detailed enough? ADRs are supposed to be light-weight and high-level, so unsure if I hit the mark or not.
* Should there be separate ADRs for the `/query` API and the UI aspect, since those are arguably different things?
* Does something like this seem useful to the team?


graft:
Looks nice! The naming recommendations suggest we should do something like "query-timur.md" instead of adr-0001.md

coleshaw:
Yeah, I think the naming depends on which document you read ... some of the other docs ([like Michael Nygard's source article](https://cognitect.com/blog/2011/11/15/documenting-architecture-decisions)) recommend the `ard-####.md` format. I'm happy to change however we agree. I like that the `verb-noun.md` format makes it obvious what the ADR describes, but the sequential numbering also lets you see how designs progress over time, kind of like data migration files. Especially since ADRs can build off of each other and provide new / additional context.

i.e. how would you  name a follow-on Timur Query ADR file?

Perhaps we combine the two, into a format like `adr-0001-query-timur.md`?

coleshaw:
The suggested `adr-0001-query-timur.md` format would also match [the referred to examples given for the Arachne framework](https://github.com/arachne-framework/architecture)...

dtm2451:
I like this combined naming scheme.

Issue #1192, opened 2023-01-30
Metis ETLs throwing exception about `/etna/Gemfile.lock`


coleshaw:
Seeing the following exception on the three Metis ETLs in Airflow:

```
[2023-01-30, 11:02:39 PST] {docker_operator_base.py:317} INFO - /bundle/ruby/2.7.0/gems/bundler-2.3.21/lib/bundler/shared_helpers.rb:105:in `rescue in filesystem_access': There was an error while trying to write to `/etna/Gemfile.lock`. It is likely that you need to grant write permissions for that path. (Bundler::PermissionError)
```

Perhaps related to the recent bump in `etna` gem version? The lock file would have changed from `0.1.46` to `0.1.47` ... at least locally it seems to have, for me. Would that cause permissions issues? Seems weird...

coleshaw:
ETLs seem to be running again, without errors, in Airflow.

PR #1216, opened 2023-02-15
expose page_size argument to materialize command

graft:
For some reason materialize requests to download ipi are dying with a 403, which results from an expired download URL. These URLs are generated by `retrieve` requests to Magma, which happen right now in pages of 20 records. A plausible explanation for how a URL could become stale before it is used might be that the contents of the page take more than a day to download. This requires a page to be > 1 TB, which seems plausible for some of these datasets.

A hopeful solution is to simply shrink the page size to 1, so each record retrieve is fresh. An expiration might STILL result if a single record takes more than a day to download, but that probably isn't going to happen here.

### DIFF
--- a/etna/Gemfile.lock
+++ b/etna/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    etna (0.1.49)
+    etna (0.1.50)
       concurrent-ruby
       curb
       jwt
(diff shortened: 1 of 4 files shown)


coleshaw (changes requested):
Looks good! One small request: I think you also need to bump the `etna` version in `etna/Gemfile.lock` to match `0.1.50`, otherwise some of the apps will fail to install `etna` correctly and complain. This was what caused #1192, I think.

dtm2451 (approved):
Not sure if a change is needed related to my commented default Q, just a consideration.
Other than Cole's .lock change request, seems good to me!

coleshaw (approved):

PR #1217, opened 2023-02-16
fix map for too many boxes

graft:
Sometimes if there are too many boxes in a row on the map they overlap. This uses a max-width to make sure that doesn't happen.

### DIFF
--- a/timur/lib/client/jsx/components/model_map/model_node.jsx
+++ b/timur/lib/client/jsx/components/model_map/model_node.jsx
@@ -1,12 +1,13 @@
 import React from 'react';
 
-const ModelNode = ({model_name, center, size, selected, handler, disabled}) =>
+const ModelNode = ({model_name, center, size, selected, handler, disabled, numSiblings}) =>
+
   center ? (
     <div
       className={`model_node ${selected ? 'selected' : ''} ${
         disabled ? 'disabled' : ''
       }`}
-      style={{top: center.y, left: center.x}}
+      style={{top: center.y, left: center.x, '--model-width': `${600 / (numSiblings + 1) - 15}px` }}
       onClick={() => {
         if (!disabled) handler(model_name);
       }}
--- a/timur/lib/client/jsx/components/model_map/tree_layout.jsx
+++ b/timur/lib/client/jsx/components/model_map/tree_layout.jsx
@@ -65,7 +65,7 @@ class LayoutNode {
     let maxpos = grid[this.depth].length;
     let maxdepth = grid.length;
     this.center = {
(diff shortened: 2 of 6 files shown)


graft:
Peek 2023-02-16 11-43

timur/lib/client/jsx/components/model_map/model_node.jsx
```diff
@@ -1,12 +1,13 @@
...
-const ModelNode = ({model_name, center, size, selected, handler, disabled}) =>
+const ModelNode = ({model_name, center, size, selected, handler, disabled, numSiblings}) =>
+
   center ? (
     <div
       className={`model_node ${selected ? 'selected' : ''} ${
         disabled ? 'disabled' : ''
       }`}
-      style={{top: center.y, left: center.x}}
+      style={{top: center.y, left: center.x, '--model-width': `${600 / (numSiblings + 1) - 15}px` }}
```
graft, line 10: This is a pretty cool trick (paired with `var(--model-width)` in the css) that lets you pass data into any css styles set on this thing without having to override the style itself (so precedence isn't destroyed).

coleshaw: oh neat. Is this a React thing, where you can pass in / use CSS variables?

timur/lib/client/jsx/components/model_map/tree_layout.jsx
```diff
@@ -87,6 +87,8 @@ class TreeLayout {
 
     let grid = [];
 
+    this.grid = grid;
```
coleshaw, line 90: Do you still need this?

coleshaw (approved):
Nice, this looks like a pretty clean approach to addressing the issue!

timur/lib/client/jsx/components/model_map/model_node.jsx
dtm2451: pretty cool!

timur/lib/client/jsx/components/model_map/tree_layout.jsx
graft: Yeah has to be retained to figure out how many siblings each node has.

timur/lib/client/jsx/components/model_map/model_node.jsx
graft: Apparently, pure CSS/html.

dtm2451 (approved):
I like this approach a lot!  IMO, it works well enough even when there are MANY models, like the 15 in this gif =)
map_edit_model_names

PR #1218, opened 2023-02-17
Adr 0003 subselect queries

coleshaw:
This is an ADR for the Magma subselect implementation, and includes some backstory and our performance testing results. The detailed subselect examples are probably not 100% accurate, but hopefully give an idea of the intention behind subselects.

### DIFF
--- a/decisions/adr-0003-subselect-magma.md
+++ b/decisions/adr-0003-subselect-magma.md
@@ -1,0 +1,104 @@
+# ADR-0003: Subselect Magma
+
+## Status
+
+Accepted
+
+## Context
+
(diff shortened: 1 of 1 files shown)


dtm2451:
I think Saurabh can speak better to if there's anything important left out, since he was more involved in scoping this task and actually reviewed the relevant PR(s).  So I'm not going to officially 'approve' myself.  But from this ADR, I do feel like I can follow what was done and why, the scope of the code changes involved, what side effects there may be (both + and -), and even what tools were useful in benchmarking the changes.  :+1: Thanks for putting it together, Cole!

PR #1229, opened 2023-03-20
Fix materialize to pull all records

graft:
Previously the model walker would bail out on retrieving if it had seen a { from, model_name } pair before while walking the data tree, where "from" was the name of a linking model (e.g., you saw the "patient" model from the "project" model). I'm not sure why this was written this way, but it seems to have skipped a number of records as a result. This PR repairs the walker so that it only skips on actually previously-seen record names that it is asked to retrieve. This more accurately consumes the tree resulting in a complete materialize.

### DIFF
--- a/etna/Gemfile.lock
+++ b/etna/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    etna (0.1.50)
+    etna (0.1.51)
       concurrent-ruby
       curb
       jwt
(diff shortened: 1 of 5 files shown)


dtm2451:
I follow your logic and can see it from the old code.

Hard to review the new code without testing since I'm not great with Ruby...  Maybe I'm missing it, but is the `list_only` flag you're adding actually called anywhere?

PR #1234, opened 2023-05-05
WIP: Allow make build to work on OSX

amadeovezz:
### Problem

On the linux, the GNU utility `readlink` supports a `-m` parameter . The `readlink` command on mac does not support this parameter.

You can install `coreutils` on mac in order to acquire this feature, however the utility is named  `greadlink` (note the `g` for GNU)  in order to avoid naming conflicts with the native `readlink`. The makefiles have historically been supported for developement on linux, therefore many `make` commands will fail if they try and use the native `readlink`. 

### Solution

Add conditionals to use `greadlink` when the OS is mac in makefiles and other supporting build scripts. 

### Problem 

Trying to build docker images on Mac when the intended target platform is `linux/amd64` causes many docker issues

For more information: https://stackoverflow.com/questions/65612411/forcing-docker-to-use-linux-amd64-platform-by-default-on-macos

### Solutions 

Whatever shell you use to run `make -C magma up`, set the following environment variables:

**bash**  

`export DOCKER_DEFAULT_PLATFORM=linux/amd64`

**fish** 🐟 

`set -xU DOCKER_DEFAULT_PLATFORM linux/amd64`

Note: is is important that you set this env var globally (notice the `U`) !

#### Other (potential) issues 

You may also need to set the following env vars:

**bash** 

`export DOCKER_BUILDKIT=0`
`export COMPOSE_DOCKER_CLI_BUILD=0`

**fish** 🐟 

`set -xU DOCKER_BUILDKIT false`
`set -xU COMPOSE_DOCKER_CLI_BUILD false`

https://github.com/docker/compose/issues/8449
https://stackoverflow.com/questions/66662820/m1-docker-preview-and-keycloak-images-platform-linux-amd64-does-not-match-th

### Other approaches

- using buildx
- adding this key to compose files: platform: linux/amd64

### specs and versions

Docker can be quite finicky when it comes to OS specs and docker settings, so this solution was tested on:

- Apple M2 Pro
- Ventura 13.2

And the `docker version` :

```
Client:
 Cloud integration: v1.0.31
 Version:           23.0.5
 API version:       1.42
 Go version:        go1.19.8
 Git commit:        bc4487a
 Built:             Wed Apr 26 16:12:52 2023
 OS/Arch:           darwin/arm64
 Context:           default

Server: Docker Desktop 4.19.0 (106363)
 Engine:
  Version:          23.0.5
  API version:      1.42 (minimum version 1.12)
  Go version:       go1.19.8
  Git commit:       94d3ad6
  Built:            Wed Apr 26 16:17:14 2023
  OS/Arch:          linux/arm64
  Experimental:     false
 containerd:
  Version:          1.6.20
  GitCommit:        2806fc1057397dbaeefbea0e4e17bddfbd388f38
 runc:
  Version:          1.1.5
  GitCommit:        v1.1.5-0-gf19387a
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
  ```
  
  I've also attached some docker-desktop screenshots of various settings:

Screenshot 2023-05-10 at 3 02 01 PM

Screenshot 2023-05-10 at 3 02 44 PM

Screenshot 2023-05-10 at 3 02 12 PM

Screenshot 2023-05-10 at 3 03 05 PM

### DIFF
--- a/airflow/Makefile
+++ b/airflow/Makefile
@@ -3,6 +3,8 @@ release::
 
 include ../make-base/stubs.mk
 include ../make-base/docker-compose.mk
+include ../make-base/utils.mk
+
 app_service_name:=airflow_ui
 
 baseTag:=$(shell basename "$$(pwd)")
@@ -15,5 +17,5 @@ run-image-test::
 black:
 	docker run -u 0:0 -v $(pwd)/opt/providers:/opt/airflow/providers --rm $(baseTag) bash -c '$$@' -- black /opt/airflow/providers/etna
 
-pdoc: $(shell readlink -m $(shell dirname $(firstword $(MAKEFILE_LIST))))/image-release.marker
+pdoc: $(call readlink_m $(shell dirname $(firstword $(MAKEFILE_LIST))))/image-release.marker
 	docker run -u 0:0 -v $(pwd)/opt/providers:/opt/airflow/providers --rm $(baseTag) bash -c '$$@' -- pdoc -o /opt/airflow/providers/etna/docs etna
--- a/docker/build_support/build_image
+++ b/docker/build_support/build_image
@@ -32,12 +32,22 @@ if [[ "$1" == "-h" ]]; then
   exit 1
 fi
 
+OS=$(uname -s)
+
+# Set the appropriate readlink command based on the operating system
+# Note: use this only for the -m flag
+if [[ "$OS" == "Darwin" ]]; then
+  READLINK_M="greadlink -m"
+else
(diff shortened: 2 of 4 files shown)


amadeovezz:
UPDATE:

While the above readlink changes do allow path proper path resolutions and the make commands appear to work, I cannot consistently get images and their dependencies to resolve and build. Tinkering with docker settings sometimes works, but the current set-up is simply too difficult to debug and reason about.  

Instead please refer to: https://github.com/mountetna/monoetna/pull/1236

PR #1235, opened 2023-05-09
Validate Magma with Gnomon

graft:
## What

 Adds Gnomon Support for `/update` requests to identifier types in Magma. 

### Additional context

There are multiple different "gnomon modes" (see below), and we need to persist the state of each mode per project somewhere. Currently, there is no table currently to hold such a flag in Magma. There *is* a project table to hold such a flag in Janus, and some operations currently phone Janus for Magma requests (relating to community projects). So: either we make a project table and attach it to the existing `Magma::Project` class (i.e., make `Magma::Project < Sequel::Model`), or we have Magma phone Janus on update requests for the necessary flag.

## How

1. Defines a gnomon configuration mode for a project in Magma. 
- `none`, `identifier`, `pattern`

2. Each mode is analogous to a type of authorization
- `none`: no auth
- `identifier`: the user posts an identifier to `/update`, magma looks for the exact identifier string in the "name" column of the identifiers table. 
-  `pattern`:  the user posts an identifier to `/update`, magma constructs a regexp from the gnomon rules and compares it to this identifier

3. Each mode also describes data is attached
- `none`:  n/a
- `identifier`: magma creates ancestors and attaches data if auth passes 
- `pattern`:  magma does not attach data but records are saved to the db if auth passes 

### Solution

- Create a flag table to store gnomon configuration mode
- Add `BaseAttributeValidation` to the `IdentifierAttribute` class for auth
- Add `find_parent_models(identifier)` and `autolink_parent_identifiers()` to `loader.rb`. See functions for additional documentation.

### A note on control flow

Updates currently work like this:

1. The loader class determines which records and parent records should be updated
2. It pushes those records to memory, stored in the `@records` instance var
3. Validations, attribute hooks and access controls are then run on the entire list of records to be pushed

Therefore, when we call `autolink_parent_identifiers()` on `identifier` or `pattern` mode, as long as a proper grammar exists with a model hierarchy, we could push to `@records`. It is optimistic in this sense. However, later down the function call stack, all of these records may be rejected if auth fails.

### A note on old identifier types

As it stands, for either `identifier` or `pattern` mode, we do NOT allow updates on out of date identifiers. An identifier is considered "out of date", when the identifier points to an old grammar. Because grammars can be used to decompose identifiers to a model hierarchy, we want to ensure that those models are also valid. 

There is a method that now exists on the `Identifier` class called `is_up_to_date?`, that checks for out of date identifiers.

## TODOS

- [x] Allow magma to support project flags
  - [x] Tests
- [x] Add flags for `GNOMON_CONFIG`
- [x] Implement identifier mode
  - [x] Auth
     - [x] Tests
  - [x] Attaching data
     - [x] Tests
- [x] Implement pattern mode
  - [x] Auth
     - [x] Tests

## FUTURE TODO's

Solve non-hierarchical parenting. Requests like: 

```
{ 'LABORS-LION-NEMEAN': { 'victim': [ 'LABORS-HYDRA-LERNEAN-H1-C1' ] } }
```

Here victim should resolve to: `LABORS-HYDRA-LERNEAN` instead of `LABORS-LION-NEMEAN` . So how do we deal with children that do not properly decompose to a parent? These would likely be user errors? Since their intention is to attach the victim to a `LABORS-HYRDA-LERNEAN`.

This would cause a bug in the system, if both identifiers could be decomposed into their own model hierarchy. Then we would have created two links something like:

`LABORS-HYDRA  -> LABORS-HYDRA-LERNEAN-> LABORS-HYDRA-LERNEAN-H1-C1`
`LABORS-LION -> LABORS-LION-NEMEAN`


magma/lib/magma/flag.rb
```diff
@@ -0,0 +1,17 @@
+class Magma
+  class Flag < Sequel::Model
+  end
+
+  module Flags
+    # All modules nested under this namespace represent flags that can be used in a production.
+    # Each module should define a NAME constant representing the name of a flag and each
+    # module should also specify all potential values that can be assigned to a flag.
+    module GnomonMode
```
amadeovezz, line 9: Not sure if we like this approach, but I think it is important to have a record somewhere of all the possible values a flag can have. This could also be stored in the db.

graft: I think i like it in code as we cannot just create modes on the fly by addung db entries so the db doesn't really make sense here. Might be easier as a Hash instead of a nested Module?

magma/spec/update_spec.rb
```diff
@@ -1262,6 +1258,234 @@ def update(revisions, user_type=:editor, params: {})
...
+        Magma.instance.db[:flags].insert(
+          project_name: "labors",
+          flag_name: Magma::Flags::GNOMON_MODE[:name],
+          value: Magma::Flags::GNOMON_MODE[:pattern],
+          created_at: Time.now,
+          updated_at: Time.now,
+          )
+
+        grammar = create(:grammar, { project_name: 'labors', version_number: 1, config: VALID_GRAMMAR_CONFIG, comment: 'update' })
+        # TODO: why does this pass? This is an invalid identifier
```
amadeovezz, line 1474: @graft is there a bug with Gnomon here?

Here this test creates an identifier using an invalid separator: `*`.

When it is validated via:

```ruby
gnomon_validation = Magma::Gnomon::Validation.new( Magma::Gnomon::Grammar::Parser.new(grammar[:config]) )
return if gnomon_validation.valid?
```
it somehow passes?

graft: Hmm, I think this is right? You inserted an invalid identifier into the db. Gnomon will complain if you try to register an invalid id via api, but once it is in the db, magma will treat it as ok without further validation.

graft: a bit more, when an identifier is created in Gnomon it notes the version of the grammar used to create the identifier (via foreign key to the grammar table i think); however most endpoints like decompose currently deal only with the current version of the grammar, so they might not decompose an old identifier correctly. It is pretty hard to do this well since an old grammars may deal in old models, which are no longer valid.

graft: After some thought, seems like we can't auto-attach old identifiers this way, since we can't necessarily decompose them or even line them up against existing models. A few options:

1. Allow the no-longer-valid identifier to be used in the update, but don't auto-attach the record using Gnomon decomposition in this case. The user can either explicitly attach this (like specify parents in the update), or leave it detached.
2. Optimistically try to decompose using the identifier's version of the grammar. This could produce a valid set of rules allowing attachment against the current set of models, in which case, why not attach? If they don't, the current method will skip anyway.
3. Raise an error for out-of-date identifiers and refuse to enter them.

magma/lib/magma/loader.rb
```diff
@@ -134,6 +136,91 @@ def explicit_revision_exists?(revisions:, parent_model:, parent_record_name:, ch
...
+      # Then we check to make sure a valid hierarchy chain exists
+      # We check that each parent_model is in the list of models available
+      arr = chain.to_a
+      valid = arr.all? do |el|
+        model_name = el[0]
+        model_name == :project || arr.any? do |other|
+          candidate_model_name = other[0]
+          el[1][:parent_model_name] == candidate_model_name
+        end
+      end
```
graft, line 183: Because Hash < Enumerable, could do like:
```
valid = chain.all? do |model_name, model|
  model_name == :project || chain.has_key?(model[:parent_model_name])
end
```

magma/lib/magma/loader.rb
```diff
@@ -134,6 +136,91 @@ def explicit_revision_exists?(revisions:, parent_model:, parent_record_name:, ch
...
+      @records.each do |model, record_set|
+        record_set.each do |record_name, record|
+
+          next unless flag_value == gnomon_mode[:identifier] and not record.includes_parent_record?
+
+          # Attempt to find parent models
+          parent_models = find_parent_models(record_name)
+          next if parent_models.empty?
+
+          parent_models.each do |model_name, hash|
```
graft, line 215: might call this `model_info` or something instead of `hash`

magma/lib/magma/loader.rb
amadeovezz: nice, much cleaner

magma/lib/magma/loader.rb
amadeovezz: ya I like that

magma/spec/update_spec.rb
amadeovezz: > Gnomon will complain if you try to register an invalid id via api, but once it is in the db, magma will treat it as ok without further validation.

Are there instances where we would not use the API to register an identifier?

graft: >Are there instances where we would not use the API to register an identifier?
Doubtful, but since the grammar may change over time, we will undoubtedly accumulate no-longer-valid identifiers anyway.

magma/lib/magma/loader.rb
```diff
@@ -134,6 +136,92 @@ def explicit_revision_exists?(revisions:, parent_model:, parent_record_name:, ch
...
+      # TODO: eventually refactor to address the above comment.
+      #
+      gnomon_mode = Magma::Flags::GNOMON_MODE
+      flag_value = @flags[gnomon_mode[:name]]
+      updates = Set.new([])
+
+      @records.each do |model, record_set|
+        record_set.each do |record_name, record|
+
+          next unless flag_value == gnomon_mode[:identifier]
```
amadeovezz, line 205: @graft note that here, I've dropped the conditional here - previously `record.includes_parent_record?`.

I think this is more reasonable, that we just and build any model hierarchy we can - regardless if it is with the grand-parent, parent or child record. The `updates` var avoids duplicates

magma/spec/update_spec.rb
amadeovezz: For now, maybe let's raise an error for out-of-date identifiers. 

Conceptually the simplest and prevents potential data integrity issues. If it becomes cumbersome to users we can re-evaluate and allow a more flexible model

amadeovezz: Given you agree with my previous comment, this is something we likely want to do for both identifier and pattern mode?

amadeovezz:
@graft I've gotten the PR in a state where it is ready to be fully reviewed.

Potential remaining items:

- finalize discussion on old identifiers 
- additional tests for edge cases we haven't thought about?

magma/lib/magma/loader.rb
```diff
@@ -134,6 +136,92 @@ def explicit_revision_exists?(revisions:, parent_model:, parent_record_name:, ch
...
+          parent_models.each do |model_name, model_info|
+            next if model_info[:parent_model_name].nil?
+            parent_identifier = parent_models[model_info[:parent_model_name]][:identifier]
+            updates.add([model_info[:model], model_info[:identifier].to_s, model_info[:parent_model_name], parent_identifier])
+          end
+        end
+      end
+
+      updates.each do |el|
+        push_record(el[0], el[1], el[2] => el[3])
```
graft, line 220: fyi you can also use splat to do this like:
```
updates.add([ model_info[:model], model_info[:identifier].to_s, { model_info[:parent_model_name] => parent_identifier }])
...
push_record(*el)
```

magma/lib/magma/gnomon/identifier.rb
```diff
@@ -13,6 +13,12 @@ def to_hash
         }
       end
 
+      def is_up_to_date?
+        most_recent_grammar = Magma::Gnomon::Grammar.for_project(project_name).to_hash
+        identifier_grammar = Magma.instance.db[:grammars].where(project_name: project_name, id: grammar_id).first
+        most_recent_grammar[:version_number] == identifier_grammar[:version_number]
```
graft, line 19: could avoid another db call here by doing most_recent_grammar.id == grammar_id - wish the grammar could also be memoized somehow

PR #1237, opened 2023-05-22
Vulcan Workflow for Visualization of sc_seq_dataset

dtm2451:
### Primarily:
- Utilizes and expands the composability of the vizualization setup UI to include some dittoSeq plots.
- Utilizes and expands archimedes-r to run plot creation with dittoSeq directly.
- Workflow builds off of planned `sc_seq_dataset` model:
  - makes use of the .Rds `object` that's the central component of the model -- the dataset itself.
  - pulls from known `metadata_*` and `umap_name` attributes to power recommendations of metadata or dimensionality reductions within the object that the user can use for plotting.  (Non-restrictive... any dimensionality reductions and metadata contained in the object can be used for plotting.  These recommendations are simply a way to give a standardized pointer to the final umap, clustering resolution, cell annotations, etc. even though such pieces might have any name within the object.)

### Remaining ToDos:
- [x] Fix tests: CI failures from python tool updates not touched by this PR!
- [x] Finish dealing with Recommendations in `make_dittoSeq_plot.r`.  I designed the UX to give standard choices for dimensionality reduction and metadata to use in the plot setup ui, but the internally used names for these recommendations must still be parsed back to their original target values before dittoSeq can use them.  Dimensionality reduction recommendation parse-back is complete.  A scan for, and replacement of, `plot_setup` values matching `^_Recommended_` is still required for the metadata recommendation parse-back.
- [x] *maybe* Code ReorderPiece that only offers the 'custom' option (checkbox for on/off instead of the current dropdown) and utilization of custom order definitions into `make_dittoSeq_plot.r` so can finish implementing `var_order` and `group_order` inputs.

### Additional changes to the larger UI System encoded here:
- Plot outputting:
	- Converts original `PlotOutput` output function into a wrapper that determines the input plot type and then passes along to proper of the below...
	- Renames old code to `PlotlyOutput`
	- Adds `PngOutput` output path. 
- Adds optional labeling to the `nested_select_autocomplete` input so could carry this over to new `nestedDropdownPiece`.
- Adds new `nestedDropdownPiece` which wraps the `nested_select_autocomplete` input.
- Adds new **but unused, remove?** `nestedDropdownFullPathPiece` which I'd planned to use for gene and metadata selections, but it turned out the `nestedDropdownPiece` piece described above worked better.
- Adds new `reductionSetupPiece` for selecting e.g. UMAP vs PCA and components 5&6 instead of 1&2. 
- Adds `stepSize` control to the `sliderPiece`.
- Adds `can_randomize` control to the `ReorderPiece` because dittoSeq plotters sometimes allow `order = "randomize"`.
- Adds new `ReorderCustomOnlyPiece` to handle reordering of things that can't rely on an "increasing"/"decreasing"/ "randomize" options.
  - for disabling control (when no target data has been chosen yet), exposed 'disabled' feature through other components
- Added 'disabled' input to `BooleanInput`, false by default, and `checkboxPiece` .
- Updates the `subsetDataFramePiece` to work from an abstracted view of the target data so that the same piece can work for subsetting single-cell data without needing to flatten the object / directly provide data for all genes.


dtm2451:
Similar to the universal linker PR that's also in review now, testing this branch requires some particular development environment system setup:
1. Connect your dev vulcan to speak to the 'example' project in production magma -- that's where I created the `sc_seq_dataset` model and seeded it with Seurat's mini test dataset.  (Similar to the universal linker PR, using dev<->dev client/API interactions would require extra hacking.  But for vulcan testing, Cole built a dev-vulcan<-production-magma pathway which has dual bonuses of getting around "strict = False" connection issues andddd also allows us to test vulcan with real data directly.  We'll set that up here!) 
    - Set up your vulcan config.yml to have it connect with production magma by giving a production token in line 11 and the ucsf edu magma url in line 13.
    - Create a project via dev janus with "example" as its short name.
    - Give your dev user access to that project.  Viewer perms should be enough here.
2. Run the workflow!
    - (I'll try to give minimal required details here because your initial experience notes on the Vulcan UX would be pretty valuable on their own, Amadeo!)  
    - Click into the example project from the dev-vulcan homepage
    - Click the `scViz` workflow's card and then the `Create Figure` button.
    - *Give the example datset's record id to the single primary input -- EXAMPLE-SCD1 -- then `Commit` your choice and hit `Run`.
    - The next ui steps and everything that follows are where the real meat of this PR comes in...  Play around with the visualization options and see if you can break anything!

*Re-design of the primary input entry point for the workflow is something I plan to focus on during alpha testing of the workflow.  The current basic string entry method should be sufficient for the small group of alpha testers we'll have, and I have a few specific ideas already, so I'd like to consider that particular UX piece to be **out-of-scope for this PR** and not bog down any discussion there?  (We'll probably switch to starting the workflow with an automatically run magma query followed by population of a purpose-built UI for selecting among all the projects' available sc_seq_datasets in a more descriptive way!)

PR #1243, opened 2023-06-12
Fix bugs and performance issues with the vulcan scViz workflow

dtm2451:
Bug: Fixes a bug that were unfortunately introduced in the final features I'd tried to squeeze in to the original PR:
- var/group_order variables processing in the make_dittoSeq_plot.R script needed to come after recommendation replacement & 'left-as-make' removal.
  - means can't plot scatter or dim plots when the 'Primary Data' or 'Grouping Data' choice was from the recommend metadata rather than a true metadata name, cuz bugs out, AND can't leave the feature 'off', cuz grouping data will be made all equal to 'make'!

Performance Issue: Also fixes a performance issue where too many cells / dots in a scatter plot kills the browser when trying to plot in the plotly / interactive style.
- changed the default for this option to false, a.k.a. output as a flat png
- also set a cutoff of 25000 cells, beyond which the value is always treated as false (within make_dittoSeq_plot.R script, and with existence of a 'too many data points' limit told to user in the input label)

Previously Missing Features:
- Interpretation of which assay (ex: "RNA" vs "ADT") gene/protien data was coming from, and that's a required input for dittoSeq.  Relied on dittoSeq defaulting, which targets "RNA", so... ADT features could not actually have been plotted with the original version.
  - PR adds determination of how to fill in all 'assay'-defining dittoSeq inputs so ADTs can now be plotted!
- Recommended metadata only got suggested for inputs scoped to discrete data, but not for inputs that could recieve either discrete or continuous data
  - PR adds recommended metadata to 'all_opts', not just 'discrete_opts', in 'get_dataset_and_summarize.r'
- 'opacity' input not exposed
  - now added for both dittoDimPlot and dittoScatterPlot


graft:
Re: the "too many points" problem, I would make the default option for output to be interactive and set what you think is a decent default count as the hard cutoff, just so the interactive option is more visible. Also there is maybe something called scattergl that you might have available in plotlyjs?

dtm2451:
> Re: the "too many points" problem, I would make the default option for output to be interactive and set what you think is a decent default count as the hard cutoff, just so the interactive option is more visible. Also there is maybe something called scattergl that you might have available in plotlyjs?

Could go back to have interactive checked on as the default for scatter and dim plots, but the cutoff I chose in testing, and already worked in =) [a1eb0d1](https://github.com/mountetna/monoetna/pull/1243/commits/a1eb0d106240acdee5a6832d80b3cb93ddb27501), is so low that I don't think it makes much difference what the default value is... almost all of our datasets will be more than 25,000 cells and will thus have interactivity off no matter what.

Interactive is, however, still the default for bar plots, so it's not completely hidden

dtm2451:
The plots are made in ggplot in R (just optionally converted to plotly format with plotly::ggplotly) so it may be hard to work in plotly-side manipulations.  But it also may be worth doing and looking into the scattergl you mention as there's noticeable lower performance for the 'ggplotly' scatters compared to the version made by scatter_plotly.py in the scRNAseq workflow.  100k+ cell scatterplots run efficiently in that workflow, whereas I noticed choppy bowser slowness and updates below 100k points in this workflow (which is why I chose 25k as the cutoff).

PR #1244, opened 2023-06-28
Fix metis linker

graft:
Bunch of stuff happening here:
1. Switches to using globs to match files in the airflow metis linker. This hopefully lines up with the same in Ruby.
2. Add a "match" button which leads to a "touch" button, to determine if your regexp is valid. The touch touches everything, which may not be what you want, so it is something of a big hammer.
3. Adds a bulk touch_files endpoint to metis to avoid spamming.
4. Fixes some issues with the run pane
5. Fixes the original typo that was causing the linker to fail


dtm2451 (approved):
successfully touched and connected a file in testing.  Looks good to me!

PR #1246, opened 2023-07-03
Fix urls in tsv download from Magma

### LINKED ISSUE #1239: Cannot download links via Query page
If you query for a `file` attribute, and try to download a TSV of urls this way, you will end up with a column of `[]`, rather than the links that are visible in the browser.

graft:
This should resolve #1239, seems to have just been a simple fix as suspected.


amadeovezz (approved):
lgtm 👍

PR #1247, opened 2023-07-06
fix globbing

graft:
The glob implementation in the MetisLinker improperly matches one or more characters, while the glob in the find api uses SQL %, which may match zero-length characters. This produces a discrepancy in matching, here resolved.


dtm2451 (approved):
very nice, simple fix =)

PR #1248, opened 2023-07-06
Minor fixed for the Makefile-mac build path

dtm2451:
- Fixes a typo where `build .` was incorrectly ordered `. build`
- Also adds a target for the deployer-base image.  Likely, this is not necessary for using make targets within this file, but it was useful for me when I was mixing the build calls from this file with then using `make up` from the "base" Makefile.

### DIFF
--- a/Makefile-mac
+++ b/Makefile-mac
@@ -43,7 +43,12 @@ build-dev-archimedes-images:
 	 echo "Building archimedes images..."
 	 cd docker/archimedes-base && docker build . -t archimedes-base
 	 cd archimedes-r-base && docker build . -t archimedes-r-base # takes a really long time!
-	 cd docker/archimedes-node-base && docker . build -t archimedes-node-base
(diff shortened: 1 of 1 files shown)


amadeovezz:
its funny I had a comment about deployer base:

```
cd docker/development-psql && docker build . -t development-psql
	# Is this needed? cd docker/deployer_base && docker build . -t deployer-base
```

do you know what it is?

amadeovezz (approved):
lgtm!

dtm2451:
Seems an older automated deployment system that is currently replaced with 'Swarm Keeper'. I see messages from `"deployer on ${HOST_HOSTNAME}"` ([deployer/deployer.sh line 20](https://github.com/mountetna/monoetna/blob/61e9d0491ab9388d50c1e411a401733f94a027cf/deployer/deployer.sh#L20)) in the watchtower-ping slack from 2021 into Feb 2022 when the sender switches over to Swarm Keeper

PR #1249, opened 2023-07-10
File / Folder UIs for Vulcan

dtm2451:
Overall:
- **Adds new Visualization 'from File' workflow, because *optional* primary inputs are not possible with our current setup**
- **Adds ui-input components for selecting a metis bucket + path to a file, folder, or either ("path").**
- Adds validators for this input type which will complain if no bucket is chosen or if the user chooses a file/folder when the input requires folder/file (the opposite).  The validator can also be given a regex to additionally require a path_regex match.
- Edits to underlying etna-js `PickBucket` and/or `Pick(FileOr)Folder` components:
  - To allow type-awareness of the vulcan ui-inputs, adds target type knowledge to the `Pick(FileOr)Folder` and optional send up of this knowledge.
  - To let the ui-inputs extend past input group boundaries which otherwise cut off dropdowns, often breaking usability, makes diablePortal={false} behavior optional for all involved Autocompletes.
  - Cleanup:
    - Fixes `PickFileOrFolder` allowFiles input to actually work
    - Use placeholder of disabled Autocomplete in `Pick(FileOr)Folder` when bucket not chosen, rather than just null / nothing
    - Fixes `PickBucket` to not give error before a bucket is chosen (via adding hidden '' option)
    - Have `PickBucket` only add paddingTop when given a label (Standard for all similar components I've made) 
- Vulcan/Archimedes Orchestration:
  - Adds pass through of metis_host to archimedes orchestrator and runners.***
  - So can use as primary inputs in vulcan workflow CWLs, adds "MetisFile", "MetisFolder", "MetisPath", and "MetisCSVorTSV" nominal types to etna cwl definition
  - minor cleanup of a variable-name typo, `commit_res*e*t_buttons`

PR To Dos:
- [x] actually add folder & file or folder variant inputs
- [x] Manual Testing - Fix dev connections
- [x] CI Tests for input validators
- [x] Implement properly in the Visualization workflow
  - [x] allow 'File' / 'Folder' type as primary inputs
    - [x] Let the Primary input box grow to space needed
    - [x] Remove need to double-commit sometimes or re-commit after reload
    - [x] ensure placeholder text-box stays aligned

***ToDo via portainer:
- Ensure metis host exists in the production vulcan config


dtm2451:
One BIG note (reminder to check, and do, if needed!): may need to add metis_host to the production vulcan config in portainer!  Otherwise, deploy likely breaks all vulcan steps that run a script.

dtm2451:
I think we discussed giving pointers to reviewers as focus/entry points...
Primarily, I'd love if you ran the workflow:
1. Put some tsv or csv in dev-metis for a project
2. Follow code edits in [this slack](https://immunox.slack.com/archives/C011CDG6469/p1694187951542769) to let the dev systems communicate and use proper targets.
3. Try out the "Visualization from File" workflow!

Some minor questions:
- I think these are fine/good, but could be worth a double-check: some changes here make slight alterations to the look of the polyphemus metis loader UI (definitely a change when model added before bucket choice: placeholder change for PickFolder; unlikely a change, but would be always: PickBucket paddingTop now only if label is given, but is 'Bucket' by default)
- For `metis_path_validators.tsx`, should validator for 'MetisPath' input still check that type is either 'file' or 'folder' and not a `null` meaning undetermined by the component?  My implementation skips the check, reasoning that the GUIs only allow for building of valid paths, rendering a check not required.

etna/packages/etna-js/components/metis_exploration.tsx
```diff
@@ -45,12 +45,19 @@ function simplifyFolderListReturn(fullList: any) {
...
+    folderContents.folders
+}
+
+export function PickBucket({ project_name=CONFIG.project_name, setBucket, bucket, label="Bucket", className, disablePortal=true}: {
   project_name?: string;
   setBucket: any;
   bucket: string;
   label?: string;
   className?: string;
+  disablePortal?: boolean;
```
maflaven, line 60: This will always be defined now that there's a default?
```suggestion
  disablePortal: boolean;
```

etna/packages/etna-js/components/metis_exploration.tsx
```diff
@@ -146,42 +163,82 @@ function FileOrFolderInner({ optionSet, path, target, setTarget, onEmpty, placeh
   />
 }
 
-export function PickFileOrFolder({ project_name=CONFIG.project_name, bucket, setPath, path, allowFiles=true, label, basePath, topLevelPlaceholder='', className}: {
+export function PickFileOrFolder({ project_name=CONFIG.project_name, bucket, setPath, path, useTargetType, allowFiles=true, label, basePath, topLevelPlaceholder='', className, disablePortal=true}: {
```
maflaven, line 166: Probably out of scope for this PR, but I tried adding a new file to a bucket after I already selected a file, and it didn't show up. Maybe add a refresh button somewhere in the input?

vulcan/lib/client/jsx/components/workflow/user_interactions/inputs/metis_items.tsx
```diff
@@ -0,0 +1,76 @@
...
+          setBucket={ (b: string) => {
+            updateKeyToVal('bucket', b)
+          }}
+          bucket={value.bucket}
+          label={label? label+', '+'Bucket' : undefined}
+          disablePortal={false}
+        />
+      </Grid>
+      <Grid>
+        <PickFileOrFolder
```
maflaven, line 52: Think it doesn't filter non-CSV/TSV files yet.

maflaven:
I tried changing the metis file to see if it triggers a refresh on the "Review your data frame" step. It only refreshes if the review step is committed.

image

Review step won't refresh in this case.

maflaven (approved):
The happy path looks great! I'm not familiar enough yet to speak to workflow setup. Just some minor UX edge cases.

etna/packages/etna-js/components/metis_exploration.tsx
dtm2451: fair!  Refreshing the page achieves this, but there's machinery built in to where this may not be hard to add in to the component itself so that you can refresh the `contentsSeen` without loosing all the other intermediate state that's lost with a page referesh.

vulcan/lib/client/jsx/components/workflow/user_interactions/inputs/metis_items.tsx
dtm2451: Right!  It doesn't.  Had meant to give an error via the python script that reads in the file, but forgot.  Not the best way anyway.

I wonder if there's a way to pass in a path regex from the CWL so that we don't have to make new input_type for all unique file typings we might want...

For now, I'll add the single one for this csv/tsv case & the check machinery via the validator, before merge.

dtm2451:
I'm confused by the error case you are seeing in https://github.com/mountetna/monoetna/pull/1249#issuecomment-1716105336 @maflaven.

Perhaps we can slack or zoom to figure out how you got to the state you are in as I can't understand how you got there.  It's confusing to me because the way things should work, and do for me in the gif attached, is that as soon as you 'commit' to using that 'Pancan_samp...' file, the workflow should refresh and clear or fill in next steps based on matching hashes with the new workflow state.  Admittedly, I could understand how this part of the state system works better, so I am very interested in understanding your workflow state as it seems like it does point to a bug.

For me, if I change the path / 'File or Folder' target & commit... Changing from the original "*.tsv" to the "*.csv" I had also previously used successfully, when I hit 'Commit' the "Review your data frame" remains because there is a past state matching the '*.csv' file.  (And it looks the same because the "*.tsv" and "*.csv" files are equivalent.)  If I instead change to the 'xneo1_diag__test__new.csv' file that I'd never used before, when I hit 'Commit' the "Review your data frame" step completely disappears as I expect it should.
file-choice-update

dtm2451:
Ohhhhhhh!  not a bug case... the state did update successfully =).  Notice in your image that it refers to 224 rows and 2 columns.  I think that matches your file?  Certainly doesn't match the test data I'd shared!

PR #1250, opened 2023-07-10
Fix polyphemus revision history

graft:
Revisions in polyphemus UI don't work currently because the UI chokes on lack of a date to display. Since "updated_at" was dropped from these histories (in favor of `version_number` and retaining `created_at`) the revisions had no date associated with them. Changed to use `created_at` instead.


dtm2451 (approved):
lgtm code-wise and testing -- confirmed I see the bug in my dev setup from 'master', and then that revisions successfully render with code from this branch.

PR #1251, opened 2023-07-11
clean Makefile-mac clutter mistakenly added in last PR

dtm2451:
Removes the unnecessary 'build-dev-deployer-images' build target that I'd added in my last PR

Also removes the comment about whether deployer building is needed, though we could revert.


dtm2451:
Also wanna note, I just didn't want this to go forgotten, but feel free to manage this however you'd like Amadeo!!!

amadeovezz:
thanks for putting this up dan!

PR #1253, opened 2023-07-20
Magma Gnomon Support -> Small fixes

amadeovezz:
# TODO
 
- [x] drop is_up_to_date restriction
- [x] allow pattern mode to auto find parent models
- [x] remove check for identifier presence for pattern mode validation
- [x] only run BaseAttributeValidation for no flags or none mode
- [x] refactor tests
- [x] stop auto parenting when a user explicitly wants to detach
- [x] manually test

## BONUS

- Add two additional make targets that help with local model creation
- Remove legacy Mac docs and add new instructions

### DIFF
--- a/magma/lib/magma/attributes/identifier_attribute.rb
+++ b/magma/lib/magma/attributes/identifier_attribute.rb
@@ -15,41 +15,34 @@ class Magma
     class Validation < Magma::Validation::Attribute::BaseAttributeValidation
       def validate(value, &block)
 
-        # Some identifier attributes have regexp validations, so we call super() to run those validations first
-        super if @attribute.validation
-
         gnomon_mode = Magma::Flags::GNOMON_MODE
         flag_value = @flags[gnomon_mode[:name]]
 
         # We skip validation if we have the gnomon flag or if it's value is set to none
-        return if flag_value.nil? or flag_value == gnomon_mode[:none]
+        if flag_value.nil? or flag_value == gnomon_mode[:none]
+          # Some identifier attributes have additional validations, so we call super() to run those validations first
+          super if @attribute.validation
+          return
+        end
 
         begin
         # Identifier mode - we just care about the existence of the identifier in the identifier table
           if flag_value == gnomon_mode[:identifier]
-            get_identifier(value)
+            identifier = Magma::Gnomon::Identifier.where(project_name: @model.project_name.to_s, identifier: value).first
+            raise "The identifier '#{value}' has not been assigned in Gnomon." unless identifier
           # Pattern mode - We must make sure the identifier conforms to a grammar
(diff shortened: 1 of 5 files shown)


magma/lib/magma/attributes/identifier_attribute.rb
```diff
@@ -15,43 +15,36 @@ def after_magma_model_set
...
+            raise "The identifier '#{value}' has not been assigned in Gnomon." unless identifier
           # Pattern mode - We must make sure the identifier conforms to a grammar
           elsif flag_value == gnomon_mode[:pattern]
-            identifier = get_identifier(value)
             grammar = Magma::Gnomon::Grammar.for_project(@model.project_name.to_s)
             raise "Grammar not found, identifier is: '#{value}'" unless grammar
 
-            rule = grammar.parser.rules[identifier.rule]
-            raise "No such #{identifier.rule} for #{@model.project_name.to_s}" unless rule
+            identifier_model = grammar.model_name(value)
```
graft, line 38: this method seems to do some parsing/ unrolling, why not just `@model.name` here?

amadeovezz: good catch, I was looking at stale test case, and thought I needed to infer it

magma/lib/magma/attributes/identifier_attribute.rb
```diff
@@ -15,43 +15,35 @@ def after_magma_model_set
...
           # Pattern mode - We must make sure the identifier conforms to a grammar
           elsif flag_value == gnomon_mode[:pattern]
-            identifier = get_identifier(value)
             grammar = Magma::Gnomon::Grammar.for_project(@model.project_name.to_s)
             raise "Grammar not found, identifier is: '#{value}'" unless grammar
 
-            rule = grammar.parser.rules[identifier.rule]
-            raise "No such #{identifier.rule} for #{@model.project_name.to_s}" unless rule
+            rule = grammar.parser.rules[@model.model_name.to_s]
+            raise "No such rule for #{@model.project_name.to_s}" unless rule
```
graft, line 39: We seem to have lost the rule name in this error which seems crucial.

amadeovezz: I don't know if we can still access the `identifier.rule`? In pattern mode, identifiers don't need to exist. The previous commit had access to `rule` cause we directly queried the `Identifier` table.  

In this case our rules are in the following hash from our `grammar`:

```ruby
 {"project"=> RuleDefinintion,
   "labor"=>RuleDefinition,
   "village"=>RuleDefinition,
   "victim"=>RuleDefinition}
```

And if we can't find a rule associated to the model, we inform the user that no rule exists for a specific model.

PR #1257, opened 2023-08-09
fix for missing models in Metis Linker

graft:
We noticed that if someone tries to run an empty config the metis linker barfs, a simple fix to change this.

Also changes `wait_for_downstream` and `depends_on_past` to be False, so failed tasks will not block subsequent runs (I think it will still wait for completion, i.e. it won't run two jobs at once). This might result in missed failures in the logs but seems better than having to scoop out failed tasks all the time. Individual config errors should still triage to the polyphemus logs.


amadeovezz:
lgtm

amadeovezz (approved):
lgtm

dtm2451 (approved):

PR #1260, opened 2023-08-14
Fix metis glob

graft:
This changeset repairs the Metis find api, which allows a "glob" search to be performed against file paths (although the api says "name", it matches full file paths). Previously this implementation seemed to only cover a subset of possible glob patterns, and thus did not appear to work in some instances.

The previous implementation made use of SQL queries and some Ruby logic to identify folder ids to search. The current implementation instead builds a folder tree for each bucket, constructs paths using this tree, and matches folder ids using regular expressions built from the globs. This is the same method I was using in Python in the MetisLinker code, so this works well. In practice it appears construction of a folder tree for thousands of folders is not at all expensive, so this is a fairly efficient method.

The glob implementation here supports `**` to match any number of folders and `*` to match any number of characters within a folder or file name.

Still requires manual testing to validate this works as expected in the Polyphemus UI.


amadeovezz:
@graft do you mind adding some manual testing instructions?

amadeovezz:
Metis:

Screenshot 2023-08-21 at 4 05 45 PM

Screenshot 2023-08-21 at 4 06 07 PM


File matching appeared to work for both directory `*.fastq.gz` and sub directory patterns `**.fastq.gz` . However, upon any subsequent attempts (`*.fastq.gz` or `**.fastq.gz`) both yield

Screenshot 2023-08-21 at 4 11 16 PM


Not sure if the results are cached, but I'd expect that glob patterns of `*.fastq.gz` should only give me files in the current directory?

metis/lib/globber.rb
```diff
@@ -1,6 +1,11 @@
 class Metis
   class Globber
-    def initialize(glob_string, is_file_glob)
+    STAR_MAGIC=SecureRandom.hex
```
amadeovezz, line 3: lol `STAR_MAGIC` ⭐

amadeovezz:
> > Not sure if the results are cached, but I'd expect that glob patterns of *.fastq.gz should only give me files in the current directory?
> 
> The glob is constructed like folder_path + '**/' + file_match, so it will match any files in the tree below folder_path.

I see, with the file hierarchy I listed above, I was able to use match patterns to just search the current directory though.
`*.fastq.gz` -> `EXAMPLE-HS1.etc.fastq.gz`and not `dummy2.fastq.gz`

PR #1264, opened 2023-08-22
Split vulcan ruby tests into separate workflows

dtm2451:
Goal: stop the constant test failures of our all our commits owing to hitting github action constraints rather than actually broken code

### DIFF
--- a/.github/workflows/branches.yml
+++ b/.github/workflows/branches.yml
@@ -47,3 +47,22 @@ jobs:
           IMAGES_POSTFIX: :master
         run: |
           make -C ${{ matrix.test-app }} release
+  test_vulcan:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: docker login
+        env:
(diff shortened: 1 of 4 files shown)


.github/workflows/branches.yml
```diff
@@ -47,3 +47,27 @@ jobs:
...
+          IS_CI: 1
+          CI_SECRET: ${{ secrets.CI_SECRET }}
+          IMAGES_POSTFIX: :master
+          RUN_E2E: 1
+        run: |
+          sudo rm -rf /usr/share/dotnet
+          sudo rm -rf /opt/ghc
+          sudo rm -rf /usr/local/lib/android
+          sudo rm -rf "/usr/local/share/boost"
+          sudo rm -rf "$AGENT_TOOLSDIRECTORY"
```
dtm2451, line 72: per same code in master.yml and also production.yml workflows.  4/5 of these lines align with https://github.com/actions/runner-images/issues/2840#issuecomment-790492173 workaround for getting extra space in runner images.

graft (approved):

PR #1266, opened 2023-09-12
Fix magma hold file

graft:
Currently Magma cannot start up correctly if hold_file is configured, this PR repairs this issue and adds some tests.

To review:
1. start magma up with hold_file defined in config.yml
2. test bin/healthcheck works as expected after calling Magma.instance.write_hold_file

### DIFF
--- a/magma/spec/magma_spec.rb
+++ b/magma/spec/magma_spec.rb
@@ -78,4 +78,22 @@ describe Magma do
       Labors::Monster.attributes.delete(:size)
     end
   end
+
+  context 'hold file' do
+    it 'writes the hold file' do
+      Magma.instance.write_hold_file
+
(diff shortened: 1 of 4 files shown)


magma/spec/magma_spec.rb
```diff
@@ -78,4 +78,50 @@ def reset_models project
...
+
+    it 'removes the hold file' do
+      Magma.instance.write_hold_file
+
+      Magma.instance.remove_hold_file
+
+      expect(::File.exists?(Magma.instance.config(:hold_file))).to be_falsy
+    end
+
+    it 'passes the healthcheck if hold file exists' do
```
maflaven, line 99:
```suggestion
    it 'passes the healthcheck if hold file exists and is not expired' do
```

maflaven (approved):

PR #1267, opened 2023-09-12
Gnomon - Batch Create and Browse

maflaven:
TODO
- [ ] batch create (probably replaces single create aka "compose") 
  - [ ] toolbar (leave export and filter to browse?) (collapsable?) 
    - [ ] add names
    - [ ] add from selection
    - [ ] replace value in selection
    - [ ] create all
    - [ ] delete selection
  - [ ] rule groups
    - [ ] rule toolbar (no filter?)
      - [ ] grow/shrink
      - [ ] add name
      - [ ] delete
      - [ ] name count/status
    - [ ] name form
      - [ ] toolbar
        - [ ] copy
        - [ ] delete
        - [ ] edit/lock button?
      - [ ] token read only
      - [ ] token dropdown
      - [ ] token counter
        - [ ] numeric input
        - [ ] +1 fn: needs to respect DB and what's on page
  - [ ] create dialogue
    - [ ] data table
      - [ ] name info
      - [ ] name creation status
    - [ ] create request to magma (needs new endpoint?)
    - [ ] bonus: look into sliding in view and fading in from right and former view fades out as it slides to left
- [ ] browse MVP
  - [ ] toolbar
    - [ ] export
    - [ ] filter
  - [ ] data table (needs new magma endpoint?)
- [ ] move starting points on project home page to navbar – home (includes decompose?), browse, create, decompose (better word), rules

### DIFF
--- a/Makefile-mac
+++ b/Makefile-mac
@@ -43,6 +43,9 @@ build-dev-archimedes-images:
 	 cd docker/archimedes-base && docker build . -t archimedes-base
 	 cd archimedes-r-base && docker build . -t archimedes-r-base # takes a really long time!
 	 cd docker/archimedes-node-base && docker build . -t archimedes-node-base
+	 cd archimedes && docker build . -t archimedes
+	 cd archimedes-r && docker build . -t archimedes-r
+	 cd archimedes-node && docker build . -t archimedes-node
 
 ### Project level targets ###
 


dtm2451:
One bit from the demo today that you mentioned yourself, when a name already exists the error note currently refers to "already exists in **magma**", but should refer to in the "database" instead.

gnomon/lib/client/jsx/components/names-create/rule-group-composer.tsx
```diff
@@ -0,0 +1,332 @@
...
+    readyCounts: {
+        fontWeight: 'bold',
+        textAlign: 'right',
+        '& .count-not-ready': {
+            color: 'red',
+        },
+    },
+    composersContainer: {
+        maxHeight: '50vh',
+        overflow: 'scroll',
```
maflaven, line 98:
```suggestion
        overflow: 'auto',
```

dtm2451:
Some quick testing went well for me just now after pulling the current code =)

Just one set of nitty new ideas regarding the grow / shrink buttons for both the create and browse pages that I think make their usage a bit more clear:
  - perhaps the icons can be swapped for their horizontal form, as that's the direction more affected.  Not super important... I think MUI only has the one direction pre-baked.
  - perhaps change the tooltip to "Show Button Labels" from the current "Grow"?
  - perhaps change the label to "Shrink Buttons" or "Hide Button Labels" on the "Shrink" button?

dtm2451 (approved):
did some minor code review too, seems pretty well organized to me!  I didn't look in too deep, but no specific quibbles.

maflaven:
> Some quick testing went well for me just now after pulling the current code =)
> 
> Just one set of nitty new ideas regarding the grow / shrink buttons for both the create and browse pages that I think make their usage a bit more clear:
> 
>     * perhaps the icons can be swapped for their horizontal form, as that's the direction more affected.  Not super important... I think MUI only has the one direction pre-baked.
> 
>     * perhaps change the tooltip to "Show Button Labels" from the current "Grow"?
> 
>     * perhaps change the label to "Shrink Buttons" or "Hide Button Labels" on the "Shrink" button?

I changed to "Show Labels" and "Hide Labels". Including "Button" would've made only that button have 3 lines of text, which makes all the other buttons a bit too big. Also rotated the icons 90° as suggested.

PR #1268, opened 2023-09-20
Add data_frame type to polyphemus metis loaders

dtm2451:
Adds ability to fill in attributes for multiple records from csv/tsv files + a map of column name <-> attribute pairs.

ToDo:
- [x] Build UI
  - [x] SelectAttributes input (in polyphemus)
    - [x] etna MUI Autocomplete input with property 'multiple' set to true
- [x] json schema for this loader type
- [x] python executor
  - [x] tests for this
- [x] Also work for Table Models
  - [x] UI additions
    - replace/append switch
    - identifier pointers  --> parent
  - [x] python additions
    - alternative paths / error catches


graft:
Testing by trying to put data into ipi demographics model, which is a table (somewhat of a strange case), I get:
`Job failed with error: Exception('MetisLoaderError("demographics-1.csv is missing column(s) targetted by demographic data_frame loader \'column_map\': id.")')`

I'm curious to see how the successful result arrives; I suspect it will be append-only (i.e., it will add items to the demographics table, not overwrite them). Might need an option to allow replacement-of or append-to existing table rows, although I suspect there will remain some further edge cases about this.

PR #1270, opened 2023-09-26
allow update pagination in etna-py magma client + paginate metis linker

graft:
This adds pagination to the `update` method in the Python magma client, and makes use of this in the Metis linker in airflow.

A few issues:
One problem is, I'm not sure what a decent page_size is, or whether this should be configurable by users. I guessed at 1000, but I'm not sure what this number should be. Different kinds of updates probably also have different update times; files, which require Magma phoning Metis, generally take a lot longer to handle than data attributes, so different scripts might have different page limits.

The update pagination also doesn't have a very nice sad path, i.e., if there are errors in the update, only some of them may be reported, and an incomplete/partial update may occur (that is, the pages, as far as Magma is concerned, are independent, and the notion of "pages" is only on the client side).

To test:
1. Run the Metis Linker on a set of files and validation completion of linking. See [here](https://github.com/mountetna/monoetna/pull/1175#issuecomment-1563250651) for how to setup Metis Linker in dev.
2. Change page_size at airflow/opt/providers/etna/etna/etls/metis_linker.py:268 to be less than the number of files and validate completion of linking.


amadeovezz:
is the motivation behind this PR to ease load of magma by chunking/paging updates together?

etna/packages/etna-py/src/mountetna/magma.py
```diff
@@ -287,7 +298,83 @@ def consume_pages():
 
     batch_size = 300
 
-    def update(self, update: UpdateRequest):
+    def _paged_update(self, update: UpdateRequest, page_size:int):
```
amadeovezz, line 301: This code is a bit tricky to follow. Perhaps it would be useful to:

- add some comments
- separate out some functionality, such as encapsulating the gathering all records to be updated and performing the actual updates
- refactoring how `results` is used. Maybe can be clarified with some renaming? For instance it is passed into `update_page()`, which is then fed into `add_to_page()`, but now `results` is treated as a `page` parameter?  Why are we re-adding to the page after we've performed the update?


The way I was originally trying to think about this code was:

1.  I have some records I want to update via pagination
2. I need to gather those records
3. I will then start to chunk those records and then send them to magma
4. I will collect the results of the updates (success/failures)
5. I will continue to paginate until I am done
6. Return the results

There are some elements of this, but I it could be improved?

I could be missing something though.

graft:
Yes; to clarify, when loading data for a particular project I found some file collections numbering in the thousands would result in the operation timing out. I was able to load the data by splitting up the upload using match/touch, but the process is difficult. This PR attempts to smooth that over by chunking updates to magma so updates do not time out.

etna/packages/etna-py/src/mountetna/magma.py
```diff
@@ -287,7 +298,83 @@ def consume_pages():
...
+
+                page = {}
+                page_count = 0
+
+        self._update_page(results, update.project_name, page)
+
+        return results
+
+    def _update_page(self, results, project_name, page):
+        page_results = self.update(UpdateRequest(project_name="ipi", revisions=page))
```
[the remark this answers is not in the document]
amadeovezz: oops to clarify, was wondering why we need `project_name="ipi"`

etna/packages/etna-py/src/mountetna/magma.py
graft: Ok, refactored a bunch. I also realized that as written the code would not deal correctly with tables, which are the main thing making this update difficult.

Basically, there are two ways to attach data in Magma. One is to set the `parent` attribute on a record, the other is to set the reciprocal `table` or `collection` attribute on the parent record. In the case of a `table` attribute, this latter kind of update will replace any existing table contents. In such a case, the update will look like: `{ parent_model: { parent_record: { table_att: [ table_record1, table_record2 ] }, table_model: { table_record1: {}, table_record2 }` - so, to attach successfully in these cases the update must include all of the table records at once. Thus, we have to ensure that all of a table falls in a single "page" of our update. In practice this means we can't really guarantee that there are `page_size` records in the page, since a table could have any number of records, so we simply do our best (i.e. stop when the page count is exceeded). Therefore, when we flatten, we group tables with their parent records and only paginate over the parents.

amadeovezz: ahh i see

amadeovezz: code looks cleaner!

amadeovezz:
Manual testing also passed!

Screenshot 2023-10-03 at 4 53 10 PM
Screenshot 2023-10-03 at 4 43 47 PM

PR #1274, opened 2023-10-23
Add flag api and tests

amadeovezz:
## What

This adds a Flag API to magma. Currently it supports:

- `get` requests. You can view all the available flags for a project
  - Via: `/flags/{project_name`
- `post` requests. You can either create or update flags. 
  - Via: `/flags/{project_name}` and submitting a parameter: `flags: [{flag_name: value}])`
  - We can delete flags when `{flag_name: nil}`

### Other noteworthy changes

We've decided that in order to create any flags in the Flag API, they must be registered in the `Flags::Module`.

There is a function called `is_registered?` that verifies this.

#### Testing impacts

The `is_registered` constraint makes testing the flag functionality a bit tricky. In particular - depending on what you are trying to test - you may need to use a helper function in `spec_helper` called `register_flag` that takes in a hash and registers it as a constant in `Magma::Flags`. 

If you use this function it is important to use `unregister_flags` at the end of your tests (or use it a `before block`)

## How to test

`rspec ./spec/flags_spec.rb`

### DIFF
--- a/magma/lib/magma/flag.rb
+++ b/magma/lib/magma/flag.rb
@@ -31,7 +31,7 @@ class Magma
       none: 'none'
     }
 
-    def self.flag_valid?(flag_name, flag_value)
+    def self.is_registered(flag_name, flag_value)
       # flag_hash: a string, flag_value: a string
       # ensures that a flag has the proper name and the values are defined
       constants.each do |constant|
--- a/magma/lib/magma/flag.rb
+++ b/magma/lib/magma/flag.rb
@@ -1,16 +1,20 @@
 class Magma
   class Flag < Sequel::Model
     # update_or_create is natively implemented in a later version of Sequel - I believe it can also be installed as a plugin
-    def self.update_or_create(project_name, flag_list)
-      flag_list.each do |flag|
-        flag.each do |flag_name, value|
-          existing_flag = self.where(project_name: project_name, flag_name: flag_name.to_s).first
-          if existing_flag
-            existing_flag.update(value: value)
-          else
-            self.create(project_name: project_name, flag_name: flag_name.to_s, value: value)
-          end
-        end
+    def self.update_or_create(update_hash, value_hash)
+      # handles updates, creates and deletes. Deletes when {value: nil}.
+      #
+      # update_key: a hash with two keys: {project_name: , flag_name: }
+      # value: a hash with one key: {value: }
+
+      existing_flag = self.where(update_hash).first
+      if existing_flag
+        # We delete flags set as nil and update the flags otherwise
+        value_hash[:value].nil? ? existing_flag.delete : existing_flag.update(value_hash)
+      else
+        # create a new row, make sure the value is not nil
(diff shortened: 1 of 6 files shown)


magma/lib/magma/server.rb
```diff
@@ -25,6 +26,9 @@ def initialize
 
     post '/update_model', action: 'update_model#action', auth: { user: { is_admin?: :project_name } }
 
+    get '/flags/:project_name/', action: 'flags#get', auth: { user: { can_view?: :project_name } }
+    post '/flags/:project_name/', action: 'flags#set', auth: { user: { is_admin?: :project_name } }
+
```
graft, line 31: not sure what if any effect of trailing / is here, but usually routes don't have this

magma/lib/magma/server/flag.rb
```diff
@@ -0,0 +1,18 @@
...
+    success_json({flags: @flags})
+  end
+
+  def set
+    require_param(:flags)
+    Magma::Flag.update_or_create(project_name, @params[:flags])
+    success_json({success: 200})
+  end
+
+  def project_name
```
graft, line 14: probably redundant with @project_name, which is already set

magma/lib/magma/flag.rb
```diff
@@ -1,5 +1,18 @@
 class Magma
   class Flag < Sequel::Model
+    # update_or_create is natively implemented in a later version of Sequel - I believe it can also be installed as a plugin
+    def self.update_or_create(project_name, flag_list)
```
graft, line 4: This interface is not what i'd expect for update_or_create - in particular the flag_list probably belongs back in the #set controller so the format of params[:flags] is evident in that controller; you might reduce this to lines 7-12 and just take a single hash argument, which would be the standard interface for #update_or_create i think.

graft:
Looks nice, the tests are great. I suggested a few refactors for legibility/convention.

graft:
One question - were you going to add post NULL to remove the flag?

magma/lib/magma/server/controller.rb
```diff
@@ -3,6 +3,7 @@ class Controller < Etna::Controller
     def initialize(request, action=nil)
       super
       @project_name = @params[:project_name]
+      @flags = Magma::Project.flags(@project_name)
```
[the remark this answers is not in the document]
amadeovezz: good call

amadeovezz:
yep I was going to make it such that any `post` request:

`{flag_name: 'my_flag', value: null}` should be interpreted as a delete.

Thoughts?

magma/lib/magma/flag.rb
maflaven, line 4: Should this also include some check that the flag name and value are valid for production (i.e., anything in `module Flags` on l.18)?

amadeovezz: this is a good question. Lets chat about this briefly in stand-up.

maflaven (approved):
lgtm

maflaven (approved):

PR #1276, opened 2023-10-25
Update data-frame loader

dtm2451:
## This PR:
- adds a 'hole_value' textfield to the data-frame loader config
  - Relatedly, turns off pandas' default na filtration when the data_frame is loaded in to not create additional, hidden to the user, 'hole_value' ignorance.  Importantly, this means that this loader can functionally be used to fill attributes as "NA" / "N/A" / "null" if wanted by the user.  All of these examples are treated as missing data by pandas.read_table in the default behavior that's turned off here.
- Adjusts the labeling of the 'blank_table' switch

### 'hole_value' UI:
image

### Previous 'blank_table' look:
non-table model
image
table model
image

### NEW 'blank_table' look:
non-table model
image
table model
image

### DIFF
--- a/airflow/opt/providers/etna/etna/etls/metis_linker.py
+++ b/airflow/opt/providers/etna/etna/etls/metis_linker.py
@@ -121,7 +121,7 @@ class MetisLoaderConfig(EtlConfigResponse):
             if isTable:
                 data['__temp__']=['::temp-id-' + str(temp) for temp in data.index]
                 data = data.set_index('__temp__', drop=True)
-                if script.get(['blank_table'], False):
+                if script.get('blank_table', False):
                     for parent_name in set(list(data[template.parent])):
                         if not self.get_identifier(template.parent, parent_name):
                             continue
--- a/airflow/opt/providers/etna/etna/etls/metis_linker.py
+++ b/airflow/opt/providers/etna/etna/etls/metis_linker.py
@@ -102,8 +102,13 @@ class MetisLoaderConfig(EtlConfigResponse):
                     else:
                         # Attempt to use pandas' automated detection
                         separator = None
-                # Maybe Future Feature: expose additional blanker value in config ui
-                data = pandas.read_table(file_reader, sep=separator, engine = 'python').replace({numpy.nan: None})
+                # Load dataframe
+                data = pandas.read_table(file_reader, sep=separator, engine = 'python', na_filter=False)
+                # Blank data equaling hole_value or empty string (None's ignored from the update-build later on)
+                replacements = {'': None}
+                if 'hole_value' in script:
+                    replacements[script['hole_value']] = None
+                data=data.replace(replacements)
             if len(data.columns) < 2:
                 raise MetisLoaderError(f"{file.file_name} seems to have fewer than 2 columns. Check the 'format' configuration for this data_frame loader.")
             if not set(columns).issubset(data.columns):
--- a/airflow/opt/providers/etna/etna/etls/metis_linker.py
+++ b/airflow/opt/providers/etna/etna/etls/metis_linker.py
@@ -122,7 +122,7 @@ class MetisLoaderConfig(EtlConfigResponse):
             if isTable:
                 data['__temp__']=['::temp-id-' + str(temp) for temp in data.index]
                 data = data.set_index('__temp__', drop=True)
-                if script['blank_table']:
+                if script.get(['blank_table'], False):
                     for parent_name in set(list(data[template.parent])):
                         if not self.get_identifier(template.parent, parent_name):
                             continue
--- a/airflow/opt/providers/etna/etna/etls/metis_linker.py
+++ b/airflow/opt/providers/etna/etna/etls/metis_linker.py
@@ -93,12 +93,12 @@ class MetisLoaderConfig(EtlConfigResponse):
                 if script['format']=="csv":
                     separator = ","
                 if script['format']=="tsv":
-                    separator = "\\t"
+                    separator = "\t"
                 if script['format']=="auto-detect":
                     if re.search("csv$", file.file_path) is not None:
                         separator = ","
                     elif re.search("tsv$", file.file_path) is not None:
-                        separator = "\\t"
+                        separator = "\t"
                     else:
                         # Attempt to use pandas' automated detection
                         separator = None
--- a/airflow/opt/providers/etna/etna/etls/metis_linker.py
+++ b/airflow/opt/providers/etna/etna/etls/metis_linker.py
@@ -104,11 +104,6 @@ class MetisLoaderConfig(EtlConfigResponse):
                         separator = None
                 # Load dataframe
                 data = pandas.read_table(file_reader, sep=separator, engine = 'python', na_filter=False)
-                # Blank data equaling hole_value or empty string (None's ignored from the update-build later on)
-                replacements = {'': None}
-                if 'hole_value' in script:
-                    replacements[script['hole_value']] = None
-                data=data.replace(replacements)
             if len(data.columns) < 2:
                 raise MetisLoaderError(f"{file.file_name} seems to have fewer than 2 columns. Check the 'format' configuration for this data_frame loader.")
             if not set(columns).issubset(data.columns):
@@ -118,6 +113,10 @@ class MetisLoaderConfig(EtlConfigResponse):
                 raise MetisLoaderError(f"{file.file_name} has unexpected NA values after all parsing. Data rows may be shorter than the column row indicates.")
             # Trim to mapped columns and convert to attribute names
             data = data.rename(columns={v: k for k,v in column_map.items()})[attributes]
+            # Blank data equaling values_to_ignore by setting as None (None's are ignored when updates are constructed)
+            if 'values_to_ignore' in script:
+                replacements = {k: None for k in script['values_to_ignore'].split(',')}
+                data=data.replace(replacements)
             # Determine Updates
             if isTable:
                 data['__temp__']=['::temp-id-' + str(temp) for temp in data.index]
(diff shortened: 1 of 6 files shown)


airflow/opt/providers/etna/etna/etls/metis_linker.py
```diff
@@ -102,8 +102,13 @@ def data_frame_update(self, model_name, script, tail, update, metis, model):
                     else:
                         # Attempt to use pandas' automated detection
                         separator = None
-                # Maybe Future Feature: expose additional blanker value in config ui
-                data = pandas.read_table(file_reader, sep=separator, engine = 'python').replace({numpy.nan: None})
+                # Load dataframe
+                data = pandas.read_table(file_reader, sep=separator, engine = 'python', na_filter=False)
+                # Blank data equaling hole_value or empty string (None's ignored from the update-build later on)
+                replacements = {'': None}
+                if 'hole_value' in script:
```
amadeovezz, line 109: I think it is more explicit to have `data=data.replace(replacements)` nested under the if statement. It is also a bit unusual to set a Dict key as a empty string? Are you worried about bad input data? 

```python
if 'hole_value' in script:
    replacements = {script['hole_value']: None}
    data=data.replace(replacements)
```

amadeovezz (commented):
A couple of thoughts:

- I'm not sure `hole_value` is a very intuitive term for a user. Something like `value_to_skip`, `avoid_value`, `ignore_value`
- I actually think it would be useful if the user could pass a comma separated list of values they want to ignore. Not sure if this is out of scope for a V1 though.
- I am running into issues testing this locally, I've never linked a data frame before so it is probably something on my end. Getting: `MetisLoaderError('Invalid type for script data_frame for model population')`.

airflow/opt/providers/etna/etna/etls/metis_linker.py
graft: These seem like slightly different behaviors: the first is, always replace `''` with `None`, and also replace hole_value with `None`. The revision is only replace hole_value with `None`. However, the issue may be moot if pandas is already transforming `''` (i.e. what is between `,,` in a csv) into `None`

dtm2451: > * I'm not sure `hole_value` is a very intuitive term for a user. Something like `value_to_skip`, `avoid_value`, `ignore_value`

Good point.  It was called `hole_value

> * I actually think it would be useful if the user could pass a comma separated list of values they want to ignore. Not sure if this is out of scope for a V1 though.
> * I am running into issues testing this locally, I've never linked a data frame before so it is probably something on my end. Getting: `MetisLoaderError('Invalid type for script data_frame for model population')`.

dtm2451: > * I'm not sure `hole_value` is a very intuitive term for a user. Something like `value_to_skip`, `avoid_value`, `ignore_value`
> * I actually think it would be useful if the user could pass a comma separated list of values they want to ignore. Not sure if this is out of scope for a V1 though.

Fair points!  It was called `hole_value` in the etna gem version (`etna administrate models attributes update_from_csv __project_name__ __model_name__ __filepath__ --hole-value`), but I'm on board to give a better name, and like `values_to_skip`!  Multi-value was a thought I'd had at one point too.  Good poke -- I think I can code it easily enough -- and I like the comma separated list as the ui.

dtm2451: > * I am running into issues testing this locally, I've never linked a data frame before so it is probably something on my end. Getting: `MetisLoaderError('Invalid type for script data_frame for model population')`.

Do you have an empty loader definition for the population model?  This seems like the catch for a barely initialized script: https://github.com/mountetna/monoetna/blob/6f6f89b17490d90bb6a97f87ab438ba2220c2fd1/airflow/opt/providers/etna/etna/etls/metis_linker.py#L187

dtm2451: Re this and Saurabh's addition:
> I think it is more explicit to have `data=data.replace(replacements)` nested under the if statement. It is also a bit unusual to set a Dict key as a empty string? Are you worried about bad input data?

Originally, I was wanting to just ignore empty values, ie what is between `,,` in a csv as Saurabh put it.  AFAICT from the read_table docs, and now also from double-checking, emptiness is not ignored by default with a switch to having `na_filter=False` in the line 106 at the top of the preview here.

BUT re-thinking, the empty string is our blank value for string type data so always ignoring '' would block functionality of blanking string attributes from this path.

Testing of the behavior:
```
>>> df = pd.read_table('table.csv', sep = ',')
>>> df
          rec  parent comp
0  XX-YY1-ZZ1     NaN  abc
1  XX-YY1-ZZ2     NaN  abc
2  XX-YY1-ZZ3     NaN  abc
>>> df = pd.read_table('table.csv', sep = ',', na_filter=False)
>>> df
          rec parent comp
0  XX-YY1-ZZ1         abc
1  XX-YY1-ZZ2         abc
2  XX-YY1-ZZ3         abc
```

dtm2451: Per this discussion:
- changed the name to values_to_ignore
- moved the `data.replace(replacements)` into the if statement as it will only be done if the config has this element. (a.k.a. empty strings are not always removed & allows for old configs that don't have this non-required element to still work)
- updated the usage to be a comma separated list
- added a display of the parsed targets where 'empty string' is used to stress that target =)

## New UI look
`,NA,NULL,<space>`
image

amadeovezz: oh my bad - I now see what your intention was. 

> BUT re-thinking, the empty string is our blank value for string type data so always ignoring '' would block functionality of blanking string attributes from this path.

Good point, if you wanted to making ignoring empty values optional, you could make a toggle?

airflow/opt/providers/etna/etna/etls/metis_linker.py
```diff
@@ -93,29 +93,35 @@ def data_frame_update(self, model_name, script, tail, update, metis, model):
...
-                data = pandas.read_table(file_reader, sep=separator, engine = 'python').replace({numpy.nan: None})
+                # Load dataframe
+                data = pandas.read_table(file_reader, sep=separator, engine = 'python', na_filter=False)
             if len(data.columns) < 2:
                 raise MetisLoaderError(f"{file.file_name} seems to have fewer than 2 columns. Check the 'format' configuration for this data_frame loader.")
             if not set(columns).issubset(data.columns):
                 missing = ', '.join([col for col in columns if col not in data.columns])
                 raise MetisLoaderError(f"{file.file_name} is missing column(s) targetted by {model_name} data_frame loader 'column_map': {missing}.")
+            if pandas.isna(data).values.any():
+                raise MetisLoaderError(f"{file.file_name} has unexpected NA values after all parsing. Data rows may be shorter than the column row indicates.")
```
dtm2451, line 113: Encountered a couple other errors during testing:
- Discovered from an unfortunately mis-formatted target file that `pandas.read_table()` does not error if your data rows are shorter than your header.  Instead, it fills the missing columns with `numpy.NaN`.  So **added a check and error message for presence of na values**.

airflow/opt/providers/etna/etna/etls/metis_linker.py
```diff
@@ -93,29 +93,35 @@ def data_frame_update(self, model_name, script, tail, update, metis, model):
...
                     separator = ","
                 if script['format']=="tsv":
-                    separator = "\\t"
+                    separator = "\t"
                 if script['format']=="auto-detect":
                     if re.search("csv$", file.file_path) is not None:
                         separator = ","
                     elif re.search("tsv$", file.file_path) is not None:
-                        separator = "\\t"
+                        separator = "\t"
```
dtm2451, line 101: Encountered a couple other errors during testing:
- For whatever reason, this "\\t" separator mostly works but can cause the rightmost data column to fill as NaN even with `na_filter=False`.  I'm not sure why I'd had "\\t" rather than "\t" anymore, so since "\t" seems to work better, switching to that here and for the analogous vulcan code too.

airflow/opt/providers/etna/etna/etls/metis_linker.py
amadeovezz: or empty string works in the description as you have! 

Was worried users would pass in a range of different inputs to mean `''` , but it is pretty explicit in your description

airflow/opt/providers/etna/etna/etls/metis_linker.py
```diff
@@ -93,29 +93,35 @@ def data_frame_update(self, model_name, script, tail, update, metis, model):
...
+            # Blank data equaling values_to_ignore by setting as None (None's are ignored when updates are constructed)
+            if 'values_to_ignore' in script:
+                replacements = {k: None for k in script['values_to_ignore'].split(',')}
+                data=data.replace(replacements)
             # Determine Updates
             if isTable:
                 data['__temp__']=['::temp-id-' + str(temp) for temp in data.index]
                 data = data.set_index('__temp__', drop=True)
-                if script['blank_table']:
+                if script.get(['blank_table'], False):
```
dtm2451, line 124: Just noting: because 'blank_table' was not made required in the json schema for these loader configs (see bottom of [this change](https://github.com/mountetna/monoetna/pull/1276/files#diff-6a2c12f626c0043442f4035894b6e7b85b533cefa2652f0c51250d05a21aeaa4)), better to use the get function and have an explicit default.

airflow/opt/providers/etna/etna/etls/metis_linker.py
dtm2451: > Was worried users would pass in a range of different inputs to mean '' , but it is pretty explicit in your description

Same!  Primarily came up with the targets display because I was also worried users might not understand either 1) how to target empty strings, or 2) that the empty field default translates to an empty string!

amadeovezz (approved):
LGTM!

PR #1277, opened 2023-10-27
bulk create API

graft:
This adds a bulk generate API for Gnomon, which mostly behaves like the existing generate API, except that it accepts a list of names and rules and returns only names that were generated (or various 422s).

### DIFF
--- a/magma/spec/gnomon_spec.rb
+++ b/magma/spec/gnomon_spec.rb
@@ -439,6 +439,83 @@ describe GnomonController do
     end
   end
 
+  context 'bulk generate API' do
+
+    it 'throws exception when no grammar for project' do
+      auth_header(:admin)
+      json_post('/gnomon/labors/generate', names: [])
+      expect(last_response.status).to eq(422)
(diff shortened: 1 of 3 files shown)


maflaven (approved):
lgtm

magma/spec/gnomon_spec.rb
```diff
@@ -439,6 +439,83 @@ def app
...
+
+    it 'throws exception when no grammar for project' do
+      auth_header(:admin)
+      json_post('/gnomon/labors/generate', names: [])
+      expect(last_response.status).to eq(422)
+      expect(json_body[:error]).to eq('No grammar found for project labors.')
+    end
+
+    context 'with grammar' do
+      before(:each) do
```
amadeovezz, line 452: is `:each` needed here? Isn't the default behavior for each spec? Unless you like the explicitness

graft: hmm, is that the default behavior? I didn't know that... but i like the explicitness anyway

PR #1280, opened 2023-11-08
WIP: Vulcan & Gnomon ADRs

graft:
First pass at a Vulcan ADR, Gnomon ADRs coming soon. Please comment and ask questions.


decisions/adr-0007-vulcan-workflows.md
```diff
@@ -0,0 +1,60 @@
...
+
+### UI Queries
+
+Currently the Vulcan orchestrator has special hooks for "ui query" steps. Since we cannot expect the Snakemake orchestator to understand this logic, we will instead separate these out into a new set of Vulcan endpoints, which allow a workflow to set a requirement for a UI query associated with a particular step in the workflow. To make use of this interface, a step in a Snakemake workflow which expects to output a UI query JSON (i.e., the JSON response from the user) will use a vulcan client interface to set a query requirement and exit, halting the workflow due to the unsatisfied output requirement. The Vulcan UI will then be able to determine an unresolved query and will show it to the user in the UI, making use of the existing query interfaces to resolve the query. This results in the response being written to the appropriate output location, and the Snakemake workflow resumes with the requirement satisfied.
```
dtm2451, line 40: > a step in a Snakemake workflow which expects to output a UI query JSON (i.e., the JSON response from the user) will use a vulcan client interface to set a query requirement and exit.

Better if the run blockage can come from the other end -- reaching a step that requires **input** from UI query's output JSON.  There are cases currently where inputs for UI queries are prepped by an early step, yet the ui-query output is not needed until additional steps are completed.  Stopping the execution only at this stage might save a significant amount of users' time.

Perhaps getting to the point where a UI query could be presented could still trigger the UI to be presented to the user?  And execution would only be stopped if the user doesn't confirm UI selections before the workflow gets to a step that requires them?

decisions/adr-0007-vulcan-workflows.md
```diff
@@ -0,0 +1,60 @@
...
+
+### Project workflows
+The Vulcan API will allow creating and listing project workflows. Each workflow can be also be browsed via the Vulcan UI. Workflows can still be run to produce Figures in Vulcan, which allow the workflow to be parameterized and run via the Vulcan UI. The Vulcan "steps" UI will be improved to monitor progress of a running workflow and show logging.
+
+The Vulcan API will also expose a git remote endpoint for each project workflow. When the workflow is created, it will be initialized to an empty tree or forked from an existing workflow. This is a normal git remote and can be pushed/pulled from as desired; the endpoint can possibly validate the push using an update hook.
+
+### Using Snakemake for workflows
+
+The format of a Vulcan workflow is a Snakemake workflow. This may include steps in python, R, etc. Workflows may define containers via a Snakefile and conda yaml. Vulcan will keep track of containers and re-use them across workflows.
```
dtm2451, line 32: perhaps nit-level, and perhaps not the exact intent here, but conda recipes for R packages can lag behind what's released in CRAN and Bioconductor.  conda maintenance of R setups isn't as common as for python.

decisions/adr-0007-vulcan-workflows.md
```diff
@@ -0,0 +1,60 @@
...
+
+### Using Vulcan via the Command-Line
+
+Although it may eventually be attractive to run a given workflow in the Vulcan UI, the development loop in the UI is difficult. Files cannot be inspected directly, and other utilities, REPLs, etc, may not be accessible. Although the use of Snakemake workflows makes it easier to transfer Vulcan workflows to other environments and run them there, many dependencies (containers and inputs) may be more difficult to achieve than a git clone. To allow easier development, Vulcan will include a command-line client, which will run the workflow's Snakefile from the command-line in a given environment, allowing users to use Vulcan containers and attach inputs in local folders. Similarly the command-line client will allow a user to resolve a UI query step in the shell by posting the appropriate output JSON to the correct location.
```
dtm2451, line 44: >Similarly the command-line client will allow a user to resolve a UI query step in the shell by posting the appropriate output JSON to the correct location

These JSONs can sometimes be VERY hard to create from scratch, and creation from scratch obfuscates all gotcha-catches and restrictions that an actual UI might have built in.  See below for an example.  **I think a browser-based (or something else that can render react) tool for UI input generation might be required.**  This could still be launched from the command line, and inherit its inputs by default.

<details>
 <summary>E.g. plot setups can be VERY large and complex:</summary>

AND, to name one of many quirks that the VizUI system handles, the `"plot_type": "dittoFreqPlot"` selection gets used by the VizUI system to determine what all the other fields can be because each plot types uses a different subset of the possible fields

```
plot_setup/plot_setup: {
  "plot_type": "dittoFreqPlot",
  "var": "wsnn.harmony.int_res.0.1",
  "group_by": "arm",
  "color_by": "make",
  "sample_by": "CoLabs_patient",
  "plots": [
    "jitter",
    "boxplot"
  ],
  "scale_by": "fraction",
  "plot_title": "make",
  "plot_subtitle": "make",
  "legend_title": "make",
  "xlab": "make",
  "ylab": "make",
  "group_order": "make",
  "vars_use": [],
  "cells_use": {},
  "vlnplot_lineweight": 1,
  "vlnplot_width": 1,
  "vlnplot_scaling": "area",
  "boxplot_width": 0.2,
  "boxplot_color": "black",
  "boxplot_fill": true,
  "boxplot_lineweight": 1,
  "jitter_size": 1,
  "jitter_width": 0.2,
  "jitter_color": "black",
  "ridgeplot_lineweight": 1,
  "legend_show": true,
  "split_adjust_free_y": true
}
```

</details>

dtm2451:
I've added a lot of my open Qs now, but **I do wanna convey that I am behind this idea in general**.  Just wanna point out all the holes I can so that we can think through them properly!

decisions/adr-0007-vulcan-workflows.md
graft: Currently there are empty "ui-query" steps which contain no code, but cause the UI to show the inputs to the step and massage through a query. I'm imagining that instead of being blank, this step would have some code calling a client to set the input requirement; the halting would thus occur in the same place in the workflow.

decisions/adr-0007-vulcan-workflows.md
graft: Yeah, I'm fretting about this. I'd say R package management and containerization is "not good" in general. The conda yml has some advantages, but nothing precludes some other mechanism for defining containers being used and the resulting image referenced in the Snakefile.

decisions/adr-0007-vulcan-workflows.md
graft: Fair - I thought that the command-line client could provide some help here, like an empty JSON template you can fill out. We can also work on simplifying/clarifying this format so it is easier to generate.

decisions/adr-0007-vulcan-workflows.md
dtm2451: Definitely a fair read on R package menagement and containerization.  I think conda could work out okay, and with other languages in mind I agree it's the right place to start.  

The primary bit I'd like to be sure we can do, is grab most recent versions when I'm doing side-by-side workflow & dittoVerse dev.  This generally would extend past 'released' versions anyway, so an Rscript system that installes dittoViz & dittoSeq, when needed, from github could suffice.

graft:
@amadeovezz do you think this is current with the WIP vulcan stuff? If so I will merge for the sake of record. The lines about a git remote might be updated, perhaps.

amadeovezz:
I think this is mostly current and captures many of the high level ideas of Vulcan V2. I would merge for record (and we can always iterate later).

My issue (https://github.com/mountetna/monoetna/issues/1311) about the Vulcan config is more specific, but I'm happy to convert it / or even combine it with this ADR once we discuss and converge on a decision.

PR #1282, opened 2023-11-16
make minimum settable run interval 600 seconds for polyphemus loaders

dtm2451:
Pretty minimal PR.

Implemented by using `max(user-input, 600)` as the new value, rather than allowing any user input.

Alternative strategies that could let users type more freely would be (downsides in parentheses):
1. adding a debounce before the new value is attempted to be used (might give a feeling of lag?)
2. letting users type freely but adding a "Set" or "confirm" button (requires an extra click that might be forgotten -> probably need to add red error coloring or some other highlight to ensure the user knows they haven't set this value)

Edit: After some discussion, decided instead on:
- essentially 2 above, but making use of the `Save` button that the user already needs to click as the "confirm".  When clicked, if the value is too small, saving is blocked and an error / text message appears to the right of the Save and Reset buttons.


dtm2451:
EDIT: 600 seconds = 10minutes :facepalm:
Intention is **300** / 5min

polyphemus/lib/client/jsx/etl/run-pane.tsx
```diff
@@ -247,18 +244,18 @@ const RunPane = ({
...
                 endAdornment: (
-                  <InputAdornment position='end'>seconds</InputAdornment>
+                  <InputAdornment position='end'>{'seconds (300 minimum)'}</InputAdornment>
                 )
               }}
               value={runIntervalTime}
               onChange={(e) => {
-                setRunIntervalTime(parseInt(e.target.value as string));
+                setRunIntervalTime(Math.max(300,parseInt(e.target.value as string)));
               }}
```
maflaven, line 259: Could put the restriction on the input instead.
```suggestion
                setRunIntervalTime(parseInt(e.target.value as string));
              }}
              inputProps={{
                min: 300,
              }}
```

maflaven:
> Pretty minimal PR.
> 
> Implemented by using `max(user-input, 600)` as the new value, rather than allowing any user input.
> 
> Alternative strategies that could let users type more freely would be (downsides in parentheses):
> 
> 1. adding a debounce before the new value is attempted to be used (might give a feeling of lag?)
> 2. letting users type freely but adding a "Set" or "confirm" button (requires an extra click that might be forgotten -> probably need to add red error coloring or some other highlight to ensure the user knows they haven't set this value)

re: 2, I think this is what you're suggesting: https://v4.mui.com/components/text-fields/#validation. Is this input part of a form? Could disable form submission if there are any errors.

maflaven (approved):

dtm2451:
> > Pretty minimal PR.
> > Implemented by using `max(user-input, 600)` as the new value, rather than allowing any user input.
> > Alternative strategies that could let users type more freely would be (downsides in parentheses):
> > 
> > 1. adding a debounce before the new value is attempted to be used (might give a feeling of lag?)
> > 2. letting users type freely but adding a "Set" or "confirm" button (requires an extra click that might be forgotten -> probably need to add red error coloring or some other highlight to ensure the user knows they haven't set this value)
> 
> re: 2, I think this is what you're suggesting: https://v4.mui.com/components/text-fields/#validation. Is this input part of a form? Could disable form submission if there are any errors.

Not part of a form, no.

And thinking it's not really that annoying here to not be able to, even just temporarily, enter anything less than 300.

polyphemus/lib/client/jsx/etl/run-pane.tsx
maflaven: Just tested this and what you had before is better. `min` in inputProps only enforces if you press the up/down buttons or arrow keys to change the value.

maflaven (approved):

dtm2451:
@graft requested you here for thoughts on how the new methodology (error on click of the save button and block actual save) does not utilize the greater error reporting method of EtlConfig.  Merging in with that larger system would take a bit of extra work, but could be done.

PR #1283, opened 2023-11-20
Add dittoDotPlot to Vulcan scViz workflow

dtm2451:
This PR:
- Creates new versions of `SelectAutocompleteInput` and `NestedSelectAutocompleteInput` inputs which make use of the mui Autocomplete's `multiple` option.  The value output changes from string to string[] with this change, so though it easier to keep cleanly separated as new functions than to manage via an new input to the original functions.  Named `SelectAutocompleteMultiPickInput` and `NestedSelectAutocompleteMultiPickInput`.
- Implements the `NestedSelectAutocompleteMultiPickInput` as a "piece" so could add to the VizUI
- Then implements dittoDotPlot in the VizUI to seemlessly add it to vulcan's scViz workflow.
- Reorder-ability:
  - Renames the VizUI-specific `ReorderPiece` as `ReorderVizPiece`
  - Creates a new `ReorderCollapsiblePiece` and implements that alongside the multi-pick piece as `nestedDropdownMultiPickAndReorderPiece`.
    - This is what's actually used for the Feature / Primary Data selection of dittoDotPlot in the VizUI


dtm2451:
I think this is ready for review!

To test:
1. You'll need a project with an sc_seq_dataset model and an sc_seq_dataset to test with.  Ping me for one, or use [this dataset from xneo1](https://timur.ucsf.edu/xneo1/browse/sc_seq_dataset/XNEO1-SCD2#overview) by recreating the record in a dev project. (OR you can have dev vulcan talk with production metis/magma, but I'm rusty about the details of that setup at the moment)
2. Create a scViz figure in vulcan, for that project, and select this dataset in the 'Object Selection' Primary Input.
3. In the Set Plot Options step, choose 'dittoDotPlot' as plot type, then test away!
    - the new input is used in selection of 'Primary Data', but only for the  'dittoDotPlot' plot type

One note / limitation I'm feeling 'okay' with:
In my opinion, the current form is usable but does feel a little clunky because feature selection is slower than ideal.  I'd like to create some ways to fix the slow/clunk, but they're not small adds.  Feel free to contradict or suggest alternatives, but these are my additional ideas for **a future PR**:
- a copy/paste entry method via a button & pop-out modal
- a DGE/dataframe-based entry method where a particular column's data can be used as values, perhaps after trimming based on values of other columns (e.g. a fold change column being greater than 1.5, or a p-value column being less than 0.05)
- both would require a confirm button followed by a parse step that warns about any values not found among the available options.

maflaven (approved):
One nit: it wasn't obvious that I could expand the `Primary Data - reorder` section. Also not obvious that you can drag+drop each item. Maybe add a drag handle icon beside each item? 
image

Everything else looks good re UI! I can't really speak to plotting capabilities though.

PR #1284, opened 2023-11-22
"bug"-fix for timur-map reparent button - disable rather than hide

dtm2451:
Another quick one, this PR just switches the reparent button behavior from starting out as hidden -> starting out as disabled (until after a magma request allows confirmation of whether the model can be moved).

Why? Both Saurabh and Dan separately thought the button was bugged when it didn't appear in certain cases.  And we should know better than users =p.  I think the 'shown, but disabled' behavior makes it a bit clearer that the action is known and either being checked for usability or deemed non-usable, rather than just being bugged out when not shown!


maflaven (approved):

graft:
Maybe add a tooltip if disabled explaining why it is disabled, like "Cannot reparent a model containing records"

dtm2451:
Updated the past tests for this new disabled behavior, but will need to wait til after Thanksgiving to add a test that the button stays disabled when a model has records.

PR #17, opened 2020-07-09
Zc/uploader fixes

corps:
Extension of previous PR. 

Bit of a larger refactor of uploader here, but it hums a lot better.

1.  Much deeper testing of all cases, including error handling, retry, queueing, and worker integration.
2.  Removal of a ton of code no longer necessary
3.  uploader is source of truth for all upload behavior and state; clients only need to subscribe and follow state changes, not much else.
4.  Pauses and other interactions aid dispatched to the service worker, who will dispatch back the updated state itself.

PR #1286, opened 2023-12-11
This PR attempts to fix a build issue within swarm-keeper

dtm2451:
Details:
A new build error surfaced today after @dtm2451 merged a new PR into master, [erroring here](https://github.com/mountetna/monoetna/actions/runs/7170020662/job/19521772280#step:4:13924) in a line of the swarm-keeper Dockerfile.  Investigation led towards a suggestion that the issue may come from the alpine image starting point newly adopting [PEP668](https://peps.python.org/pep-0668/).  Indeed, there was a new alpine image version pushed 4 days ago, with major version number incremented, between our [last successful build-and-push job run](https://github.com/mountetna/monoetna/actions/runs/7118436169/job/19381384344) 5 days ago, and the current run.

Attempted fix then:
- pinning on the previous image version number

An alternative fix we can try if we want to continue receiving alpine image updates would be to instead try adding `--break-system-packages` to the problematic pip call.

### DIFF
--- a/swarm/swarm-keeper/Dockerfile
+++ b/swarm/swarm-keeper/Dockerfile
@@ -4,7 +4,7 @@ FROM etna-base-dev
 FROM alpine
 RUN mkdir /app
 RUN apk add --no-cache bash curl jq httpie py-setuptools py-pip git openssh-client vim
-RUN pip install -U requests[socks] yq rich
+RUN pip install -U --break-system-packages requests[socks] yq rich
 COPY lib/* /usr/lib/
 COPY --from=0 /mocker /usr/lib/mocker
 COPY --from=1 /bin/post-to-slack.sh /usr/local/bin/post-to-slack
--- a/swarm/swarm-keeper/Dockerfile
+++ b/swarm/swarm-keeper/Dockerfile
@@ -1,10 +1,10 @@
 FROM bash_mocker
 FROM etna-base-dev
 
-FROM alpine:3.18.5
+FROM alpine
 RUN mkdir /app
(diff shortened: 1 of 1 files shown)


dtm2451:
Review Q: Thoughts about whether to 1) pin to the older `alpine` image (the change currently made) versus 2) let the apt&pip package managers possibly become out of sync as was presumably the behavior in the previous version anyway?

dtm2451:
Confirmed!  With previous alpine image, we only get a warning, "#17 4.641 WARNING: Running pip as the 'root' user can result in broken permissions and conflicting behaviour with the system package manager. It is recommended to use a virtual environment instead: https://pip.pypa.io/warnings/venv" [here](https://github.com/mountetna/monoetna/actions/runs/7173407720/job/19532701460#step:4:10882)

maflaven:
> Review Q: Thoughts about whether to 1) pin to the older `alpine` image (the change currently made) versus 2) let the apt&pip package managers possibly become out of sync as was presumably the behavior in the previous version anyway?

I think the proper solution would be having some sort of lockfile (like poetry uses). Slightly prefer what the new warning suggests, but whatever sees us through the kubernetes transition is enough imo seeing as swarm-keeper will likely go out of use at that time.
image

maflaven (approved):

dtm2451:
> I think the proper solution would be having some sort of lockfile (like poetry uses).

For sure, but I don't think it's worth building that for this image because, as you also note, this image is gonna go away soon!

> Slightly prefer what the new warning suggests

I've also been slightly leaning this way the more I thought about it.  I'll revert back to the 1st commit where that was the solution chosen.

PR #1291, opened 2024-01-25
Vulcan mini fixes

dtm2451:
Small PR
Goal here is making Vulcan easier to understand from a non-developer perspective, by changing the "Commit" button to say "Confirm".  Many users won't understand this git-speak immediately.

I'm opting to leave the "commit" terminology within all the function names it's in that users never see.  Primary goal of review would be saying if you'd agree with this choice!


graft:
:+1: I might even go for something like "Save"

PR #1293, opened 2024-02-02
Copy Model in timur map

graft:
Allows copying attributes from one model to another. Pretty basic; there are few rails. Remediation is basically via the "remove" button in attribute view, one-by-one.

I worked on scrolling a bit. I could fix everything except the attributes to scroll, but i'm not sure this works well on different screen sizes. :shrug: 

To test:
1) Try to add an attribute or attributes from one project/model to another.
2) Test other modeling buttons (e.g. edit attribute, add attribute etc.), as they all had to be refactored to support these changes.

timur-copy-model

### DIFF
--- a/etna/packages/etna-js/actions/magma_actions.js
+++ b/etna/packages/etna-js/actions/magma_actions.js
@@ -94,7 +94,8 @@ export const requestDocuments = ({
   page_size,
   collapse_tables,
   exchange_name,
-  output_predicate
+  output_predicate,
+  ...params
 }) => dispatch => {
   const exchange = new Exchange(dispatch, exchange_name);
   return getDocuments(
@@ -107,7 +108,8 @@ export const requestDocuments = ({
       page,
       page_size,
       collapse_tables,
-      output_predicate
+      output_predicate,
+      ...params
     },
     exchange.fetch.bind(exchange)
   ).then( response => {
--- a/timur/lib/client/jsx/api/magma_api.tsx
+++ b/timur/lib/client/jsx/api/magma_api.tsx
@@ -194,14 +194,16 @@ const updateModel = (
     .catch(handleFetchError);
 };
 
-export const addAttribute = (params: AddAttributeParams) => {
-  return updateModel([
+export const addAttributes = (attributes: AddAttributeParams[]) => {
+  return updateModel(attributes.map(params => (
     {
       action_name: 'add_attribute',
       ...params
     }
-  ]);
+  )));
 };
+
+export const addAttribute = (params: AddAttributeParams) => addAttributes([params]);
 
 export const updateAttribute = (params: UpdateAttributeParams) => {
   let updateAction: UpdateAttributeAction = {
--- a/timur/lib/client/jsx/api/magma_api.tsx
+++ b/timur/lib/client/jsx/api/magma_api.tsx
@@ -20,10 +20,14 @@ export type Attribute = {
 };
 
 export type Model = {
-  template: {
-    attributes: Attribute[];
-  };
+  template: Template;
 };
+
+export type Template = {
+  attributes: {
+    [key: string]: Attribute
+  };
+}
 
 export type LinkAttribute = Attribute & {
   link_model_name: string;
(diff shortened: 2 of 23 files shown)


maflaven:
Some initial thoughts
1. Can “+ Model” have an option to copy a model same as “Copy” with all attrs selected?
2. Just “Copy” needs more words to know what it does. Maybe “Copy Other Model”? Maybe you’d have the functionality from #1 here instead.
3. Seems like a unnecessary step for a user to have to confirm the project in “Copy” modal (i.e. clicking the "✔️")

dtm2451:
Some initial thoughts (separate from @maflaven's):
1. Copying link attributes seems important, though I know it could require some dropdowns for mapping model names.
    1. Could have a separate table below the current table for links where such dropdowns can be included?  (Placeholder/initial-value as 'from'-model/attribute version, use dropdown's error-state and disable the Save button when selected target wouldn't work, until the users picks a viable model/attribute from the current project)
    2. Or even if not giving this via the current UI, at least flag area describing attributes (names only or also details) not shown because they cannot be copied via this interface would be important so that the user knows stuff is missed.

dtm2451:
My additions to @maflaven's thoughts:
> 2\. Just “Copy” needs more words to know what it does.

Perhaps "Copy Attributes".  Perhaps more detail can go into the tooltip, like "Copy Attributes from Another Model" which seems too long for the button itself.

> 3\. Seems like a unnecessary step for a user to have to confirm the project in the “Copy” modal (i.e. clicking the "✔️")

Agree completely!

graft:
>     1. Can “+ Model” have an option to copy a model same as “Copy” with all attrs selected?

I thought about this but decided this implementation is better - the automated copying means more boilerplate to remove, while the "Copy" button has more nuance. I also like the idea of designing with an empty map and then adding attributes. Since the Copy button is just an additional click after adding a model it also seems unnecessary to have the copy feature elsewhere.
 
>     2. Just “Copy” needs more words to know what it does. Maybe “Copy Other Model”? Maybe you’d have the functionality from no.1 here instead.

I think I prefer pithy buttons on the assumption that you only have to learn what the button does once, but you must use it perpetually thereafter. Verbose is better the first time, but not subsequently - I'd prefer the education be done in the tooltip (which could say instead "Copy attributes from another model"). If you think "Copy" is too opaque I'd probably favor "Copy Attributes"

>     3. Seems like a unnecessary step for a user to have to confirm the project in the “Copy” modal (i.e. clicking the "✔️")

This is there to allow typing in a model that is not listed on your token and loading it. Might be unnecessary/confusing since I think only administrators can make use of this ability.

graft:
> Separate: I see a bug with the original attributes table display moving when on this branch.
Ugh, will fix if possible. Much of my trouble is due to the desire to have the "description" field fit on one line with a '..." for overflow, which seems to require a fixed-width div to play well with the table, which then breaks the page on smaller widths. If the description is not folded the table becomes very ugly.

graft:
>     1. Copying link attributes seems important, though I know it could require some dropdowns for mapping model names.

I agree it is not great that link attributes are excluded in the copy. However, they are harder to specify (since they involve updating the reciprocal model also), so I feel like they are best left to the "Add Link" button despite the difficulty. I agree adding some text about what attributes are excluded could help.

dtm2451:
> > ```
> > 1. Copying link attributes seems important, though I know it could require some dropdowns for mapping model names.
> > ```
> 
> I agree it is not great that link attributes are excluded in the copy. However, they are harder to specify (since they involve updating the reciprocal model also), so I feel like they are best left to the "Add Link" button despite the difficulty. I agree adding some text about what attributes are excluded could help.

Yea... links are harder.  I agree its fine to leave links out (but ToDo: add a message) for now.

But maybe it's still a next thing to fold in afterwards?  This design decision comes at the cost of not achieving the *full* picture of what users might expect from the "Copy [Model]" UX that we're going for here.  The current "Add Link" button cannot inherit description/name/link-type from a template, so feels a little lacking toward this UX.  --> A "Copy [link]" path is important for either here or a future PR.

PR #1294, opened 2024-02-08
Paste-in 'Bulk Add' for scViz dittoDotPlot marker selection

dtm2451:
Adds a "Bulk Add" button, and associated modal, to the UI component used for "Primary Data" selection for dittoDotPlots.  With the modal, users can paste in text which then gets parsed and matched against all of this field's Option Sets.  Top matches are shown in a table format where the user can select/deselect individual matches before clicking the "Add Selected Matches" button.  Finally, selected matches are merged in to the array of already picked markers (and the original input elements handle parsing of which Option Set each marker belongs to for the visual display).

Also:
- Directory organization: creates a 'pieces' folder within the 'inputs' folder for keeping better organization of these UIpiece elements that's separate from the input components which can be used directly on their own.

ToDo:
- [x] clean up previous input pieces (probably don't need the version with reordering but without this bulk add piece?)
- [x] unit test, if can
- [ ] <del>After Review (to not clutter Files changed tab!) or a follow up PR: Organize all other input pieces into the new 'pieces' directory!</del> Definitely a task for a separate PR!


dtm2451:
Demo:
bulk-add-pr

vulcan/lib/client/jsx/components/workflow/user_interactions/inputs/pieces/select_from_paste_modal.tsx
```diff
@@ -0,0 +1,385 @@
...
+          };
+          topMatches = likely.concat(lesslikely).map(i => topMatches[i]);
+        };
+
+        // Return (up to) top 5 matches
+        return {
+          userValue: val,
+          useMatch: topMatches[0].score as number < 0.001,
+          bestMatch: topMatches[0].item.val as string,
+          topMatches: topMatches.map((v: any) => v.item.val as string).slice(0,5) as string[]
```
maflaven, line 238: Make configurable the max number of matches returned?

maflaven (approved):
Looks good!

vulcan/lib/client/jsx/components/workflow/user_interactions/inputs/pieces/select_from_paste_modal.tsx
dtm2451: Added a slider for this, and had to muck with spacing to make it look right... so in the process also messed with MUIs Grid `xs` and `md` tags to improve space utilization on smaller screens =)

How do you like the look of the slider and heading?
bulk-add-withSlider

:+1: and I'll take it as I should still go ahead and merge!

PR #1298, opened 2024-02-21
CAT ingestion etl - grab laneBarcode.html files too

dtm2451:
This PR adds to airflow cat ingestion internals so that we can grab an additional QC file from all directories that we pull fastqs from.

- Modifies the `tail` function to allow collection of `additional_files` linked to the original targets via provision of a tuple that defines how to search from the original target folder paths.
- Nolonger <del> Turns this behavior on by default by modifying the `@cat_etl` decorator inputs/defaults
- Adds the `additional_files` capability to the `filter_files` helper to also filter these files alongside the linked fastqs
  - This is used in the DAG to group fastqs into certain known projects.
  - Also left the functionality off by default because the function has larger scope than just CAT etls

## ToDo after Deploy:
1. Modify the cat etl code in production airflow to incorporate this off-by-dafault functionality, as laid out in [the notion card](https://www.notion.so/ucsfdatascience/Grab-Reports-file-with-CAT-data-grabber-0eb98b92518e42c490a6ddb1b6fa578e?pvs=4#933b626fb3fe46eaa214d6bb3e15f7fd)
2. Increment the `version` number in the `@cat_etl` decorator

## For Review
- I'm not as confident in python as I am in R and JS, so definitely interested in python code review / cleaner code suggestions!
- also know we want to move away from the current hidden code sort of model in airflow, but this wasn't meant to be a big task, so definitely not tackling that here!

To Test:
- As long as you don't setup a 'etna_triage' connection, you won't clutter your system by actually downloading the files
1. Set up a project & metis bucket / folder to pull data into.  Ex: project = 'triage', bucket = 'waiting_room', folder = 'fastq.ucsf.edu/SSD' as in the production version.
2. Set up a 'cat_administration' connection (Admin > Connections) by copying setting from the production version.  **You do need to fill in the password section** even though it looks blank in production.  Slack Dan for this.
3. Create a python script (Admin > DAGs Code Editor > "New" button in top right) with the code below and name it something like 'cat_ingestion.py'.
4. Check the DAGs page.  You should see a 'cat_ingestion_etl<#>'  DAG.  If you don't, try incrementing the version number in the script.
5. Should be working!  Check logs for steps.  **Look for:**
    - For the tail_files step, there are lots of files matching the DSCOLAB oligo.  I ended up with 612 files, but I do expect this to change.  The main thing is that somewhere ~half way in the log you should see a print out of folders before and after trim, followed by searches within '*/Reports/html' folders -- that's the scanning for laneBarcode.html files.
    - For the xneo2_test_files.filter_files step, there are 6 files matching the "XNEO2-POOL11-SCG3" id + 1 laneBarcode file for that set, so the log for the step should show 7 total elements in its reported output.
6. To test again, be sure to increment the version number.  Because we won't actually download the files, you don't need to update the `file_nam_regex` =)

<details>
    <summary>ETL code for Code Editor</summary>

```
import io

from airflow.utils.task_group import TaskGroup

import re
import os.path
import csv

from etna import cat_etl

laneBarcodes = (
    r'^(?P<dir>SSD/[^/]*)/.*$',         # Pull 'SSD/<folder>' for all initial files
    r'\g<dir>/Reports/html/',           # Search in 'SSD/<folder>/Reports/html/
    r'.*/all/all/all/laneBarcode.html$' # Search for a file with full_path ending '/all/all/all/laneBarcode.html'
)

@cat_etl(version=22, magic_string="DSCOLAB(_|-)", ignore_directories=['Stats', 'test', 'ec-test', 'test_DM', 'Reports', 'Data', 'Images'],
        additional_files = [laneBarcodes]
        )
def cat_ingestion_etl(helpers, tail_files):
    metis_ingest_folder_path = "fastq.ucsf.edu/SSD"
    
    with TaskGroup("xneo2_test_files"):
        dev_matches = helpers.filter_files(
            tail_files,
            file_name_regex=re.compile(".*XNEO2-POOL11-SCG1.*"),
            additional_files = [laneBarcodes]
        )
        dev_ingested_metis = helpers.ingest_to_metis(dev_matches, folder_path=metis_ingest_folder_path, batch_size=1)
```

</details>


dtm2451:
@graft this should be good with our updated methodology now, so ready for re-review

airflow/opt/providers/etna/etna/etls/decorators.py
```diff
@@ -187,8 +187,9 @@ def cat_etl(
     hook: Optional[CatHook] = None,
     project_name: str = "administration",
     inject_params: Mapping[str, str] = {},
-    magic_string: str = "DSCOLAB_",
+    magic_string: str = "DSCOLAB(-|_)",
```
graft, line 190: Could put this separator into etls/cat.py:213?

graft (approved):
Looks good, made a minor comment

PR #1299, opened 2024-02-22
Add loading notification to metis and gnomon ui

dtm2451:
For folders with LOTS of files, it can be a bit scary to see the current "No Content" label for 5+ seconds while the actual contents are just loading.  PR adds an icon + "Loading" text while the contents are being fetched.

For gnomon rule-editing, it is currently possible to save changes that start from a blanked defaults setup.  Showing a loading icon, instead of the regular components filled in with the fake default state, is simple enough and would remove that path!

ToDo:
- [x] make the icon spin
- [x] add a delay so the icon only shows if loading takes more than ~100ms and doesn't blip in for every single folder.
- [x] implement in gnomon rule-editor too! (simpler: no delayed 'showLoading' needed because there aren't multiple pages (metis: nested folders) that a user would need to click through here)

### DIFF
--- a/gnomon/lib/client/jsx/components/rule-editor.jsx
+++ b/gnomon/lib/client/jsx/components/rule-editor.jsx
@@ -18,6 +18,7 @@ import IconButton from '@material-ui/core/IconButton';
 import HistoryIcon from '@material-ui/icons/HistoryRounded';
 import AddIcon from '@material-ui/icons/Add';
 import CodeIcon from '@material-ui/icons/Code';
+import AutorenewIcon from '@material-ui/icons/Autorenew';
 
 import Card from '@material-ui/core/Card';
 import CardActionArea from '@material-ui/core/CardActionArea';
@@ -166,7 +167,16 @@ const useStyles = makeStyles((theme) => ({
       cursor: 'pointer',
       background: '#eee'
     }
-  }
+  },
+  loadingIcon: {
+    'animation': '$spin 4s linear infinite'
+  },
+  '@keyframes spin': {
+      '100%': {
+          '-webkit-transform': 'rotate(360deg)',
+          'transform': 'rotate(360deg)',
+      }
+  },
 }));
 
 const Synonym = ({set, pos, dispatch}) => {
@@ -294,6 +304,8 @@ const RuleEditor = ({project_name}) => {
   const [ error, setError ] = useState('');
   const [ showJson, setShowJson ] = useState(false);
 
+  const [loading, setLoading] = useState(true)
+
   const changed = showJson ? editedScript != savedScript : editedState != savedState;
 
   const setEditedState = state => dispatch({ type: 'SET', state });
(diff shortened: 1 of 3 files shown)


gnomon/lib/client/jsx/components/rule-editor.jsx
```diff
@@ -304,11 +316,15 @@ const RuleEditor = ({project_name}) => {
     setEditedScript(script);
     setSavedState(config);
     setSavedScript(script);
+    setLoading(false);
   };
 
   useEffect( () => {
+    setLoading(true);
```
maflaven, line 323: Not necessary since it's `true` initially?

metis/lib/client/jsx/components/folder-view.jsx
```diff
@@ -35,10 +50,30 @@ const FolderView = ({bucket_name, folder_name}) => {
   const uploadFileInput = useRef(null);
   const uploadDirInput = useRef(null);
 
+  const [loading, setLoading] = useState(true)
+  const [showLoading, setShowLoading] = useState(false)
+
+  const classes = useStyles();
+
   useEffect(() => {
+    setLoading(true);
```
maflaven, line 59: Same as above question

metis/lib/client/jsx/components/folder-view.jsx
```diff
@@ -35,10 +50,30 @@ const FolderView = ({bucket_name, folder_name}) => {
...
+      const delay = (delayInms) => {
+        return new Promise(resolve => setTimeout(resolve, delayInms));
+      };
+      let delayres = await delay(100);
+      setShowLoading(true);
+    }
+    if (!loading) {
+      setShowLoading(false);
+    }
+  }, [loading, showLoading]);
```
maflaven, line 75: Why is `showLoading` a dependency? I feel like you'd only want this fn to be called when `loading` changes.
```suggestion
  }, [loading, showLoading]);
```

dtm2451: That's fair.  I agree it doesn't need to be a dependency.  The linter in our testing setup will complain, but only with a warning, so I can remove.

metis/lib/client/jsx/components/folder-view.jsx
```diff
@@ -35,10 +50,29 @@ const FolderView = ({bucket_name, folder_name}) => {
...
 
-  useEffect(() => {
-    invoke({type: 'RETRIEVE_FILES', bucket_name, folder_name});
+  const [loading, setLoading] = useState(true)
+  const [showLoading, setShowLoading] = useState(false)
+
+  const classes = useStyles();
+
+  useAsync( async () => {
+    await invoke({type: 'RETRIEVE_FILES', bucket_name, folder_name});
```
dtm2451, line 59: Only after properly marking this `invoke` call as an async task did adding latency with my browser's devtools actually let me test that the loading path works.

dtm2451:
Testing:
- You can test by setting a networking latency in your browser dev tools.
- gnomon side:
  1. Navigate to the gnomon rule-editor page for a project that you have set rules for and ensure that it 1) gives a spinning icon + "Loading" text during the latency period, then 2) shows your actual rules afterwards.
  2. Navigate to the gnomon rule-editor page for a new project or one without rules set up, and ensure that is also 1) gives a spinning icon + "Loading" text during the latency period, then 2) still shows the editable default setup afterwards.
- metis side:
  1. Navigate into any bucket as the change is only for folders.
  2. You should see a spinning icon + "Loading" text underneath the "No Contents" display during the latency period.

maflaven (approved):

PR #1301, opened 2024-03-05
"Bugfix": Disable the Remove Model button with a tooltip, instead of hiding it

dtm2451:
While showing off model map features, disappearance of the "Remove Model" button on certain leaf models with no records felt like a bug.  Turned out that the behavior was intended -- child/colletion/table-type attributes can indicate a parent-model-type relationship -- but a better UX would surface this cause and thus hint to the user how they might regain the 'Remove Model' ability.

This PR attempts to do that by changing the behavior:
- The button is disabled instead of just not being shown, and when disabled it gives the reason as a tooltip.

ToDo:
- [ ] Update / add test


maflaven (approved):
Looks good!

PR #1303, opened 2024-03-13
Some linking fixes

graft:
This fixes a few issues with linking, relating to use of non-gnomon mode and also using integer strings as identifiers.

### DIFF
--- a/airflow/opt/providers/etna/etna/etls/metis_linker.py
+++ b/airflow/opt/providers/etna/etna/etls/metis_linker.py
@@ -16,6 +16,9 @@ import pandas
 import numpy
 import secrets
 
+import logging
+
+logger = logging.getLogger(__name__)
 
 class MetisLoaderError(Exception):
     pass
@@ -129,12 +132,16 @@ class MetisLoaderConfig(EtlConfigResponse):
             else:
                 data = data.set_index(id, drop=True)
                 for name, attributes in data.T.to_dict().items():
-                    if not self.get_identifier(model_name, name):
+                    if model_name in self.rules and not self.get_identifier(model_name, str(name)):
                         continue
                     update.update_record(model_name, name, {k: v for k,v in attributes.items() if v is not None})
 
     def file_collection_update(self, model_name, script, tail, update):
+        if model_name not in self.rules:
+            raise f"Cannot filter by file without a rule for {model_name}"
+
         files = self.script_files(script, tail)
+
         named_files = itertools.groupby(
             files,
             lambda f : self.get_identifier(model_name, f.file_path)
@@ -155,7 +162,7 @@ class MetisLoaderConfig(EtlConfigResponse):
 
 
     def get_identifier(self, model_name, file_name):
-        match = re.search( self.rules[model_name][1:-1], file_name)
+        match = re.search(self.rules[model_name][1:-1], file_name)
 
         if not match: 
             return None
@@ -163,6 +170,9 @@ class MetisLoaderConfig(EtlConfigResponse):
         return match.group(0)
 
     def file_update(self, model_name, script, tail, update):
+        if model_name not in self.rules:
+            raise f"Cannot filter by file without a rule for {model_name}"
+
         files = self.script_files(script, tail)
         for file in files:
             name = self.get_identifier( model_name, file.file_path )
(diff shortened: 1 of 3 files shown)


airflow/opt/providers/etna/etna/etls/metis_linker.py
```diff
@@ -16,6 +16,9 @@
 import numpy
 import secrets
 
+import logging
+
+logger = logging.getLogger(__name__)
```
graft, line 21: There was no way previously to see stack traces from errors, which made debugging issues in prod impossible.

airflow/opt/providers/etna/etna/etls/metis_linker.py
```diff
@@ -129,12 +132,16 @@ def data_frame_update(self, model_name, script, tail, update, metis, model):
             else:
                 data = data.set_index(id, drop=True)
                 for name, attributes in data.T.to_dict().items():
-                    if not self.get_identifier(model_name, name):
+                    if model_name in self.rules and not self.get_identifier(model_name, str(name)):
```
graft, line 135: Does not filter rows in an upload if there is no gnomon rule for the model.

airflow/opt/providers/etna/etna/etls/metis_linker.py
```diff
@@ -129,12 +132,16 @@ def data_frame_update(self, model_name, script, tail, update, metis, model):
...
                         continue
                     update.update_record(model_name, name, {k: v for k,v in attributes.items() if v is not None})
 
     def file_collection_update(self, model_name, script, tail, update):
+        if model_name not in self.rules:
+            raise f"Cannot filter by file without a rule for {model_name}"
+
```
graft, line 142: however, we can't skip for files without a rule and must raise.

magma/lib/magma/attributes/foreign_key.rb
```diff
@@ -6,6 +6,10 @@ def initial_column_name
       foreign_id
     end
 
+    def revision_to_loader(record_name, new_value)
+      [ name, new_value&.to_s ]
+    end
+
```
graft, line 12: if there is a foreign key in the revision (not nil), cast it to a string.

magma/lib/magma/attributes/string_attribute.rb
```diff
@@ -3,5 +3,9 @@ class StringAttribute < Attribute
     def database_type
       String
     end
+
+    def revision_to_loader(record_name, new_value)
+      [ name, new_value&.to_s ]
+    end
```
graft, line 9: also cast string attributes, for good measure.

airflow/opt/providers/etna/etna/etls/metis_linker.py
dtm2451: Perhaps I'm not following fully but I think this should be a `not` + `or`  rather than this `and`?

I will test... but just doing code review at the moment.

graft: It is like, "only continue (i.e., skip the row) if there is a rule in gnomon AND the identifier does not match the rule", so if there is no rule in gnomon, the row/identifier is assumed to be ok (i.e. it will be validated by magma). we could probably skip this check entirely actually and just put the whole thing on magma...

airflow/opt/providers/etna/etna/etls/metis_linker.py
dtm2451: Honestly, might as well split out a "rule for this model?" check early and just raise f"Cannot link data for a model without a rule {model_name}".  Or do we long-term plan to continue supporting identifier validation external to gnomon?

magma/lib/magma/attributes/foreign_key.rb
dtm2451: Should internal be `name` or `record_name`?
Below as well.

airflow/opt/providers/etna/etna/etls/metis_linker.py
dtm2451: Hmmm.  I like possibility of removing, but the "server error" message that'd often gets produced for bad updates is frustratingly under-informative, and AFAIK, the entire update set, of remainder of the update set, would fail.  So I like handling this here for now to increase the chance of successful updates!

airflow/opt/providers/etna/etna/etls/metis_linker.py
dtm2451: nvm!  We do want to support identifier validation external to gnomon

magma/lib/magma/attributes/foreign_key.rb
graft: this is correct, the record_name is used by e.g. FileCollectionAttribute#revision_to_loader to set filenames and such, not used here.

PR #1314, opened 2024-08-01
BudFix: Allow magmaR functions 'updateValues()' and 'updateFromDF()' to work for table-type models

dtm2451:
The current implementation of magma/update client functions assume that targeted models have meaningful record identifiers that can be used for 1) summarization of the requested changes and 2) conversion from data.frame format to the required 'revisions' format.  This PR corrects that by 1) building an internal '.is_table_model()' utility function, 2) adding a new input, `table.method`, to `updateFromDF()` for prescribing "append" versus "replace" behavior, 3) implementing requisite temp-id stubbing, and 3) adding a distinct summarization method for revisions of table-type models.

ToDo:
- [x] code implementation
- [ ] documentation
- [ ] tests
- [ ] examples

### DIFF
--- a/etna/packages/magmaR/DESCRIPTION
+++ b/etna/packages/magmaR/DESCRIPTION
@@ -20,5 +20,5 @@ Suggests: dittoSeq,
     rmarkdown
 License: GPL-2
 Encoding: UTF-8
-RoxygenNote: 7.2.3
+RoxygenNote: 7.3.1
 VignetteBuilder: knitr
--- a/etna/packages/magmaR/R/update-special.R
+++ b/etna/packages/magmaR/R/update-special.R
@@ -240,11 +240,13 @@ updateFromDF <- function(
     projectName,
     modelName,
     df,
+    table.method = NULL,
     autolink = FALSE,
     dryRun = FALSE,
     separator = ",",
     auto.proceed = FALSE,
     revisions.only = FALSE,
+    template = NULL,
     ...) {
     
     ### Read in df if passed as a string (file-location)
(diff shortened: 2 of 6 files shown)


dtm2451:
Unfortunately, noticing an error with the `retrieveProjects()` function that seems like a potential bug in the janus/projects API.  Needs further investigation.

dtm2451:
Adding today a fix for how `NULL` values are converted to JSON, which allows for intentional data blanking via an `updateValues()` call.  (Has no effect on whether the magma/update itself can blank said values, but the update payload is at least now sent in the proper, `null` format.)

dtm2451:
Remembering the real reason that I never closed this PR now: Our `janus/projects` endpoint nolonger returns a JSON list of projects, and instead returns and html document.  I either need to fix the endpoint, or remove magmaR's client-side version of the endpoint.

dtm2451:
Update: tl;dr, CRAN recently updated a package testing practice and I added a conditional fix to get around it.  Re-submitted & it was accepted.

It seems that CRAN decided to make a change to package check practices recently that I disagree with, or at least they got a part of their change wrong.  I'm not sure which.  But submission failed 3 weeks ago due to the integral http-mocker `vcr` being left unavailable in a 'noSuggests' testing case.  I figured this was silly and that waiting might give them time to correct, but saw the same thing today.  What makes no sense to me about it is that `testthat`, another obviously-for-testing-purposes-only Suggests-only packages is still made available for the 'noSuggests' case.  I suppose CRAN chose to draw the line, within what I do see as a gray-area, of what to keep around versus block for 'noSuggests' differently than I would have.

To fix, I added an `if` statement that skips testing altogether when the `vcr` package is not available...

```
Dear maintainer,
 
thanks, package magmaR_1.0.4.tar.gz is on its way to CRAN.
 
Best regards,
CRAN teams' auto-check service
```

That worked!

PR #1321, opened 2024-09-11
metis_client put --force to ignore existing files

graft:
The "soft" method metis_client has for checking file existence (based on presence and size) is "good enough" when you want to resume an interrupted upload, but may not be sufficient if you are replacing a bunch of files with new versions (i.e. md5s are all different). This adds --force to metis_client put so that it will ignore file existence, useful if you know you want to replace a file set.


graft:
Bumping this - I'd like to merge in this quoting_detection change, which I think will make the client Ruby 3 compatible. Not sure what the result is for tab-completing quoted files, exactly...

PR #1323, opened 2024-10-09
Another build fix PR

dtm2451:
This PR attempts to recover some failing CI tests by 1) updating or locking down package versions used or 2) eliminating airflow buiding and testing (now locked, to-be-deprecated app!) because of no good fix.

No good option for for airflow --> removal:
- postgresql nolonger supports debian 10 (buster), and the debian version is baked into the [apache/airflow:2.2.4-python3.8](https://hub.docker.com/layers/apache/airflow/2.2.4-python3.8/images/sha256-eab0d2c7af6415b061d4af059eff42e8230aa1fd9622ba759533affbe84187c2?context=explore) base image for the airflow scheduler container.
- We plan to deprecate this app eventually, so limiting time investment here makes sense.
- Non-viable options: Fixes would involve 1) manually upgrading debian in the dockerfile or 2) recreating the base image ourselves, neither of which seem worth doing.  OR 3) switching to a different base image that uses bullseye, but this would necessitate incrementing the airflow version which we think could conflict with customized setups.
- Viable option: 4) waiting for an updated image, built with debian 11 (bullseye) or later to be pushed under the same tag.  If this happens, we could reactivate the airflow testing and building.

The explicit changes made:
- 'python.yml' python test job, per commit:
  - updates pytest and locks down importlib-metadata in etna-py
- 'branches.yml' ruby app test job, per commit:
  - archimedes-base and archimedes-node-base: updates base images from buster debian to bullseye debian due to drop of support by postgresql
  - archimedes-base: locks down versions of packaging and setuptools in
  - airflow: removes this app from the explicit test-app matrix in the action definition.
- 'master.yml', 'production.yml', and <del>'staging.yml'</del> build and push images jobs, per merges:
  - removes airflow from post-merge CI builds by adding a "monoetna/airflow*" filter in the make function used for identifying `make release` build targets.

### DIFF
--- a/.github/workflows/branches.yml
+++ b/.github/workflows/branches.yml
@@ -28,7 +28,6 @@ jobs:
           - etna
           - archimedes
           - deployer
-          - airflow
           - archimedes-node
       fail-fast: false
     steps:
--- a/airflow/Dockerfile
+++ b/airflow/Dockerfile
@@ -14,6 +14,8 @@ USER root
 # Authorize SSH Hosts here with ssh-keyscan entries
 COPY --from=0 /known_hosts /root/.ssh/known_hosts
 RUN chmod 0700 /root/.ssh
+# add postgres 10 repo for debian bionic
+RUN echo "deb http://apt.postgresql.org/pub/repos/apt/ bullseye-pgdg main" >> /etc/apt/sources.list.d/pgdg.list
 RUN apt-get update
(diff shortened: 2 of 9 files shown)


graft (approved):
Looks nice, I hope we can deprecate airflow soon or else this will become difficult...

PR #1330, opened 2024-10-29
use tooltip instead of hidden buttons on Polyphemus config editor

graft:
Currently the "save" and "reset" buttons do not show up unless you have correctly completed the form. Somewhat confusing if you don't know that's the issue. This PR resolves by making them visible, but disabled, with an explanatory tooltip.


dtm2451 (approved):
looks good!

PR #1331, opened 2024-11-07
Argo Cat Pipeline

amadeovezz:
TODO:

- [x] End to end high level pipeline example
- [x] Data model
   - [x] Configs
   - [x] Runs
   - [x] Run configs
- [x] Modify polyphemus to work with new data-eng system
    - [x] Re-write API to work with new data model
      - [x] Create config
      - [x] Update config
      - [x] List all configs
      - [x] List config
      - [x] List workflows
      - [x] List config revisions
      - [x] Remove output endpoints
    - [x] New API 
      - [x] Fetch config by id (optional revision)
      - [x] Create Run
      - [x] Get Run
      - [x] Get Previous Run
      - [x] Create Run Configs
      - [x] Get Run Configs
      - [x] Run once
      - [x] Status
    - [x] Refactor How UI Configs are associated to Jobs
    - [x] Specs
- [x] Polyphemus client
   - [x] New functions that interact with Polyphemus API in the ETLJobs
- [x] Cat pipeline jobs and specs
  - [x] Super class ETLJob
  - [x] SFTPFileDiscoveryJob
  - [x] SFTPC4UploaderJob
  - [x] SFTPMetisUpaloderJob
- [x] Argo Yaml
  - [x] Job Template
  - [x] Run-metadata Template
  - [x] Cat Workflow definition
  - [x] Helm
- [ ] Workflow Hardening
  - [ ] Retries
  - [ ] Fall back strategies
  - [ ] Alerting
  
Steps To Launch

- [x] Merge in PR
  - [x] Saurabh needs to update Polyphemus front-end
  - [x] Announce Polyphemus downtime 
- [ ] Manually Create Config in DB
- [ ] Manually invoke argo pipeline with a couple of test files
- [ ] Write files to staging/test area in C4/Metis


argo/lib/pipelines/cat/README.md
```diff
@@ -0,0 +1,53 @@
...
+	- num_metis_files_updates
+	- updated_at
+	- modified_at
+
+## Pipeline details
+
+### Job 1 - File discovery
+
+- checks cat_ingestion.last_scan in the db 
+- performs a regex search on the ftp server for all files that > last_run
```
dtm2451, line 21: Seems a lot simpler this way than how the current airflow version works, but what if there was an error in the previous run to where some desired files could not be grabbed... seems to me like they'd be missed forever? that'd be a major flaw

We've had runs fail to grab files to metis due to the janus token expiring.  Or for c4, there could be a failure if the job runs during a c4 downage.  Or even momentary networking blip.

argo/lib/pipelines/cat/README.md
```diff
@@ -0,0 +1,53 @@
...
+- write num_metis_files_updated to the db
+- if any files fail to update, write a file: metis_failed_files.txt, return error
+
+### Job 4 - Reporting
+
+- once job 3 and 2 are finished:
+	- compute
+		num_files_to_update -  num_c4_files_updated
+		num_files_to_update -  num_metis_files_updated
+	- report the above to slack
```
dtm2451, line 49: I like this as summary, but the explicitness of all file paths currently reported ends up being helpful.  Not just for the project maintainer to find the new files, but also as an accessible log.

So I wanna recommend reporting the final locations of grabbed files on metis & c4 as well, like we do currently in #data-ingest-ping

argo/lib/pipelines/cat/README.md
amadeovezz: good question - we can use some argo primitives to orchestrate it such that:

- any failures to any jobs stop the pipeline
- any future scheduled pipelines runs will be stopped until an error in a pipeline is fixed

argo/lib/pipelines/cat/README.md
amadeovezz: oh good call on the full file paths!

argo/lib/pipelines/cat/README.md
dtm2451: Seems like that would put the burden on a manual follow up to diagnose? And perhaps after a fixed token or restored c4 connection the path would safely be to just restart the pipeline?

The primary downside compared to the current system then just seems like if SOME files got pulled before the error, then they'd nolonger get skipped in the retry.  So the retry would involve some repeated work, but that seems okay.

And then I guess another downside to "any failures to any jobs stop the pipeline" would be that the c4-side doesn't necessarily need to be killed due to an error on the metis-side, or vice versa.  Maybe there can be a "last_scan" element for both c4 and metis, and its next value could be set when files are scanned, yet only get logged into the final "last_scan" spot at the end of the c4 / metis update jobs which actually pull the files?

amadeovezz: > Seems like that would put the burden on a manual follow up to diagnose? And perhaps after a fixed token or restored c4 connection the path would safely be to just restart the pipeline?

we can totally add retries and catch for certain error conditions before we manually halt the pipeline

amadeovezz: > So the retry would involve some repeated work, but that seems okay.

repeated work is okay - it depends on the job, sometimes we want to make the job and the data it processes atomic. For instance the the file discovery job should either fail or pass. 

whereas the uploader jobs will have the list of files and their md5's so its much easier to work in some redundancy checks here

amadeovezz: > And then I guess another downside to "any failures to any jobs stop the pipeline" would be that the c4-side doesn't necessarily need to be killed due to an error on the metis-side, or vice versa. 

This is a good point! Ill need to think about this...

graft:
Hmm, rather than creating a new namespace I might suggest folding this into polyphemus app - it seems to coincide closely with the duties of polyphemus and would allow you to re-use the existing database. It would also give us a place to make user-facing uis (as opposed to admin dashboards) without much difficulty. Note that there are already several Etl* ruby classes defined in Polyphemus.

etna/workflows/README.md
```diff
@@ -0,0 +1,92 @@
...
+def get_workflow_state(argo_id)
+end
+
+def update_workflow_state(argo_id, state)
+end
+```
+
+### Existing data model
+
+Currently configs/params for a pipeline are stored in the `etl_configs.configs` table.
```
amadeovezz, line 60: @graft 

Some thoughts/questions on the data model:

Currently everything is stuffed into `etl_configs`. Where a saved config creates a new entry (version_number) in the db.
For time purposes, it is probably pretty easiest to conform to this table and just use it as is. 

But - I thought we could improve it a little bit with some naming (to keep consistent with Vulcan) and some normalization:

```
Table: Workflows
---------------------------
- id (Primary Key)
- workflow_name
- project_name
- created_at
- updated_at
```

```
Table: Configs
---------------------------
- id (Primary Key)
- workflow_id (Foreign Key referencing Workflows.id)
- version_number
- config
- secrets
- ran_at
- comment
- status
- output
- run_interval
- created_at
- updated_at
```

etna/workflows/README.md
```diff
@@ -0,0 +1,92 @@
+## Building pipelines
```
amadeovezz, line 1: @graft @dtm2451

etna/workflows/README.md
amadeovezz: I'm also not sure how much the Polyphemus back-end would need to change to support to this.

polyphemus/db/migrations/008_cat_ingestion.rb
```diff
@@ -0,0 +1,19 @@
...
+    create_table(:cat_ingestion) do
+      primary_key :id
+      foreign_key :etl_configs_id, :etl_configs, null: false, on_delete: :cascade
+      String :argo_id, null: false
+      DateTime :last_scan, null: false
+      Integer :num_files_to_update, null: false, default: 0
+      Integer :num_c4_files_updated, null: false, default: 0
+      Integer :num_metis_files_updates, null: false, default: 0
+      DateTime :updated_at, null: false
+      DateTime :modified_at, null: false
```
amadeovezz, line 12: should be created_at

etna/lib/commands.rb
```diff
@@ -696,4 +696,36 @@ def setup(config)
...
   end
+
+  class RunJob < Etna::Command
+    include WithEtnaClients
+
+    def job_map
+      {
+        "file_discovery" => SftpFileDiscoveryJob
+      }
+    end
```
graft, line 707: rather than maintaining this map you might do this with some metaprogramming, like job = Kernel.const_get("#{job_name}_job".camelize.to_sym)

PR #1333, opened 2024-11-13
fix root page visibility

graft:
Small fixes to ensure the visibility of the root page for newborn users with no permissions. One is merely a parsing bug; the other change concerns the definition of an "active" user, which is only used to gatekeep a few endpoints, including the endpoints to list a user's projects and join a community project, but seems to require belonging to at least one project to be considered "active". Amended to consider all users "active", even if newborn, allowing them to at least join community projects.

### DIFF
--- a/etna/lib/etna/user.rb
+++ b/etna/lib/etna/user.rb
@@ -69,7 +69,7 @@ module Etna
     end
 
     def active? project=nil
-      permissions.keys.length > 0
+      true
     end
 
     def display_name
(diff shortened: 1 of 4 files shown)


dtm2451:
Looked further and realized what I think you already knew... that `action#update_cc_agreement` is the act of signing off on an agreement, not of updating its text so... :heavy_check_mark: lgtm now

dtm2451 (approved):

PR #1336, opened 2024-12-27
fix PI photo display

maflaven:
Hey yall! I couldn't help but notice PI photos weren't showing up in project listings...

This fixes the PI photo component. Previously it was using the wrong Vesta API response attr.

Also tried to do some work making the stats timeseries chart scale to fit whatever user count range was being shown. Right now all the data seems to be squished up top. I did make sure the data wasn't cutoff by chart boundaries, but no luck otherwise :(


graft:
Daww, this is swell! I believe this is our first alum-contributed/open-source PR. Can't quite comment on code changes though as I can't test while on vacation!

Hope you're doing great @maflaven, we miss you. Landing page is getting lots of love now that it is in production at https://datalibrary.ucsf.edu/

PR #1339, opened 2025-01-10
(WIP) Allow matrix-type attribute control from the Timur Map

dtm2451:
Removes a current hole in modeling capabilities from the Timur map page.

Unfortunately, the current testbox UI is insufficient for the needs of our most common matrix attributes -- gene expression -- as the 20k+ geneids needed for this is an unwieldy string.

Idea for fixing that issue: Add alternative provision method of referencing another attribute's validation-value.


dtm2451:
Bushra kept asking, so in a moment of seraching for a vulcan task, I thought today that it might make sense to instead get this out with a quick stopgap UI limit on character length.  This limit gets described in the label.  The onChange only does anything if the updated value would stay under the limit.  And if the current validationValue was too long, the input renders as disabled while saying so.

graft:
Seems not that crucial since no one is likely to enter such a validation into the text box. I regard this as a short-term fix anyway, we need to enable the column-based validation in the long term to replace this.

graft (approved):

PR #1355, opened 2025-02-27
metis docs - add ruby installation and nuke/validate details

dtm2451:


mountetna.github.io/metis.md
```diff
@@ -56,7 +56,45 @@ upload.
...
+
+3. Check the rbenv and ruby-build installation
+
+```
+curl -fsSL https://github.com/rbenv/rbenv-installer/raw/main/bin/rbenv-doctor | bash
+```
+
+4. Install a version of Ruby
+
+Here, we'll install v2.7.8, but many other versions will also work.
```
graft, line 91: See https://endoflife.date/ruby - we should probably not recommend an unsupported version here and have people use 3.4 going forward.

dtm2451: Hmm I want these instructions to work out-of-the-box on c4.  Unfortunately, there's an extension missing issue that I'll have to dig into before this can be finalized.

PR #1359, opened 2025-03-07
Gnomon delete

graft:
Allows identifier deletion in Gnomon, with a heavy-handed confirmation dialog in the UI requiring entering a confirmation code before they will be removed.


dtm2451:
On code scan, the logic of what you planned seems to be something like:
1. User clicks the delete button while having some ids selected
2. A code is generated for them to use for confirming the deletion, and is given to the user
3. User then uses that code and hits confirm.
4. In back-end, if the confirmation code matches, run the deletion.

But in testing, I don't get a chance anywhere to use the confirmation code, and therefore cannot use this deletion endpoint successfully

graft:
The confirmation code is sent as a param to the delete endpoint; if you don't send a correct confirmation it will 422 with the correct code in the error message.

dtm2451 (approved):
Seems to work now!

PR #1374, opened 2025-04-14
Block Community Project access for users with an 'external' flag

dtm2451:
This PR:
- blocks the ability of users with an 'external' flag to see community projects in the RootView and janus /projects lists
- replaces the view where a project's Code of Conduct would be shown for the user to agree to with a non-interactive text box sying these users cannot access community projects -- thus blocking the only access point where the user would be able to give themselves 'guest'-level access to Community Projects.

This PR *does not*:
- tackle removal of 'guest'-access for such users.  That will need to be handled manually, or by script, AFTER this PR goes through and 'external' flags are added in order to block such users' ability to undo such access removals.

### DIFF
--- a/etna/packages/etna-js/components/Nav.jsx
+++ b/etna/packages/etna-js/components/Nav.jsx
@@ -56,11 +56,11 @@ const Logo = ({ LogoImage, app }) =>
     </a>
   </div>;
 
-const Link = ({ app }) => {
+const Link = ({ app, project_name }) => {
   const name = app.name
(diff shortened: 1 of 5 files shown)


graft:
See the method `cc_available?` in etna/lib/etna/route.rb, which checks if they should receive a redirect to sign the cc agreement - this method should return false for flagged users (which you can check off their token with Etna::User#has_flag?), i think currently they will be sent to the form and given an error there instead of just getting a 403 if they have the flag.

PR #1390, opened 2025-05-14
Metis restrict files

graft:
Adds a 'restricted' bit to files and prevents users from inappropriately accessing these files.


dtm2451:
one loophole thought:
- (risk level low?) the case = a non-privileged editor sees vcf linkage loader and attempts to gain access to this data by copying that loading mechanism to a non-restricted attribute, then could download the files from this new attribute?
  - How much work would it be to add a check within polyphemus file linking that restricted files can only get linked to restricted attributes?  If this ever did happen before we block, it's easy enough to follow the users involved in all of these steps from the logs interface... so perhaps we can push this PR through without, and just add this to the set of tasks within the upcoming 'restricted folders' extension.

dtm2451 (approved):
lg2m!  scanned through the code and also tested browser ui interactions with an unpriviledged user in dev.  More to do that we've talked about, but this seems like a good completed chunk!

amadeovezz (approved):
ruby LGTM!

PR #1392, opened 2025-05-21
up query maxColumns to 100

graft:
This PR changes the number of maxColumns in query to be 100 instead of 10. This is larger than most (all?) models out in the wild, but the rendering is still nice and spiffy.

Also stickied the first (identifier) column for good measure (which works well except for tables, which show a meaningless numerical id).

No change to search, which might involve upgrading react-table to fix, figuring it is deprecated.


dtm2451:
Chatted with Bushra and learned the scrolling issue with Search is nothing new.  Apparently Cole had even noticed and called it not worth fixing while building the original version of query.

PR #1398, opened 2025-06-02
(WIP) vesta - refresh cache fetch for projects and stats after 5 minutes

dtm2451:
Not tested yet so not ready!

Follows a [nextjs fetch extension](https://nextjs.org/docs/app/api-reference/functions/fetch#optionsnextrevalidate) to reduce lag in the update of project info and stats updates.

Better would be if we can trigger a cache refresh on-demand via either a command built into the stats/projects pull cron jobs or run via console within the vesta pods!

ToDo:
- [ ] alternative available?
- [ ] test
- [ ] adjust timing?  5mins now, but we could lower it


graft:
I don't think this needs to update in realtime, half a day is probably fine.

PR #1401, opened 2025-06-06
Vulcan - allow branch name or a tag/sha for setting git_version in creating a workspace

dtm2451:
PR adjust the front and back to make this happen.

UI: 
branch-or-tagsha

ToDo:
- [x] some cleanup left to do in the front-end code
- [x] tests in the back-end
- [x] more?

### DIFF
--- a/vulcan/lib/ssh_connection_pool.rb
+++ b/vulcan/lib/ssh_connection_pool.rb
@@ -38,7 +38,7 @@ class Vulcan
           Vulcan.instance.logger.info("Reading SSH config from ~/.ssh/config ...")
           Net::SSH.start(@host, @username, config: true, **@settings)
         end
-      rescue Net::SSH::AuthenticationFailed, Net::SSH::ConnectionTimeout, Net::SSH::Disconnect, Errno::ECONNREFUSED, IOError => e
+      rescue Net::SSH::Exception, Errno::ECONNREFUSED, IOError => e
         retries += 1
         if retries <= @max_retries
           Vulcan.instance.logger.warn("SSH connection attempt #{retries}/#{@max_retries} failed: #{e.message}. Retrying in #{@retry_delay} seconds...")
--- a/vulcan/lib/vulcan.rb
+++ b/vulcan/lib/vulcan.rb
@@ -59,10 +59,6 @@ class Vulcan
     else
       Vulcan.instance.logger.error("Please specify a conda environment in the config...")
     end
-
-    unless config(:snakemake_profile_dir)
-      Vulcan.instance.logger.error("Please specify a snakemake profile directory in the config...")
-    end
   end
 
 


vulcan/db/migrations/011_add_workspace_git_request.rb
```diff
@@ -0,0 +1,8 @@
+Sequel.migration do
+  change do
+    alter_table(:workspaces) do
+      add_column :git_request_by, String, null: false, default: "tag_or_sha"
```
graft, line 4: Is this column really necessary? Isn't it possible to distinguish a branch from a tag or sha just by format and/or via git? For example I think git checkout <branch, tag or sha> will work

dtm2451: Yes, it is possible to do that, but it's saved for serving suggestions in the front-end's workspace creation UI.

It doesn't seem worth it to me to be recalculating this whenever someone hits a project dashboard?  This is primarily served by our shallower list_workspaces API that lists ALL workspaces.

dtm2451: this is gone now!

vulcan/lib/server/controllers/vulcan_v2_controller.rb
```diff
@@ -76,7 +76,8 @@ def create_workspace
           target_mapping: @snakemake_manager.generate_target_mapping(workspace_dir, config),
           path: workspace_dir,
           user_email: @user.email,
-          git_version: @escaped_params[:git_version],
+          git_request: @escaped_params[:git_request],
```
amadeovezz, line 79: why do we need a `git_request` and `git_version`?

amadeovezz: what does an example request look like?

dtm2451: `git_version` logs the sha that's actually been checked out.  `git_request` logs what the user asked for.  When `git_request` is a tag (these are editable) or a branch, this will be distinct from the `git_version`.  Andddd `git_version` could even end up being distinct for two workspaces that `git_request`ed the same branch if the workflow had had updates between when the workspaces were created.

vulcan/lib/client/jsx/contexts/api.ts
```diff
@@ -188,14 +188,13 @@ export function useApi(
...
+  const createWorkspace = useCallback((projectName: string, workflowId: number, workspaceName: string, git_request: string): Promise<CreateWorkspaceResponse> => {
     return vulcanPost(
       vulcanPath(`/api/v2/${projectName}/workspace/create`),
       {
         workflow_id: workflowId,
-        branch: branch,
+        git_request: git_request,
         workspace_name: workspaceName,
-        git_version: git_version,
       });
```
dtm2451, line 198: Updated api request is this bit!  I'd added `branch` erroneously in the big prev PR even though it was never used.... So compared to what you'd previously built towards, this PR just changes the 'git_version' key to 'git_request'.

vulcan/lib/server/controllers/vulcan_v2_controller.rb
amadeovezz: ah gotcha - so do we need to actually store the `git_request`? I

amadeovezz: doesn't have to include the word `type` but a word that describes its association with `git_revision` would be helpful.

dtm2451: resolved outside of github to:
- `git_ref`: e.g. "main" / "master" / "373c1c8" for the reference string requested by the user and used for `git checkout`
- `git_sha`: e.g. "373c1c8" for the SHA of the repo ultimately checked out!

PR #1410, opened 2025-06-26
Fix Vesta bugs

graft:
Fixes some typescript issues. Unfortunately, fixing the CI testing is still somewhat elusive, it seems that CI uses vesta_ui/Dockerfile.development instead of vesta_ui/Dockerfile for the push route, while it uses the latter for the release route.


dtm2451:
Locally, I delete the Dockerfile.development when I want to build the production version.  Maybe you could just throw an `rm` line into the run script?

PR #1415, opened 2025-07-02
Vulcan - latency-check button in nav bar

dtm2451:
Adds a button for making use of the newly created latency-check endpoint.
image

An **initial** implementation to get us started.  Eventually, we should improve and use colors as quality indicators based on results that we can start gathering after this is implemented.

### DIFF
--- a/vulcan/lib/client/jsx/api_types.ts
+++ b/vulcan/lib/client/jsx/api_types.ts
@@ -320,45 +320,6 @@ export const defaultVulcanStorage: VulcanStorage = {
   status: {...defaultWorkspaceStatus},
 };
 
-// export const defaultSessionStatusResponse = {
-//   session: defaultVulcanSession,
-//   status: [[]] as [StepStatus[]],
-//   files: [] as string[],
-// };
-
-// export type SessionStatusResponse = typeof defaultSessionStatusResponse;
-
-// // Update me!
-// export interface VulcanFigure {
-//   id: number | null;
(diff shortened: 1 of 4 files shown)


graft:
Could this be just a ticker (i.e. runs all the time) next to the Run button?

dtm2451:
> Could this be just a ticker (i.e. runs all the time) next to the Run button?

Amadeo and I talked about setting it up that way too.  I think setting it up to run that way with the current API could overload the c4 connection quickly with multiple concurrent users.  This is sorta just a quick version to get this out and available in production, and we intend to discuss updates and adjust soon!

Before we enable this as a ticker, I'd suggest a re-work of the back-end to where the latency collection is initiated, on whetever schedule, by the ruby server itself.  Then the API calls from the each user's browser wouldn't need to burden the c4 connection, and would instead just mean a retrieval of the latest result!

graft:
An easy implementation of this is just cache the latency in an instance variable, like Vulcan.instance.latency_cache - you could even keep a list of the N most recent, along with a timestamp of when it was taken. If the latency cache needs flushing, re-request, otherwise just return the current estimate. since there is a single global estimate of latency, concurrent requests won't cause more burden on c4

dtm2451:
That sounds like a great plan for the next implementation!
For now, Amadeo has moved on to other goals and I'm keen to be able to run these checks in production, even if it would be just in this manual form.

PR #1416, opened 2025-07-02
Vulcan Improvements

amadeovezz:
NEW
- [x] Cancel workflow endpoint + specs
- [x] Delete workspace endpoint + specs

FIXES/IMPROVEMENTS:
- [x] Do not shut down app if SSH connection fails


dtm2451:
> @dtm2451 in terms of the cancel endpoint:
> 
> * after you hit `/cancel`, its typically a good idea to let snakemake do its thing and cleanup - this takes a couple of seconds. So before any subsequent user interaction - it's a good idea to hit `get("/api/v2/#{PROJECT}/workspace/#{workspace.id}/running")` and wait until `running=False` before proceeding

Easy enough to do!

vulcan/lib/server/controllers/vulcan_v2_controller.rb
```diff
@@ -368,6 +410,45 @@ def cluster_latency
     end
   end
 
+  def delete_workspace
```
amadeovezz, line 413: @graft @dtm2451 thoughts around db deletes?

Not sure if you guys have any opinions or conventions around deletes for similar endpoints or just mark objects as deleted.

amadeovezz: we would delete the workspace, run and config records. Which I think is okay. Those records really just store a lot of metadata about the workspace anyways. And since we seem to all agree that we are deleting the workspace directory - those records would just point to stale filesystem paths / data.

amadeovezz: maybbewe can at least start with this approach and then if a use case comes up - move to soft deletes

PR #1419, opened 2025-07-08
Vulcan Status Report

dtm2451:
Adds a Connection Status Check on visit, with display in the Nav Bar.

This API includes 3 elements:
1. a real check of if there is an active c4 conenction
2. sourced from the app config, a call of whether downage might be 'expected' or not at this time
3. sourced from the app config, a message for describing current or future maintenance windows

The UI in the NavBar then uses the MUI Alert component where icon&color can be easily adjusted based on the current situation.  Scroll is used to keep this element constrained within our standard NavBAr height.

## Some examples:
Known Maintenance Coming:
image
Connection Failure that's expected:
expected-failure
Failure that's NOT expected:
unexpected-failure

### DIFF
--- a/vulcan/lib/client/jsx/api_types.ts
+++ b/vulcan/lib/client/jsx/api_types.ts
@@ -325,6 +325,7 @@ export type LatencyReturn = {
 };
 
 export type ClusterStatusReturn = {
+  connection_success: boolean;
   expected_down: boolean;
   message: string; // of form "#{median_latency}ms"
 };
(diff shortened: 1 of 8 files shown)


amadeovezz:
backend code looks good to me! although the code is super simple - its probably worth just adding one quick spec. Happy to help here!

dtm2451:
Updated the UI to just make use of our normal messaging system in all cases:

Known Maintenance Coming:
image
Connection Failure that's expected:
image
Failure that's NOT expected:
image

PR #1430, opened 2025-08-04
Production Deploy

graft:
Deploys vulcan fixes to cancel and delete workflows along with project dashboard widget in Timur.


dtm2451:
FYI, only gets out the back-end APIs for vulcan workspace deletion and work canceling.  I'll be building the front-end hookups for 'em this week!

PR #1436, opened 2025-08-12
Some hot fixes/features for Vulcan

amadeovezz:
BUG FIXES
- [x] Use config values as params instead of named snakemake params
- [x] Comment out flakey tests (until race condition is fixed on CI)

NEW FEATURES
- [x] Modify what files are returned when downstream files are stale


graft:
Don't forget to disable health_check at `lib/ssh_connection_pool.rb:36`, I did not do this on my latency branch.

PR #1448, opened 2025-08-26
vulcan FE - project dashboard tweaks

dtm2451:
This PR makes various tweaks to the Vulcan Project Dashboard page aimed at better UX on the way from entry to having a workspace running.

It does not tackle Workflow vs Workspace confusion much, but is still WIP so perhaps more will be added on that front after we brainstorm some more ideas.

This PR:
- [X] Shows disabled Create Workspace button and makes it BIGGER.  (with a tooltip explaining to pick a workflow when it is disabled)
- [X] Hides the Create Work*flow* button unless superuser
- [X] Improves the workflow card's information content by:
    - [X] making the repo url a clickable link
    - [X] removes the currently ununsed image portion (which also makes these card visibly distinct from workspace cards)
- [X] Adjusts the Create Workspace Modal:
    - [X] to have less scary, and shorter, text describing what happens when you click 'Create Workspace'
    - [X] makes the fact that there is a "Default" workflow version more obvious so that the user can be less scared about this decision!
- [X] Improves workspace cards by adding info:
    - [X] "version": a.k.a. the git_ref (branch/tag/sha used to request the workspace version during creation)
    - [X] id number
    - [X] created at date

The Dashboard:
image


dtm2451:
Workspace Creation Modal:
image

PR #1496, opened 2026-04-17
Cross Project Search - Part 1 - Infra for standardizing models 

amadeovezz:
## Problem

The flexibility of the library allows users to model data in a project however they choose. This is extremely powerful but also means it is trickier to search across projects, as some models have different names, different attributes, etc. We need a way to standardize models in the library such that we can search across projects.

## Solution

Overtime we have realized that many models across projects are similar, Ie: our `sc_seq` model. Therefore, we have made it easier for users to instantiate these models by creating a set of `template` models. But - we can just repurpose these template models as our set of standardized models! And if we have a way to associate project models to our templates, we now have a way to:

- Query across the library
- Determine what user models are "out of date" and non conforming (auditing)

From an implementation level, all we need is a pointer from user models to template models.

## Code

We need a few things to make this work:

- [x] 2 columns on `magma.models` (`template_project_name`, `template_model_name`) - point project models back to template models.
- [x] Allow writes to `template_project_name` and `template_model_name` when create new models and for existing models


magma/lib/magma/commands.rb
```diff
@@ -108,6 +111,70 @@ def options
     end
   end
 
+  class LinkTemplateModels < Etna::Command
```
amadeovezz: @graft 


thoughts on a first pass way to create a bunch of links?

etna/lib/etna/clients/magma/models.rb
```diff
@@ -58,7 +58,7 @@ def add_action(action)
         end
       end
 
-      class AddModelAction < Struct.new(:action_name, :model_name, :parent_model_name, :parent_link_type, :identifier, :date_shift_root, keyword_init: true)
+      class AddModelAction < Struct.new(:action_name, :model_name, :parent_model_name, :parent_link_type, :identifier, :date_shift_root, :template_project_name, :template_model_name, keyword_init: true)
```
graft, line 61: Incidentally this Struct.new() situation is a bad idea and seems to break in later Ruby versions; a refactor (e.g. a base class with a "params" method) is required.

magma/lib/magma/commands.rb
graft: Probably overkill to have a command for this, the backfill itself is pretty simple, the hard part of this would seem to me to be just creating the map.

amadeovezz: kk ill remove the command

etna/lib/etna/clients/magma/models.rb
amadeovezz: do we want to just keep this code consistent for now - so that when we do the refactor we change em all?

graft: yes, just pointing it out

magma/lib/magma/project.rb
```diff
@@ -131,7 +131,7 @@ def load_project
 
     def load_models
       Magma.instance.db[:models].where(project_name: @project_name.to_s).
-        reject { |model| project_container.const_defined?(model[:model_name].camelize) }.
+        reject { |model| project_container.const_defined?(model[:model_name].camelize, false) }.
```
amadeovezz, line 134: @graft 

okay so this change was discovered because I named a model "Random" in the tests and then it conflicted with Ruby's internal `Random` class. There is a low probability that someone names a model after a ruby class, but it might be nice guard for us. Passing `false` means don't check the inheritance ladder.

PR #1514, opened 2026-06-01
Metis storage lockers

graft:
A surprisingly small PR that allows Metis to shard into multiple data stores, which work like this:

Each data store is mounted into the metis app at some mount point, which is specified as a named "storage" location, with a designated `path` in Metis config. Each data store holds a collection of datablocks in /a/b folders based on the first two chars in md5 hash.

One such storage is the "active" storage and has the flag "active"; uploads and new data blocks go here. When a data store fills up, a new empty storage is added and the "active" flag is moved to the new storage. The old storage continues as normal; blocks with this storage name as their "storage" field will continue to be served (downloaded) from here.

Note that once an md5 is in a store, it does not move, that is its permanent "home" - duplicates still resolve in the same way, and newer versions of this file will be removed, so the file remains in place.


dtm2451 (approved):
Tested locally and it worked seemlessly.  Old files accessible, new files could be added, confirmed storage db column got filled

PR #1519, opened 2026-06-08
Revert "Magma project name fixes"

dtm2451:
Temporarily reverts mountetna/monoetna#1516, which was perhaps not quite finished, so that other elements might be deployed this week.


graft:
Err no need for this, it all works as expected.

dtm2451:
Oh good!  Thanks for confirming!
I wasn't sure... I got blocked trying to do my own testing, so was just aiming to be safe!

